PR #1, opened 2017-03-24
Support search by trace IDs

yurishkuro:


src/components/SearchTracePage/TraceSearchForm.js
```diff
@@ -246,6 +253,7 @@ const mapStateToProps = state => {
     maxDuration,
     minDuration,
     lookback,
+    traceID: traceIDParams,
```
saminzadeh: What is the format of the URL here?
`traceID=1&traceID=2` or `traceID=1,2`

If the later, let's use `traceIDs`

yurishkuro: the format is `traceID=1&traceID=2`, same as with tags

PR #2, opened 2017-04-11
Adding Travis Integration

saminzadeh:


yurishkuro:
it seems the lint is generating a lot of complaints, yet not causing any errors.

saminzadeh:
Those are only React library propTypes errors. We should fix them but they don't affect run time.

Also, that does not have anything to do with the Travis integration and should be fixed outside this PR request. The test are still passing accurately.

yurishkuro:
What about these logs from the tests?

image

image

I don't mind if we fix those separately, but my question related to this PR is whether those should be considered errors and fail the unit tests.

saminzadeh:
We should clean those up so messages so they don't show up, but the asserts are still passing for what we are testing. I think we should do that in a separate PR and have travis and coveralls setup independently.

PR #14, opened 2017-05-09
[wip] Added Dockerfile and instructions on how to build it.

jpkrohling:


jpkrohling:
This PR adds a simple `Dockerfile` to the UI repository. Unfortunately, on my tests, I couldn't get the UI to properly load, as I get only a blank screen. I remember reading about a similar issue, so, I'm opening this PR anyway, so that others can help me pin point the problem.

cc @pavolloffay

yurishkuro:
what's the intention here?  We normally include UI as part of the jaeger-query build, so that both static assets and API endpoints are served from jaeger-query.

jpkrohling:
Should it, though? Isn't it better to have the UI being served by a proper HTTP server, able to do caching, HTTP/2 and so on?

yurishkuro:
It can work both ways, but it seems running a separate web server is, in many cases, an overkill, it's not getting that much traffic to worry about caching etc. In our internal setup it was certainly a lot more hustle to set up infra to run ui and query as split services.

ymmv

README.md
```diff
@@ -45,13 +45,25 @@ npm start
...
+
+To generate a Docker image, run:
+
+```bash
+docker build -t yournamespace/jaeger-ui .
+```
+
+You can run it as:
+```bash
+docker run --rm -p 8080:8080 yournamespace/jaeger-ui
```
yurishkuro: does it not need extra instructions for proxying API requests?

jpkrohling: Right, that might be the part I'm missing.

saminzadeh:
I agree, running it standalone would be a bit overkill in my opinion. Also, to avoid CORS, the UI expects `/api` to be  on the same URL. 

We should be able to add the same caching rules in the Go static server.

jpkrohling:
I see no big issues in using the query service to deliver the static assets as well. From a "microservices" perspective, though, it would be better to keep each service with its own concerns.

In any case, the question at this point if is it should at least work when the UI is separate from the `query` service. The final decision, as I see it, should belong to whoever is deploying it.

yurishkuro:
it works for us in development, the UI server proxies the API request to the query service. I am not sure how exactly to run it in this mode, @saminzadeh can help.

jpkrohling:
Alright. For a first iteration, I'll keep it simple and just embed the static files into the query, as it's commonly done. We can split it later, as it would require a CORS handler at the query side and an ability to configure the query server URL at the UI side.

I'd like to leave this PR open, though, as I'll eventually come back to this.

jpkrohling:
I'll close this for now, as there are no plans to come back to this task at the moment.

PR #15, opened 2017-05-10
Fix issue 9 - Span timeline not compatible with Safari

tiffon:
Resolves the issue with the span timeline view in Safari.

Edited UI view in Chrome and Safari are attached.

Chrome:
jaeger_ui_issue_9_chrome

Safari:
jaeger_ui_issue_9_safari

### DIFF
--- a/src/components/TracePage/TraceTimelineViewer/index.js
+++ b/src/components/TracePage/TraceTimelineViewer/index.js
@@ -198,10 +198,8 @@ TimelineRow.Left.propTypes = {
 TimelineRow.Right = props => {
   const { children, ...rest } = props;
   return (
-    <div className="col-xs-9" {...rest}>
-      <div className="relative" style={{ height: '100%' }}>
-        {children}
-      </div>
+    <div className="col-xs-9 relative" {...rest}>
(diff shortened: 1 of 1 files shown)


src/components/TracePage/TraceTimelineViewer/index.js
```diff
@@ -198,10 +198,8 @@ TimelineRow.Left.propTypes = {
 TimelineRow.Right = props => {
   const { children, ...rest } = props;
   return (
-    <div className="col-xs-9" {...rest}>
-      <div className="relative" style={{ height: '100%' }}>
```
yurishkuro: was the `height` not necessary?

yurishkuro:
@saminzadeh could you please review?

src/components/TracePage/TraceTimelineViewer/index.js
tiffon: On the 

```
<div className="relative" style={{ height: '100%' }}>
```

It was necessary (due to the `relative`). But, it is not needed on the `<div className="col-xs-9 relative" {...rest}>`

PR #18, opened 2017-05-13
Fix #6 - Hierarchical error span highlighting

### LINKED ISSUE #6: Hierarchical error span highlighting
Something we discussed in the past.
  1. Use a red icon (like a stop sigh) next to span name to indicate errors, instead of highlighting the whole row
  1. When the spans are collapsed, the error indicator should bubble up to the first visible ancestor span, but go back to the proper level when spans are expanded again.

tiffon:
Adds a red warning icon to span with error tags. Removes the light red background from them.

Full UI showing the icon next to the span name because the span has an error tag:
_jaeger_ui_006-0

Detail showing the icon because a collapsed span has a descendant with an error tag:
_jaeger_ui_006-1

Detail showing the icon because a collapsed span has a descendant with an error tag:
_jaeger_ui_006-2

### DIFF
--- a/src/components/TracePage/TraceTimelineViewer/index.js
+++ b/src/components/TracePage/TraceTimelineViewer/index.js
@@ -38,6 +38,7 @@ import {
   formatDuration,
   findServerChildSpan,
   isErrorSpan,
+  spanContainsErredSpan,
 } from './utils';
 import { transformTrace } from './transforms';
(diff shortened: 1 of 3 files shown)


yurishkuro:
Please see https://github.com/uber/jaeger-ui/blob/master/CONTRIBUTING.md about the PR title / commit message

tiffon:
Fixes #6

PR #19, opened 2017-05-13
Fixes #12 - TraceView header expander not working correctly

### LINKED ISSUE #12: TraceView header expander not working correctly
When the header of the trace view is collapsed, it leaves a bunch of whitespace. The goal of the expander was to allow for more screen real estate for the main timeline view.

image

image

cc @saminzadeh

tiffon:
This fixes #12, but it is a hack—`setTimeout()` with `forceUpdate()` are used to ensure the `<Sticky />` component has the latest dimensions of its content when calculating its layout.

IMO, a better long-term fix is to remove `react-sticky`. Using `position: fixed`, directly, should suffice because the elements that need to be "sticky" are flush with the top at all times and therefore can just be position fixed.

Note: I wrapped the `<Sticky />` element with a `<StickyContainer />`, per [the intended use of `react-sticky`](https://github.com/captivationsoftware/react-sticky/blob/master/examples/basic/basic.js#L25), but that did not resolve the issue.

### DIFF
--- a/src/components/TracePage/index.js
+++ b/src/components/TracePage/index.js
@@ -121,6 +121,7 @@ export default class TracePage extends Component {
   toggleSlimView(slimView) {
     this.setState({ slimView });
     // fix issue #12 - TraceView header expander not working correctly
+    // TODO: evaluate alternatives to react-sticky
     setTimeout(() => this.forceUpdate(), 0);
   }
 
--- a/src/components/TracePage/index.js
+++ b/src/components/TracePage/index.js
@@ -120,6 +120,8 @@ export default class TracePage extends Component {
 
   toggleSlimView(slimView) {
     this.setState({ slimView });
+    // fix issue #12 - TraceView header expander not working correctly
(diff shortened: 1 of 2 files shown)


src/components/TracePage/index.js
```diff
@@ -120,6 +120,8 @@ export default class TracePage extends Component {
 
   toggleSlimView(slimView) {
     this.setState({ slimView });
+    // fix issue #12 - TraceView header expander not working correctly
+    setTimeout(() => this.forceUpdate(), 0);
```
saminzadeh: Agreed this is not ideal, so we'll looking into replacing ReactSticky. Lets just add a // `TODO:` comment here and we can merge it

tiffon: Sounds good.

yurishkuro:
Thanks, @tiffon , I am going to merge it.

tiffon:
No problem, @yurishkuro

PR #35, opened 2017-06-13
Fixes #3: Update react to 15.5

### LINKED ISSUE #3: Update to React 15.5
The latest version of React has been released: 
https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html

saminzadeh:
This fixes #3. 

**Summary:**
React 15.5 introduce some important changes:
- Moving createClass and propTypes into separate packages, so I ran the codemode to fix those.

More importantly, I need to create sub tickets to upgrade packages using older versions of React to allow for easier upgrades in the future.


saminzadeh:
> lgtm. I noticed s/.array/.arrayOf/ change, but only in one place, is that expected?
Yes, it was a typo that happened to work with early versions of React. I double checked to make sure this was the only case.

PR #36, opened 2017-06-14
Update react scripts to 1.0.7

saminzadeh:


saminzadeh:
I realized that newest version has a breaking change of adding `jsx-a11y/img-has-alt` eslint rule. Will re-open when that is fixed.

PR #46, opened 2017-07-31
Fix #43 - Search page UI responsiveness sluggish when many spans in results

### LINKED ISSUE #43: UI responsiveness of the search page is sluggish when many spans in results
**Description**

The responsiveness of the search results page slows as the summed number of spans in the search results increases. For instance, with ~20,000 spans, the search form takes several seconds to register the focus on one of the text inputs.

1. Using the HotROD app, register 400 traces
    ```
    var i = 100; var btns = [].slice.call(document.querySelectorAll('.hotrod-button')); while (i--) btns.forEach(btn => btn.click());
    ```
2. On the search page of the Jaeger UI, select Service = `frontend` and Limit Results = `400`
3. Submit the search and let the page finish loading
4. Click the Min Duration text input
5. Note: Several seconds elapse before the input receives focus
6. Click the Max Duration text input
7. Note: Several seconds elapse before the input receives focus

Several seconds elapse before the form inputs register focus.

The form inputs register focus immediately.

I've identified the underlying issue, have made changes to address it and will issue a pull request within the next couple of days.

tiffon:
Fix #43

Fix an issue with the search page that shows up when there are many spans (cumulatively) in the search results. Issue at hand is the `mapStateToProps` function calls `.toJS()` on some of the immutableJS state values, transforms the resultant values, then passes the transformed values to reselect. The `.toJS()` invocation prevents the calculations from being memoized by reselect because it always results in new objects or arrays.

The change is to wrap the `.toJS()` call and subsequent transformation in a function, the result of which is cached based on the state value. The caching resolves the issue. Only the last result needs to be cached because these are top-level state values (vs transforming items in an array) and therefore a state value will never be revisited.

### DIFF
--- a/src/components/SearchTracePage/index.js
+++ b/src/components/SearchTracePage/index.js
@@ -42,6 +42,7 @@ import {
   MOST_RECENT,
 } from '../../selectors/search';
 import { getPercentageOfDuration } from '../../utils/date';
+import getLastXformCacher from '../../utils/get-last-xform-cacher';
 
 /**
  * Contains the dropdown to sort and filter trace search results
@@ -196,6 +197,7 @@ export default class SearchTracePage extends Component {
 
 SearchTracePage.propTypes = {
   isHomepage: PropTypes.bool,
+  // eslint-disable-next-line react/forbid-prop-types
   traceResults: PropTypes.array,
   numberOfTraceResults: PropTypes.number,
   maxTraceDuration: PropTypes.number,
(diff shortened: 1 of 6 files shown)


src/utils/get-last-xform-cacher.js
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Given a transformer function, returns a function that invokes the
+ * transformer on the argument and caches the transformation before returning
+ * it. If the source value changes, the transformation is recalculated, cached
+ * and returned.
+ *
+ * @param  {function} xformer The transformer function, the most recent result
+ *                            of which is cached.
+ * @return {function} A wrapper around the transformer function which which
```
yurishkuro: which which

src/components/SearchTracePage/index.js
```diff
@@ -218,20 +220,32 @@ SearchTracePage.propTypes = {
   errorMessage: PropTypes.string,
 };
 
+const stateTraceXformer = getLastXformCacher(stateTrace => {
+  const { traces: traceMap, loading, error: traceError } = stateTrace.toJS();
```
yurishkuro: Q (curious): how is `stateTrace` represented in the first place? What does the call to `toJS()` achieve?

src/utils/get-last-xform-cacher.js
```diff
@@ -0,0 +1,48 @@
...
+ *                    caches the last transformation, returning the cached
+ *                    value if the source value is the same.
+ */
+export default function getLastXformCacher(xformer) {
```
saminzadeh: Shouldn't reselect or lodash memoize do this?

src/utils/get-last-xform-cacher.js
tiffon: Thanks.

src/components/SearchTracePage/index.js
tiffon: The state is an [Immutable.js](https://facebook.github.io/immutable-js/) [Map](https://facebook.github.io/immutable-js/docs/#/Map). [`.toJS()`](https://facebook.github.io/immutable-js/docs/#/Map/toJS) converts it to a plain JS object. The `.toJS()` call is probably superfluous; I'll go over why I didn't take it out in our discussion Wednesday.

src/utils/get-last-xform-cacher.js
tiffon: @saminzadeh Most likely not as we only need the last calculated value to be cached rather than the function to be fully memoized.

src/utils/get-last-xform-cacher.js
```diff
@@ -0,0 +1,48 @@
...
+ *                    the source value is the same.
+ */
+export default function getLastXformCacher(xformer) {
+  let lastArgs = null;
+  let lastXformed = null;
+
+  return function getOrCache(...args) {
+    const sameArgs = lastArgs &&
+      lastArgs.length === args.length &&
+      lastArgs.every((lastArg, i) => lastArg === args[i]);
```
saminzadeh: If .toJS() is returning a new reference, would `lastArg === args[i]` fail the test?

Also, _.memoize does the same thing but stores any existing computations rather than just the last computation. You can also pass a custom cache resolver if we need to
https://lodash.com/docs/4.17.4#memoize

tiffon: @saminzadeh The immutable JS values are the params to the [transform functions](https://github.com/uber/jaeger-ui/blob/master/src/components/SearchTracePage/index.js#L223,L240); `.toJS()` is called within the transform functions, so the caching with the param as the key should be fine.

In this case, we only care about caching the last calculation. Params to the transform function will never be revisited, so we don't need to cache anything other than the most recent calculation.

The reason the params will never be revisited is because they are top level values on the state, `traces` and `services`. When the search results change, `traces` will be different, and the previous value will never be revisited even if an identical search is executed. Same thing with `services`.

PR #48, opened 2017-08-08
Remove ImmutableJS, refactor search selectors

tiffon:
**Removed ImmutableJS (T1077867)**

- Issues with how the library was utilized (#43)
- Incorrect usage can result in perf issues and bugs
- Not ubiquitous knowledge - barrier to entry for open source contributors

**Refactor search selectors**

Data transformation for the search page was simplified and performs better. Added flow types for the derived data. Will eventually remove `get-last-xform-cacher.js`, but the way the components are organized needs to be adjusted, first.


src/components/SearchTracePage/index.js
```diff
@@ -18,29 +18,23 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
+import _values from 'lodash/values';
```
saminzadeh: Is this a common way of naming single lodash functions? Historically, this is generally meant for private variables.

Are we opposed to bring in the whole lib because of size? If so, I suppose we could do this style as long as we are consistent throughout the application.

src/model/order-by.js
```diff
@@ -0,0 +1,5 @@
+export const MOST_RECENT = 'MOST_RECENT';
```
saminzadeh: I think this file needs license info, `npm run add-license`

src/model/search.js
```diff
@@ -0,0 +1,117 @@
...
+
+  let traceName = '';
+  let minTs = Number.MAX_SAFE_INTEGER;
+  let maxTs = Number.MIN_SAFE_INTEGER;
+  let numErrorSpans = 0;
+  // serviceName -> { name, numberOfSpans }
+  const serviceMap = {};
+
+  let i = 0;
+  for (; i < spans.length; i++) {
```
saminzadeh: Not familiar with this style? Is there an advantage than doing `for (let i = 0; i < spans.length; i++)`

src/reducers/trace.js
```diff
@@ -18,35 +18,52 @@
...
+}
+
+function fetchTraceDone(state, { meta, payload }) {
+  const trace = enforceUniqueSpanIds(payload.data[0]);
+  const traces = { ...state.traces, [meta.id]: trace };
+  return { ...state, traces, loading: false };
+}
+
+function fetchTraceErred(state, { meta, payload }) {
+  const traces = Object.assign({}, state.traces, { [meta.id]: payload });
```
saminzadeh: Spread operator instead of Object.assign ?

src/components/SearchTracePage/index.js
tiffon: @saminzadeh I have mixed feelings about the name, myself, for the same reason. I went with `_values` because 
- `values` would be an odd name for it as it's a function not a collection of values
  - Come to think of it, though, `_values` has basically the same issue
- `_values` is reminiscent of `_.values` 

I noticed another [project](https://github.com/lodash/babel-plugin-lodash#example) using the same approach, but I'm definitely open to alternatives... Maybe `valuesFn`? (or `lodashValues`? Long names are a bother, though.)

Yes, importing by function (also called cherry-picking, I think) yields smaller builds. Consistency in the project, in this respect, is a WIP.

src/model/order-by.js
tiffon: Thanks. Done.

src/model/search.js
tiffon: Personal preference. Changed to the general style.

src/reducers/trace.js
tiffon: I find it easier to read. For comarison, [the diff](https://github.com/uber/jaeger-ui/pull/48/commits/661359d086e2b97ce2cb48d6272c44051eeb533b).

Additional discussion on it from the redux crew: [docs](http://redux.js.org/docs/recipes/UsingObjectSpreadOperator.html)

tiffon: Misread your comment. Good catch, done.

src/components/SearchTracePage/index.js
saminzadeh: Ok cool, im open to the lodash function style of _values as an alias for _.values. It would be sweet if we could use this babel plugin but then we'd have to eject. https://github.com/lodash/babel-plugin-lodash

PR #53, opened 2017-08-11
Refactor trace detail

tiffon:
### TODO
- [x] Maintenance of tests
- [x] Cleanup unused vars, imports
- [x] Refine utils
- [x] Many styles will be moved to CSS

### Deferred Tasks
- [ ] (Deferred) ~Finish shifting to transformed trace, exclusively, instead of raw trace~
- [ ] (Deferred) ~Cleanup ordering of vars, imports~
- [ ] (Deferred) ~Refine trace detail components~
- [ ] (Deferred) ~Refine trace selectors~
- [ ] ~Will likely remove `model/trace-viewer.js`~

### SpanGraph

- Fix #49 - Span position in graph doesn't not match its position in the detail
- Ticks in span graph made to match trace detail (in number and formatting)
- Span graph refactored to trim down files and DOM elements
- Styling adjustments

### TracePageHeader

- `trace` prop removed
- Added props for various title values instead of deriving them from `trace`

### Trace Detail

- Several components split out into separate files
- `transformTrace` to use already created span tree to determine span depth
- Fix #59 - "Span Name" to "Service & Operation"

### Span Bar / Detail

- Fix uber/jaeger#326: extraneous scrollbars in trace views
- Fix #55: Some tags were not being rendered due to clashing keys (observed in a log message)
- Tall content scrolls via entire table instead of single table cell
- Horizontal scrolling for wide content (e.g. long log values) (Fix #58)
- Full width of the header is clickable for tags, process, and logs headers (instead of header text, only)
- Service and endpoint are shown on mouseover anywhere in span bar row
- Label on span bars no longer off-screen
- Clip or hide span bars when zoomed in (instead of flush left)
- Add shadow to left / right boundary when span bar view is clipped
- Darkened span name column to differentiate from span bar section
- Span detail left column color coded by service
- Clicking span detail left column collapses detail
- Clicking anywhere left of parent span name toggles children visibility
- Prevent collision of logs in log entries table

### SearchTracePage
- Scatterplot dots are sized based on number of spans
- Scatterplot dots mouseover shows trace name

### Misc
- Several TraceTimelineViewer / utils removed
- `TreeNode` `.walk()` method can now be used to calculate the depth, avoiding use of less efficient `.getPath()`
- Removed several `console.error` warnings caused by React key issues
- `yarn upgrade --latest`
- Removed `react-sticky`
- Fix #42 - Support URL prefix via homepage in package.json

PR #64, opened 2017-08-22
Reorganize span detail components

tiffon:
For now, target branch is `refactor-trace-detail-selectors`, PR #53 is a WIP.

- Split sub-components of `SpanDetail` out into separate files.
- Move styles into CSS (addresses [this PR comment](https://github.com/uber/jaeger-ui/pull/53#discussion_r134318660))
- Misc style and DOM tweaks  (see attached screenshot)
  - Expanded table styling
  - Tag summary styling
  - Background on hover of expandable / collapsable elements
  - Debug info

span-detail-pr-marked


jcorbin:
Awesome, love the screenshot btw, makes reviewing this at a high/user level much easier.

My only feedback is: what if instead of hiding (how even do we show?) the span id, why not float it in the upper right (across from the Big Bold operation name)?  I'd even be fine if you wanted to soften the grey contrast a bit so it doesn't draw as much attention.

yurishkuro:
I suggest keeping "SpanID:" label before the hex value in bottom right.

Otherwise looks great.

tiffon:
@jcorbin 

> My only feedback is: what if instead of hiding (how even do we show?) the span id, why not float it in the upper right (across from the Big Bold operation name)?

Currently, the span ID is always showing in the lower right corner of the detail panel (the down arrow points to it). Do you mean moving it to the upper right instead of the lower right, or were you thinking it wasn't being shown at all?

tiffon:
Screenshot of "SpanID:" label added.
screen shot 2017-08-22 at 2 43 10 pm

jcorbin:
Ahh okay then, I was only suggesting that upper right seems easier to get to, rather than scroll past one or more expanded-or-collapsed sections. It's also a top-level property, similar to tags or operation name in my mind. Maybe needing span ids isn't a common enough thing to worry about tho... not sure.

yurishkuro:
> Maybe needing span ids isn't a common enough thing to worry about tho..

it isn't, imo, only used by "experts", e.g. when you want to deep dive into trace JSON

src/components/App/App.css
```diff
@@ -24,6 +24,21 @@ THE SOFTWARE.
   height: 100%;
 }
 
+.u-no-float {
+  float: none;
+}
+
+.u-cursor-pointer {
```
saminzadeh: what does the `u` prefix mean?

src/components/TracePage/TraceTimelineViewer/SpanDetail/AccordianKeyValues.js
```diff
@@ -0,0 +1,102 @@
...
+
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import KeyValuesTable from './KeyValuesTable';
+import toggleEnhancer from './toggle-enhancer';
+
+import './AccordianKeyValues.css';
+
+const dataPropType = PropTypes.arrayOf(
```
saminzadeh: Flowtype?

src/components/TracePage/TraceTimelineViewer/SpanDetail/AccordianLogs.js
```diff
@@ -0,0 +1,76 @@
...
+  logs: PropTypes.arrayOf(
+    PropTypes.shape({
+      timestamp: PropTypes.number,
+      fields: PropTypes.array,
+    })
+  ).isRequired,
+  timestamp: PropTypes.number.isRequired,
+};
+
+export default toggleEnhancer(AccordianLogs);
```
saminzadeh: Good use of HOC here

src/components/App/App.css
tiffon: @saminzadeh util

src/components/TracePage/TraceTimelineViewer/SpanDetail/AccordianLogs.js
tiffon: @saminzadeh Thanks

src/components/App/App.css
saminzadeh: Got it, I recommend bringing this in for CSS utils:
http://basscss.com/

src/components/TracePage/TraceTimelineViewer/SpanDetail/AccordianKeyValues.js
tiffon: Adding flow typing, which is handy for props, definitely reminded me why I'm skeptical — bugs.

- [Changes](https://github.com/uber/jaeger-ui/pull/64/commits/5c7f896ed4b39836af91b81995bd0b18f0cf0f60#diff-cbab2020bb1863553129838da2e79e95) that gave me trouble
- [Underlying issue](https://github.com/facebook/flow/issues/1511)
- [Simplified example](https://goo.gl/4Z23MJ)

PR #66, opened 2017-08-24
Revert installed package versions respect .nvmrc

tiffon:
New version of react-scripts requires node >= v6. Reverting react-scripts and several eslint related packages to respect the `.nvmrc` of `4`.


saminzadeh:
I'm ok upgrading.
Node 4 was the min version create-react-app supported at the time.

tiffon:
>I'm ok upgrading.
>Node 4 was the min version create-react-app supported at the time.

Oh man! Forgot we build this locally then commit the bundled files! Nice (sort of)!

saminzadeh:
We don't commit locally. We just need to make sure whatever process is doing the build has Node >= 4

tiffon:
Closing as building on node >= 6 is fine.

PR #71, opened 2017-09-09
[trave view] Mouseover expands truncated text to full length in left column

### LINKED ISSUE #65: [trave view] Mouseover expands truncated text to full length in left column
Show full text show of clipped Service + Operation text in left column on mouseover. Text should bleed into the right column (with a background added to it). 

Re discussion of #63.

span-name-truncation-marked

tiffon:
Fix #65.

See attached screenshot (text is blurry due to scaling down retina dpi).

Mouseover in the span name column expands the service/operation text if it is truncated.

Also, fixed regression in header for left column ("Span Name" -> "Service & Operation").

span-name-hover-marked

### DIFF
--- a/src/components/TracePage/TraceTimelineViewer/ListView/__snapshots__/index.test.js.snap
+++ b/src/components/TracePage/TraceTimelineViewer/ListView/__snapshots__/index.test.js.snap
@@ -42,7 +42,6 @@ ShallowWrapper {
                 style={
                     Object {
                         "height": 2,
-                        "overflow": "hidden",
                         "position": "absolute",
                         "top": 0,
                       }
@@ -53,7 +52,6 @@ ShallowWrapper {
                 style={
                     Object {
                         "height": 4,
-                        "overflow": "hidden",
                         "position": "absolute",
                         "top": 2,
                       }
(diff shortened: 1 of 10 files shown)


src/components/TracePage/TraceTimelineViewer/ListView/__snapshots__/index.test.js.snap
```diff
@@ -42,7 +42,6 @@ ShallowWrapper {
                 style={
                     Object {
                         "height": 2,
-                        "overflow": "hidden",
```
yurishkuro: what is the purpose of these `__snapshots__` files? Are they auto-generated?

tiffon:
Snapshots are a feature of jest, the unit test framework used in the
project. Yes, they're machine generated.

https://facebook.github.io/jest/docs/en/snapshot-testing.html

They serve as a thorough reference point for assertions.

src/components/TracePage/TraceTimelineViewer/ListView/__snapshots__/index.test.js.snap
saminzadeh: Yep, https://facebook.github.io/jest/docs/en/snapshot-testing.html

tiffon: Oh man, my reply was posted to the wrong thread :/

The snapshots are for tests.

tiffon:
@saminzadeh Thanks for the giphy tip! Will keep it in mind...

PR #72, opened 2017-09-11
Use canvas instead of SVG for trace mini-map 

### LINKED ISSUE #61: Use canvas instead of SVG for trace mini-map
To improve performance on traces with many spans, use the [canvas API](https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API) to render the trace mini-map instead of SVG elements.

tiffon:
Fix #61.

- Render span graph via canvas instead of SVG
- Zoom range changed to [0, 1], e.g. time and trace agnostic (allowed removal of some utils)
- "Timeline" -> 0ms
- Use props instead of context to provide span graph with viewing range
- Move all span graph related classes into same folder
- Misc CSS cleanup


src/components/TracePage/SpanGraph/index.css
```diff
@@ -20,19 +20,23 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
...
-  stroke-width: 1px;
+.SpanGraph--zlayer {
+  position: relative;
+  z-index: 1;
 }
 
-.span-graph--tick-header {
+.SpanGraph--graph {
+  border: 1px solid #999;
+  overflow: visible !important;
```
saminzadeh: Is there a particular reason we need to use `!important` here? 
Might be better to inline it on the element to be more explicit if it is a functional requirement

tiffon: Yes, we need it to overcome this from semantic UI:

```css
svg:not(:root) {
    overflow: hidden;
}
```

Will add a comment.

Thanks.

src/components/TracePage/SpanGraph/CanvasSpanGraph.js
```diff
@@ -0,0 +1,80 @@
...
+        className="CanvasSpanGraph"
+        ref={this._setCanvasRef}
+        width={CV_WIDTH}
+        height={this.props.items.length}
+      />
+    );
+  }
+}
+
+CanvasSpanGraph.propTypes = {
```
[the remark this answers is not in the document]
tiffon: Done. 👍

PR #74, opened 2017-09-14
Make left column adjustable in trace detail

### LINKED ISSUE #63: Make width of left column adjustable in trace detail view
The width of the left column in the trace detail view can be made adjustable to accommodate deeply nested spans, e.g. the same way the width of a column in a table can be adjusted.

span-name-truncation-marked

tiffon:
Fix #63 

Column width of span name is now adjustable, within the bounds of [15%, 85%]. It's not clear in the gif, below, but the reduced opacity highlight color is present when dragging the column in either direction.

64-01-resize-left-column


src/components/TracePage/TraceTimelineViewer/SpanBarRow.js
```diff
@@ -68,7 +69,7 @@ export default function SpanBarRow(props) {
         ${isFilteredOut ? 'is-filtered-out' : ''}
       `}
     >
-      <TimelineRow.Left className="span-name-column">
+      <TimelineRow.Cell className="span-name-column" width={columnDivision}>
```
saminzadeh: Nice, this is better with `width`

src/components/TracePage/TraceTimelineViewer/Ticks.js
```diff
@@ -18,42 +20,52 @@
...
-import React from 'react';
+import * as React from 'react';
+
+import { formatDuration } from './utils';
 
 import './Ticks.css';
 
-export default function Ticks(props) {
-  const { labels, ticks } = props;
+type TicksProps = {
```
saminzadeh: Seeing a mixture of flowTypes and propTypes in this review. Let's stick with one

src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow.css
```diff
@@ -0,0 +1,124 @@
...
+  border-left: 1px solid #999;
+}
+
+.TimelineColumnResizer--wrapper.isDraggingLeft > .TimelineColumnResizer--dragger {
+  border-left: 1px solid #808;
+  /* #808 === rgb(136, 0, 136) */
+  background: rgba(136, 0, 136, 0.05);
+}
+
+/* .TimelineColumnResizer--dragger.isDraggingRight,
```
saminzadeh: Remove?

src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow.js
```diff
@@ -0,0 +1,190 @@
...
+
+  _onWindowMouseMove = function _onWindowMouseMove({ clientX }) {
+    const { rootX } = this.state;
+    this.setState({ rootX, currentDragPosition: clientX - (rootX || 0) });
+  };
+
+  _onWindowMouseUp = function _onWindowMouseUp({ clientX }) {
+    window.removeEventListener('mousemove', this._onWindowMouseMove);
+    window.removeEventListener('mouseup', this._onWindowMouseUp);
+    if (document && document.body && document.body.style) {
```
saminzadeh: You can simplify with lodash `_.get(document, 'style.body')`

src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow.js
```diff
@@ -0,0 +1,190 @@
...
+      (document.body.style: any).userSelect = 'none';
+    }
+  };
+
+  _onWindowMouseMove = function _onWindowMouseMove({ clientX }) {
+    const { rootX } = this.state;
+    this.setState({ rootX, currentDragPosition: clientX - (rootX || 0) });
+  };
+
+  _onWindowMouseUp = function _onWindowMouseUp({ clientX }) {
```
saminzadeh: We seem to be doing a lot of stuff with dragging and complex calculations inline react components that can be hard to follow in future development.

Is there a solid library we can use that abstracts this stuff out or an HOC pattern we can reuse here?

src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow.css
tiffon: 👍

src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow.js
tiffon: Went ahead and extracted the calculations out into their [own function](https://github.com/uber/jaeger-ui/pull/74/commits/00e83d37a3f558f931415725783a8267c7e688ec#diff-7c366cc806a433631c209ceb514c9877R89) and [simplified](https://github.com/uber/jaeger-ui/pull/74/commits/00e83d37a3f558f931415725783a8267c7e688ec#diff-7c366cc806a433631c209ceb514c9877R154) the way the dimensions of the highlight are calculated (via using `left`, `right` instead of `width`—simpler calculations).

If you think it needs more changes, can you elaborate?

yurishkuro:
@tiffon Please use "good commit message" guidelines in `CONTRIBUTING.md` when merging. The easiest way to do that is to make the PR titles follow the same rule, i.e. instead of "Fix #63 - Left column adjustable in trace detail", say "Make left column adjustable in trace detail", and "Fix #63" in the PR description (it doesn't need to be in the commit title since it's meaningless purely within Git, only makes sense in the context of GitHub and the PR).

yurishkuro:
The s/s looks great!

src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow.js
tiffon: Thanks. I'd like to use [idx](https://github.com/facebookincubator/idx), but it's not worth ejecting...

src/components/TracePage/TraceTimelineViewer/Ticks.js
tiffon: The `PropTypes` are holdovers from before. I've moved all of `TraceTimelineViewer/*` to use flow.

tiffon:
@yurishkuro Thanks for adjusting the title and description.

Great to hear the s/s looks good 👍

src/components/TracePage/TraceTimelineViewer/Ticks.js
saminzadeh: Ok gotcha

PR #78, opened 2017-09-21
Custom menu via /api/config with project links as defaults

### LINKED ISSUE #44: Add configurable, persistent links to the header
A site administrator should be able to add global, persistent links to the header.

Current needs for the links are:

- Support one level of grouping, e.g. Help -> { Wiki, Create Ticket, FAQs }
- Each link should have a label and a URL

The configuration mechanism should not be specific to this task, but general enough to support other scenarios, such as modifying the defaults for the search panel, etc.

Design of the configuration mechanism is TBD — details to be added after a design discussion.

tiffon:
Fix #44.

- Configurable top-level links
- Attempts to download `/api/config`, errors cause fall-back to default
- Supports one level of nesting
- Supports a `label` and `url` for each link
- Groups support a `label` and `items` where `items` is an array of links
- **All links open to external tabs**
- **Does not currently support a "selected" state for links**

Default menu config:

- About Jaeger
    - [GitHub](https://github.com/uber/jaeger)
    - [Docs](http://jaeger.readthedocs.io/en/latest/)
    - [Twitter](https://twitter.com/JaegerTracing)
    - [Discussion Group](https://groups.google.com/forum/#!forum/jaeger-tracing)
    - [Gitter.im](https://gitter.im/jaegertracing/Lobby)
    - [Blog](https://medium.com/jaegertracing/)
- ~Create a ticket~

**Screenshot updated**
screen shot 2017-09-21 at 2 37 09 pm

### DIFF
--- a/src/actions/jaeger-api.js
+++ b/src/actions/jaeger-api.js
@@ -25,6 +25,8 @@ import JaegerAPI from '../api/jaeger';
  * async wrapper to get the api object in case we're in demo mode.
  */
 
+export const fetchConfig = createAction('@JAEGER_API/FETCH_CONFIG', () => JaegerAPI.fetchConfig());
+
 export const fetchTrace = createAction(
   '@JAEGER_API/FETCH_TRACE',
   id => JaegerAPI.fetchTrace(id),
--- a/src/api/jaeger.js
+++ b/src/api/jaeger.js
@@ -53,6 +53,9 @@ export const DEFAULT_DEPENDENCY_LOOKBACK = moment.duration(1, 'weeks').asMillise
 
 const JaegerAPI = {
   apiRoot: DEFAULT_API_ROOT,
+  fetchConfig() {
+    return getJSON(`${this.apiRoot}config`).catch(err => ({ error: err }));
+  },
   fetchTrace(id) {
     return getJSON(`${this.apiRoot}traces/${id}`);
   },
--- a/src/components/App/TopNav.css
+++ b/src/components/App/TopNav.css
@@ -20,10 +20,18 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 */
 
-.ui.menu.jaeger-ui--topnav {
-  border-radius: 0;
+.TopNav {
+  /* !important is to override styles from semantic UI */
+  border-radius: 0 !important;
+  margin: 0 !important;
   padding: 5px;
   position: fixed;
   width: 100%;
   z-index: 1000;
+}
+
+.TopNav--DropdownItem {
+  display: block;
+  margin: -1em;
+  padding: 1em;
 }
--- a/src/components/App/TopNav.js
+++ b/src/components/App/TopNav.js
@@ -82,21 +82,21 @@ export default function TopNav(props: TopNavProps) {
       <Link to={prefixUrl('/')} className="header item">
         Jaeger UI
       </Link>
-      {menuItems.map(item => {
-        if (item.items) {
-          return <CustomNavDropdown key={item.label} {...item} />;
-        }
-        return <CustomNavItem key={item.label} {...item} />;
-      })}
+      <div className="ui input">
+        <TraceIDSearchInput />
+      </div>
+      {NAV_LINKS.map(({ key, to, text }) =>
+        <Link key={key} to={to} className="item">
+          {text}
+        </Link>
+      )}
       <div className="right menu">
-        <div className="ui input">
-          <TraceIDSearchInput />
-        </div>
-        {NAV_LINKS.map(({ key, to, text }) =>
-          <Link key={key} to={to} className="item">
-            {text}
-          </Link>
-        )}
+        {menuItems.map(item => {
+          if (item.items) {
+            return <CustomNavDropdown key={item.label} {...item} />;
+          }
+          return <CustomNavItem key={item.label} {...item} />;
+        })}
       </div>
     </Menu>
   );
--- a/src/components/App/TopNav.js
+++ b/src/components/App/TopNav.js
@@ -1,3 +1,5 @@
+// @flow
+
 // Copyright (c) 2017 Uber Technologies, Inc.
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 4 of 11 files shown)


yurishkuro:
Nice! 

* Could you please add a section to the README about how people can use a custom config without rebuilding the UI? We talked about ENV variable. Or does this have to happen in jaeger-query?
* Can we alter the "Jaeger UI" title? For example, I would like to add a datacenter name to it.
* It seems odd that the primary parts of the menu (Search/Deps/Search box) are flushed to the right, while auxiliary menu items are on the left. Usually it's the other way around.

saminzadeh:
Great work.

I'm a bit impartial about the "Create Ticket" as a default config. What do you guys think? If you provide your own, will it override existing or merge the two?

saminzadeh:
> Could you please add a section to the README about how people can use a custom config without rebuilding the UI? We talked about ENV variable. Or does this have to happen in jaeger-query?

I think thats a good idea, the build process can create an .env file and put a JSON list of the config so its technically static. Just like the GA Analytics key. If we need it to be dynamic without rebuilding, then the `/config` endpoint makes sense.

black-adder:
> I'm a bit impartial about the "Create Ticket" as a default config. What do you guys think? If you provide your own, will it override existing or merge the two?

I think for external users, it makes sense for them to ping their own tracing/infra team first before opening a ticket against us so even if we agree to have the "create ticket" option, I don't think it should be so prominent in the UI.

tiffon:
>- Could you please add a section to the README about how people can use a custom config without rebuilding the UI? We talked about ENV variable. Or does this have to happen in jaeger-query?

@yurishkuro It needs to happen in jaeger-query. I'll add info to the README when it's implemented.

>- Can we alter the "Jaeger UI" title? For example, I would like to add a datacenter name to it.

@yurishkuro That will need to be provided as part of the config that is loaded. Would providing something like the following at the `/api/config` endpoint work?

```text
{
	"menu": [etc...],
	"name": "Jaeger - Some DC"
}
```

>- It seems odd that the primary parts of the menu (Search/Deps/Search box) are flushed to the right, while auxiliary menu items are on the left. Usually it's the other way around.

@yurishkuro I'll move them around.

>I'm a bit impartial about the "Create Ticket" as a default config. What do you guys think? If you provide your own, will it override existing or merge the two?

@saminzadeh Right now, if the user provides their own config, it will fully replace the current config; they're not merged.

>I think thats a good idea, the build process can create an .env file and put a JSON list of the config so its technically static. Just like the GA Analytics key. If we need it to be dynamic without rebuilding, then the /config endpoint makes sense.

@saminzadeh We discussed using a `/api/config` endpoint, the payload can be defined via an env var for the query-service which would point to a config file. It can default to either `{}` or `404`. If we default it to the former, I need to make a small change...

>I think for external users, it makes sense for them to ping their own tracing/infra team first before opening a ticket against us so even if we agree to have the "create ticket" option, I don't think it should be so prominent in the UI.

@black-adder Currently there is only one level of promience, pretty much everything is the same. So, if we want to deemphasize the create ticket links, we should take them out.

@yurishkuro Any thoughts on taking out the create ticket links?

saminzadeh:
> We discussed using a /api/config endpoint, the payload can be defined via an env var for the query-service which would point to a config file.

@tiffon If the payload is stored in the .env file, why not just have create-react-app build with it so its available as a static JSON object in the UI code?

tiffon:
> If the payload is stored in the .env file, why not just have create-react-app build with it so its available as a static JSON object in the UI code?

@saminzadeh The current thinking is to parameterize the jaeger query-service so a custom `/api/config` can be served. This will allow users to customize the menu and whatever else ends up in the config without needing to rebuild the UI. @yurishkuro suggested the jaeger query-service can consume an environment variable which points to a custom config JSON or YAML file.

yurishkuro:
> Any thoughts on taking out the create ticket links?

I would take them out, completely agree with Won's comment. Internally we'll have Create Ticket link that goes to Phab.

saminzadeh:
> @saminzadeh The current thinking is to parameterize the jaeger query-service so a custom /api/config can be served. This will allow users to customize the menu and whatever else ends up in the config without needing to rebuild the UI. @yurishkuro suggested the jaeger query-service can consume an environment variable which points to a custom config JSON or YAML file.

Make sense but if we don't plan on supporting changing the config dynamically through an endpoint, we should use the env file since query service builds it anyways right?

```sh
jaeger-query$: echo 'MENU={ "name": "Root", "menu": [..] }' > ui/my-custom-ui-config.env
jaeger-query$: cd ui && npm run build # create-react-app reads .env file on build and injects it
```

If we plan on making a config endpoint in the query service, might as well move all options there including the Google analytics key (which is currently done through .env file)
```
HTTP PUT /config
{ "uiMenu": { "name": "Root", "menu": [..] } }
```
It would make sense via GET /config

My point is, we should pick one for all UI config options

tiffon:
@saminzadeh We are planning to support dynamic config through the endpoint.

It would be `GET /api/config` because anything other than `/api` is served as a static files off the disk whereas the config needs to check the env-var and possibly serve some other static file.

Moving the Google Analytics key to the config is a great suggestion, nice :+1:

saminzadeh:
> We are planning to support dynamic config through the endpoint.

Ok, in that case you're approach sounds good.

tiffon:
I've taken out the create ticket links and moved the custom links to the right side of the nav.

tiffon:
@saminzadeh Will look into moving the google analytics to the config in a separate ticket.

Right now the GA integration is not working, I think it might be related to the upgrading react-router. Secondly, moving it to the config will kill GA integration (once it's working again) until the query-service supports the config endpoint.

tiffon:
@yurishkuro This change implies a change in the query service. It can function without the change, but I need to make sure it will handle the default response of the query service, properly. 

What will the default response be from the query service? We discussed it being `{}`.

yurishkuro:
it can be that, or a 404

saminzadeh:
I would avoid a 404 (browser shows it as an error in the console, kinda annoys me). I'm more a fan of returning some default object, maybe version number and other useful information then having the config be a property on that main object even if people don't set it up.

ex:
```
{
  version: "xx",
  commit: "xxx",
  menu: null
}

PR #81, opened 2017-09-22
Fix Google Analytics tracking

### LINKED ISSUE #80: Google Analytics integration is not working
**Description**

Custom builds of the UI that provide the Google Analytics key are not reporting to google analytics. The library is downloaded, but page views are not sent.

1. Provide a Google Analytics ID at build-time
2. Observe the google analytics library is downloaded when the page is loaded
3. Observe there aren't any calls to google analytics to track page-views

Tracking is not enabled.

Tracking is enabled.

n/a

**Additional environment details (Browser, etc.):**
n/a

tiffon:
Fix #80.

Created an HOC that checks the state of the router and tracks a page-view whenever the URL or query string changes. This is necessary because react-router v4 [did away with the onUpdate](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/guides/migrating.md#on-properties) lifecycle event.


src/components/App/index.js
```diff
@@ -30,20 +29,22 @@ import 'semantic-ui-css/semantic.min.css';
...
-import metricConfig from '../../utils/metrics';
 import prefixUrl from '../../utils/prefix-url';
 
 import './App.css';
 
-const PageWithMetrics = metrics(metricConfig)(Page);
-
 const defaultHistory = createHistory();
 
+const TrackedSearchPage = trackedComponentEnahncer(ConnectedSearchTracePage);
```
saminzadeh: Why break this up? If we do it at the top level container, we don't have to add it to each page.

If our concern is that the Page only mounts once, I would hook into React router page change event and do it there.

saminzadeh: Scratch my comment, didn't realize we were on react router v4.

saminzadeh: I'm wondering if we can do something similar to this helper library 
https://github.com/pshrmn/rrc/blob/master/docs/OnUpdate.md

It would be nice if we could listen to all changes on the location without explicitly putting it on each page component. That way, when a new page is added, it will automatically start tracking it without having to use this HOC. Thoughts?

tiffon: Hm, could possibly connect it to the store and use `state.location` instead... I'll look into that.

tiffon: Moved tracking into `Page`.

src/components/App/Page.js
```diff
@@ -18,26 +20,54 @@
...
+  location: Location,
+  children: React.Node,
+};
+
+class Page extends React.Component<PageProps> {
+  props: PageProps;
+
+  componentDidMount() {
+    const { pathname, search } = this.props.location;
+    trackPageView(pathname, search);
```
saminzadeh: Nice, I think this is more robust. 

One comment, given that the GA key will come through the HTTP Config service (async), we need to make sure the key is initialized beforehand so we don't miss any page view events

tiffon: Yeah, I might make a separate ticket for that. Don't want to hold up fixing GA tracking or getting the menu working.

PR #83, opened 2017-09-24
Fix traceIDs query params clobbering form search

### LINKED ISSUE #82: Jaeger UI carries over traceID URL parameters into future searches
**Description**

When navigating to search page with a traceID in the query parameters, all future searches via the search form only return the specified trace.

1. Go to a page with a traceID search, e.g. http://localhost:3000/search?traceID=abc123
2. Use the search functionality on the left to search for anything
3. Only the trace that we searched for in step 1 shows up instead of the search results
  - The URL is http://localhost:3000/search?end=1506232861152000&limit=20&lookback=1h&maxDuration&minDuration&service=someservervice&start=1506229261152000&traceID=abc123

Only the trace specified by the traceID is present in the search results.

Fresh search results. The trace specified by the original traceID query parameter should not be carried into the search results unless it is in the result set.

_T1071845_

tiffon:
Fix #82.

Changed functionality is: traceIDs are ignored when the search form is triggering the search.

### DIFF
--- a/src/components/SearchTracePage/TraceSearchForm.js
+++ b/src/components/SearchTracePage/TraceSearchForm.js
@@ -285,8 +285,8 @@ const mapDispatchToProps = dispatch => {
         minDuration,
         maxDuration,
         lookback,
-        traceIDs,
       } = fields;
+      // Note: traceID is ignored when the form is submitted
 
       store.set('lastSearch', { service, operation });
 
@@ -317,7 +317,6 @@ const mapDispatchToProps = dispatch => {
         tag: tagsToQuery(tags) || undefined,
         minDuration: minDuration || null,
         maxDuration: maxDuration || null,
-        traceID: traceIDsToQuery(traceIDs) || undefined,
       });
     },
   };


src/components/SearchTracePage/TraceSearchForm.js
```diff
@@ -317,7 +317,6 @@ const mapDispatchToProps = dispatch => {
         tag: tagsToQuery(tags) || undefined,
         minDuration: minDuration || null,
         maxDuration: maxDuration || null,
-        traceID: traceIDsToQuery(traceIDs) || undefined,
```
yurishkuro: shouldn't this remain? When the search is performed by external URL that contains trace IDs in it, we definitely want them to be used, but when the search is performed via the search form in the UI we don't want the IDs from the previous search to be used

tiffon: @yurishkuro No, this doesn't need to be here. This is only taking them out of the form submission.

The LOC that handles traceIDs from the URL: [src/components/SearchTracePage/index.js#L72](https://github.com/uber/jaeger-ui/blob/master/src/components/SearchTracePage/index.js#L72)

saminzadeh: +1

PR #84, opened 2017-09-25
Improve search dropdowns

### LINKED ISSUE #30: The Services dropdown search should be case-insensitive
The actual service names on the backend are case-sensitive, but the auto-complete in the dropdown should not be.

### LINKED ISSUE #31: Change auto-complete boxes to "contains"
Currently both the input for service name and operation name are matching the prefix of the service/operation. I think that it would be more convenient if instead it was testing if the typed text is a substring of any service/operation. It would be more convenient to just be able to type a fragment of the name, especially when multiple operations share the same prefix.

#30 (case-insensitive matching) is related

A real-life example with a long common prefix:
Alt text

### LINKED ISSUE #79: Sort operations for a service in TraceSearchForm
**Description**
The TraceSearchForm currently shows all Operations as an unsorted list.
It would be really helpful if that list was sorted by name (ascending?)

I believe that this line has to have an additionally sort of the elements being passed?!
[TraceSearchForm.js#L106](https://github.com/uber/jaeger-ui/blob/master/src/components/SearchTracePage/TraceSearchForm.js#L106)

1. Open the UI.
2. Pick a service.
3. Try finding a operation.

screenshot from 2017-09-21 17-24-08

I expect the dropdown to be ordered by name in an ascending order to more easily find the operation I'm looking for.

**Additional environment details (Browser, etc.):**

Chrome, but probably every browser.

tiffon:
- Fix #79 - Sort services and operations operations (case insensitive)
- Fix #31 - Filter options based on contains instead of starts with
- Fix #30 - Filter options based on case insensitive match

### DIFF
--- a/src/reducers/services.js
+++ b/src/reducers/services.js
@@ -21,6 +21,7 @@
 import { handleActions } from 'redux-actions';
 
 import { fetchServices, fetchServiceOperations as fetchOps } from '../actions/jaeger-api';
+import { baseStringComparator } from '../utils/sort';
 
 const initialState = {
   services: [],
@@ -35,6 +36,9 @@ function fetchStarted(state) {
 
 function fetchServicesDone(state, { payload }) {
   const services = payload.data;
+  if (Array.isArray(services)) {
+    services.sort(baseStringComparator);
+  }
   return { ...state, services, error: null, loading: false };
 }
 
@@ -49,6 +53,9 @@ function fetchOpsStarted(state, { meta: { serviceName } }) {
 
 function fetchOpsDone(state, { meta, payload }) {
   const { data: operations } = payload;
+  if (Array.isArray(operations)) {
+    operations.sort(baseStringComparator);
+  }
   const operationsForService = { ...state.operationsForService, [meta.serviceName]: operations };
   return { ...state, operationsForService };
 }
--- a/src/components/SearchTracePage/SearchDropdownInput.js
+++ b/src/components/SearchTracePage/SearchDropdownInput.js
@@ -48,9 +48,9 @@ export default class SearchDropdownInput extends Component {
   }
   onSearch(_, searchText) {
     const { items, maxResults } = this.props;
-    const rxStr = regexpEscape(searchText);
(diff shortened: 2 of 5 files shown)


yurishkuro:
while at it, I sometimes have an issue where I bring up empty Jaeger UI (e.g. from a fresh all-in-one Docker image), so the Services dropdown is supposed to be empty, but the screen renders with a value from the previous runs on the same URL. I know we store the previously selected service name in local storage, but when we retrieve it we should probably check if the value is actually in the list of currently available services. Feel free to move to a separate issue if needed.

src/reducers/services.js
```diff
@@ -35,6 +36,9 @@ function fetchStarted(state) {
 
 function fetchServicesDone(state, { payload }) {
   const services = payload.data;
+  if (Array.isArray(services)) {
+    services.sort(baseStringComparator);
```
saminzadeh: In my opinion, I think we should use lodash as much as we can on util stuff. 

It's well tested, easier to read and immutable.
```js
const sortedServices = _.sortBy(services, s => s)
// or if it was an obj { "name": "svc1" }
const sortedServices = _.sortBy(services, 'name')

```

tiffon: @saminzadeh I would generally go with `arr.sort()` instead of `_.sortBy(arr, value => value)`. You prefer the lodash util in this case?

The sort util function is using [`string#localCompare(...)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare) with the option [`sensitivity`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare#Using_options). It's ignores accents, case, etc.

I don't think lodash gets us anywhere, here, because it takes a transform function instead of a comparator. 

Can you elaborate?

saminzadeh: Other than keeping things immutable (redux approach), I think its just about keeping it consistent. Both work, just my 2 cents.

tiffon:
@yurishkuro Created ticket #88 to track the issue you mentioned.

PR #103, opened 2017-10-23
Add a changelog, dates to 2017-08-23

tiffon:
Non-functional change.

Adds a `CHANGELOG.md` document that details merged PRs.

Starts from late August, 2017.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,302 @@
+# Changes merged into master
+
+
+
+## [#93](https://github.com/jaegertracing/jaeger-ui/pull/93) Keyboard shortcuts and minimap UX (2017-10-20)
+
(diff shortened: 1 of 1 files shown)


tiffon:
@yurishkuro Any idea why DCO is not happy with this PR?

yurishkuro:
This info is very useful, but it's not a typical format of the CHANGELOG, which usually has just a few bullet points between releases. In fact, this doc looks more like release notes, maybe we should have a folder for those and record individual files for each release? 

Are all these changes since 0.8.0 backend release?

yurishkuro:
> Any idea why DCO is not happy with this PR?

Did you sign off all commits?  DCO bots checks each commit separately. You can squash them into one and re-upload.

tiffon:
Had to delete remote branch because one of the commits was not signed.

PR #107, opened 2017-10-27
Embed UI config

### LINKED ISSUE #85: Configure Google Analytics from the /api/config
Currently, the Google Analytics key is a bundle-time env-var, which means to use Google Analytics a custom UI bundle must be generated.

Change to use the `/api/config` data, instead.

Also, consider switching `/api/config` data to be injected into the HTML when the page is served from the query-service.

tiffon:
Fix #85.

Instead of using `/api/config` to retrieve the UI config from the query service, the query service embeds it into index.html. The query service replaces `DEFAULT_CONFIG` with the actual config:

```
JAEGER_CONFIG = DEFAULT_CONFIG;
```

This resolves #85 because it allows the UI to use any GA tracking ID without needed to be rebuilt (e.g. the admin can use the all-in-one docker image and still enable Google Analytics).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 # Changes
 
 
+### [#107](https://github.com/jaegertracing/jaeger-ui/pull/107) Embed UI config
+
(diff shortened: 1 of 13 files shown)


yurishkuro:
Needed for https://github.com/jaegertracing/jaeger/issues/489

PR #108, opened 2017-10-27
Add Apache license header to CSS files

tiffon:
- Change license header in CSS files to use Apache v2
- Update license check to also check CSS files
- Removed uber-license package


tiffon:
> good catch. Have you verified that your change to the check script will find files without copyright?

@yurishkuro I took the license header out of a CSS file and verified it caught it.

PR #118, opened 2017-11-02
Handle FOLLOWS_FROM ref type

### LINKED ISSUE #115: Rendering traces with spans containing a 'FOLLOWS_FROM' reference seems broken
**Description**
I have a test program for a new client library which generates traces and sends these into Jaeger. Since a couple of days, the UI fails to properly display these traces. Initially I noticed this because the search page, upon a search request, fails to load (spinner keeps spinning).

At first I suspected my client lib, but the generated traces seem fine, and the corresponding JSON is served properly to the UI.

After some more testing I believe the issue is due to a `FOLLOWS_FROM` span in my trace, triggering the issue. Hence I bisected `jaeger` (between `v0.6.0` and current `master`), which pointed at https://github.com/jaegertracing/jaeger/commit/162470d

Hence I bisected the `jaeger-ui` submodule (using `jaeger` `master`), which pointed at https://github.com/jaegertracing/jaeger-ui/commit/69d84b901

Indeed, when using `jaeger-ui` `69d84b901^` things work fine.

I'm not familiar with the `jaeger-ui` codebase, but https://github.com/jaegertracing/jaeger-ui/commit/69d84b901#diff-8a263673c6caa913587233948bfb8af2R79 (and the `Error` being thrown for a non-`CHILD_OF` reference) raises suspicion. This `Error` isn't propagated in  the Chrome devtools console, but when setting a breakpoint, this code path *is* reached.

Note: all testing is done using `standalone-linux`, memory back-end.

1. Create a trace with a `FOLLOWS_FROM` reference
2. Load the Jaeger UI
3. Click `Find Traces`

I'll try to create a minimal trace exposing this, and attach the corresponding JSON to the ticket.

UI 'hangs'. Also, browsing to the `/trace/<traceId>` page fails to load the trace properly.

My trace being displayed, as older versions of `jaeger-ui` do.

tiffon:
Fixes #115.

Fixes a regression from #53 where `FOLLOWS_FROM` references throw an error. They're now treated the same as `CHILD_OF`.


2992f2a5b5d037a8aabffd08ef384237.json
```diff
@@ -0,0 +1,52 @@
+{
```
NicolasT: Accidental addition of this file?

tiffon: @NicolasT Great catch, thanks!

PR #120, opened 2017-11-06
Add help page

tiffon:
Per feedback on this PR, help for the keyboard shortcuts is added via a modal triggered from a `⌘` button.

screen shot 2017-11-16 at 3 25 07 pm

screen shot 2017-11-16 at 3 25 10 pm

## Edit — Info below this header is outdated

Add help documentation to the UI. 

Currently, it has a few links and covers the keyboard shortcuts for the trace timeline.

One of the issues I ran into with this change was loading markdown files. It's not well supported in create-react-app. The recommended approach is to fetch the markdown at runtime and parse it in the browser (which is absurd):

>There's no longer a way to directly import Markdown files. In case anyone finds their way here through Google, you could always use fetch() to accomplish this asynchronously:
>```
>// return the URL, e.g. /static/media/changelog.1fb9caf0.md
>import myMarkdownFile from '../../../changelog.md';
>
>fetch(myMarkdownFile)
>  .then(response => response.text())
>  .then(text => {
>    // logs a string of Markdown content
>    console.log(text);
>  });
>```

However, outside the scope of condoned approaches is [inline webpack loader configuration](https://webpack.js.org/concepts/loaders/#inline):

https://github.com/facebookincubator/create-react-app/issues/1944#issuecomment-316081906

>[...]
>```
>/* eslint import/no-webpack-loader-syntax: off */
>import QuickStart from '!raw-loader!./Sections/QuickStart.md';
>```

Inline webpack loader configuration is considered ["not permitted"](https://github.com/facebookincubator/create-react-app/issues/1944#issuecomment-316088150) in create-react-app.

But, it gets the job done and can be changed to some other approach, in the future, if that becomes necessary.

IMO, even with these trade-offs, it is still preferable to use markdown over regular HTML for docs.

Definitely open to feedback on this approach.


yurishkuro:
how is this help page going to be invoked?  I don't think we want to start keeping real documentation in the UI project, that's what readthedocs is for. For the key bindings I was thinking some help button in the Trace View that shows a popup - markdown might be an overkill for that.

src/components/Help/help.md
```diff
@@ -0,0 +1,43 @@
+# Jaeger UI Help
+
+
+## Work In Progress
+
+This document is a work in progress. To request additional documentation of the Jaeger UI, please create a [documentation ticket](https://github.com/jaegertracing/jaeger-ui/issues/new?labels=documentation).
+
+
+## Keyboard Shortcuts
```
saminzadeh: I think this should go in a modal for the Trace page, since its specific to that. A little tooltip can open it up

tiffon:
I prefer the help button and the page, but a modal sounds fine. 

@saminzadeh, @yurishkuro - any suggestions on what invokes the modal?

To answer your question, @yurishkuro, I added a "Help" button to the top menu (which I'll now take out).

saminzadeh:
I was thinking a little Question mark icon on the view trace page. Something similar to codepen:

image

tiffon:
Great. I was thinking something similar, but with the consensus approach being so far removed from my initial approach, seemed prudent to invite input, preemptively.

PR #122, opened 2017-11-14
Make dependencies tab configurable

pavolloffay:
cc @objectiser

### DIFF
--- a/src/components/App/TopNav.js
+++ b/src/components/App/TopNav.js
@@ -64,7 +64,7 @@ const NAV_LINKS = [
   },
 ];
 
-if (getConfig().depsMenuEnabled === true) {
+if (getConfig().dependenciesMenuEnabled) {
   NAV_LINKS.push({
     key: 'dependencies',
     to: prefixUrl('/dependencies'),
--- a/src/components/App/TopNav.js
+++ b/src/components/App/TopNav.js
@@ -20,6 +20,7 @@ import { Dropdown, Menu } from 'semantic-ui-react';
 
 import TraceIDSearchInput from './TraceIDSearchInput';
 import type { ConfigMenuItem, ConfigMenuGroup } from '../../types/config';
+import getConfig from '../../utils/config/get-config';
 import prefixUrl from '../../utils/prefix-url';
 
 import './TopNav.css';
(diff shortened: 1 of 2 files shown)


src/components/App/TopNav.js
```diff
@@ -79,11 +83,11 @@ export default function TopNav(props: TopNavProps) {
       <div className="ui input">
         <TraceIDSearchInput />
       </div>
-      {NAV_LINKS.map(({ key, to, text }) =>
+      {NAV_LINKS.map(({ key, to, text }) => (
```
pavolloffay: `husky > npm run -s precommit (node v6.11.2)`

does this change

yurishkuro:
@pavolloffay could you please explain the objective / use case? It would also go to documentation of the new property.

pavolloffay:
The motivation behind this is that dependency screen in production deployment requires an external process. Some deployments don't use it. In other words it's optional.

src/components/App/TopNav.js
tiffon: @pavolloffay I don't follow... can you elaborate?

src/components/App/TopNav.js
```diff
@@ -57,17 +58,20 @@ function CustomNavDropdown({ label, items }: ConfigMenuGroup) {
...
-    text: 'Dependencies',
-  },
-  {
     key: 'search',
     to: prefixUrl('/search'),
     text: 'Search',
   },
 ];
 
+if (getConfig().depsMenuEnabled === true) {
```
tiffon: This can just be
```js
if (getConfig().depsMenuEnabled) {
```

Or is there a reason to fail if truthy but not strictly `true` ?

src/constants/default-config.js
```diff
@@ -46,4 +46,5 @@ export default deepFreeze({
       ],
     },
   ],
+  depsMenuEnabled: true,
```
tiffon: Can this be renamed to `dependenciesMenuEnabled` ?

src/components/App/TopNav.js
pavolloffay: that this changes it done automatically via git hooks not by me

src/components/App/TopNav.js
pavolloffay: no, updated

pavolloffay:
PR updated

PR #131, opened 2017-12-06
Increase test coverage

tiffon:
Should be > 90% now.


.travis.yml
```diff
@@ -9,5 +9,5 @@ script:
   - npm run coverage
   - npm run build
 after_success:
-  - npm install -g coveralls
-  - cat coverage/lcov.info | coveralls
+  - npm install -g codecov
```
saminzadeh: Does Jest not have this built in? Or do we need this for Travis?

tiffon: This is for travis.

PR #133, opened 2017-12-14
Better HTTP error messages

### LINKED ISSUE #127: Show better error messages for failed HTTP requests
## Description

Currently, failed HTTP requests trigger error messages that are ok in the case of `404`s but are useless, otherwise.

## Steps to reproduce the issue:

1. Try to view a trace that does not exist
2. Execute a search with the tag parameter as "a"
3. Run the UI (via checking out this repo) without a query-service running on the proxy port

## Describe the results you received:

Step 1 yields a `404` HTTP response and the message "Resource not found in the Jaeger Query Service."

Step 2 yields a `400` HTTP response and the message "Bad JSON returned from the Jaeger Query Service."

Step 3 yields a `500` HTTP response and the message "Bad JSON returned from the Jaeger Query Service."

## Describe the results you expected:

The `404` message is fine but should probably be handled in a more generic fashion.

The other responses take two forms: with JSON data describing the error and without JSON data.

Seems like all error messaging should work as follows:

- If there is JSON data, show the response status code and the message from the JSON data
- Otherwise, show the status code and the response text
- If there is no response text, show the status code and the message `An unknown error was encountered`

The only exception being the scenario described in #51, which will be addressed by that ticket.

## Additional information you deem important (e.g. issue happens only occasionally):

Related issues:

- jaegertracing/jaeger#418
- jaegertracing/jaeger#412
- jaegertracing/jaeger#401

tiffon:
Fixes #127.

Fixes jaegertracing/jaeger#418 and jaegertracing/jaeger#401. 

Improved HTTP error messages. No longer always showing the same often incorrect error message.

"Response body" is the first 1021 characters (with "..." appended when truncated).

screen shot 2017-12-14 at 11 48 36 am
screen shot 2017-12-14 at 11 48 43 am


tiffon:
@saminzadeh Done. I reverted and then isolated the changes.

PR #136, opened 2017-12-16
Fix endless trace HTTP requests

### LINKED ISSUE #128: When trace id is invalid, Jaeger UI send this request forever
Related to jaegertracing/jaeger#556. 

The UI loops endlessly if the HTTP request to get a trace succeeds but does not have the expected trace. This can happen in the scenario described in the ticket and also when the URL has uppercase letters in the trace ID because the HTTP response always uses lowercase.

- ~~Limit the number of times the same request is issued~~
- Do not reissue the same request
- If the URL contains uppercase update the URL to be lowercase, then issue the request
    - A permanent redirect would be preferred, but that's in the hands of the query-service (cc @yurishkuro , @black-adder)

tiffon:
Fix #128 - problem of endless HTTP requests when there are caps in the trace ID specified in the URL or when the trace data returned has `""` as the `traceID`.

I'm not able to reproduce the second scenario, but I made an effort to address it—an error will be shown instead of issuing endless HTTP requests.

### DIFF
--- a/src/model/transform-trace-data.js
+++ b/src/model/transform-trace-data.js
@@ -26,10 +26,11 @@ type SpanWithProcess = SpanData & { process: Process };
  * generally requires.
  */
 export default function transfromTraceData(data: TraceData & { spans: SpanWithProcess[] }): ?Trace {
-  const traceID = data.traceID.toLowerCase();
+  let { traceID } = data;
   if (!traceID) {
     return null;
   }
+  traceID = traceID.toLowerCase();
 
   let traceEndTime = 0;
   let traceStartTime = Number.MAX_SAFE_INTEGER;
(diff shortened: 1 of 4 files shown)


src/model/transform-trace-data.js
```diff
@@ -25,7 +25,12 @@ type SpanWithProcess = SpanData & { process: Process };
  * NOTE: Mutates `data` - Transform the HTTP response data into the form the app
  * generally requires.
  */
-export default function transfromTraceData(data: TraceData & { spans: SpanWithProcess[] }): Trace {
+export default function transfromTraceData(data: TraceData & { spans: SpanWithProcess[] }): ?Trace {
+  const traceID = data.traceID.toLowerCase();
+  if (!traceID) {
```
yurishkuro: this looks like a check for traceID==null, shouldn't it be before L29?

tiffon: Checking for empty string, but to be cautious will adjust.

PR #140, opened 2017-12-18
Encode service names in API calls

### LINKED ISSUE #138: Cannot find operations if there is '/' char in serviceName
Re-booking https://github.com/jaegertracing/jaeger/issues/586

The UI must url-encode service name when calling the API to retrieve operation names:  `/api/services/{service}/operations`. It does _some_ encoding, but not correctly, e.g.
  * one of our internal services is emitting a bad service name `${artifactId}`, and the API request looked like this `/api/services/$%7BartifactId%7D/operations`, which is ok
  * however, when the service name contains a slash like `hotrod/customer`, the slash is not encoded correctly: `http://localhost:16686/api/services/hotrod/customer/operations`

Incidentally, the API supports an alternative endpoint `/api/operations?service={service}`, we should decide which one we want to keep.

tiffon:
Fix #138.

### DIFF
--- a/src/api/jaeger.js
+++ b/src/api/jaeger.js
@@ -55,7 +55,7 @@ const JaegerAPI = {
     return getJSON(`${this.apiRoot}services`);
   },
   fetchServiceOperations(serviceName) {
-    return getJSON(`${this.apiRoot}services/${serviceName}/operations`);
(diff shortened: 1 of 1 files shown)


tiffon:
@black-adder I don't think so. The other endpoint names are static.

PR #142, opened 2017-12-19
Enlarge "Logs" header in span detail

### LINKED ISSUE #86: Make the font for Logs section in the span view larger
image

Currently Tags and Process sections are a lot more prominent that Logs, which is shown in smaller font and inside a box. Can we make it look similar to the headers of the other sections?

tiffon:
Fixes #86.

screen shot 2017-12-19 at 3 21 37 am

### DIFF
--- a/src/components/TracePage/TraceTimelineViewer/SpanDetail/AccordianLogs.css
+++ b/src/components/TracePage/TraceTimelineViewer/SpanDetail/AccordianLogs.css
@@ -14,6 +14,10 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+.AccordianLogs--header {
+  font-size: 1rem;
+}
+
 .AccordianLogs--footer {
   color: #888;
   display: block;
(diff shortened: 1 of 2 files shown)


yurishkuro:
It still looks stylistically different from tags/process, is it possible to use the same font and `[+]/[-]`

saminzadeh:
> It still looks stylistically different from tags/process, is it possible to use the same font and [+]/[-]

Maybe it should just use the same component?

tiffon:
>Maybe it should just use the same component?

The component serves a different purpose. It groups the messages together, vs a collection of key/value pairs.

I think the visual distinction is a benefit as it implies something is different, which is the case.

OTH, adopting the +/- but keeping the darkened background (which establishes the logs group) looks fine to me.

screen shot 2017-12-19 at 2 07 01 pm

screen shot 2017-12-19 at 2 07 19 pm

@yurishkuro Is this what you had in mind? @saminzadeh Any thoughts on this?

As an aside, I think a deeper look at the aesthetics would be useful. I'd like explore that as part of investigating [ant.design](https://ant.design/) (or [rc-components](http://react-component.github.io/badgeboard/)).

tiffon:
@yurishkuro Done. The logs group heading is now using the +/- icons instead of the double-angle icon.

yurishkuro:
:shipit:

PR #147, opened 2017-12-21
Use logfmt for search tag input format

### LINKED ISSUE #11: Document allowed operators on tag search
image

The tag search prefilled text shows that a `|` is allowed, but there is no documentation about operators.

### LINKED ISSUE #145: Support logfmt for tags text input in the search form
Ref jaegertracing/jaeger#550.

Add a `tags` query string parameter that is URL encoded JSON.

Tags text input to accept [logfmt](https://brandur.org/logfmt). Front-end parses this, converts all values to strings, serializes as JSON, and passes over the wire as the value for `tags` form value (via GET, as-is currently used). 

Add a tooltip next to the tags text field that mentions logfmt, etc.

tiffon:
Fixes #145.

Fixes #11.

Changes the input format for tag matching in the search for to logfmt. Now populates a `tags` query string parameter when submitting the search HTTP request. The `tags` GET param is JSON with all values converted to strings (per [this comment](https://github.com/jaegertracing/jaeger/issues/550#issuecomment-352850811)).

This change is backward compatible with URLs that have the former format. When the old format (`key:value|key2:value2`) is in the URL, it is still parsed but added to the form in the logfmt format.

Added a tooltip that links to information on [logfmt](https://brandur.org/logfmt).

screen shot 2017-12-21 at 3 41 28 am

### DIFF
--- a/src/components/SearchTracePage/TraceSearchForm.js
+++ b/src/components/SearchTracePage/TraceSearchForm.js
@@ -13,12 +13,15 @@
 // limitations under the License.
 
 import React from 'react';
+import logfmtParser from 'logfmt/lib/logfmt_parser';
+import { stringify as logfmtStringify } from 'logfmt/lib/stringify';
 import moment from 'moment';
 import PropTypes from 'prop-types';
 import queryString from 'query-string';
 import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
 import { Field, reduxForm, formValueSelector } from 'redux-form';
+import { Popup } from 'semantic-ui-react';
 import store from 'store';
 
 import SearchDropdownInput from './SearchDropdownInput';
@@ -34,11 +37,20 @@ export function getUnixTimeStampInMSFromForm({ startDate, startDateTime, endDate
   };
 }
 
-export function tagsToQuery(tags) {
+export function convTagsLogfmt(tags) {
   if (!tags) {
     return null;
   }
-  return tags.split('|');
+  const data = logfmtParser.parse(tags);
+  Object.keys(data).forEach(key => {
+    const value = data[key];
+    // make sure all values are strings
+    // https://github.com/jaegertracing/jaeger/issues/550#issuecomment-352850811
+    if (typeof value !== 'string') {
+      data[key] = String(value);
+    }
+  });
+  return JSON.stringify(data);
 }
 
 export function traceIDsToQuery(traceIDs) {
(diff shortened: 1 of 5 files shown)


src/components/SearchTracePage/TraceSearchForm.js
```diff
@@ -128,6 +140,7 @@ export function TraceSearchFormImpl(props) {
   const selectedServicePayload = services.find(s => s.name === selectedService);
   const operationsForService = (selectedServicePayload && selectedServicePayload.operations) || [];
   const noSelectedService = selectedService === '-' || !selectedService;
+  // const tagInfoIcon = <i className="info circle icon" />;
```
yurishkuro: dead code?

src/components/SearchTracePage/TraceSearchForm.js
```diff
@@ -154,14 +167,24 @@ export function TraceSearchFormImpl(props) {
...
+            Tags{' '}
+            <Popup
+              on="click"
+              trigger={<i className="info circle icon grey" />}
+              content={
+                <span>
+                  Values should be in the{' '}
+                  <a href="https://brandur.org/logfmt" rel="noopener noreferrer" target="_blank">
+                    logfmt
+                  </a>{' '}
```
yurishkuro: perhaps we could give an example, save people a click

> e.g. `db.statement="select * from User"`. Use space for conjunctions: `error=true http.status_code:503`

src/components/SearchTracePage/TraceSearchForm.js
tiffon: Yes, thanks, will fix.

src/components/SearchTracePage/TraceSearchForm.js
tiffon: The placeholder has two examples. But, I can add a few to the info tooltip, including one that uses quotes.

tiffon:
There is one more thing I want to change before this is merged: how tag names without values in the old format are handled.

src/components/SearchTracePage/TraceSearchForm.js
tiffon: Nevermind, I see your note now. I'll add that to the tooltip :+1:

tiffon:
@yurishkuro Updated tooltip. Slight variation from your suggestion. Look ok?

screen shot 2017-12-21 at 10 24 17 pm

yurishkuro:
great!

PR #151, opened 2017-12-23
[Blocked] Query-Service config and runtime configuration of URL path-prefix

### LINKED ISSUE #42: No support for Jaeger behind a reverse proxy
Jaeger does not run behind a reverse proxy due to how the css and js is being loaded.

```html
<link href="/static/css/main.123456789.css" rel="stylesheet">
```

This can be corrected with something like 
```html
<link href="./static/css/main.123456789.css" rel="stylesheet">
```

Similarly with the js file.

Thanks

tiffon:
Fix #42.

Adds config data from the query service that is separate from the UI config, dubbed `querySvcConfig`.

Presently, this contains a single property, `pathPrefix`, which causes the UI to handle being served from a URL prefix, at runtime.

When set, the `pathPrefix` should be formatted as `/xyz`.

## Query-service changes

Changes in the query-service this PR relies on:

### Config injection

- Related to the [UI config regex](https://github.com/jaegertracing/jaeger/blob/master/cmd/query/app/static_handler.go#L32):
    - Rename `DEFAULT_CONFIG` to `__INJECTED_UI_CONFIG__`
    - Change the regex to now handle the `return __INJECTED_UI_CONFIG__;` replacement
- Treatment similar to the UI config for the query-service config
    - The variable name is: `__INJECTED_QS_CONFIG__`
- If the URL path-prefix is set, it needs to be passed into `index.html` in a manner similar to the UI config, e.g. as JSON:
    - `{"pathPrefix":"/xyz"}`

### Static assets that need URL rewrites in their content

In order for the URL path prefix to work, the all URLs need to have the prefix pre-pended. This essentially amounts to the reference to the favicon and references to static assets (including references in static assets).


yurishkuro:
> Incidentally, the URLs are not constants but instead are derived from expressions. Therefore, while possible, rewriting the static asset URLs in the JS files looks precarious and error-prone.

wouldn't they still have a fixed prefix `/static/js` that is easy to search & replace? Doesn't seem that much different from how the query service would have to pre-process the other files.

public/index.html
```diff
@@ -14,13 +14,24 @@
...
-      // service should be an empty object or it can leave `DEFAULT_CONFIG`
+      // src/utils/config#getUiConfig. The default provided by the query
+      // service should be an empty object or it can leave `DEFAULT_UI_CONFIG`
       // unchanged.
       function getJaegerUiConfig() {
-        const DEFAULT_CONFIG = null;
-        const JAEGER_CONFIG = DEFAULT_CONFIG;
-        return JAEGER_CONFIG;
+        const DEFAULT_UI_CONFIG = null;
+        const JAEGER_UI_CONFIG = DEFAULT_UI_CONFIG;
```
yurishkuro: why do we need DEFAULT_ constants? When I originally proposed them I thought they would have some concrete values, and it was easier to search&replace the string `JAEGER_UI_CONFIG = DEFAULT_UI_CONFIG` than to replace the JS expression on the right hand side. But these are always null anyway, we might as well have `JAEGER_UI_CONFIG = null`, it's just as easy to search&replace.

src/components/SearchTracePage/index.js
```diff
@@ -116,7 +116,7 @@ export default class SearchTracePage extends Component {
             !hasTraceResults && (
               <div className="ui middle aligned center aligned grid" style={{ marginTop: 100 }}>
                 <div className="column">
-                  <img className="js-test-logo" alt="presentation" src={JaegerLogo} width="400" />
+                  <img className="js-test-logo" alt="presentation" src={prefixUrl(JaegerLogo)} width="400" />
```
yurishkuro: was it using relative url before?

tiffon: No, it uses an absolute URL. 

The image is imported in JS:

https://github.com/jaegertracing/jaeger-ui/blob/7779eb56f485f8d60c2ff158d7a7bfaa9f6252a5/src/components/SearchTracePage/index.js#L25

That does two things:

- Causes the asset to be stored in `/static/media/` with a filename derived from the file's hash
- Stores the eventual absolute path of the media asset in the variable `JaegerLogo`

The absolute path is then used in the JSX as you excerpted. 

The `prefixUrl(...)` adds the path-prefix, if it is configured.

public/index.html
tiffon: We could do that.

I think tweaking the name a bit to imply it might be over-written has some value, though:

```javascript
// use `null` as a fallback value if the query-service does not embed a value
const __INJECTED_UI_CONFIG__ = null;
const JAEGER_UI_CONFIG = __INJECTED_UI_CONFIG__;
```

tiffon:
@yurishkuro 

>wouldn't they still have a fixed prefix /static/js that is easy to search & replace? Doesn't seem that much different from how the query service would have to pre-process the other files.

It's not that much different, and we can go down that road. It would spare any impact to the initial page-load time and remove the need for react-app-rewired. The expression(s) in `main.NNNN.js` that would need to be updated is:

```javascript
u.src=t.p+"static/js/"+({}[e]||e)+"."+{0:"03cf6b4e"}[e]+".chunk.js";
```

The filesize of `main...js` is around 2 meg.

public/index.html
yurishkuro: +1 for `__INJECTED_UI_CONFIG__`. I don't think you need the second const, you can simply `return __INJECTED_UI_CONFIG__;`.

tiffon: Great. Done.

config-overrides.js
```diff
@@ -0,0 +1,30 @@
...
+// eslint-disable-next-line import/no-extraneous-dependencies
+const webpack = require('webpack');
+
+module.exports = function override(config, env) {
+  if (env === 'production') {
+    config.plugins.push(
+      // prevent code-splitting to allow the URL path-prefix for the site to be
+      // configurable at runtime
+      // https://github.com/jaegertracing/jaeger-ui/issues/42
+      new webpack.optimize.LimitChunkCountPlugin({
```
yurishkuro: so which way are we going, global search&replace of the path in the query service or using this plugin to not split the js files?

tiffon: @yurishkuro Still up for debate.

The main considerations:

- **Primarily** - The exact form the loading of JS chunks takes (which is the site of the search/replace) has an unknown level of risk of changing
- The search/replace in the JS files is not as "clean" as the others
- The main.js filesize is ~2M, on disk (not sure to what extent that is much of a consideration)

Note: every *.js file, not just main.NNNN.js, must be checked for the `static/js` path and prefixed if found

So, my main concern is around maintainability. 

The above inconveniences allow us to avert a likely hit on initial page-load time, possibly a ~5% increase.

The ideal scenario, IMO, is to do the global replace and if maintaining it is a PITA then revert and go with the plugin approach to not split the files. Not sure how feasible that approach is on the query-service side.

Would that work for you?

yurishkuro: Yes, I would suggest trying global replace first.

tiffon: Sounds good. This PR needs some rework, then.

tiffon:
Closing, will re-open when I've reverted the chunkless changes.

tiffon:
Reopening this PR. Main change necessary, from the previous state, was to revert the removal of the chunk generation. 

It's worth noting, this PR requires several changes to the query-service, as noted in the PR description.

saminzadeh:
Looking back at this, create react app [documents a section ](https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#injecting-data-from-the-server-into-the-page) on this. 

Looks like we are doing this already +1

ys:
Can you please stop mentioning me all the time here! I think it is about someone else

tiffon:
@ys Apologies! I'll block you so it doesn't happen again!

tiffon:
@yurishkuro Doesn't look like the work needed in the query-service for this PR is going to be on the docket anytime soon... Think it makes sense to close this PR?

tiffon:
This PR is waiting on work from the query-service, which is currently not scheduled. So, I'm closing and will mark the ticket as blocked.

yurishkuro:
> This PR is waiting on work from the query-service, which is currently not scheduled. 

Do we have a ticket in the main repo w/ description of what needs to be done? We can tag it "help wanted".

tiffon:
@yurishkuro I'll create an issue in the main repo once I create the PR for the `<base>` approach.

PR #153, opened 2017-12-23
Add View Option for raw/unadjusted trace

### LINKED ISSUE #152: Add View Option for raw/unadjusted trace
Per https://github.com/jaegertracing/jaeger/issues/617, the backend will support `/api/traces/{traceID}/?raw=true` URL. We should add it to the View Options dropdown as "Trace JSON (unadjusted)".

Also, `s/View Trace JSON/Trace JSON/`

image

yurishkuro:
Fixes #152

### DIFF
--- a/src/components/TracePage/TracePageHeader.js
+++ b/src/components/TracePage/TracePageHeader.js
@@ -20,6 +20,7 @@ import { Dropdown, Menu } from 'semantic-ui-react';
 import KeyboardShortcutsHelp from './KeyboardShortcutsHelp';
 import { FALLBACK_TRACE_NAME } from '../../constants';
 import { formatDatetime, formatDuration } from '../../utils/date';
+import prefixUrl from '../../utils/prefix-url';
 
 type TracePageHeaderProps = {
   traceID: string,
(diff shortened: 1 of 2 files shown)


src/components/TracePage/TracePageHeader.js
```diff
@@ -104,7 +104,12 @@ export default function TracePageHeader(props: TracePageHeaderProps) {
               <Dropdown.Menu>
                 <Dropdown.Item>
                   <a rel="noopener noreferrer" target="_blank" href={`/api/traces/${traceID}`}>
```
yurishkuro: @tiffon I don't see this file in #151, it probably needs adjustment with the prefix as well

black-adder:
add this to the change log so it can go out with 1.1?

src/components/TracePage/TracePageHeader.js
tiffon: @yurishkuro Yes, this needs the `prefixUrl(...)`.

Also, is the trailing `/` in the raw version URL intentional?

```
/api/traces/${traceID}/?raw=true
```

Seems like the only difference should be the query string is appended?

```
/api/traces/${traceID}?raw=true
```

tiffon:
@yurishkuro The `prefixUrl(...)` needs to be in master in the event of `package.json#homepage` is changed; it's not just relevant to the query-config PR (#151). 

I went ahead and added it (and fixed a failing unit test).

PR #161, opened 2017-12-24
Add timezone tooltip to custom lookback form-field

### LINKED ISSUE #154: Explain time zone of the lookback parameter
Moved from https://github.com/jaegertracing/jaeger/issues/372

When "Custom Time Range" is specified, it would be good to show in which time zone the entered values are interpreted, either via `(?)` icon or explicitly "Start Time (TZ=New York [UTC-0500])".

image

tiffon:
Fix #154.

EST Timezone:

screen shot 2017-12-24 at 3 51 46 pm

PST Timezone

screen shot 2017-12-24 at 12 52 32 pm

### DIFF
--- a/src/components/SearchTracePage/TraceSearchForm.js
+++ b/src/components/SearchTracePage/TraceSearchForm.js
@@ -222,7 +222,7 @@ export function TraceSearchFormImpl(props) {
                 on="click"
                 wide="very"
                 trigger={<i className="SearchForm--hintTrigger info circle icon grey" />}
-                content={<h5>Times are expressed in {tz} timezone and converted to UTC</h5>}
+                content={<h5>Times are expressed in {tz} timezone</h5>}
               />
             </label>
             <div>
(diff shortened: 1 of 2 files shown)


src/components/SearchTracePage/TraceSearchForm.js
```diff
@@ -229,7 +238,15 @@ export function TraceSearchFormImpl(props) {
...
         {selectedLookback === 'custom' && (
           <div className="search-form--end-time field js-test-end-input">
-            <label htmlFor="service">End time</label>
+            <label htmlFor="service">
+              End time{' '}
+              <Popup
+                on="click"
+                wide="very"
+                trigger={<i className="SearchForm--hintTrigger info circle icon grey" />}
+                content={<h5>Times are expressed in {tz} timezone and converted to UTC</h5>}
```
yurishkuro: I would remove the part "and converted to UTC" from the tool tip, I don't think it's relevant to the user experience.

PR #168, opened 2018-01-13
fixes  #167: 2 digit lookback (12h, 24h) parsing

### LINKED ISSUE #167: 12 and 24 hour search lookbacks not converted to start timestamp correctly
**Description**

A JS bug causes start timestamp for the 12 hour and 24 hour lookback to be the same as the end timestamp.

1. Set the search lookback to either 12 hour or 24 hour
2. Execute a search

The start and end timestamps GET query parameters are the same.

The start timestamp should be the adjusted based on the lookback.

golonzovsky:
fixes #167

current code assumes 1 digit format of [lookback options](https://github.com/jaegertracing/jaeger-ui/blob/master/src/components/SearchTracePage/TraceSearchForm.js#L209-L212)

### DIFF
--- a/src/components/SearchTracePage/TraceSearchForm.js
+++ b/src/components/SearchTracePage/TraceSearchForm.js
@@ -106,7 +106,7 @@ export function submitForm(fields, searchTraces) {
   let start;
   let end;
   if (lookback !== 'custom') {
-    const unit = lookback[1];
+    const unit = lookback.substr(-1);
     const now = new Date();
     start =
       moment(now)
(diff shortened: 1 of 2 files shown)


src/components/SearchTracePage/TraceSearchForm.js
```diff
@@ -106,7 +106,7 @@ export function submitForm(fields, searchTraces) {
   let start;
   let end;
   if (lookback !== 'custom') {
-    const unit = lookback[1];
+    const unit = lookback.substr(-1);
```
yurishkuro: seems like we need a unit test

golonzovsky: added double-digit parse [test](https://github.com/jaegertracing/jaeger-ui/pull/168/commits/5db4d6a251aec6880831a1ec5b57ac23f73f5a0c#diff-7478a808394cdea4c7237c63c8cc13b4R165)

tiffon:
LGTM

PR #169, opened 2018-01-17
Use Ant Design instead of Semantic UI 

### LINKED ISSUE #69: Missing endpoints in jaeger ui dropdown
Currently, dropdowns in the Search page are limited to 250 results. This is implemented for performance reasons. Find a dropdown that has sufficient performance to meet our needs.

Ref: T1123163

### LINKED ISSUE #164: Use Ant Design instead of Semantic UI
Switch to [Ant Design UI Components](https://ant.design/docs/react/introduce).

### LINKED ISSUE #165: Search results are shown without a date
**Description**

For any given search result, only the time is shown. 

screen shot 2018-01-03 at 7 13 33 pm

1. Search for traces that are a day or two old

The result is listed with only the time of the trace, the date is omitted

Result items older than a certain amount should show the date of the timestamp.

tiffon:
Fix #165. Fix #164. Fix #69.

**Note: The blue highlights in the search form in some of the screenshots are outdated. Please refer to the screenshots marked with "†"**

Screenshot of the search page showing the custom menu dropdown.
ant-00-search

**†** Service and operation dropdowns now use react `react-virtualized-select` and are able to show the full list of options.
01--ant-01-service-virtual-select

Hint for tags input.
ant-02-tags-hint

Hint for start date and time input.
ant-03-start-hint

**†** No results.
01--ant-04-no-results

Result list, with a hovered item.
ant-05-results

Trace detail view.
ant-06-trace

Trace name is clickable to collapse the mini-map and overview information.
ant-07-trace-name-hover

Collapsed view.
ant-08-trace-slim

Keyboard shortcut info.
ant-09-kbd-info

Parent spans can be collapsed.
ant-10-parent-hover

A collapsed parent span.
ant-11-parent-collapsed

A span detail row.
ant-12-detail-row

Various details expanded.
ant-13-details-expanded

The force directed graph of the services.
ant-14-force-graph

The DAG.
ant-15-dag


ekelleyv:
Caveat that my comments are entirely based on the screenshots, posted, but just wanted to share a couple of thoughts. Let me know if this is worth discussing somewhere else rather than this PR.

It may be helpful to test some of these designs with applications that have many internal spans. The hot rod application doesn't have too many levels of indentation and uses several different "application tracers" resulting in spans that are more visually legible. Here is a short segment from a trace in our infra::

image

Since everything is in the same service and there are several levels of indentation, it is very hard to see what elements are children of the others.

A couple suggestions:

1. Add guidelines down each indentation level like code editors do for whitespace
image

2. When a parent is hovered, every element that is a child of that element should have some sort of hover state. At the very least, the first and last child should have some marker. I would suggest just extending   that hover state down to cover its children (mockup based on posted screenshots):
image


Aside from that, the screenshots posted here look great.

tiffon:
@ekelleyv Thanks for the feedback and suggestion! We've discussed guidelines, internally, but have yet to create a ticket for it. I'll add one now.


@saminzadeh The color change sounds good to me. :+1:

src/components/DependencyGraph/index.js
```diff
@@ -67,66 +70,44 @@ export default class DependencyGraphPage extends Component {
...
-      return (
-        <div className="m1">
-          <div className="ui active centered inline loader" />
-        </div>
-      );
+      return <LoadingIndicator className="u-space-top-vast" centered />;
     }
     if (error) {
-      return <NotFound error={error} />;
+      return <ErrorMessage className="u-space" error={error} />;
```
saminzadeh: Curious what `u-space` does? Is it a utility?

I notice that we removed [Basscss](http://basscss.com/). IMO, I think we should use basscss as a CSS utility library since its small (~2kb), low on side effects and documented.

src/components/common/utils.css
```diff
@@ -0,0 +1,162 @@
+/*
```
saminzadeh: See common above, should we use Basscss? Believe it will help with developer productivity since its OSS.

tiffon:
@saminzadeh Thanks for the suggestion re [`basscss`](https://github.com/basscss/basscss). I prefer the util css class names to be prepended with `u-*`, so I forked `basscss` ((`u-basscss`)[https://github.com/tiffon/u-basscss]). I also made it so sub-sets of the package can be imported, instead of only the full package.

PR #179, opened 2018-02-04
Resolve perf issues on the search page

### LINKED ISSUE #178: Performance regression - Search page
A performance regression on the search page was introduced with the Ant Design changes.

While shifting the search page to Ant, I broke it out into a series of smaller components. Things are now rendering too often. The issue is aggravated when there are many search results.

tiffon:
Fix #178.

Main issue was two too many stateless functional components, now `React.PureComponent`s.

Also:
- No longer mutating the array of search results when sorting it
- Added a couple of missing util css imports

### DIFF
--- a/src/components/SearchTracePage/SearchResults/ResultItem.js
+++ b/src/components/SearchTracePage/SearchResults/ResultItem.js
@@ -14,7 +14,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import React from 'react';
+import * as React from 'react';
 import { Col, Divider, Row, Tag } from 'antd';
 import { sortBy } from 'lodash';
 import moment from 'moment';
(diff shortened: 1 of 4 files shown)


black-adder:
just curious, are there automated ways to detect regressions like this in the UI?

tiffon:
@black-adder [jaegertracing/jaeger-client-node/benchmarks](https://github.com/jaegertracing/jaeger-client-node/tree/master/benchmarks) is one approach. Another is testing through a browser with something like [Puppeteer for Chrome](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#class-tracing) — similar to e2e testing but you'd want to capture a [perf trace](https://developers.google.com/web/tools/chrome-devtools/rendering-tools/#tools_1) via the dev tools. The former is granular while the latter has sweeping coverage but variance is likely to be a problem.

What's the approach taken with the various Jaeger benchmarks to have some sort of historical perspective or expectation relative to the current benchmark results? E.g. the current perf is X, the new perf is Y.

saminzadeh:
> Main issue was two too many stateless functional components, now React.PureComponents.

To clarify, the issue is not necessarily stateless components but rather unnecessary render comparisons right? The PureComponent will add the `shouldComponentUpdate` lifecycle method to do a shallow comparison of props and state to avoid unnecessary renders.

Another alternative would be to keep the components stateless and then use a simple utility like  recompose's [pure](https://github.com/acdlite/recompose/blob/master/docs/API.md#pure) or [onlyUpdateForKeys](https://github.com/acdlite/recompose/blob/master/docs/API.md#onlyupdateforkeys) HOCs that would only result in a few lines of code change.

The result would be a smaller code change and easier to remove when we optimize further up the state tree:
```js
export function ResultItem(props) { /*... */ }

export default pure(ResultItem);
```

PR #188, opened 2018-02-12
[Non-functional change] - Run prettier on *.md files

tiffon:
Also made the pre-commit hook a bit more stringent.


tiffon:
@yurishkuro Done.

Issue #113, opened 2017-10-31
Expose version or commit SHA of UI


tiffon:
Purpose is to be able to determine which version of the UI is currently deployed.

If version is used, it needs to not always be [`0.0.1`](https://github.com/jaegertracing/jaeger-ui/blob/4e02e451c7cd3f62154d4b5a6c8a0a0668281454/package.json#L3).

Thanks to @saminzadeh for suggesting this.

yurishkuro:
+1. Is there an npm module that can extract the commit during the build?

tiffon:
It would also be nice for this to show the version of the Jaeger services, not just the UI.

Related jaegertracing/jaeger#145

tiffon:
@yurishkuro Re npm package, we might be able to use [git-repo-version](https://github.com/cibernox/git-repo-version) to generate (package-version, SHA, date) strings and [jest-changed-files](https://github.com/facebook/jest/tree/master/packages/jest-changed-files) to capture which files, if any, have been modified for the build.

**git-repo-version:** `1.5.0+a1b2c3d4e5 2016-10-24T18:26:53.000Z`

**jest-changed-files:**

```javascript
import {getChangedFilesForRoots} from 'jest-changed-files';

getChangedFilesForRoots(['/path/to/test'], {
  lastCommit: true,
  withAncestor: true,
}).then(files => {
  /*
  {
    repos: [],
    changedFiles: []
  }
  */
});
```

yurishkuro:
during the build we can easily get the SHAs for both query service and the ui submodule using git commands directly. The UI build could read them from env vars.

yurishkuro:
Resolved by #606

PR #189, opened 2018-02-12
Track JS errors in GA

### LINKED ISSUE #39: Log js client side errors in our server side logs
We don't track errors in the UI, and rely on users to contact us when things break.

Adding such logging would allow for better discovery of js errors.

tiffon:
Fix #39.

TLDR: Using [raven-js](https://github.com/getsentry/raven-js) to capture errors and context info, then sending it to GA instead of sentry. It's a little out there, but it should be informative.

I did a pretty thorough write-up in [`src/utils/tracking/README.md`](https://github.com/jaegertracing/jaeger-ui/blob/issue-39-track-js-errors/src/utils/tracking/README.md), which lays out the changes in this PR. It might make sense to move some of the README to the Jaeger docs. They need to be updated to reflect `gaTrackingID` ↠ `tracking.gaID` and the addition of `tracking.trackErrors`, anyway.

The changes in the `*.md` files that aren't related to GA tracking can be ignored; they're captured in #188.

The version info generated for GA can also be used to satisfy #113. I'll add a footer in another PR.

I'm currently working on UX event tracking in a different branch.


tiffon:
@simonrobb Can you look this over when you get the chance?

src/utils/tracking/conv-raven-to-ga.js
```diff
@@ -0,0 +1,357 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/* eslint-disable camelcase */
+
+import prefixUrl from '../prefix-url';
+
+const UNKONWN_SYM = { sym: '??', word: '??' };
```
simonrobb: May as well fix this typo (`UNKONWN`) now.

src/index.js
```diff
@@ -12,27 +12,30 @@
...
 import 'u-basscss/css/padding.css';
 import 'u-basscss/css/position.css';
 import 'u-basscss/css/typography.css';
 
-initTracking();
-
 const UI_ROOT_ID = 'jaeger-ui-root';
 
 /* istanbul ignore if */
 if (document && process.env.NODE_ENV !== 'test') {
```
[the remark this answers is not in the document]
tiffon: :+1:

src/utils/tracking/conv-raven-to-ga.js
tiffon: Done.

src/utils/tracking/README.md
```diff
@@ -0,0 +1,202 @@
+# Google Analytics (GA) Tracking In Jaeger UI
```
saminzadeh: Great write up Joe

Although I'm all for documentation my only comment would be, is it really worth to maintain all this stuff for Jaeger UI? Although this is nice to have, my concern is this information will go stale as its not really part of the core of what Jaeger provides.

src/utils/tracking/conv-raven-to-ga.js
```diff
@@ -0,0 +1,357 @@
+// @flow
```
saminzadeh: I feel like this could be a library in itself outside of Jaeger

src/utils/tracking/README.md
tiffon: That's a valid point. But, it really shouldn't need to change much. Seems like low-maintenance docs, to me. 

What do you suggest?

scripts/get-tracking-version.js
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env node
+
+const spawnSync = require('child_process').spawnSync;
```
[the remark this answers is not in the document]
tiffon: Done.

PR #192, opened 2018-03-01
Change fallback trace name to be more informative

### LINKED ISSUE #190: cannot-find-trace-name
@tiffon I suggest we change this string to `trace-without-root-span`

tiffon:
Fix #190.

### DIFF
--- a/src/constants/index.js
+++ b/src/constants/index.js
@@ -13,4 +13,4 @@
 // limitations under the License.
 
 export const FALLBACK_DAG_MAX_NUM_SERVICES = 100;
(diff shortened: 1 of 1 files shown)


black-adder:
Can we also add one of them scroll over info thingies (I'm a UX professional as you can tell) that says what root-span is missing means? ie It happens because the root service is poorly instrumented or the root span hasn't been persisted yet because it's the last span in a trace to be finished and persisted.

tiffon:
@black-adder Added a ticket from your comment, #193.

PR #194, opened 2018-03-06
Support archiving traces, enabled via UI config

### LINKED ISSUE #7: Add support to "archive traces"
Users often ask for this feature. The trace retention period is usually short, and often people would put a link to the trace into some issue tracker to improve service performance, but by the time someone starts working on the ticket the trace has already expired.

The functionality is already supported in the backend, but missing UI changes. The idea is to add a button "Archive Trace" on the trace detail view (next to View Options dropdown), which opens (expands?) a form with:
 * a dropdown to select the TTL (options: 1 month, 3 month, 1yr, indefinite)
 * submit button that does a POST to `/api/v1/archive/{trace_id}?ttldays=30` (or whichever is the best REST API, maybe move trace_id/ttldays to POST body)
 * once the trace is archived, the form displays the permanent URL: `http://.../archive/{trace_id}`

The `/archive/` route needs to make the webapp query `/api/v1/archive/{trace_id}` endpoint instead of the usual `/traces` endpoint. And it must also change View JSON to the same URL.

@saminzadeh wdyt? how hard is it to do?

tiffon:
Partially addresses #7.

## TODO

- ~Add a tooltip on the archive button that states it's for archiving a trace~
- [x] See about getting an icon that is more iconic of archiving

## Enabling Archiving

The archive functionality is hidden behind the UI config flag `archiveEnabled`, which is `false`, by default as archiving only works with Cassandra.

## Normal UX

Clicking the archive button raises a notification with a loading spinner.

When the HTTP call completes, the spinner notification is closed (if the user did not close it) and a second notification is raised showing the outcome. 

If the outcome is success, then a success message is shown.

If the outcome is an error, then error details are shown.

## Small Edge Case

The user should acknowledge the result notification (success or error). This is done by closing the notification. If the user doesn't close the notification but exits the trace view, the notification will be redisplayed if the user re-enters that trace detail (for that particular trace) during the session. The same applies if the user exits the trace view while the archive is in progress.

### The archive button

screen shot 2018-03-08 at 11 04 31 am

--------

### Button hover

screen shot 2018-03-08 at 11 04 42 am

--------

### In progress notification

02-in-progress

--------

### Success notification

03-success

--------

### Error notification
Note: This error is shown just as an example of an error to show. This particular error should be very easy to avoid as the system administrator needs to simple keep archiving disabled if they're not using Cassandra.

Also, the error message styling (global) was tweaked a bit.

04-error-message

--------

### Button is hidden due to UI config

05-button-disabled

### DIFF
--- a/src/api/jaeger.js
+++ b/src/api/jaeger.js
@@ -33,10 +33,11 @@ export function getMessageFromError(errData, status) {
   }
 }
 
-function getJSON(url, query) {
-  return fetch(`${url}${query ? `?${queryString.stringify(query)}` : ''}`, {
-    credentials: 'include',
-  }).then(response => {
+function getJSON(url, options = {}) {
+  const { query = null, ...init } = options;
+  init.credentials = 'same-origin';
+  const queryStr = query ? `?${queryString.stringify(query)}` : '';
+  return fetch(`${url}${queryStr}`, init).then(response => {
     if (response.status < 400) {
       return response.json();
     }
@@ -78,8 +79,11 @@ const JaegerAPI = {
   fetchTrace(id) {
     return getJSON(`${this.apiRoot}traces/${id}`);
   },
+  archiveTrace(id) {
+    return getJSON(`${this.apiRoot}archive/${id}`, { method: 'POST' });
+  },
   searchTraces(query) {
-    return getJSON(`${this.apiRoot}traces`, query);
+    return getJSON(`${this.apiRoot}traces`, { query });
   },
   fetchServices() {
     return getJSON(`${this.apiRoot}services`);
@@ -88,7 +92,7 @@ const JaegerAPI = {
     return getJSON(`${this.apiRoot}services/${encodeURIComponent(serviceName)}/operations`);
   },
   fetchDependencies(endTs = new Date().getTime(), lookback = DEFAULT_DEPENDENCY_LOOKBACK) {
-    return getJSON(`${this.apiRoot}dependencies`, { endTs, lookback });
+    return getJSON(`${this.apiRoot}dependencies`, { query: { endTs, lookback } });
   },
 };
 
--- a/src/components/TracePage/ArchiveNotifier/duck.js
+++ b/src/components/TracePage/ArchiveNotifier/duck.js
@@ -1,0 +1,83 @@
+// @flow
+
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { createActions, handleActions } from 'redux-actions';
+
+import { archiveTrace } from '../../../actions/jaeger-api';
+import type { ApiError } from '../../../types/api-error';
+import type { TracesArchive } from '../../../types/archive';
+import generateActionTypes from '../../../utils/generate-action-types';
+
+type ArchiveAction = {
+  meta: {
+    id: string,
+  },
+  payload?: ApiError | string,
+};
+
+const initialState: TracesArchive = {};
+
+const actionTypes = generateActionTypes('@jaeger-ui/archive-trace', ['ACKNOWLEDGE']);
+
+const fullActions = createActions({
+  [actionTypes.ACKNOWLEDGE]: traceID => traceID,
+});
+
+export const actions = { ...fullActions.jaegerUi.archiveTrace, archiveTrace };
+
+// reducers
+
+function acknowledge(state: TracesArchive, { payload }: ArchiveAction) {
+  const traceID = typeof payload === 'string' ? payload : null;
+  if (!traceID) {
+    // make flow happy
+    throw new Error('Invalid state, missing traceID for archive acknowledge');
+  }
+  const traceArchive = state[traceID];
+  if (traceArchive && traceArchive.isLoading) {
+    // acknowledgement during loading is invalid (should not happen)
+    return state;
+  }
+  const next = { ...traceArchive, isAcknowledged: true };
+  return { ...state, [traceID]: next };
+}
+
+function archiveStarted(state: TracesArchive, { meta }: ArchiveAction) {
+  return { ...state, [meta.id]: { isLoading: true } };
+}
+
+function archiveDone(state: TracesArchive, { meta }: ArchiveAction) {
+  return { ...state, [meta.id]: { isArchived: true, isAcknowledged: false } };
+}
+
+function archiveErred(state: TracesArchive, { meta, payload }: ArchiveAction) {
+  if (!payload) {
+    // make flow happy
+    throw new Error('Invalid state, missing API error details');
+  }
+  const traceArchive = { error: payload, isArchived: false, isError: true, isAcknowledged: false };
+  return { ...state, [meta.id]: traceArchive };
+}
+
+export default handleActions(
+  {
+    [actionTypes.ACKNOWLEDGE]: acknowledge,
+    [`${archiveTrace}_PENDING`]: archiveStarted,
+    [`${archiveTrace}_FULFILLED`]: archiveDone,
+    [`${archiveTrace}_REJECTED`]: archiveErred,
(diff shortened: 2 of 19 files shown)


src/components/TracePage/ArchiveNotifier/duck.js
```diff
@@ -0,0 +1,83 @@
+// @flow
+
+// Copyright (c) 2017 Uber Technologies, Inc.
```
yurishkuro: 2018

src/components/TracePage/ArchiveNotifier/duck.js
```diff
@@ -0,0 +1,83 @@
+// @flow
```
yurishkuro: why is it called `duck.js`?

yurishkuro:
nit: why clock icon for archive? More commonly it's a file cabinet or a row of books on a shelf.

src/components/TracePage/ArchiveNotifier/duck.js
tiffon: Seems like the name of this approach (combining the reducer and actions into a single file) is a play on word(s) involving redux: https://github.com/erikras/ducks-modular-redux 

Derivative stuff uses the ducks word: https://github.com/erikras/ducks-modular-redux#implementation

tiffon:
@yurishkuro The clock was just the first thing I thought of as emblematic of the action, maybe something to do with causing the retention to not be bound by the limits of time.

Did you have something like this icon in mind?

screen shot 2018-03-06 at 10 26 51 am

tiffon:
@simonrobb Can you look this over when you get the chance?

yurishkuro:
more like these

image
image

yurishkuro:
timer is fine too though

tiffon:
I can see the timer being better suited to something like history (e.g. search history, etc).

src/types/config.js
```diff
@@ -25,7 +25,11 @@ export type ConfigMenuGroup = {
 };
 
 export type Config = {
+  archiveEnabled: ?boolean,
   dependencies?: { dagMaxServicesLen?: number, menuEnabled?: boolean },
-  gaTrackingID?: ?string,
+  tracking?: {
+    gaID: ?string,
+    trackErrors: ?boolean,
+  },
```
tiffon: This changes to add `tracking` to config is something I missed when adding the GA error tracking.

src/components/common/ErrorMessage.js
```diff
@@ -14,64 +14,101 @@
...
-    <tr>
+    <tr className="ErrorMessage--detailItem">
       <td className="ErrorMessage--attr">{name}</td>
       <td className="ErrorMessage--value">{value}</td>
     </tr>
   );
 }
 
+function Message(props: SubPartProps) {
+  const { className, error, wrap } = props;
```
saminzadeh: This seems like we can simplify. What is the point of wrap?

```js
function Message({ error, className }) {
  const errorMsg = typeof error === 'string' ? error : error.message;
  return (
    <h3 className={`ErrorMessage--msg ${className || ''}`}>{errorMsg}</h3>
  );
}
```

src/components/common/ErrorMessage.css
```diff
@@ -25,22 +23,42 @@ limitations under the License.
 .ErrorMessage--msg {
   color: #c00;
   margin: 0;
-  padding: 1rem;
+  padding-bottom: 1rem;
 }
 
 .ErrorMessage--details {
+  background: #f5f5f5;
+  border: 1px solid #e6e6e6;
```
saminzadeh: Is there a way to reuse things like border? Seems like its not specific to the message details, but just a general UI theme we use. Suggestions: Use a utility class (basscss), css vars, or styled components in JS

src/api/jaeger.js
```diff
@@ -78,8 +79,11 @@ const JaegerAPI = {
   fetchTrace(id) {
     return getJSON(`${this.apiRoot}traces/${id}`);
   },
+  archiveTrace(id) {
```
saminzadeh: How do you find out if a trace is archived already? Is it in the trace response from `GET /trace/:traceId` ?

tiffon: Basically, you can't. International trace of mystery. Not ideal, and in #7 we discuss that.

tiffon:
No, that's the thing, right now that functionality is not implemented. This
is a stop-gap solution, as I understand it.

src/api/jaeger.js
saminzadeh: @yurishkuro IMO, I think this should be a requirement of the API before this is implemented.
 
When the user refreshes the page, you won't know the trace has been archived. It will confuse the user when they click the button and its toggled to "Archived", then refresh the page and its no longer toggled.

yurishkuro: I think perfect is the enemy of good. I'd rather have this feature in prod helping users than block on implementing the additional APIs.

The act of archiving already has pretty active notifications, so unlikely to confuse the user who's doing the archiving.

src/components/common/ErrorMessage.js
tiffon: `wrap` is so the component can be split between the message and the details but still have consistent layout behavior.

In the notification, the message is used as the notification title and the details are the notification body. `wrap` keeps the layout the same whether the parts are split out and used individually like that or not.

src/components/common/ErrorMessage.css
tiffon: CSS vars aren't supported in IE11, which is ~3% of browsers. I try to avoid stuff that's not supported by IE. 

Another approach might be to embrace a CSS preprocessor now that we have react-app-rewired going. What do you think of PostCSS with [these plugins](https://jf.io/2015/10/26/a-selection-of-postcss-plugins#plugin-selection)?

src/api/jaeger.js
tiffon: Let's leave it undocumented until it's fully implemented.

tiffon:
@yurishkuro I switched the icon to a box, per the screenshots in the PR description.

src/api/jaeger.js
saminzadeh: Maybe even explain to the user when they click the button or hover over with a tooltip.

tiffon: I was thinking that too. Not sure what to say, though. 

> This trace has been archived. But, if you refresh this page, the archive button will still be enabled. For now (unfortunately) we don't disable the button when you have already archived the trace. This feature is a WIP, you can follow the implementation at #7. Please contact @yurishkuro with questions or to verify the trace has been archived.

Ok, not that last part, but the other stuff...? Kind of verbose, but not sure how to put it succinctly.

src/components/common/ErrorMessage.css
saminzadeh: I'm good with that. We need something to manage this CSS better.

src/api/jaeger.js
saminzadeh: SGTM.

tiffon: @yurishkuro What do you think of adding this to the success notification?

>This trace has been archived. But, if you refresh this page, the archive button will still be enabled. For now (unfortunately) we don't disable the button when you have already archived the trace. This feature is a WIP, you can follow the implementation at #7. [Please contact @yurishkuro with questions or to verify the trace has been archived.]

Exact phrasing on that last part is TBD...

yurishkuro: > What do you think of adding this to the success notification?

I think we should not do that. It should only say "This trace has been archived." I think you guys are arguing for a feature that (a) nobody asked for yet, and (b) costs way more to implement than its benefits. If someone needs this feature in the future, let them open a ticket, get the votes, and get someone to prioritize that above all the other work & improvements we could be spending time on.

simonrobb: @yurishkuro I'm not sure which feature you're referring to, when you say nobody asked for it - but I'd say that displaying the archived status is an integral part of the archive feature. Shipping this without displaying the status or at least providing messaging feels half-baked to me.

yurishkuro: we ARE providing a message on successful save.

src/components/common/ErrorMessage.css
tiffon: Created a ticket for this #200

tiffon:
@saminzadeh @simonrobb Any further remarks about the implementation?

src/api/jaeger.js
simonrobb: Confirmation of the save is one thing, it's another provide feedback on an ongoing basis. "Is this trace going to disappear tomorrow or not?" 

I disagree the costs outweigh the benefits - UX is important. Certainly less so when this is an undocumented feature (this was some context I was missing) but still important. I personally think the UX needs more work before the feature is released to the public but the technical implementation looks good to me so the PR has a thumbs up from me.

Issue #51, opened 2017-08-10
Multi-trace traceID search does not show 404s


black-adder:
Cf. https://github.com/uber/jaeger/issues/310

When a user retrieves multiple traces, a single trace that doesn't exist (due to TTLing) will return an error for the entire batch. This has been fixed on the query side to return as many traces as possible + errors for the traces that couldn't be fetched. 

Current version of the UI does not parse the errors portion of the response, it only seems to react to the status code.

tiffon:
Something along these lines should work:

<img width="871" alt="screen shot 2018-03-11 at 11 01 20 pm" src="https://user-images.githubusercontent.com/2304337/37263816-459872ca-2580-11e8-9fc5-d6c905b8d4cf.png">

The markup for the error list item:

```html
<div class="ResultItem" style="
    border: 1px solid #fbb;
"><div class="ResultItem--title clearfix" style="
    background: hsla(0, 100%, 96%, 1);
    border: none;
"><h3 class="ub-m0 ub-relative" style="
    color: #c00;
                                     "><span style="
    /* color: rgba(0,0,0,0.75); */
">Trace not found:</span><span style="
    color: rgba(0,0,0,0.75);
"> dad19e3d79b8af</span></h3></div></div>
```

(Needs to be cleaned up, etc.)

shubhamkumar9199:
Hey @yurishkuro,

I’d like to take this up.

I went through the codebase and it looks like the backend and reducer already handle per-trace errors properly — the only thing missing is how we show this in the UI.

Right now, in SearchTracePage, errored traces are being filtered out here:

results.map(id => traceMap[id].data).filter(t => t !== undefined)

So when a trace isn’t found, it just disappears from the results without any indication to the user.

What I’m thinking:

Instead of dropping those entries, we pass them through to SearchResults
Then render a simple error card (with a red border) saying something like:
"Trace not found: {traceID}", alongside the successful results — similar to the design Joe shared earlier in the thread

I’ll keep the change small and focused.

WasThatRudy:
Tracing this: backend returns `{ data, errors: [{msg, traceID}] }` for multi-trace search, but `searchDone` in `reducers/trace.ts` only iterates `payload.data`, so failed IDs never make it to the page. `fetchMultipleTracesDone` in the same file already handles the errors array correctly, so this is essentially a parity gap.

[...]

PR #195, opened 2018-03-12
Handle Error stored in redux trace.traces

### LINKED ISSUE #166: JS error on search page after viewing 404 trace
**Description**

Navigating from a trace page that is 404 (trace not found) to the search page results in a JS error.

1. Visit a trace page that does not exist, e.g. `/trace/000`
2. Click the "Search" button in the top nav

The page does not change and the following JS error is logged:

```
Uncaught TypeError: Cannot read property 'length' of undefined
    at o (search.js:43)
    at Array.map (<anonymous>)
    at a (search.js:88)
    at index.js:209
    at get-last-xform-cacher.js:38
    at Function.a [as mapToProps] (index.js:232)
    at r (wrapMapToProps.js:43)
    at Function.r.mapToProps (wrapMapToProps.js:52)
    at r (wrapMapToProps.js:43)
    at o (selectorFactory.js:26)
```

The search page is presented without error.

tiffon:
Fix #166.

HTTP responses >= 400 (.e.g. 404s) cause an `Error` to be added to the redux store in `trace.traces[traceID]`. This PR prevents the search page from treating them as a normal trace.

### DIFF
--- a/src/model/search.js
+++ b/src/model/search.js
@@ -30,10 +30,7 @@ const isErrorTag = ({ key, value }) => key === 'error' && (value === true || val
  * @param trace Trace data in the format sent over the wire.
  * @return {TraceSummary} Summary of the trace data for use in the search results.
  */
-export function getTraceSummary(trace: Trace | Error): ?TraceSummary {
-  if (trace instanceof Error) {
-    return null;
-  }
+export function getTraceSummary(trace: Trace): TraceSummary {
   const { processes, spans, traceID } = trace;
 
   let traceName = '';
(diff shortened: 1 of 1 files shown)


tiffon:
@simonrobb can you look this over when you get the chance?

src/model/search.js
```diff
@@ -83,7 +86,7 @@ export function getTraceSummary(trace: Trace): TraceSummary {
  * @return {TraceSummaries} The `{ traces, maxDuration }` value.
  */
 export function getTraceSummaries(_traces: Trace[]): TraceSummaries {
-  const traces = _traces.map(getTraceSummary);
+  const traces = _traces.map(getTraceSummary).filter(Boolean);
```
yurishkuro: not sure I am following the typing here: `_traces` is an array of `Trace` objects, how could `getTraceSummary` function be called with an `Error` object?

tiffon: It's an inefficiency or irregularity in the state of static typing in the repo. Some files have flow-types and some do not. 

The type here is really `(Trace | Error)[]`.

yurishkuro: so maybe the fix should be (using Scala-ish notation)
```
_traces.filterNot(_ instanceof Error).map(getTraceSummary);
```
?

or even do the filtering somewhere upstream, instead of propagating the mixed type deeper.

simonrobb: Agreed. At the least, if it's possible for there to be `Error` objects present, the parameter type should be annotated like you described, `getTraceSummaries(_traces: (Trace | Error)[]): TraceSummaries`.

tiffon: @yurishkuro Great point. I will filter then get the summary. I don't want to filter it further upstream because the presence of the error will be useful in resolving #51.

@simonrobb 👍 

Note, the following code was not possible due to Flow flagging it as an error:

```js
const traces = _traces.filter(item => !(item instanceof Error)).map(getTraceSummary);
```

PR #198, opened 2018-03-19
Use <base> and config webpack at runtime to allow path prefix

### LINKED ISSUE #42: No support for Jaeger behind a reverse proxy

tiffon:
Fix #42.

Allows a path prefix to be used through four changes:

- `package.json#homepage` is set to `"."`. When create-react-app generates the build index.html the static assets are linked via relative paths.
- Added a [`<base>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base) element to the index.html template. All relative paths (e.g. the static assets referenced in HTML) are relative to the value of `<base>`, which is set to `"/"` in index.html, but the query-service can replace it (see jaegertracing/jaeger#745). It's worth noting, this only affects relative paths.
- The `publicPath` value in webpack is configured at runtime via the [`__webpack_public_path__`](https://webpack.js.org/guides/public-path/#on-the-fly) global. The value is derived from the `<base>`, and ensures any chunks or other assets (like the logo) are loaded relative to the site prefix.
- The `pathPrefix` used for `fetch()` calls is based on the `<base>` value instead of `package.json#homepage`

As a side note, this PR can be merged before any work is done in the query-service.


yurishkuro:
what's the issue with code coverage? the numbers are always confusing does `88.88% of diff hit` refer to coverage of the new/changed code?

tiffon:
> what's the issue with code coverage? the numbers are always confusing does 88.88% of diff hit refer to coverage of the new/changed code?

@yurishkuro I think that means of the modified lines 88.88% of them were hit by code coverage.

PR #205, opened 2018-04-15
Fix link to commit message guidelines

tiffon:
Re jaegertracing/jaeger#771


yurishkuro:
The proposal was to inline the rules as well.

tiffon:
Doh! Also, I made the change online and forgot the sign-off.

PR #208, opened 2018-05-02
Development of a directed graph component

tiffon:
## High level

I'm breaking this down into chunks by targeting the `develop-directed-graph` branch instead of `master`.

The repo has been reorganized to use [Lerna](https://lernajs.io/) with the main Jaeger UI app moved into `packages/jaeger-ui` and the directed graph is being developed in `packages/plexus`. Lerna is being used to make consumption of the component in Jaeger UI easier during development. The component is being broken out into it's own NPM package because it will be used in other projects, as well.

## "plexus"

>1. A network or interwoven mass, especially (anatomy) of nerves, blood vessels, or lymphatic vessels.
>
>2. (mathematics) The system of equations required for the complete expression of the relations which exist between a set of quantities.

[Wiktionary - plexus](https://en.wiktionary.org/wiki/plexus#Noun)

## Component

The directed graph will be used in:

- Service dependency graphs
- Viewing a trace as a DAG
- Comparing two traces via comparing two DAGs

This PR implements the functionality to generate the layout (positions and edges) of a graph, but does not have the presentation functionality (React component), yet. 

Trying to move briskly due to time constraints, so comments are _very_ light and unit tests are TODO. I wanted to get some eyes on this while moving onto the presentation layer.

PR #209, opened 2018-05-02
Use Lerna

tiffon:
The PR reorganizes the repo to use [Lerna](https://lernajs.io/) with the main Jaeger UI app moved into `packages/jaeger-ui`. This is in preparation for developing UI components as independent NPM packages, particularly the directed graph component (#208), so they can be used in other projects as well as Jaeger UI. Lerna eases the development of Jaeger UI as it depends on the separated components.

This also upgrades react to 16.3.2.


tiffon:
@simonrobb Please look this over when you get the chance.

tiffon:
cc @whistlinwilly

saminzadeh:
Will we need to update the README to develop and run the application?

tiffon:
@saminzadeh Great catch, thanks. I've updated the readme.

PR #212, opened 2018-05-11
Directed graph layout and presentation

tiffon:
## High level

The overall effort for creating the directed graphs is being broken into multiple PRs and merged into an intermediary branch, `develop-directed-graph`, instead of `master`.

## "plexus"

>1. A network or interwoven mass, especially (anatomy) of nerves, blood vessels, or lymphatic vessels.
>
>2. (mathematics) The system of equations required for the complete expression of the relations which exist between a set of quantities.

[Wiktionary - plexus](https://en.wiktionary.org/wiki/plexus#Noun)

## Component

The directed graph will be used in:

- Service dependency graphs
- Viewing a trace as a DAG
- Comparing two traces via comparing two DAGs

This PR implements the functionality to generate the layout (positions and edges) of a graph, ~~but does not have the presentation functionality (React component), yet.~~ Also implemented in this PR is a React component to present the directed graph. It implements the MVP functionality that is required for kaiju.

Comments are very light and tests are TODO, but I wanted to get eyes on this as soon as possible.


packages/plexus/src/LayoutManager/Coordinator.js
```diff
@@ -0,0 +1,346 @@
...
+  idleWorkers: LayoutWorker[];
+  busyWorkers: LayoutWorker[];
+  callback: LayoutUpdate => void;
+
+  constructor(callback: LayoutUpdate => void) {
+    this.callback = callback;
+    this.currentLayout = null;
+    this.nextWorkerId = 1;
+    this.isDisposed = false;
+    this.idleWorkers = [this._initWorker()];
```
simonrobb: Why initialize with a worker? It looks like you just create one as they're required. Seems like introducing unnecessary state.

packages/plexus/src/LayoutManager/Coordinator.js
```diff
@@ -0,0 +1,346 @@
...
+  nextWorkerId: number;
+  isDisposed: boolean;
+  idleWorkers: LayoutWorker[];
+  busyWorkers: LayoutWorker[];
+  callback: LayoutUpdate => void;
+
+  constructor(callback: LayoutUpdate => void) {
+    this.callback = callback;
+    this.currentLayout = null;
+    this.nextWorkerId = 1;
```
simonrobb: A zero-based index might be more intuitive.

packages/plexus/src/LayoutManager/Coordinator.js
```diff
@@ -0,0 +1,346 @@
...
+    this.isDisposed = false;
+    this.idleWorkers = [this._initWorker()];
+    this.busyWorkers = [];
+  }
+
+  getLayout(id: number, edges: Edge[], vertices: SizeVertex[]) {
+    if (this.isDisposed) {
```
simonrobb: Any reason for having a disposed state? You're almost returning it to its initial state in the dispose method; it doesn't seem dangerous to allow a `getLayout` call on a disposed instance.

packages/plexus/src/LayoutManager/Coordinator.js
tiffon: It takes some time to initialize the worker. A [quick test](https://fiddle.jshell.net/1ukjdxpb/show/light/) showed around 7 - 10ms. (Most of it is in generating the blob.) Kind of point-less to not do it lazily, though.

Will change.

As a side note: the rationale was driven by the fact that the workers are pooled.

packages/plexus/src/LayoutManager/Coordinator.js
tiffon: Yeah, I guess the main thing is to interrupt any pending worker tasks, for instance if the view changes and the graph component unmounts.

Seems like having a `reset()` method to clear any current tasks would suffice. 

On the other hand, I don't know much about garbage collecting workers, so I'll leave a flag for killing idle workers, too.

tiffon:
@simonrobb Thanks for the feedback, I made updates per your remarks (finally). Let me know if you see anything else that needs some work.

packages/plexus/src/LayoutManager/Coordinator.js
simonrobb: This looks better to me 👍

packages/plexus/src/LayoutManager/Coordinator.js
simonrobb: Were you intending to fix this?

Issue #216, opened 2018-06-06
Add links to make values in tags or log properties clickable


divdavem:
Hello,

I am opening this issue to discuss the following feature that we would like to add: in the list of tags or log properties, we would like to replace some values by clickable links, so that it is possible to click on those values to open another tool in another browser tab with more information related to the item that was clicked.

This feature would have to be configurable: there should be a list of link templates. Each template would be a URL with tokens that would be replaced by a value, looking first in the current set of key-values (tags/log properties) and if the corresponding token is not found, going up in the tags of the parent spans in the CHILD_OF hierarchy until a value is found for that key.

If we open a pull request with such a feature, would you be interested in integrating it? Or would it be better to have some kind of extension point that anyone could use to customize the way tags and log properties are displayed, which would allow us to also implement our feature in a kind of plugin? The way such plugins could be implemented would then need to be defined, and in that case we should have enough contextual information from the parent spans.

Please tell us what you think.

Thank you in advance for your answer!

black-adder:
I think the plugin route sounds more feasible, @tiffon weren't you working on plugin designs? Would something like this fit into that model?

yurishkuro:
+1, we wanted this feature from the start, never had a chance to implement.

I am not sure it requires a plugin framework vs. one general framework that is config-file based, if it can be made flexible enough.

divdavem:
@black-adder @yurishkuro Thank you for your answer!

If the configuration can be put in a JS file (cf #123), I think it can be made very flexible.
For example, it would be possible to accept a function from the config which would be called to decide whether each tag should be displayed as a link and how (maybe it could return the React markup to display that tag, if React itself is passed as a parameter). Now the interface of such a function would need to be clearly defined. Passing enough context would be needed for us (including the full current trace and a reference to the current span and log item).

tiffon:
@divdavem Thanks for opening this ticket!

I agree with @yurishkuro, the config file approach (#123) is more attractive as the current thoughts around plugins are on more of a macro level, so to speak.

Seems like there are a few scenarios outlined... or just one general scenario?

>replace some values by clickable links, looking first in the current set of key-values (tags/log properties) and if the corresponding token is not found, going up in the tags of the parent spans in the CHILD_OF hierarchy until a value is found for that key.

[...]

Issue #123, opened 2017-11-14
Change UI configuration to a Javascript file


yurishkuro:
@tiffon you proposed changing the UI config file from a JSON file to a Javascript file, to allow comments, etc. Does it require any changes in the index.html?

tiffon:
@yurishkuro No, it does not. It will require additional guidance, though, as not any valid JavaScript file would be valid.

For instance, these two code blocks would be valid JavaScript for the UI config:

```javascript
{
  // this is some setting
  someSetting: 'abc'
}
```

Or

```javascript
(function() {
  const regex = /abc/i;
  function someUtil(str) {
    return regex.test(str);
  }
  return {
    someSetting: someUtil
  };
})();
```

On the other hand, this file would be invalid:

```javascript
const regex = /abc/i;
function someUtil(str) {
  return regex.test(str);
}
const config = {
  someSetting: someUtil
};
```

The guidance for this is simple, it should either be an object or an IIFE (**without** a preceding semicolin).

yurishkuro:
why not give the instructions that the JS file should have the following statement at the end
```
JAEGER_CONFIG = {
    // config here
};
```

tiffon:
That works, too.

It should have `const`:
```javascript
const JAEGER_CONFIG = {
  // config here
};
```

In which case, from the index.html file:

```html

    <script>
      // Jaeger UI config data is embedded by the query-service. This is
      // later merged with defaults into the redux `state.config` via
      // src/utils/config/get-config.js. The default provided by the query
      // service should be an empty object or it can leave `DEFAULT_CONFIG`
      // unchanged.
      function getJaegerUiConfig() {
        const DEFAULT_CONFIG = null;
        const JAEGER_CONFIG = DEFAULT_CONFIG;
        return JAEGER_CONFIG;
      }
    </script>
  </head>
```

The following should be wholly replaced by the contents of the UI config JavaScript file:

```javascript
const JAEGER_CONFIG = DEFAULT_CONFIG;
```

tiffon:
@yurishkuro The above would not require any code change to the UI repo.

tiffon:
@yurishkuro One downside of requiring `const JAEGER_CONFIG = ` is JSON will no longer be valid.

yurishkuro:
why not? If the input file is JSON it will replace the `DEFAULT_CONFIG` on the right hand side of the assignment. If the input is JS then it will replace the whole statement.

th3M1ke:
[...]

PR #223, opened 2018-07-04
[WIP] Add links to make values in tags or log properties clickable

divdavem:
This pull request contains an implementation of the feature described in #216.

It allows to define in the configuration some patterns so that the values of keys from the `Tags`, `Process` and `Logs` sections become clickable links that open the configured URL.

This has to be configured in the new `linkPatterns` property. For example:

```json
{
  "linkPatterns": [{
    "type": "process",
    "key": "jaeger.version",
    "url": "https://github.com/jaegertracing/jaeger-client-go/releases/tag/#{jaeger.version}",
    "text": "Information about Jaeger release #{jaeger.version}"
  }]
}
```

When this is declared in the configuration, every time the `jaeger.version` key appears in the `process` section, the value of that key will be displayed as a clickable link, with an icon on the right to show that the value is clickable, for example:

jaeger link

If several patterns match the same tag, clicking on the value will display a popup with all the links:

jaeger multiple links

Links are opened in a new window or tab.

In the configuration of a link, there are two kinds of properties:
- test properties: they determine whether the link should be displayed. These properties can contain a string or an array of strings specifying the possible values of that property that lead to the display of the link. All test properties must match for the link to be displayed. If a test property is not defined, it is considered to match.
  - `type` (possible values: `process`, `logs` or `tags`)
  - `key`
  - `value`
- template properties: they determine the url and text of the link. They are simple templates where each occurrence of `#{key}` is replaced by the value of the given `key`.
  - `url`: each replacement in the URL is wrapped in a call to `encodeURIComponent`
  - `text`: if there is only one link, the text of the link is displayed as a tooltip, if there are multiple links the text is displayed in the popup menu

  Template properties can depend on the value of the following keys:
  - for type `process`, only the keys in the current process can be used
  - for other types, the search order is the following:
    - for type `logs`: the keys present in the log  
    - the keys of the tags or process in the span on which the log or tag being rendered can be found
    - the keys of the tags or process in the parent span according to the CHILD_OF hierarchy
    - other parents in the CHILD_OF hierarchy...

**Commented extra features**

Note that this PR originally also implemented the following possibilities that rely on the implementation of #123 (which allows the use of any JS expression instead of only JSON in the configuration). Those features have been commented in the code of the PR.
- test properties can contain functions or regular expressions instead of containing only strings or arrays of strings
- template properties can be implemented as an object containing two properties:
  - parameters: an array of strings specifying which keys are needed
  - template: a function that receives as its only argument an object with the keys declared in the `parameters` array and returns the url or the text of the link

For example, once #123 is implemented and the corresponding code is uncommented, the previous example can also be configured in the following equivalent way (using functions):

```js
const JAEGER_CONFIG = {
  linkPatterns: [{
    type: (type) => type === "process",
    key: (key) => key === "jaeger.version",
    url: {
      parameters: ["jaeger.version"],
      template: data => `https://github.com/jaegertracing/jaeger-client-go/releases/tag/${encodeURIComponent(data["jaeger.version"])}`
    },
    text: {
      parameters: ["jaeger.version"],
      template: data => `Information about Jaeger release ${data["jaeger.version"]}`
    }
  }]
}
```

Thank you in advance for your comments.

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/AccordianKeyValues.js
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/AccordianKeyValues.js
@@ -30,7 +30,7 @@ type AccordianKeyValuesProps = {
   highContrast?: boolean,
   isOpen: boolean,
   label: string,
-  linksGetter: ({ key: string, value: any }[], number) => { url: string, text: string }[],
+  linksGetter: ?({ key: string, value: any }[], number) => { url: string, text: string }[],
   onToggle: () => void,
 };
 
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/AccordianKeyValues.js
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/AccordianKeyValues.js
@@ -30,6 +30,7 @@ type AccordianKeyValuesProps = {
   highContrast?: boolean,
   isOpen: boolean,
   label: string,
+  linksGetter: ({ key: string, value: any }[], number) => { url: string, text: string }[],
   onToggle: () => void,
 };
 
@@ -59,7 +60,7 @@ KeyValuesSummary.defaultProps = {
 };
 
 export default function AccordianKeyValues(props: AccordianKeyValuesProps) {
-  const { className, data, highContrast, isOpen, label, onToggle } = props;
+  const { className, data, highContrast, isOpen, label, linksGetter, onToggle } = props;
   const isEmpty = !Array.isArray(data) || !data.length;
   const iconCls = cx('u-align-icon', { 'AccordianKeyValues--emptyIcon': isEmpty });
   return (
@@ -80,7 +81,7 @@ export default function AccordianKeyValues(props: AccordianKeyValuesProps) {
         </strong>
         {!isOpen && <KeyValuesSummary data={data} />}
       </div>
-      {isOpen && <KeyValuesTable data={data} />}
+      {isOpen && <KeyValuesTable data={data} linksGetter={linksGetter} />}
     </div>
   );
 }
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/AccordianKeyValues.js
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/AccordianKeyValues.js
@@ -21,21 +21,22 @@ import IoIosArrowRight from 'react-icons/lib/io/ios-arrow-right';
 
 import * as markers from './AccordianKeyValues.markers';
 import KeyValuesTable from './KeyValuesTable';
+import type { KeyValuePair, Link } from '../../../../types';
 
 import './AccordianKeyValues.css';
 
 type AccordianKeyValuesProps = {
   className?: ?string,
-  data: { key: string, value: any }[],
+  data: KeyValuePair[],
   highContrast?: boolean,
   isOpen: boolean,
   label: string,
-  linksGetter: ?({ key: string, value: any }[], number) => { url: string, text: string }[],
+  linksGetter: ?(KeyValuePair[], number) => Link[],
   onToggle: () => void,
 };
 
 // export for tests
-export function KeyValuesSummary(props: { data?: { key: string, value: any }[] }) {
+export function KeyValuesSummary(props: { data?: KeyValuePair[] }) {
   const { data } = props;
   if (!Array.isArray(data) || !data.length) {
     return null;
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.js
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.js
@@ -14,7 +14,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import React from 'react';
+import * as React from 'react';
 import jsonMarkup from 'json-markup';
 import { Dropdown, Icon, Menu } from 'antd';
 
@@ -30,6 +30,24 @@ function parseIfJson(value) {
   } catch (_) {}
   return value;
 }
+
+const LinkValue = (props: { href: string, title?: string, children: React.Node }) => (
+  <a href={props.href} title={props.title} target="_blank" rel="noopener noreferrer">
+    {props.children} <Icon className="KeyValueTable--linkIcon" type="export" />
+  </a>
+);
+
+const linkValueList = (links: { url: string, text: string }[]) => (
+  <Menu>
+    {links.map(({ text, url }, index) => (
+      // `index` is necessary in the key because url can repeat
+      // eslint-disable-next-line react/no-array-index-key
+      <Menu.Item key={`${url}-${index}`}>
+        <LinkValue href={url}>{text}</LinkValue>
+      </Menu.Item>
+    ))}
+  </Menu>
+);
 
 type KeyValuesTableProps = {
   data: { key: string, value: any }[],
@@ -53,31 +71,15 @@ export default function KeyValuesTable(props: KeyValuesTableProps) {
             if (links && links.length === 1) {
               valueMarkup = (
                 <div>
-                  <a href={links[0].url} title={links[0].text} target="_blank" rel="noopener noreferrer">
-                    {jsonTable} <Icon className="KeyValueTable--linkIcon" type="export" />
-                  </a>
+                  <LinkValue href={links[0].url} title={links[0].text}>
+                    {jsonTable}
+                  </LinkValue>
                 </div>
               );
             } else if (links && links.length > 1) {
-              const menuItems = (
-                <Menu>
-                  {links.map((link, index) => {
-                    const { text, url } = link;
-                    return (
-                      // `index` is necessary in the key because url can repeat
-                      // eslint-disable-next-line react/no-array-index-key
-                      <Menu.Item key={`${url}-${index}`}>
-                        <a href={url} target="_blank" rel="noopener noreferrer">
-                          {text}
-                        </a>
-                      </Menu.Item>
-                    );
-                  })}
-                </Menu>
-              );
               valueMarkup = (
                 <div>
-                  <Dropdown overlay={menuItems} placement="bottomRight" trigger={['click']}>
+                  <Dropdown overlay={linkValueList(links)} placement="bottomRight" trigger={['click']}>
                     <a>
                       {jsonTable} <Icon className="KeyValueTable--linkIcon" type="profile" />
                     </a>
@@ -101,3 +103,5 @@ export default function KeyValuesTable(props: KeyValuesTableProps) {
     </div>
   );
 }
+
+KeyValuesTable.LinkValue = LinkValue;
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.js
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.js
@@ -31,17 +31,6 @@ function parseIfJson(value) {
   return value;
 }
 
-function markupAsDiv(markup) {
-  // eslint-disable-next-line react/no-danger
-  return <div dangerouslySetInnerHTML={{ __html: markup }} />;
-}
-
-function markupAsSpan(markup) {
-  const spanMarkup = markup.replace(/^<div /i, '<span ').replace(/<\/div>$/i, '</span>');
-  // eslint-disable-next-line react/no-danger
-  return <span dangerouslySetInnerHTML={{ __html: spanMarkup }} />;
-}
-
 type KeyValuesTableProps = {
   data: { key: string, value: any }[],
   linksGetter: ?({ key: string, value: any }[], number) => { url: string, text: string }[],
@@ -54,14 +43,18 @@ export default function KeyValuesTable(props: KeyValuesTableProps) {
       <table className="u-width-100">
         <tbody className="KeyValueTable--body">
           {data.map((row, i) => {
-            const markup = jsonMarkup(parseIfJson(row.value));
+            const markup = {
+              __html: jsonMarkup(parseIfJson(row.value)),
+            };
+            // eslint-disable-next-line react/no-danger
+            const jsonTable = <div className="ub-inline-block" dangerouslySetInnerHTML={markup} />;
             const links = linksGetter ? linksGetter(data, i) : null;
             let valueMarkup;
             if (links && links.length === 1) {
               valueMarkup = (
                 <div>
                   <a href={links[0].url} title={links[0].text} target="_blank" rel="noopener noreferrer">
-                    {markupAsSpan(markup)} <Icon className="KeyValueTable--linkIcon" type="export" />
+                    {jsonTable} <Icon className="KeyValueTable--linkIcon" type="export" />
                   </a>
                 </div>
               );
@@ -86,13 +79,13 @@ export default function KeyValuesTable(props: KeyValuesTableProps) {
                 <div>
                   <Dropdown overlay={menuItems} placement="bottomRight" trigger={['click']}>
                     <a>
-                      {markupAsSpan(markup)} <Icon className="KeyValueTable--linkIcon" type="profile" />
+                      {jsonTable} <Icon className="KeyValueTable--linkIcon" type="profile" />
                     </a>
                   </Dropdown>
                 </div>
               );
             } else {
-              valueMarkup = markupAsDiv(markup);
+              valueMarkup = jsonTable;
             }
             return (
               // `i` is necessary in the key because row.key can repeat
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.js
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.js
@@ -31,6 +31,22 @@ function parseIfJson(value) {
   return value;
 }
 
+function markupAsDiv(markup) {
+  // eslint-disable-next-line react/no-danger
+  return <div dangerouslySetInnerHTML={{ __html: markup }} />;
+}
+
+function markupAsSpan(markup) {
+  // eslint-disable-next-line react/no-danger
+  return (
+    <span
+      dangerouslySetInnerHTML={{
+        __html: markup.replace(/^<div /i, '<span ').replace(/<\/div>$/i, '</span>'),
+      }}
+    />
+  );
+}
+
 type KeyValuesTableProps = {
   data: { key: string, value: any }[],
   linksGetter: ?({ key: string, value: any }[], number) => { url: string, text: string }[],
@@ -43,24 +59,16 @@ export default function KeyValuesTable(props: KeyValuesTableProps) {
       <table className="u-width-100">
         <tbody className="KeyValueTable--body">
           {data.map((row, i) => {
-            const jsonTable = (
-              // eslint-disable-next-line react/no-danger
-              <div dangerouslySetInnerHTML={{ __html: jsonMarkup(parseIfJson(row.value)) }} />
-            );
-            let valueMarkup = jsonTable;
+            const markup = jsonMarkup(parseIfJson(row.value));
             const links = linksGetter ? linksGetter(data, i) : null;
+            let valueMarkup;
             if (links && links.length === 1) {
               valueMarkup = (
-                <a
-                  className="KeyValueTable--link"
-                  href={links[0].url}
-                  title={links[0].text}
-                  target="_blank"
-                  rel="noopener noreferrer"
-                >
-                  <Icon className="KeyValueTable--linkIcon" type="export" />
-                  {jsonTable}
-                </a>
+                <div>
+                  <a href={links[0].url} title={links[0].text} target="_blank" rel="noopener noreferrer">
+                    {markupAsSpan(markup)} <Icon className="KeyValueTable--linkIcon" type="export" />
+                  </a>
+                </div>
               );
             } else if (links && links.length > 1) {
               const menuItems = (
@@ -80,13 +88,16 @@ export default function KeyValuesTable(props: KeyValuesTableProps) {
                 </Menu>
               );
               valueMarkup = (
-                <Dropdown overlay={menuItems} placement="bottomRight" trigger={['click']}>
-                  <a className="KeyValueTable--link">
-                    <Icon className="KeyValueTable--linkIcon" type="profile" />
-                    {jsonTable}
-                  </a>
-                </Dropdown>
+                <div>
+                  <Dropdown overlay={menuItems} placement="bottomRight" trigger={['click']}>
+                    <a>
+                      {markupAsSpan(markup)} <Icon className="KeyValueTable--linkIcon" type="profile" />
+                    </a>
+                  </Dropdown>
+                </div>
               );
+            } else {
+              valueMarkup = markupAsDiv(markup);
             }
             return (
               // `i` is necessary in the key because row.key can repeat
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.js
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.js
@@ -33,7 +33,7 @@ function parseIfJson(value) {
 
 type KeyValuesTableProps = {
   data: { key: string, value: any }[],
-  linksGetter: ({ key: string, value: any }[], number) => { url: string, text: string }[],
+  linksGetter: ?({ key: string, value: any }[], number) => { url: string, text: string }[],
 };
 
 export default function KeyValuesTable(props: KeyValuesTableProps) {
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.js
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.js
@@ -37,14 +37,9 @@ function markupAsDiv(markup) {
 }
 
 function markupAsSpan(markup) {
+  const spanMarkup = markup.replace(/^<div /i, '<span ').replace(/<\/div>$/i, '</span>');
   // eslint-disable-next-line react/no-danger
-  return (
-    <span
-      dangerouslySetInnerHTML={{
-        __html: markup.replace(/^<div /i, '<span ').replace(/<\/div>$/i, '</span>'),
-      }}
-    />
-  );
+  return <span dangerouslySetInnerHTML={{ __html: spanMarkup }} />;
 }
 
 type KeyValuesTableProps = {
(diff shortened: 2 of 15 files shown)


yurishkuro:
When are the evaluations performed, only when the span is expanded?

I think we'll need some visual indication that the tag/value have links attached, maybe the hyperlink 🔗 symbol somewhere.

divdavem:
@yurishkuro Thank you for your answer.

> When are the evaluations performed, only when the span is expanded?

Yes, exactly, just before they are displayed.

> I think we'll need some visual indication that the tag/value have links attached, maybe the hyperlink 🔗 symbol somewhere.

There are already some icons on the right, but maybe they are not visible enough or not the right ones?
I have used icons from [antd](https://ant.design/components/icon/): the `profile` icon when there are multiple links and the `export` icon when there is a single link. I can change them if you think it is better .
The `link` icon usually means there is an anchor to link back to this position in the page, which is not exactly the case here, but maybe it is clearer than the icons I have used.

divdavem:
I have added a commit to make icons more visible:
- they are now just next to the value, on the right
- they are in bold

jaeger link icons

Also, when there are multiple links, the popup is no longer on the whole column width, but has a width adapted to its content.

yurishkuro:
+1

I would stick with 
this icon: 
image 

One concern with showing it on the right is when the value of the tag is long & truncated.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.js
```diff
@@ -30,27 +31,75 @@ function parseIfJson(value) {
...
+function markupAsDiv(markup) {
+  // eslint-disable-next-line react/no-danger
+  return <div dangerouslySetInnerHTML={{ __html: markup }} />;
+}
+
+function markupAsSpan(markup) {
+  const spanMarkup = markup.replace(/^<div /i, '<span ').replace(/<\/div>$/i, '</span>');
+  // eslint-disable-next-line react/no-danger
+  return <span dangerouslySetInnerHTML={{ __html: spanMarkup }} />;
+}
```
tiffon: Instead of manipulating the HTML seems like adding `ub-inline-block` to the `<div>` would work?

```js
            const jsonTable = (
              // eslint-disable-next-line react/no-danger
              <div className="ub-inline-block" dangerouslySetInnerHTML={{ __html: jsonMarkup(parseIfJson(row.value)) }} />
            );
```

I think that might also need

```css
.KeyValueTable--body > tr > td {
    padding: 0.25rem 0.5rem;
    vertical-align: top;
}
```

tiffon:
@divdavem, Awesome! :tada: Thank you for putting this together! :)

I have a few questions but want to spend more time digesting it and should be able to follow up tomorrow.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.js
divdavem: Thank you for your suggestion. I have added a commit with this change.

divdavem:
@tiffon Thank you. For info, I am on holidays tonight for a bit more than two weeks, so I may not be able to answer your questions soon. I am planning to add some more tests this afternoon before my holidays.

divdavem:
There seems to be a failing test in Travis which is unrelated to my changes: https://travis-ci.org/jaegertracing/jaeger-ui/builds/400936060#L1378

> Summary of all failing tests
>  FAIL  src/selectors/trace.test.js (122.854s)
>   ● getTraceServiceCount() should return the length of the service list
>     expect(received).toBe(expected)
>     
>     Expected value to be (using ===):
>       5
>     Received:
>       6
>       
>       at Object.<anonymous>.it (src/selectors/trace.test.js:168:63)
>       at Promise.resolve.then.el (../../node_modules/p-map/index.js:46:16)
>       at process._tickCallback (internal/process/next_tick.js:109:7)

Between the commit of the previous successful build (https://travis-ci.org/jaegertracing/jaeger-ui/builds/400922649) (136c149060637b33342cd8aa04baea9774709508) and the commit of this failing travis build (a159459c548770100f8b7e4970845017f19f69b9), I only added some tests and did not change any other code: https://github.com/divdavem/jaeger-ui/compare/136c149060637b33342cd8aa04baea9774709508...a159459c548770100f8b7e4970845017f19f69b9

The data used in the failing test seem to be randomly generated, that's probably why such a failure suddenly happened in one build and does not happen in most builds, but, it may underline a real issue, either in the random data generation or in the code.

packages/jaeger-ui/src/model/link-patterns.js
```diff
@@ -0,0 +1,183 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import _uniq from 'lodash/uniq';
+import { getConfigValue } from '../utils/config/get-config';
+
+const parameterRegExp = /\$\{([^{}]*)\}/;
+
+export function processTemplate(template, encodeFn) {
```
tiffon: For the string interpolation, what do you think of using something like [twitter-text/stringSupplant](https://github.com/twitter/twitter-text/blob/34dc1dd9f10e9171100cdff0cb2b7a9ed9ea2bd6/js/src/lib/stringSupplant.js)? 

So, to get the parameter names, something like:

```js
function getParamNames(str) {
  const names = [];
  str.replace(/#\{([^\}]+)\}/g, function (_, name) {
    names.push(name);
  });
  return names;
}
```

And to apply them:

```js
function stringSupplant(str, encodeFn, map) {
  return str.replace(/#\{([^}]+)\}/g, function (_, name) {
    const v = map[name];
    return v ? encodeFn(v) : '';
  });
}
```

I think this would simplify `processTemplate()` and the resulting `template()` function.

```js
template: stringSupplant.bind(null, template, endcodeFn),
```

packages/jaeger-ui/src/model/link-patterns.js
```diff
@@ -0,0 +1,183 @@
...
+}
+
+export function createTestFunction(entry) {
+  if (typeof entry === 'string') {
+    return arg => arg === entry;
+  }
+  if (Array.isArray(entry)) {
+    return arg => entry.indexOf(arg) > -1;
+  }
+  if (entry instanceof RegExp) {
```
tiffon: Planning ahead in preparation for the JS config file (#123) is awesome but it's tough to validate functionality that relies on or is impacted by the JS config before it's implemented. That increases the risks when switching to using the JS config. Maybe the regular expression and function variants of the test function should be kept on ice until a PR for #128 is merged?

packages/jaeger-ui/src/model/link-patterns.js
```diff
@@ -0,0 +1,183 @@
...
+export function getParameterInArray(name, array) {
+  if (array) {
+    return array.find(entry => entry.key === name);
+  }
+  return undefined;
+}
+
+export function getParameterInAncestor(name, spans, startSpanIndex) {
+  let currentSpan = { depth: spans[startSpanIndex].depth + 1 };
+  for (let spanIndex = startSpanIndex; spanIndex >= 0; spanIndex--) {
```
tiffon: This will not exit early when a root level ancestor is encountered, e.g.

```js
for (...) {
  ...
  if (currentSpan.depth === 0) {
    break; // as there will be no more ancestors
  }
}
```

packages/jaeger-ui/src/model/link-patterns.js
```diff
@@ -0,0 +1,183 @@
...
+  if (processTags) {
+    type = 'process';
+  }
+  const spanTags = span.tags === items;
+  if (spanTags) {
+    type = 'tags';
+  }
+  const result = [];
+  linkPatterns.forEach(pattern => {
+    if (pattern.type(type) && pattern.key(item.key, item.value, type) && pattern.value(item.value)) {
```
tiffon: Why is `pattern.key(...)` getting `item.value` and `type`?

packages/jaeger-ui/src/model/link-patterns.js
```diff
@@ -0,0 +1,183 @@
...
+      let parameterValues = {};
+      pattern.parameters.every(parameter => {
+        let entry = getParameterInArray(parameter, items);
+        if (!entry && !processTags) {
```
tiffon: I believe ancestors are also being searched when type === 'log', which seems to contradict your comment on the PR description?

packages/jaeger-ui/src/model/link-patterns.js
```diff
@@ -0,0 +1,183 @@
...
+          // do not look in ancestors for process tags because the same object may appear in different places in the hierarchy
```
tiffon: Given this comment, why are process tags always searched when looking through ancestors? Seems like the only process tags that are skipped are those on the current span.

Also, why is it relevant that process tags can be repeated? A process will match the first time it's encountered or it will fail every time it's tested?

packages/jaeger-ui/src/model/link-patterns.js
```diff
@@ -0,0 +1,183 @@
...
+          url: callTemplate(pattern.url, parameterValues),
+          text: callTemplate(pattern.text, parameterValues),
+        });
+      }
+    }
+  });
+  return result;
+}
+
+const linkPatterns = (getConfigValue('linkPatterns') || []).map(processLinkPattern).filter(value => !!value);
```
tiffon: This can be shortened to `.filter(Boolean)`.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.js
```diff
@@ -16,6 +16,7 @@
 
 import React from 'react';
 import jsonMarkup from 'json-markup';
+import { Dropdown, Icon, Menu } from 'antd';
```
tiffon: This file looks great! Awesome.

(Not sure how to add a comment to a file instead of a line...)

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.js
```diff
@@ -32,25 +33,66 @@ function parseIfJson(value) {
...
+            } else {
+              valueMarkup = jsonTable;
+            }
             return (
               // `i` is necessary in the key because row.key can repeat
               // eslint-disable-next-line react/no-array-index-key
               <tr key={`${row.key}-${i}`}>
                 <td className="KeyValueTable--keyColumn">{row.key}</td>
-                <td>{jsonTable}</td>
+                <td>{valueMarkup}</td>
```
tiffon: Seems like the code above is sufficiently intricate to warrant being broken down into sub-components at the top of the file, similar to [`CustomNavDropdown`](https://github.com/jaegertracing/jaeger-ui/blob/e31801d1a283cc40b360cfda01e4e061a124c0d8/packages/jaeger-ui/src/components/App/TopNav.js#L45). Something like `LinkValue` and `LinkValueList` where `LinkValueList` would compose `LinkValue`? 

```jsx
// Individual link
<LinkValue
  href={href}
  title={title}
  content={jsonMarkup}
/>

// List of links
<LinkValueList
  links={links}
  content={jsonMarkup}
/>
```

What do you think?

Also, I think it would make sense for the link to always show the icon?

packages/jaeger-ui/src/model/link-patterns.js
divdavem: Your suggestion is clearly more readable, even if it requires to execute the regular expression each time the template is executed (as opposed to only once, when `processTemplate` is called). The difference in performance (if any) is probably negligible, so I have updated the PR to use your suggestion.
Do you have any preference between `#{variable}` or `${variable}` ?

packages/jaeger-ui/src/model/link-patterns.js
divdavem: I have now commented the parts of the code that rely on #123.

packages/jaeger-ui/src/model/link-patterns.js
divdavem: I have now added this condition in the for loop to exit early when reaching a root level ancestor.

packages/jaeger-ui/src/model/link-patterns.js
divdavem: This allowed to have a function that can decide whether to display a link with a logic depending on those multiple parameters. I have now removed that possibility as it also depends on #123.

packages/jaeger-ui/src/model/link-patterns.js
divdavem: Thank you! I have updated the PR.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.js
divdavem: Thank you for your encouraging feedback!

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.js
divdavem: I have taken into account this suggestion and updated the PR.
Note that the `Dropdown` component in its `overlay` property seems to work differently if there is a wrapper component such as `LinkValueList` before the `Menu` component, so I don't call `LinkValueList` as a component (`<LinkValueList ... />`) but directly as a function (`{linkValueList(...)}`).

packages/jaeger-ui/src/model/link-patterns.js
divdavem: The `getLinks` function caches results in the `alreadyComputed` weak map.
The key used in that map is the object (with the `key` and `value` properties) that will be displayed as a link (if a link has to be displayed).
If that object is reused at different places in the hierarchy (which is the case for process tags, as the same process object can be reused for multiple spans), a call to `getLinks` with an object that is already in the cache could get the result of a previous call to `getLinks` from a different place in the hierarchy. If a process tag could use values from an ancestor, the ancestor from which the value could be found and the corresponding value could be different between the two places in the hierarchy, which would make `getLinks` return an invalid value.
Another solution (instead of preventing process tags links from using values from ancestors) would be to have a cache that also stores the current span (so we would need a cache with 2 keys), but that would be more complex and I am not sure if it is actually needed to refer to ancestor values from a process tag.

divdavem:
@tiffon Thank you for your review!

Now that I am back from my holidays, I have updated the PR to take your remarks into account.

To answer your question about the model, it is true that it would make the code a bit simpler to have a reference to the parent span from a child span. As you say, it would avoid having to pass the full trace and the index in the trace. Moreover it would allow `getParameterInAncestor` to be more efficient as it would not have to loop over (and skip) spans that are not ancestors.
However, I don't know the implications of this change, as it obviously makes the model recursive (preventing it from being serialized easily...).

tiffon:
@divdavem I will think more about parent references and dive into the changes, Monday!

Thank you for the changes, very thoughtful descriptions and explanations, and consideration of the parent reference.

Side note: This is the biggest PR in Jaeger UI from someone who doesn't work for Uber (by a landslide)...  MILESTONE! Thank you!

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.js
tiffon: Looks great, thank you.

packages/jaeger-ui/src/model/link-patterns.js
```diff
@@ -0,0 +1,215 @@
...
+  template: (data: { [parameter: string]: any }) => string,
+};
+
+function getParamNames(str) {
+  const names = [];
+  str.replace(parameterRegExp, (match, name) => {
+    names.push(name);
+    return match;
+  });
+  return _uniq(names);
```
tiffon: This could maybe use a `Set`?

```
const names = new Set();
// ...
return Array.from(names);
```

packages/jaeger-ui/src/model/link-patterns.js
```diff
@@ -0,0 +1,215 @@
...
+
+import _uniq from 'lodash/uniq';
+import { getConfigValue } from '../utils/config/get-config';
+import type { Span, Trace } from '../types';
+
+const parameterRegExp = /\$\{([^{}]*)\}/g;
+
+type ProcessedTemplate = {
+  parameters: string[],
+  template: (data: { [parameter: string]: any }) => string,
```
tiffon: I think this can be 

```
template: { [string]: any } => string,
```

packages/jaeger-ui/src/model/link-patterns.js
```diff
@@ -0,0 +1,215 @@
...
+  const names = [];
+  str.replace(parameterRegExp, (match, name) => {
+    names.push(name);
+    return match;
+  });
+  return _uniq(names);
+}
+
+function stringSupplant(str, encodeFn: any => string, map) {
+  return str.replace(parameterRegExp, (_, name) => encodeFn(map[name]));
```
tiffon: `encodeUriComponent` will stringify null (and undefined?) values, so maybe it makes sense to guard the encode function invocation?

```js
encodeURIComponent(null)
// -> "null"

fn = () => null;
'abbc'.replace(/b/g, fn)
// -> "anullnullc"
```

packages/jaeger-ui/src/model/link-patterns.js
```diff
@@ -0,0 +1,215 @@
...
+}
+
+export function processTemplate(template: any, encodeFn: any => string): ProcessedTemplate {
+  if (typeof template !== 'string') {
+    /*
+
+    // kept on ice until #123 is implemented:
```
tiffon: Thank you!

packages/jaeger-ui/src/model/link-patterns.js
tiffon: Also, would `mixed` fit the bill instead of `any`?

https://flow.org/en/docs/types/mixed/

packages/jaeger-ui/src/model/link-patterns.js
```diff
@@ -0,0 +1,215 @@
...
+  */
+  if (entry == null) {
+    return () => true;
+  }
+  throw new Error(`Invalid value: ${entry}`);
+}
+
+const identity = a => a;
+
+type ProcessedLinkPattern = {
```
tiffon: Can you please move type definitions to the top of the file?

packages/jaeger-ui/src/model/link-patterns.js
tiffon: Thank you. 

I'd suggest using the `#` to differentiate from template literals.

packages/jaeger-ui/src/model/link-patterns.js
divdavem: Ok

packages/jaeger-ui/src/model/link-patterns.js
divdavem: Ok

packages/jaeger-ui/src/model/link-patterns.js
divdavem: Ok!

packages/jaeger-ui/src/model/link-patterns.js
```diff
@@ -0,0 +1,215 @@
...
+    type = 'process';
+  }
+  const spanTags = span.tags === items;
+  if (spanTags) {
+    type = 'tags';
+  }
+  const result = [];
+  linkPatterns.forEach(pattern => {
+    if (pattern.type(type) && pattern.key(item.key) && pattern.value(item.value)) {
+      let parameterValues = {};
```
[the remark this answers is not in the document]
divdavem: Indeed, I did not think about using the return value, but that is a good idea, thank you!

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/AccordianKeyValues.js
```diff
@@ -30,6 +30,7 @@ type AccordianKeyValuesProps = {
   highContrast?: boolean,
   isOpen: boolean,
   label: string,
+  linksGetter: ?({ key: string, value: any }[], number) => { url: string, text: string }[],
```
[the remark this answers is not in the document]
divdavem: Ok! Indeed, it is better to use named types. I have updated the code to use this KeyValuePair type and the newly added Link one so that it is more readable.

packages/jaeger-ui/src/model/link-patterns.js
divdavem: Ok. As links are not displayed if any parameter is missing, this problem only happens when tags are explicitly added with a null or undefined value. In this case, it is true that is better to guard the encode function invocation and have an empty string rather than the strange "undefined" or "null" strings. I have updated the code.

packages/jaeger-ui/src/model/link-patterns.js
divdavem: Ok. I have updated the PR to use `#` instead of `$`.

divdavem:
@tiffon Thank you for your review!
I have added a commit to add references to the parent span in the model. It simplifies the code in many places (allowing to simply pass a span instead of passing a trace and a span index), and it makes sense, but it also adds two loops of processing in `addSpanReferences` called at the end of `transfromTraceData`.
If there is at most one reference per span, and if that reference is of type `CHILD_OF` or `FOLLOWS_FROM` (as it seems to be assumed in `getTraceSpanIdsAsTree` in `trace.js`), it would be possible to merge those two loops together (because in that case, the spans referred to are always before the spans containing the reference, in the spans array). Also, maybe it would be good to do that as part of the existing loop in `transfromTraceData`, or is it better to keep things separate (which is, maybe, a bit easier to read)?
What do you think?

packages/jaeger-ui/src/model/link-patterns.js
divdavem: I will explain in Javascript:

```js
// Let's say there is one tag like this:

let myProcessTag = {
  key: 'myProcessInfo',
  value: 'myProcessValue'
};

// The tag is used in a process:

let myProcess = {
  tags: [myProcessTag]
};

// The same process reference is used in multiple spans in the trace hierarchy:
let trace = {
  spans: [{
    depth: 0
  }, {
    depth: 1,
    tags: [{
      key: 'myParentKey',
      value: 'parentValue1'
    }]
  }, {
    depth: 2,
    process: myProcess
  }, {
    depth: 1,
    tags: [{
      key: 'myParentKey',
      value: 'parentValue2'
    }]
  }, {
    depth: 2,
    process: myProcess
  }]
};

// So myProcess is used in trace.spans[2], which is the child of trace.spans[1]
// and myProcess is also used in trace.spans[4], which is the child of trace.spans[3]

// Now, let's suppose that it is allowed for a link of a process tag to refer to tags from ancestors
// and someone defined the configuration like this:
JAEGER_CONFIG = {
  "linkPatterns": [{
    "key": "myProcessInfo",
    "url": "http://example.org/?myProcessValue=#{myProcessInfo}&myParentValue=#{myParentKey}",
    "text": "Where will this link go?"
  }]
};

// This link on the process tag refers to a tag from an ancestor
// (which is the thing I am not allowing in the current code).

// Then, if we call computeLinks with trace.spans[2] and myProcessTag, we should get:
[{
  "url": "http://example.org/?myProcessValue=myProcessValue&myParentValue=parentValue1",
  "text": "Where will this link go?"
}]

// Now, if we call computeLinks with trace.spans[4] and the same myProcessTag, we should get:
[{
  "url": "http://example.org/?myProcessValue=myProcessValue&myParentValue=parentValue2",
  "text": "Where will this link go?"
}]

// These two values are different because the same tag is at two different places in the spans hierarchy,
// and the value for myParentKey is taken
// - in the first case from trace.spans[1].tags[0]
// - in the second case from trace.spans[3].tags[0]

// Now the getLink function returned by createGetLinks uses the tag object as a key in the cache.
// In our case, the tag object is myProcessTag.

// So if we first call the getLink function returned by createGetLinks for trace.spans[2] and myProcessTag :
result = cache.get(myProcessTag); // result is undefined
result = computeLinks(...);
// which means:
result = [{
  "url": "http://example.org/?myProcessValue=myProcessValue&myParentValue=parentValue1",
  "text": "Where will this link go?"
}];
cache.set(myProcessTag, result); // result is cached

// now if we call the getLink function for trace.spans[4] and myProcessTag :
result = cache.get(myProcessTag); // result is already computed
return [{
  "url": "http://example.org/?myProcessValue=myProcessValue&myParentValue=parentValue1",
  "text": "Where will this link go?"
}]; // now the return value is wrong for trace.spans[4]
```

I hope this explanation is clearer!
I do not have the requirement of having links on process tags that use values from ancestor spans, so I chose the simplest solution of not allowing it. Do you think it would be better (for consistency or some use cases that I don't know) to allow this and then to use as the key in the cache both the span and the tag object (for example, with a cache based on nested WeakMaps)?

tiffon: Great explanation, I think I've got it... Don't look in ancestors for process tags because the same process object, and therefore tags, can be used for many spans. In this case, searching a span's ancestors could yield different results (if they weren't cached). But, as the cache is keyed on the initial, shared, process tag, all spans with that process would end up getting the same results even if they have different ancestors?

divdavem: Exactly!

divdavem:
@tiffon Thank you for your answer. I have pushed a commit to merge  `addSpanReferences` with `transfromTraceData` and make `span` optional in `SpanReference`.

Note that I had to use a type cast to `any` to make flow accept my change (is there a better solution?). That's because I am converting a `SpanWithProcess` object into a `Span` object by adding the missing properties. I had to remove the creation of a new object so that `spanMap` (from which I get the reference to put in the `SpanReference` object) contains the same references as the final `spans` array (otherwise it contains old `SpanWithProcess` references before their copy in new `Span` objects).

divdavem:
Ok! Thank you!

divdavem:
@tiffon Can you show an example of what you are surprised about (with the configuration and the sample trace)?

tiffon:
@divdavem Yes, absolutely. Thank you for asking. I got my wires crossed a bit, I believe the rest of the list of key-value-pairs is searched, for instance, the other kvps in a log message, but aside from that the rest of the span is not searched. 

For instance, if we have the following config (written in JavaScript but it would actually be in JSON):

```js
linkPatters: [{
  "type": "logs",
  "key": "errorType",
  "url": "/find-more-logs?filterLevel=error&filterErrorType=#{errorType}&filterPhase=#{phase}&filterAuthResult=#{authenticationResult}&filterModus=#{modus}",
  "text": "This links to a search on logs for errors of the same type with the same authentication result"
}]

```

If we had the following span:

```js
{
  spanID: 'abc',
  depth: 0,
  tags: [{
    { key: 'modus', value: 'operandi' },
  }]
  // etc...
  logs: [
    {
      timestamp: 999,
      fields: [
        { key: 'authenticationResult', value: 200 }
      ]
    }
    {
      timestamp: 1001,
      fields: [
        { key: '0error', value: true },
        { key: 'errorType', value: 'timeout' },
        { key: 'phase', value: 'startup' }
      ]
    }
  ]
}
```

I believe when the link pattern is processed it will not fill in the `parameterValues` because only the fields on the second log message will be checked. Not finding the `authenticationResult` or `modus` fields, ancestors would be searched, but there are none.

While `parameterValues` will not be fully populated and the link will not be generated, the example would be able to populate `errorType` and `phase`.

Initially, I incorrectly assumed `phase` would not be populated, but I see now that it would.

I was surprised that ancestor spans are searched but the rest of the initial span is not. If the logs of the initial span and the tags and process tags were checked, the link would be generated.

divdavem:
@tiffon Thank you very much for the time you spent discussing and reviewing this PR and for integrating it!

I personally do not have much experience with Jaeger, but, being part of a UI team, I developed this feature for another team that uses Jaeger and needed this improvement. I will suggest them to share their feedback with you.

Thank you again for welcoming so kindly external contributions!

PR #230, opened 2018-07-24
Fix CollapseOne action

### LINKED ISSUE #229: "Collapse One" button throws an error when trace is fully collapsed
```
0
- 1
--- 2
- 3
--- 4
- 5
```
When you click on the "Collapse One" button for 3 times using the above trace as input, the webapp throws the following error:
`Uncaught TypeError: Cannot read property 'spanID' of undefined
    at collapseOne (duck.js:174)`

yuribit:
Resolves #229 

`nearestCollapsedAncestor` should always have children otherwise in some particular cases (see test) it will be incorrectly added to `childrenHiddenIDs`.


yurishkuro:
cc @davit-y

Issue #213, opened 2018-05-16
Better trace page search


nziebart:
## Requirement - what kind of business use case are you trying to solve?

We would like the ability to search for spans by tag in the trace view page. 

A specific use case here is finding which span is making a certain SQL query. We already know the query, but if there are hundreds of spans in the trace, it's very difficult to locate the corresponding span.

## Problem - what in Jaeger blocks you from solving the requirement?

The search bar on the trace view page only searches by service and operation name (https://github.com/jaegertracing/jaeger-ui/blob/master/packages/jaeger-ui/src/selectors/span.js#L65).

## Proposal - what do you suggest to solve the problem or improve the existing situation?

It would be relatively simple to also search tags and logs, by adding them to the search string. However I came across two issues when experimenting with this:
1) Some tags/logs are too big and match too many inputs. E.g. a stacktrace
2) The fuzzy search exacerbates (1). Also, some of our tag values are hash strings (e.g. `0ca9bd1a2bfcdd3d7abdff52cb49a072`) which the fuzzy search loves to match against.

I'm not sure about the best solution for (1) - maybe have a way to configure tag keys to omit from search? Alternatively, require tag search to be of the form `key: value` (same as the homepage search feature) rather then just blindly searching all tag values against the input string.

To solve (2) I would propose either (a) removing fuzzy search (IMO it's not that useful, and users who are used to chrome search are used to needing exact matching) or (b) using fuzzy search only on operation/service name and doing exact matching for tags/logs.

## Any open questions to address

Questions are noted in proposal.

bboreham:
Came here looking for a way to find a span by tag value.

tiffon:
@nziebart Thanks for creating this ticket!

Great suggestions and discussion of issues / solutions.

>The fuzzy search exacerbates..

I agree, fuzzy search is a bit dubious... Seems fine to change to something better.

>1. Some tags/logs are too big and match too many inputs.

The pattern used in GMail's email search seems like it would work pretty well. It starts as a simple text input which works most of the time but can be used as a more advanced search when the "▾" is clicked.

One approach would be something like the following:

<img width="736" alt="screen shot 2018-06-17 at 3 57 07 am" src="https://user-images.githubusercontent.com/2304337/41506015-443923e8-71e3-11e8-83e1-62990ed10ba0.png">


[...]

Issue #91, opened 2017-10-07
In-trace search should allow navigation between results


yurishkuro:
Currently if we use the Search box on the trace view, it grays out all spans not matching the search and keeps the matching span in normal opacity. However, on a large trace this makes it difficult to find the matching spans.

Suggestion: 
  * scroll the trace view to the first matching span automatically if it's not currently in the view port
  * add the navigation buttons to the Search bar that allow jumping forwards/backwards between the matching span, similar to how in-page search works in Chrome - it shows up and down arrows, and by default the Enter key jumps to the next match.

![image](https://user-images.githubusercontent.com/3523016/31310022-851fddbc-ab5e-11e7-9375-6c4b1a1f8f8b.png)

tiffon:
Great suggestion.

FYI, something tangentially related, in #93 I've added shortcuts to scroll to prev / next visible span. When a search is in-progress, this skips spans that are filtered out.

yurishkuro:
done in #234

PR #234, opened 2018-08-13
Span Search - Add result count, navigation and clear buttons

davit-y:
- #235
- Also #91 and #213 

- Add result count
- Add up/down arrows (call current nextVisibleSpan method)
- Add clear button

## TODO:
- [x] Improve aesthetics of disabled state
- [x] Tests
- [x] Add shortcuts


black-adder:
Screen shots please

black-adder:
bonus points if you screen shot your IDE with the code instead

davit-y:
Without text filter:

screen shot 2018-08-13 at 5 48 23 pm


With text filter:

screen shot 2018-08-13 at 5 48 42 pm

packages/jaeger-ui/src/components/TracePage/TracePageHeader.js
```diff
@@ -30,14 +29,18 @@ import { formatDatetime, formatDuration } from '../../utils/date';
 import prefixUrl from '../../utils/prefix-url';
 
 import './TracePageHeader.css';
+import TracePageSearchBar from './TracePageSearchBar';
```
tiffon: Please move this import to fall after TracePageHeader.track.

Imports are sorted alphabetically with more dots under less dots and css files that are imported for the side effect go below the other imports. Also, external packages go above local resources, separated by a blank line, with react going above others.

	import extA from 'a';
	import extB from 'b';

	import a from './a';
	import z from './z';
	import b from '../b';

	import 'a.css';

packages/jaeger-ui/src/components/TracePage/TracePageHeader.js
```diff
@@ -170,13 +176,13 @@ export default function TracePageHeader(props: TracePageHeaderProps) {
...
         </a>
         <KeyboardShortcutsHelp className="ub-mr2" />
-        <div className="ub-mr2">
-          <Input
-            name="search"
-            placeholder="Search..."
-            onChange={event => updateTextFilter(event.target.value)}
-            defaultValue={textFilter}
-            data-test={markers.IN_TRACE_SEARCH}
+        <div className="ub-flex-auto ub-mr2 TracePageHeader--search">
```
tiffon: Please move this wrapping `<div>` into the `TracePageSearchBar` component.

packages/jaeger-ui/src/components/TracePage/TracePageSearchBar.js
```diff
@@ -0,0 +1,77 @@
...
+
+type TracePageSearchBarProps = {
+  updateTextFilter: string => void,
+  textFilter: string,
+  prevResult: () => void,
+  nextResult: () => void,
+  resultCount: number,
+};
+
+type TracePageSearchBarState = {
```
tiffon: Why do you need state? `this.props.textFilter` should be all you need.

packages/jaeger-ui/src/components/TracePage/TracePageSearchBar.js
```diff
@@ -0,0 +1,77 @@
...
+  };
+
+  render() {
+    const { prevResult, nextResult, resultCount } = this.props;
+
+    const count = this.state.textFilter ? <span>{resultCount.toString()}</span> : null;
+
+    return (
+      <Input.Group compact style={{ display: 'flex' }}>
+        <Input // ^ inline because compact overwrites the display
```
tiffon: Please move the comment above the line it refers to.

packages/jaeger-ui/src/components/TracePage/TracePageSearchBar.js
```diff
@@ -0,0 +1,77 @@
...
+
+  updateTextFilter = (textFilter: string) => {
+    this.props.updateTextFilter(textFilter);
+    this.setState({ textFilter });
+  };
+
+  render() {
+    const { prevResult, nextResult, resultCount } = this.props;
+
+    const count = this.state.textFilter ? <span>{resultCount.toString()}</span> : null;
```
tiffon: I believe the `.toString()` call here is unnecessary.

tiffon: I'd say it's preferable for the suffix to have less contrast than the input text:

```jsx
const count = ... ? <span className="u-tx-muted">{resultCount}</span> : null;
```

packages/jaeger-ui/src/components/TracePage/TracePageSearchBar.js
```diff
@@ -0,0 +1,77 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import * as React from 'react';
+import { Button, Input } from 'antd';
+
+import * as markers from './TracePageHeader.markers';
```
tiffon: Please set the tests up so each component only uses it's own markers, in a file named after the component, e.g. `TracePageSearchBar.markers.js`.

packages/jaeger-ui/src/components/TracePage/TracePageSearchBar.js
```diff
@@ -0,0 +1,77 @@
...
+  state: TracePageSearchBarState;
+
+  constructor(props: TracePageSearchBarProps) {
+    super(props);
+    this.state = {
+      textFilter: props.textFilter,
+    };
+  }
+
+  updateTextFilter = (textFilter: string) => {
```
tiffon: If you change this to take the event as the parameter you do not need to create a function as the handler on every render.

packages/jaeger-ui/src/components/TracePage/TracePageSearchBar.js
```diff
@@ -0,0 +1,77 @@
...
+          className="TracePageSearchBar--bar ub-flex-auto"
+          placeholder="Search..."
+          onChange={event => this.updateTextFilter(event.target.value)}
+          value={this.state.textFilter}
+          data-test={markers.IN_TRACE_SEARCH}
+          suffix={count}
+        />
+        <Button disabled={!this.state.textFilter} icon="up" onClick={prevResult} />
+        <Button disabled={!this.state.textFilter} icon="down" onClick={nextResult} />
+        <Button disabled={!this.state.textFilter} icon="close" onClick={() => this.updateTextFilter('')} />
```
tiffon: I recommend creating a method `clearFilter = () => { ... }` instead of creating a function on every render.

packages/jaeger-ui/src/components/TracePage/index.js
```diff
@@ -45,6 +45,7 @@ import type { Config } from '../../types/config';
 import prefixUrl from '../../utils/prefix-url';
 
 import './index.css';
+import getFilteredSpans from './TraceTimelineViewer/get-filtered-spans';
```
tiffon: Doesn't look like this file needs to be in the `./TraceTimelineViewer/` folder anymore?

packages/jaeger-ui/src/components/TracePage/index.js
```diff
@@ -278,6 +291,9 @@ export default class TracePage extends React.PureComponent<TracePageProps, Trace
             traceID={traceID}
             onSlimViewClicked={this.toggleSlimView}
             textFilter={textFilter}
+            prevResult={this.prevResult}
+            nextResult={this.nextResult}
```
tiffon: I don't think the use of methods is needed for the scroll props, e.g. 

```jsx
prevResult={this._scrollManager.scrollToPrevVisibleSpan}
```

packages/jaeger-ui/src/components/TracePage/index.js
```diff
@@ -201,11 +204,21 @@ export default class TracePage extends React.PureComponent<TracePageProps, Trace
     }
   };
 
-  updateTextFilter = (textFilter: ?string) => {
+  updateTextFilter = (textFilter: string) => {
     trackFilter(textFilter);
     this.setState({ textFilter });
+    const findMatchesIDs = getFilteredSpans(this.props.trace, textFilter.trim());
```
tiffon: I recommend setting `findMatchesIDs` to `null` if `!textFilter.trim()` ...

packages/jaeger-ui/src/components/TracePage/TracePageHeader.js
davit-y: oops, forgot to push my commit

packages/jaeger-ui/src/components/TracePage/index.js
davit-y: I initially did this to set up the next PR where the buttons move one result at a time, but that will probably just live in the scrollManager too so I'll get rid of the methods.

packages/jaeger-ui/src/components/TracePage/TracePageSearchBar.js
davit-y: the `.toString()` makes sure that "0" shows up, otherwise it's interpreted as null

I had wrapped with a span to add formatting, but forgot. `u-tx-muted` didn't seem to work so I created a style that drops the opacity.

davit-y:
Old disabled state:
screen shot 2018-08-13 at 5 48 23 pm

Lighter disabled state:
screen shot 2018-08-14 at 12 52 11 pm

packages/jaeger-ui/src/components/TracePage/TracePageHeader.test.js
```diff
@@ -56,8 +57,14 @@ describe('<TracePageHeader>', () => {
...
-    const props = { ...defaultProps, updateTextFilter };
-    wrapper = shallow(<TracePageHeader {...props} />);
+    const searchBarProps = {
+      updateTextFilter,
+      textFilter: defaultProps.textFilter,
+      prevResult: () => {},
+      nextResult: () => {},
+      resultCount: 0,
+    };
+    wrapper = shallow(<TracePageSearchBar {...searchBarProps} />);
```
tiffon: Seems like this should be in `TracePageSearchBar.test.js`?

packages/jaeger-ui/src/components/TracePage/TracePageSearchBar.css
```diff
@@ -0,0 +1,36 @@
...
+.TracePageSearchBar {
+  max-width: 20rem;
+}
+
+.TracePageSearchBar--bar {
+  display: flex;
+  max-width: 20rem;
+}
+
+.TracePageSearchBar--count {
```
tiffon: Was the `u-tx-muted` CSS utility class too dark?

packages/jaeger-ui/src/components/TracePage/TracePageSearchBar.css
```diff
@@ -0,0 +1,36 @@
...
+  opacity: 0.6;
+}
+
+.TracePageSearchBar--btn {
+  transition: 0.5s;
+}
+
+.TracePageSearchBar--btn-hide {
```
tiffon: For consistency, please change the selector to something like `.TracePageSearchBar--btn.is-*` where `*` describes the state, e.g. `.DiffNode--labelCell.is-added`.

packages/jaeger-ui/src/components/TracePage/TracePageSearchBar.js
tiffon: Ok, good to know. But, looks like `0` is working ok with the suffix?

https://codesandbox.io/s/lp1r4yqz8q

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/index.js
```diff
@@ -36,8 +36,8 @@ type TraceTimelineViewerProps = {
   collapseOne: (Span[]) => void,
   expandAll: () => void,
   expandOne: (Span[]) => void,
+  findMatchesIDs: Set<string>,
```
tiffon: If this value can be null, then the type should be `Set<string> | null`

packages/jaeger-ui/src/components/TracePage/index.js
```diff
@@ -99,6 +101,7 @@ export class TracePageImpl extends React.PureComponent<TracePageProps, TracePage
   state: TracePageState;
 
   _headerElm: ?Element;
+  _searchBar: any;
```
tiffon: I recommend using the following type instead of `any`:

```
import { Input } from 'antd';

// ...

ref: { current: Input | null }
```

packages/jaeger-ui/src/components/TracePage/keyboard-shortcuts.js
```diff
@@ -64,6 +67,8 @@ export const kbdMappings = {
   expandAll: '[',
   collapseOne: 'p',
   expandOne: 'o',
+  searchSpans: 'ctrl+b',
```
tiffon: For this, why not use `meta` ?

packages/jaeger-ui/src/components/TracePage/index.js
```diff
@@ -61,7 +62,8 @@ type TracePageProps = {
 type TracePageState = {
   headerHeight: ?number,
   slimView: boolean,
-  textFilter: ?string,
+  textFilter: string,
+  findMatchesIDs: ?Set<string>,
```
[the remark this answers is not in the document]
davit-y: Done

packages/jaeger-ui/src/components/TracePage/keyboard-shortcuts.js
davit-y: The keyboard handler adds the inputted character to the input. Meta+b outputs a character, while ctrl+b does not.

A work around to use meta is to remove all instances of the '∫' char from the input string, but I didn't want to clutter the code. I tried having the textFilter shortened by 1 char after the event goes off, but unfortunately the input box receives the char after the completion of the event.

tiffon:
Looks great. Thanks for the hard work!

tiffon:
@yurishkuro, can you merge this PR? I'll take care of the test coverage when I address test coverage in trace comparison.

Issue #248, opened 2018-09-17
Make jaeger-ui installable by npm


aljesusg:
## Requirement - what kind of business use case are you trying to solve?

I am from Jaeger team and [Kiali](https://github.com/kiali) in RH. 
Kiali is the UI for [Istio](https://istio.io/), we are currently integrating the jaeger with some external links.

We want to collect the information provided by jaeger and the information of Kiali and show the relation to improve the tracing and observability of the service mesh so for example if we can show the component of the traceresults with our information, will be useful for the user.

Actually we have an external link to the full UI jaeger in tabs, our idea is contribute/improve the component to import them to our project from the jaeger UI for example the last 10 traces in our view using the component TraceTimelineViewer. 

## Problem - what in Jaeger blocks you from solving the requirement?

We can't integrate components of Jaeger UI  So we are using external links to the Jaeger-UI in a new tabs.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

1. Change the package.json to publish the releases
2. Compile al build the components to export them
3. Improve/contribute in  the component

cc @jpkrohling @objectiser

yurishkuro:
@tiffon thoughts?

mtho11:
It would be great to see these components published to npm so that others could use the jaeger-ui components directly in applications.

saminzadeh:
> It would be great to see these components published to npm so that others could use the jaeger-ui components directly in applications.

+1. This would be nice as a component library.

A few initial ideas on how to approach this:

- **Stateless Components that take the standardized Trace object**. You will responsible for providing the Trace object to the component.

```js
import { TraceViewTimeline, TraceDAG } from '@jeagertracing/ui-kit';
import trace from './samples/my-jaeger-trace.json'

<TraceViewTimeline trace={trace} />
<TraceDAG trace={trace} customOverrides={...} />
```

- **React Provider Pattern (eg. Redux, Apollo GraphQL, etc)**. Similar to how [SearchKit](http://docs.searchkit.co/stable/) makes it easy to integrate with ElasticSearch. We could do something similar for the JaegerQuery API. 

```js
import { 
  TraceViewTimeline, 
  JaegerClient, 
  JaegerProvider, 
  JaegerRenderer 
} from '@jeagertracing/ui-kit';

// Init Jaeger client
const jaegerClient = new JaegerClient("http://my.jaeger.host.com")

const App = ()=> (
  <JaegerProvider jaegerClient={jaegerClient}>
[...]

PR #249, opened 2018-09-19
WIP Make jaeger-ui installable

aljesusg:

Issue https://github.com/jaegertracing/jaeger-ui/issues/248 Make jaeger-ui installable

Change package.json to be public, add babel-preset-react-app to compile components


tiffon:
I have reservations about wholesale publishing the UI as a component library. I added a comment to #248:

>One drawback for wholesale publishing the React components, in their current form, is that when all of the components become public, their combined API becomes the API contract. That's a massive contract. So much so it would essentially become impossible to maintain backwards compatibility while making any significant changes to the UI.
>
>So, if we go down the path of publishing React components, I think the approach outlined by @saminzadeh would make the API contract more manageable.

I'm not sure publishing the current site is the most manageable approach, in the long-term. Let's discuss in #248?

PR #263, opened 2018-10-08
Embedded components (SearchTraces and Tracepage )

aljesusg:
Related issue : https://github.com/jaegertracing/jaeger-ui/issues/248

The point is make a PoC about how whould this look .

We have here 2 uses of cases:

- Show the results of a search
results

- Show a Trace
trace

So I added query param called **embed**, with this params the top navigation and search are removed.
**See demo GIF**

embed_components

# How works ? pass **embed** like a query param

# TracePage View
trace

- **minimap** option
- **details** option

# SearchTraces View
search

- **disableComparision** option
- **hideGraph** option

### DIFF
--- a/packages/jaeger-ui/src/components/App/Page.js
+++ b/packages/jaeger-ui/src/components/App/Page.js
@@ -20,7 +20,7 @@ import Helmet from 'react-helmet';
 import { connect } from 'react-redux';
 import type { Location } from 'react-router-dom';
 import { withRouter } from 'react-router-dom';
-
+import { isEmbed } from '../../utils/embedded';
 import TopNav from './TopNav';
 import { trackPageView } from '../../utils/tracking';
 
@@ -56,10 +56,12 @@ export class PageImpl extends React.Component<Props> {
       <div>
         <Helmet title="Jaeger UI" />
         <Layout>
-          <Header className="Page--topNav">
-            <TopNav />
-          </Header>
-          <Content className="Page--content">{this.props.children}</Content>
+          {!isEmbed(this.props.search) && (
+            <Header className="Page--topNav">
+              <TopNav />
+            </Header>
+          )}
+          <Content className={!isEmbed(this.props.search) && 'Page--content'}>{this.props.children}</Content>
         </Layout>
       </div>
     );
--- a/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.js
+++ b/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.js
@@ -15,7 +15,7 @@
 // limitations under the License.
 
 import * as React from 'react';
-import { Select } from 'antd';
+import { Select, Button } from 'antd';
 import { Field, reduxForm, formValueSelector } from 'redux-form';
 
 import DiffSelection from './DiffSelection';
@@ -27,6 +27,7 @@ import * as orderBy from '../../../model/order-by';
 import { getPercentageOfDuration } from '../../../utils/date';
 import prefixUrl from '../../../utils/prefix-url';
 import reduxFormFieldAdapter from '../../../utils/redux-form-field-adapter';
+import { VERSION_API } from '../../../utils/embedded';
 
 import type { FetchedTrace } from '../../../types';
 
@@ -37,7 +38,10 @@ type SearchResultsProps = {
   cohortRemoveTrace: string => void,
   diffCohort: FetchedTrace[],
   goToTrace: string => void,
+  embed?: boolean,
+  hideGraph?: boolean,
   loading: boolean,
+  getSearchURL: () => string,
   maxTraceDuration: number,
   skipMessage?: boolean,
   traces: TraceSummary[],
@@ -85,8 +89,20 @@ export default class SearchResults extends React.PureComponent<SearchResultsProp
   };
 
   render() {
-    const { loading, diffCohort, skipMessage, traces } = this.props;
-    const diffSelection = <DiffSelection toggleComparison={this.toggleComparison} traces={diffCohort} />;
+    const {
+      loading,
+      diffCohort,
+      skipMessage,
+      traces,
+      goToTrace,
+      embed,
+      hideGraph,
+      getSearchURL,
+      maxTraceDuration,
+    } = this.props;
+    const diffSelection = !embed && (
+      <DiffSelection toggleComparison={this.toggleComparison} traces={diffCohort} />
+    );
     if (loading) {
       return (
         <React.Fragment>
@@ -107,28 +123,41 @@ export default class SearchResults extends React.PureComponent<SearchResultsProp
         </React.Fragment>
       );
     }
-    const { goToTrace, maxTraceDuration } = this.props;
     const cohortIds = new Set(diffCohort.map(datum => datum.id));
     return (
       <div>
         <div>
           <div className="SearchResults--header">
-            <div className="ub-p3">
-              <ScatterPlot
-                data={traces.map(t => ({
-                  x: t.startTime,
-                  y: t.duration,
-                  traceID: t.traceID,
-                  size: t.spans.length,
-                  name: t.traceName,
-                }))}
-                onValueClick={t => {
-                  goToTrace(t.traceID);
-                }}
-              />
-            </div>
+            {!hideGraph && (
+              <div className="ub-p3">
+                <ScatterPlot
+                  data={traces.map(t => ({
+                    x: t.startTime,
+                    y: t.duration,
+                    traceID: t.traceID,
+                    size: t.spans.length,
+                    name: t.traceName,
+                  }))}
+                  onValueClick={t => {
+                    goToTrace(t.traceID);
+                  }}
+                />
+              </div>
+            )}
             <div className="SearchResults--headerOverview">
               <SelectSort />
+              {embed && (
+                <label className="ub-right">
+                  <Button
+                    className="ub-mr2 ub-items-center"
+                    icon="export"
+                    target="_blank"
+                    href={getSearchURL()}
+                  >
+                    View Results
+                  </Button>
+                </label>
+              )}
               <h2 className="ub-m0">
                 {traces.length} Trace{traces.length > 1 && 's'}
               </h2>
@@ -143,9 +172,16 @@ export default class SearchResults extends React.PureComponent<SearchResultsProp
                 <ResultItem
                   durationPercent={getPercentageOfDuration(trace.duration, maxTraceDuration)}
                   isInDiffCohort={cohortIds.has(trace.traceID)}
-                  linkTo={prefixUrl(`/trace/${trace.traceID}`)}
+                  linkTo={prefixUrl(
+                    embed
+                      ? `/trace/${trace.traceID}?embed=${VERSION_API}&fromSearch=${encodeURIComponent(
+                          getSearchURL()
+                        )}`
+                      : `/trace/${trace.traceID}`
+                  )}
                   toggleComparison={this.toggleComparison}
                   trace={trace}
+                  disableComparision={embed}
                 />
               </li>
             ))}
--- a/packages/jaeger-ui/src/components/TracePage/TracePageHeader.css
+++ b/packages/jaeger-ui/src/components/TracePage/TracePageHeader.css
@@ -19,7 +19,15 @@ limitations under the License.
   display: flex;
 }
 
-.TracePageHeader--title {
+.TracePageHeader--titleRowEmbed {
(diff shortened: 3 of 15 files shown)


lambertjosh:
Would the use case be to put this in an iframe? Are there security concerns with that pattern?

aljesusg:
Kiali PoC in https://github.com/kiali/kiali-ui/pull/736 and how looks
poc kiali

objectiser:
@aljesusg Is the time duration linked to the duration used on other pages? So if navigating from a page where duration is set to 30mins, that will be used when navigating to the traces page?

Will you also be supporting the tag filter field?

tiffon:
This looks great.

I'd say for the search page, we might want to add the ability to

- Hide the scatter plot
- Disable trace comparisons

For the trace detail page, it might be nice to be able to:

- Start with the mini-map collapsed
- Have a way to go back to the search results page (if the user navigated from there)

What do you think?

For both pages, does it makes sense to show a link which would let the user open the page in a new tab (without the embed view restrictions)?

Looks great in the Kiali gif!

aljesusg:
# Demo TracePage
With option mapCollapsed the map is collapsed.
I don't like where I put the buttons to GoBack and link to a new tab with the full version. Any idea?
demo_tracepage

aljesusg:
Changes addressed @tiffon , I don't know where is the best positivo for the buttons of goBack and go to the Jaeger Site

tiffon:
@aljesusg Looks great!

Regarding the where to put the buttons, I think you're right; they kind of stick out, right now.

I've been looking at other embed styles, and they usually have a footer which gives a title and some other details:

GitHub gist

ifnuoqt2nhdpqgolgyk9


Plotly (not an embedded view, but similar)

share from feed

(The red mark in the above image is from the blog post I copied it from; I didn't add it.)

Maybe we can do something similar, but move it to the top?

The top bar can link to the full view and replace the current trace header (title, extra buttons, etc).

This would make the view more compressed and make the "Back to search results" button seem less out of place.

Here is one possible aesthetic, with the back and without:

screen shot 2018-10-18 at 1 02 36 pm
screen shot 2018-10-18 at 1 10 47 pm

What do you think?

We can apply a similar treatment to the search results view so they'd be consistent.

Thanks for your time and hard work!

aljesusg:
Ok @tiffon I am gonna make the changes asap. Thanks

aljesusg:
@tiffon for example for trace view is ok like this or do you want create a header for embed component without search and options? Where is the antd component for that example? I can't find that component in https://ant.design/
screenshot from 2018-10-18 22-17-09

tiffon:
@aljesusg I think a header without the "View Options" and keyboard help modal would probably be preferable. It might also be nice to be abel to toggle the additional details (datetime, etc) and the minimap. What do you think?

The screenshots from my previous comment are not from an Ant Design component, I use the Chrome Dev Tools to work something up from a random page to just take the screenshot.

Would it be possible to match the aesthetic from the screenshots?

aljesusg:
Hi @tiffon 

Sorry but I was on holidays this weeks. I'll try to adapt the ui to be like your screnshoots but I don't know if we can do it with antd

aljesusg:
@tiffon What dio you think about 

# embed component
embed

# embed component with minimap option
embed_minimap
# embed component with details option
embed_details

# embed component with details and minimap option 
embed_details_minimap

tiffon:
@aljesusg Can you post updated screenshots of the search page (if there were changes)?

@aljesusg, @objectiser Apologies for the delay in responding to this. I will take a closer look over the weekend.

aljesusg:
mmm Let me try it because this is generated by another div in an upper level.

packages/jaeger-ui/src/components/TracePage/TracePageHeader.css
```diff
@@ -19,11 +19,26 @@ limitations under the License.
   display: flex;
 }
 
+Divider {
```
tiffon: I don't think this targets anything...?

packages/jaeger-ui/src/components/TracePage/TracePageHeader.js
```diff
@@ -170,38 +179,74 @@ export function TracePageHeaderFn(props: TracePageHeaderProps) {
     },
   ];
 
+  const embedComponent = (
+    <div className="TracePageHeader--titleRowEmbed">
+      <a className="ub-mr2 ub-flex ub-items-center" onClick={onGoBackClicked}>
+        <IoChevronLeft className="ub-mr2" />Go back
+      </a>
+      <Divider type="vertical" />
```
tiffon: I don't think this Divider is visible?

packages/jaeger-ui/src/components/TracePage/TracePageHeader.css
```diff
@@ -19,11 +19,26 @@ limitations under the License.
...
+  display: flex;
+}
+
 .TracePageHeader--title {
   margin: 0;
   padding: 0.25rem 0.5rem;
 }
 
+.TracePageHeader--titleEmbed {
+  margin: 0;
```
tiffon: How is this different from `.TracePageHeader--title`?

packages/jaeger-ui/src/components/TracePage/TracePageHeader.css
```diff
@@ -19,11 +19,26 @@ limitations under the License.
...
+  border-bottom: 1px solid #d8d8d8;
+  display: flex;
+}
+
 .TracePageHeader--title {
   margin: 0;
   padding: 0.25rem 0.5rem;
 }
 
+.TracePageHeader--titleEmbed {
```
tiffon: Please keep related styles adjacent, e.g. `.TracePageHeader--title` and `.TracePageHeader--title:hover`

packages/jaeger-ui/src/components/TracePage/TracePageHeader.js
```diff
@@ -170,38 +179,74 @@ export function TracePageHeaderFn(props: TracePageHeaderProps) {
     },
   ];
 
+  const embedComponent = (
```
tiffon: Seems like this can be switched to a separate component? Right now all of the JSX is being calculated and then ignored most of the time.

packages/jaeger-ui/src/components/TracePage/index.js
```diff
@@ -157,6 +162,9 @@ export class TracePageImpl extends React.PureComponent<TracePageProps, TracePage
       const { trace } = nextProps;
       this._scrollManager.setTrace(trace && trace.data);
     }
+    if (nextProps.mapCollapsed) {
```
tiffon: Can the `mapCollapsed` prop be used directly in the render function instead of using state? Would `!mapCollapsed && !slimView` work?

packages/jaeger-ui/src/components/TracePage/index.js
```diff
@@ -58,6 +59,10 @@ type TracePageProps = {
   history: RouterHistory,
   id: string,
   trace: ?FetchedTrace,
+  isEmbed?: boolean,
+  minimap?: boolean,
+  details?: boolean,
+  mapCollapsed?: boolean,
```
tiffon: What is `minimap` vs `mapCollapsed`?

tiffon:
@objectiser, re 1.8, I'm not sure; I had a number of smaller questions about the diff and there is an open question of how to handle comparisons that we need to sort out.

When is 1.8 slated to ship?

packages/jaeger-ui/src/components/TracePage/TracePageHeader.css
tiffon: Side note, since these are similar and you don't want the hover, writing it as:

```css
.TracePageHeader--title,
.TracePageHeader--titleEmbed {
  margin: 0;
  padding: 0.25rem 0.5rem;
}
```

packages/jaeger-ui/src/components/App/Page.js
```diff
@@ -52,14 +52,17 @@ export class PageImpl extends React.Component<Props> {
...
-          <Header className="Page--topNav">
-            <TopNav />
-          </Header>
-          <Content className="Page--content">{this.props.children}</Content>
+          {!isEmbed && (
+            <Header className="Page--topNav">
+              <TopNav />
+            </Header>
+          )}
+          <Content className={ !isEmbed && "Page--content"}>{this.props.children}</Content>
```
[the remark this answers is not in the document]
aljesusg: Maybe because it was with double quote. I'll change it

packages/jaeger-ui/src/components/TracePage/TracePageHeader.css
aljesusg: Sorry I was testing the antd divider with the style because I can't see it. I'll remove it

packages/jaeger-ui/src/components/TracePage/index.js
aljesusg: mapCollapsed is triggered when the user click in the trace tittle this changes the arrow too.

Minimap remove the map, the arrow and this click.

I think that we should keep this, if I use mapCollapsed the user can click and open the minimap unless that we want to use this behavior. What do you think?

aljesusg:
I'll try to have the changes today @objectiser

packages/jaeger-ui/src/components/TracePage/TracePageHeader.css
aljesusg: The different is that there is not hover because there is not mapCollapsed effect

packages/jaeger-ui/src/components/TracePage/index.js
aljesusg: mmm we can remove this now after the last change

tiffon:
@aljesusg We can use the value of the `embed` query string parameter to act as a version specification. So, for now, the only valid value for `embed` would be `embed=v0` which would enable the search and/or trace detail view embedding, and the search results will never be comparable when `embed=v0`. Presently, any other value for `embed` would not trigger embed-mode, it would effectively be ignored.

Then, later, we can add trace comparisons as a possibility and use `embed=v1`. This way, older URLs which use `embed=v0` will still not have comparisons.

This also gives us versioning for embedding, in general, which we also need.

What do you think?

aljesusg:
Great Idea @tiffon .  I did the changed that you proposed but in the traceView I can't align the tittle with the buttons, i don't know if I should set in the css margin-top:10px, to fixed or if there is another way with antd to solve this.

screenshot from 2018-11-12 11-04-22

On the other hand I am going to show the GoBack button if in the history there is a search url in the previous one.

packages/jaeger-ui/src/components/App/Page.js
```diff
@@ -52,14 +53,17 @@ export class PageImpl extends React.Component<Props> {
   }
 
   render() {
+    const { embed } = queryString.parse(this.props.search);
```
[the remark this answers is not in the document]
aljesusg: Yes sorry I forgot it

aljesusg:
Hi @tiffon , here is the last update of the demo with the comments addressed. Thanks

- It's necessary to set embed=v0  (This configuration is in /utils/embedded folder)
- SearchOptions
  - hideGraph
demo_search
- TraceOptions
  - enableDetails
  - mapCollapsed
  - The Go Back button only available if the user come from a search page (see demo for more details)
demo_goback_trace

aljesusg:
Ok but I think that I addressed all the changes, sorry @tiffon I'll do the changes asap

tiffon:
@aljesusg Thanks! I didn't meant to interrupt your weekend, just wasn't sure about some of the changes.

aljesusg:
@tiffon Thanks for your comments !!! I did the changes, don't worry about interrupt my weekend, I am glad to collaborate :)

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.js
```diff
@@ -143,9 +160,12 @@ export default class SearchResults extends React.PureComponent<SearchResultsProp
                 <ResultItem
                   durationPercent={getPercentageOfDuration(trace.duration, maxTraceDuration)}
                   isInDiffCohort={cohortIds.has(trace.traceID)}
-                  linkTo={prefixUrl(`/trace/${trace.traceID}`)}
+                  linkTo={prefixUrl(
+                    isEmbed ? `/trace/${trace.traceID}?embed&fromSearch` : `/trace/${trace.traceID}`
```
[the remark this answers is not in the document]
tiffon: Captured this as [#279 [embed] Enable configuration of any view when in embed mode](#279)

PR #276, opened 2018-11-13
Add a TraceGraph to the TracePage

### LINKED ISSUE #273: Summarize Trace data by service and operation

As user I want to see where do I loose time in a big trace. This would allow me to better understand where I can optimize.

Tracepage already provides a lot of insights. But it gets a bit overwhelming when viewing traces with 1000 spans. 

Add a component which summarize data of a trace: 
Count, TotalTime, % of time grouped by service and operation.

NOTE: There is already #158. But it is quite large in context. 
This is probably a part of it. I decided to use totalTime instead of Avg. time. It helps more to see which are the big chunks. But we could also add avg. time additionaly to it.

copa2:
Resolves #273 

This PR adds a TraceGraph view to the TracePage. This is an alternate view which summarize data of a trace: Count, TotalTime, % of time, in method time grouped by service and operation
(https://github.com/jaegertracing/jaeger-ui/issues/273#issuecomment-437699454).

jaegeruitracegraph2


yurishkuro:
love it 🎉

whistlinwilly:
This is great!!

packages/jaeger-ui/src/components/TracePage/TraceGraph/TraceGraph.css
```diff
@@ -0,0 +1,99 @@
+/*
+Copyright (c) 2017 Uber Technologies, Inc.
```
yurishkuro: (c) 2018 The Jaeger Authors

yurishkuro:
looks like there are linter errors

copa2:
Sorry for this. Fixed the lint/flow errors now. Some tools are now working with some adaptions.
I do development on windows so pre-commit with lint(prettier/lint/glow/check-licence) is not working.

yurishkuro:
@tiffon please review, would love to merge this soon. 

@copa2 is it possible to add tests? Only 5.12% of the new code is tested.

copa2:
Using now DRange for calculating "self-time". Fork-join spans should work but I have no real test data.

This is now the state of the last iteration:

jaegeruitracegraph3

Unfortunately the inclusion of DRange lib will break the build :-(. I am no react build expert:
"Failed to minify the code from this file: 
 	/home/travis/build/jaegertracing/jaeger-ui/node_modules/drange/lib/index.js:5 "

Some help here would be nice.

tiffon:
@copa2 – Great changes! 

Regarding self-time, in the previous version of the branch, the self-time for `Driver::findNearest` was a large negative (e.g. -133ms), but with the current version it's a small positive (e.g. 0.37ms), which is to be expected.

**Previously - negative self-time**


self-time-negative




**Current - small self-time**

self-time-ok





To check things out, I hacked up an edit to the HotROD demo that makes the `redis:GetDriver` calls happen in parallel. The diff is inlined below.


```diff
diff --git a/examples/hotrod/services/driver/server.go b/examples/hotrod/services/driver/server.go
index 274ceb7..5212d0c 100644
--- a/examples/hotrod/services/driver/server.go
+++ b/examples/hotrod/services/driver/server.go
@@ -15,6 +15,8 @@
 package driver
 
 import (
+	"sync"
+
 	"github.com/opentracing/opentracing-go"
 	"github.com/uber/jaeger-lib/metrics"
 	"github.com/uber/tchannel-go"
@@ -78,25 +80,35 @@ func (s *Server) FindNearest(ctx thrift.Context, location string) ([]*driver.Dri
 	driverIDs := s.redis.FindDriverIDs(ctx, location)
 
 	retMe := make([]*driver.DriverLocation, len(driverIDs))
+	wg := sync.WaitGroup{}
+	rvLock := sync.Mutex{}
 	for i, driverID := range driverIDs {
-		var drv Driver
-		var err error
-		for i := 0; i < 3; i++ {
-			drv, err = s.redis.GetDriver(ctx, driverID)
-			if err == nil {
-				break
+		wg.Add(1)
+		go func(driverID string, i int) {
+			var drv Driver
+			var err error
+			for i := 0; i < 100; i++ {
+				drv, err = s.redis.GetDriver(ctx, driverID)
+				if err == nil {
+					break
+				}
+				s.logger.For(ctx).Error("Retrying GetDriver after error", zap.Int("retry_no", i+1), zap.Error(err))
+			}
+			// if err != nil {
+			// 	s.logger.For(ctx).Error("Failed to get driver after 3 attempts", zap.Error(err))
+			// 	// wg.Done()
+			// 	return
+			// }
+			rvLock.Lock()
+			retMe[i] = &driver.DriverLocation{
+				DriverID: drv.DriverID,
+				Location: drv.Location,
 			}
-			s.logger.For(ctx).Error("Retrying GetDriver after error", zap.Int("retry_no", i+1), zap.Error(err))
-		}
-		if err != nil {
-			s.logger.For(ctx).Error("Failed to get driver after 3 attempts", zap.Error(err))
-			return nil, err
-		}
-		retMe[i] = &driver.DriverLocation{
-			DriverID: drv.DriverID,
-			Location: drv.Location,
-		}
+			rvLock.Unlock()
+			wg.Done()
+		}(driverID, i)
 	}
+	wg.Wait()
 	s.logger.For(ctx).Info("Search successful", zap.Int("num_drivers", len(retMe)))
 	return retMe, nil
 }
```

yurishkuro:
NB: the calls to route service are already parallel in HotROD. It would be nice to add a cmd line option to enable parallelism of the GetDriver calls as well, using a similar executor pool as for the route-svc.

tiffon:
@copa2 – Looks great! I'm really stoked about this diff!

And, glad the `.members` was helpful :+1:

Looks great. I had kind of a long comment regarding the implementation of `mode` in the Trace Graph. The approach outlined is kind of but it's the result of `DirectedGraph` being pretty immature. The suggested approach is more typical of React, though. Please let know what you think (it's inline in OpNode.js).

**Node metrics / aggregations**

Currently, the metrics shown on the nodes are:

1. TL (Top left) - Count
    - Number of spans contained in the node
1. BL - Duration
    - Cumulative duration of the spans contained in the node
1. TR - % of trace duration
    - Duration (#2) / trace duration
1. BR - Self-time
    - I believe this is cumulative (lmk if I'm wrong)

This very much gives a sense of scale for the node with regard to how it contributed to the trace, as a whole. And, it is possible to compare groups with one another, but mainly as "How much did this group contribute to the trace vs that group", so, still as the node relates to the full trace.

Since everything is the sum, the metrics are very sensitive to the count. And, that makes it tough to compare groups of different sizes.

What do you think of the following metrics?

1. TL - Count
    - Number of spans contained in the node
    - Unchanged
1. BL - % of trace duration
    - Sum of durations as a percent of the trace duration
    - Moved from TR
1. TR - Average span duration
    - Sum of durations / count
1. BR - Average % self-time
    - Average % self-time (distinct from sum of self-time / sum of durations)

The left-of-center metrics are cumulative and speak to the contribution to the trace as a whole. The right metrics are averages.

Seems like this might make it possible to compare groups against each other as well as see how a node contributes to the trace, overall.

Screenshot of a mock:

What do you think?

**Landing the graph and following up with the metrics**

Regarding which metrics to show in the node and which to base the color coding visualization on, seems like these warrant some exploration? I'm curious to hear from others, although gathering feedback on in-development UI features has been a bit over-looked. 

Two questions:

How would you feel about presenting this work at the Jaeger bi-weekly to see if there is any feedback or suggestions?

OTOH, landing the Trace Graph view would be great. How would you feel about landing it with only the counts metric visible and the other metrics hidden behind a feature flag (query paramter)? That would allow some breathing room for gathering feedback and mulling over the metrics?

Awesome work! Thanks so much :)

tiffon:
@copa2 I just saw in #273 that metrics are a big part of your interests in adding this. No problem if it more sense to land the PR with metrics rather than breaking it into two phases.

packages/jaeger-ui/src/components/TracePage/TraceGraph/OpNode.css
```diff
@@ -0,0 +1,52 @@
...
+.OpNode {
+  border: 1px solid #111;
+  cursor: pointer;
+  white-space: nowrap;
+  border-collapse: separate;
+  border-radius: 2px;
+}
+
+.OpNode td,
+th {
```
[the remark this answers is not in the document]
copa2: Nice catch. Will fix it.

packages/jaeger-ui/config-overrides.js
```diff
@@ -29,5 +34,6 @@ module.exports = function override(_config, env) {
   let config = _config;
   config = injectBabelPlugin(['import', { libraryName: 'antd', style: true }], config);
   config = rewireLess.withLoaderOptions({ modifyVars })(config, env);
+  config = rewireBabelLoader.include(config, resolveApp('../../node_modules/drange'));
```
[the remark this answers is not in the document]
copa2: Used for react-app-rewire-babel-loader. Didn't get it through the build without it:
https://travis-ci.org/jaegertracing/jaeger-ui/builds/457711935#L1192

packages/jaeger-ui/package.json
```diff
@@ -22,6 +22,7 @@
     "enzyme-adapter-react-16": "^1.1.0",
     "enzyme-to-json": "^3.3.0",
     "less-vars-to-js": "^1.2.1",
+    "react-app-rewire-babel-loader": "^0.1.1",
```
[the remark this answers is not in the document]
copa2: see above

packages/jaeger-ui/src/components/TracePage/TraceGraph/OpNode.js
```diff
@@ -0,0 +1,135 @@
...
+  return Math.round(percent * 100) / 100;
+}
+
+export default class OpNode extends React.PureComponent<Props, State> {
+  props: Props;
+  state: State;
+
+  static helpPopover() {
+    const helpTable = (
+      <table className={`OpNode OpNode--mode-service`}>
```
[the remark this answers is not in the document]
copa2: Will fix this

packages/jaeger-ui/src/components/TracePage/TraceGraph/OpNode.js
```diff
@@ -0,0 +1,135 @@
...
+
+  constructor(props: Props) {
+    super(props);
+    this.state = {
+      nodeMode: false,
+    };
+    props.registerFn.call(this, this);
+  }
+
+  toggleNodeMode = () => {
```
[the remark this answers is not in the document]
copa2: I had previously a similar solution using props. The problem here is when the state is updated in the `TraceGraph` it will execute a whole render of the `DirectedGraph`. This will flicker (minimap is newly loaded and the zoom is reset). So I decided to go this hacky way.

packages/jaeger-ui/src/components/TracePage/TraceGraph/TraceGraph.js
```diff
@@ -0,0 +1,181 @@
...
+      percent: 0,
+      selfTime: 0,
+      registerFn: this.registerNode,
+    });
+
+    traceDag.nodesMap.forEach(n => {
+      const ntime = n.members.reduce((p, m) => p + m.span.duration, 0);
+      const childDurationsDRange = n.members.reduce((p, m) => {
+        // Using DRange to handle overlapping spans (fork-join)
+        const cdr = new DRange(m.span.startTime, m.span.startTime + m.span.duration).intersect(
```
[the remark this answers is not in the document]
copa2: Have to test this. But you are probably right.

packages/jaeger-ui/src/components/TracePage/index.js
```diff
@@ -374,18 +385,25 @@ export class TracePageImpl extends React.PureComponent<TracePageProps, TracePage
...
-              registerAccessors={this._scrollManager.setAccessors}
-              findMatchesIDs={findMatchesIDs}
-              trace={data}
-              updateNextViewRangeTime={this.updateNextViewRangeTime}
-              updateViewRangeTime={this.updateViewRangeTime}
-              viewRange={viewRange}
-            />
-          </section>
-        )}
+        {headerHeight &&
```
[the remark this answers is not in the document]
copa2: Good idea. Will change this.

copa2:
@tiffon - Thanks for the review. Really appreciate the constructive feedback.

Regarding your points:

**Node metrics / aggregations**

Why do I use sums? They are easy to understand :-) 
There might be multiple children paths which influence the value. 
E.g. An operation will call 5 children rpc for some input, but in another call with different input it will call 0 children. 
So totaltime and self-time can be quite different and the averages are not meaningful.

I haven't tested with enough real data, so I am still trying to find out which data helps most.
If you see more value in averages I will change this.

**Landing the graph and following up with the metrics**

Color coding is currently on mode. 
Service: Uses the same color scheme as used in the timeline for the same service.
Time: Red for more time used. 0-20% mapped to alpha values. This needs also more experimentation.

Personally I think the component is not quite ready for prime time. I am also unhappy with some points. 
So let me try to go with another iteration. 

More feedback and suggestions also from other people is appreciated. I hoped this would happen within this pull request. 
I can also present this on a bi-weekly but not this friday(its my niece birthday). But if you want to show it feel free.

tiffon:
@copa2 Thanks for your response.

Re feedback in the PR, I agree that's preferable, but the bi-weekly will reach folks that might not typically respond. 

I can't present at this weeks bi-weekly, either (Happy Birthday to your niece!). But, can you show it at the following bi-weekly? It's a great feature to showcase, and feedback will be useful regardless of whether the PR is merged to master by then, or not. If so, I'll add a spot to the agenda.

I'm OOO for a few days but will respond more fully when I return.

copa2:
@pavolloffay / @tiffon 
Sorry I couldn't invest the time I planned. So the current state is in WIP.

I see two possibilities: 
* Just apply minimum changes as reviewed by tiffon and release as MVP to get more feedback(could be done next weekend. I have no time this week).
* Delay this to another release when we gathered more feedback from a bi-weekly.

To the current WIP which is more a lot of experimenting from me and not in a commitable state:
jaegeruitracegraph4
* added sidebar for help and details

Also experimenting with different Node displays:
opnode1
opnode2
opnode3

@tiffon: This friday is bi-weekly. I will shortly present this WIP if you want.

yurishkuro:
It would be great to discuss at the status call. I am also in favor of merging something as experimental MVP and iterate on that.

tiffon:
@copa2 The progress and experiments look great!

I've added an item to the [bi-weekly agenda](https://docs.google.com/document/d/1ZuBAwTJvQN7xkWVvEFXj5WU9_JmS5TPiNbxCJSvPqX0/edit). Looking forward to learning more!

packages/jaeger-ui/src/components/TracePage/TraceGraph/OpNode.js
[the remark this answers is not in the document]
everett980: To avoid rerendering the `DirectedGraph` on each change to `nodeMode`, `nodeMode` could be a part of the redux state, then each `opNode` would subscribe to this piece of redux state. Since `TraceGraph` is a pure component, it should not rerender on redux state change since it receives no props from the redux state.

I also second the change from a boolean to string. It is more descriptive and easier to extend later if we feel the need for more node modes.

packages/jaeger-ui/src/components/TracePage/index.js
everett980: Fairly minor, but I feel as though this would be clearer if it read:
```
{headerHeight && (
  traceGraphView
    ? <section ...>
      <TraceTimelineViewer ... />
    </section>
    : <section ...>
      <TraceGraph ... />
    </section>
)}

**Comment by yurishkuro** [comment]
I am +1 on tagging a feature as "Experimental" (not WIP), but I am -1 on hiding it behind feature flags. Let's show it and have users raise questions. If we hide it we won't get any feedback.

File: packages/jaeger-ui/src/components/TracePage/TraceGraph/OpNode.js (hunk 6 above)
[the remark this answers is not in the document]
Comment (copa2) [reply packages/jaeger-ui/src/components/TracePage/TraceGraph/OpNode.js]: Yes this seems the best solution. Currently I have implemented the above solution with additional caching of edges and vertices so that no redraw occurs. Will change later this to the solution with redux state change.

File: packages/jaeger-ui/src/components/TracePage/TraceGraph/TraceGraph.js (hunk 7 above)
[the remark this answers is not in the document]
Comment (copa2) [reply packages/jaeger-ui/src/components/TracePage/TraceGraph/TraceGraph.js]: Tested this with differnt data. As its the inital span there is no -1 needed

File: packages/jaeger-ui/src/components/TracePage/index.js (hunk 8 above)
Comment (copa2) [reply packages/jaeger-ui/src/components/TracePage/index.js]: Ok changed this to the proposed solution. Looks clearer.

**Comment by copa2** [comment]
@tiffon / @yurishkuro / @everett980 
Applied review changes. Please take a look if this ok. 
This does not contain the NodeDetail and different alternative views. We can add them with another story later. 
It currently does not contain an "Experimental" indicator. Not sure where to apply it on the view resp. where should I direct people when they want to give feedback. The issue and pull request are then closed. Should I open a new issue with "Enhancements to TraceGraph"?

**Comment by yurishkuro** [comment]
I find the toggle button a bit confusing. I would prefer that it displays the name of the current view, and clicking on it opens a menu to select another view (including JSON view).

> Should I open a new issue with "Enhancements to TraceGraph"?

Certainly let's keep future ideas recorded in a meta-ticket.

File: packages/jaeger-ui/src/components/TracePage/TraceGraph/OpNode.test.js [hunk 9]
```diff
@@ -0,0 +1,78 @@
...
+    expect(wrapper.find('.OpNode--avg').text()).toBe('40 ms');
+    expect(wrapper.find('.OpNode--selfTime').text()).toBe('180 ms (90 %)');
+    expect(wrapper.find('.OpNode--op').text()).toBe('op1');
+    expect(wrapper.find('.OpNode--service').text()).toBe('service1');
+  });
+
+  it('it switches mode', () => {
+    mode = MODE_TIME;
+    wrapper = shallow(<OpNode {...props} mode={mode} />);
+    expect(wrapper.find('.OpNode--mode-time').length).toBe(1);
```
everett980: this test can be made more robust by adding `expect(wrapper.find('.OpNode--mode-selftime').length).toBe(0);` above this line, and `expect(wrapper.find('.OpNode--mode-time').length).toBe(0);` above `expect(wrapper.find('.OpNode--mode-selftime').length).toBe(1);` (three lines down)

and I believe `expect(wrapper.find('.OpNode--mode-service').length).toBe(0);` should be added above each existing `expect` line.

packages/jaeger-ui/src/model/trace-dag/TraceDag.js
```diff
@@ -106,6 +106,7 @@ export default class TraceDag<T = void> {
     if (!skipToChild) {
       const node = this._getDagNode(service, operation, children.size > 0, parentNodeID, data);
       node.count++;
+      node.members.push(denseSpan);
```
everett980: Instead of having the `TraceDag` mutate a property on the `DagNode`, it would be cleaner if the `DagNode` class had a function to update `this.members`, then `TraceDag` could call `node.addMember(denseSpan)`

copa2:
Added experimental ribbon which links to #293 
experimental

packages/jaeger-ui/src/components/TracePage/TraceGraph/TraceGraph.js
```diff
@@ -0,0 +1,329 @@
...
+
+  layoutManager: LayoutManager;
+
+  constructor(props: Props) {
+    super(props);
+    this.state = {
+      showHelp: false,
+      mode: MODE_SERVICE,
+    };
+    this.modes = [MODE_SERVICE, MODE_TIME, MODE_SELFTIME];
```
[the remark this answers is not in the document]
copa2: Thanks. Forgot to remove these.

packages/jaeger-ui/src/components/TracePage/TraceGraph/OpNode.test.js
copa2: Applied these additional checks

packages/jaeger-ui/src/model/trace-dag/TraceDag.js
copa2: Good idea. Looks cleaner.

copa2:
@everett980: Thanks for your feedback. Applied your requested changes.

I will merge the commits into one commit when all reviews are ok.

Open points:
- https://github.com/jaegertracing/jaeger-ui/pull/276#discussion_r236442353. In my opinion it is needed. If there is no other feedback how solve it other I will mark it as resolved.
- Should the entrypoint for the TraceGraph be a dropdown <-> button. Can you @tiffon / @yurishkuro discuss this together. In can live with both solutions :-)

packages/jaeger-ui/src/components/TracePage/TraceGraph/OpNode.js
```diff
@@ -0,0 +1,129 @@
...
+  operation: string,
+  service: string,
+  mode: string,
+};
+
+export const MODE_SERVICE = 'service';
+export const MODE_TIME = 'time';
+export const MODE_SELFTIME = 'selftime';
+
+export function round2(percent: number) {
```
tiffon: [`Number#toFixed()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed) should do the same thing, since thjs is for formatting the percents.

packages/jaeger-ui/src/components/TracePage/TraceGraph/TraceGraph.js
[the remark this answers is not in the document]
tiffon: Great, thanks!

packages/jaeger-ui/src/components/TracePage/TraceGraph/TraceGraph.js
```diff
@@ -0,0 +1,328 @@
...
+          className="TraceGraph--dag"
+          minimapClassName="TraceGraph--miniMap"
+          layoutManager={this.layoutManager}
+          getNodeLabel={getNodeDrawer(mode)}
+          setOnRoot={classNameIsSmall}
+          setOnEdgePath={setOnEdgePath}
+          edges={ev.edges}
+          vertices={ev.vertices}
+        />
+        <div className="TraceGraph--experimental">
```
tiffon: Looks great. I don't think you need the wrapping `<div>`; the styles on the div can be applied to the `<a>`.

packages/jaeger-ui/src/components/TracePage/TraceGraph/TraceGraph.css
```diff
@@ -0,0 +1,114 @@
...
+
+.TraceGraph--menu > li {
+  list-style-type: none;
+  text-align: right;
+  padding-bottom: 0.3rem;
+}
+
+.TraceGraph--sidebar {
+  cursor: default;
+  box-shadow: -1px 0 #00000030;
```
tiffon: [20% of browsers](https://caniuse.com/#feat=css-rrggbbaa) don't support `#rrggbbaa` color format.

packages/jaeger-ui/src/components/TracePage/TraceGraph/TraceGraph.js
```diff
@@ -0,0 +1,328 @@
...
+          arrowScaleDampener={0}
+          className="TraceGraph--dag"
+          minimapClassName="TraceGraph--miniMap"
+          layoutManager={this.layoutManager}
+          getNodeLabel={getNodeDrawer(mode)}
+          setOnRoot={classNameIsSmall}
+          setOnEdgePath={setOnEdgePath}
+          edges={ev.edges}
+          vertices={ev.vertices}
+        />
```
tiffon: Btw, the following function has a nice effect on the edge coloring; it reduces the opacity of the edges, when zoomed out, so they don't look heavy compared to the nodes. 

https://github.com/jaegertracing/jaeger-ui/blob/f51dfd7e1eeafc11ece483c5d74fc416dac04d53/packages/jaeger-ui/src/components/TraceDiff/TraceDiffGraph/TraceDiffGraph.js#L38-L46

https://github.com/jaegertracing/jaeger-ui/blob/f51dfd7e1eeafc11ece483c5d74fc416dac04d53/packages/jaeger-ui/src/components/TraceDiff/TraceDiffGraph/TraceDiffGraph.js#L111

With the above:
screen shot 2018-12-17 at 3 16 59 pm


With black edges:
screen shot 2018-12-17 at 3 16 53 pm

packages/jaeger-ui/src/components/TracePage/TraceGraph/OpNode.js
copa2: As I understand toFixed(2) will always add 2 decimal places even when not needed.
So 40 % will render 40.00 %. Do you really want this?

packages/jaeger-ui/src/components/TracePage/TraceGraph/TraceGraph.js
```diff
@@ -0,0 +1,328 @@
...
+    if (errorTag) {
+      return errorTag.value;
+    }
+  }
+  return false;
+}
+
+function helpContent() {
+  return (
+    <div className="TraceGraph--help-content">
```
tiffon: Since this isn't dynamic, it can be a const.

packages/jaeger-ui/src/model/trace-dag/DagNode.js
```diff
@@ -36,7 +37,12 @@ export default class DagNode<T = void> {
     this.parentID = parentID;
     this.id = DagNode.getID(service, operation, hasChildren, parentID);
     this.count = 0;
+    this.members = [];
```
tiffon: Per @everett980's comment, `.count` can be turned into a getter to avoid having to maintain it:

```js
set count(value) {
  throw new Error('Illegal operation, use .addMember(...)');
}
get count() {
  return this.members.length;
}
```

packages/jaeger-ui/src/components/TracePage/TraceGraph/TraceGraph.js
tiffon: LMK if you don't want to add this, I'll add it to the ticket you created for enhancements.

tiffon:
>   So if you are ok with this new solution. I will commit this.

The gif you showed looks great! Sorry for the mixed signals!

copa2:
> Sorry for the mixed signals!

No problem. UIs are always a matter of taste :-) More discussion normaly bring better results.

Open now:
- toFixed vs round2
- count in DagNode

Will merge and rebase to master when when reviews are through.

packages/jaeger-ui/src/model/trace-dag/DagNode.js
tiffon: Great point. We can leave dealing with that for another day.

packages/jaeger-ui/src/components/TracePage/TraceGraph/OpNode.js
tiffon: You're right. I'd say the only downside of `40 %` is that it will look smaller than most of the values, e.g. `12.34 %`, at first glance. I'd suggest changing to `value.toFixed(1)` but seems fine either way as it was a nit to start with 😄

copa2: Will keep the current implementation. Self time in leafs are always `100 %` so this would look like `100.0 %` or `100.00 %`.

copa2:
Rebased and squashed commits.
 
@everett980: Let me know if you want further changes. Otherwise we should release this and apply improvements with #293. 

:laughing:  Altered Beast - my first game I remember was [Galaga](http://www.free80sarcade.com/galaga.php)

PR #278, opened 2018-11-17
Ability to open additional menu links in same tab (Resolves #275)

### LINKED ISSUE #275: Open additional menu links in the same tab
### tl;dr
Possibility to open [additional menu links in global nav](https://www.jaegertracing.io/docs/1.8/deployment/#UIConfiguration) in the same tab (`target="_self"`)

### Detailed use case

I'm using [keycloak gatekeeper](https://github.com/keycloak/keycloak-gatekeeper) to add login page to jaeger query and I want to add "Logout" link in top nav which will point to `/oauth/logout` which will end keycloack session and redirect to login page.

In current state this will open new tab, perform logout and redirect to login screen, however this all happens in a new tab and jaeger ui tab is still open. I need this all to happen in the same tab. 

Hardcoded `target="_blank"` in [here](https://github.com/jaegertracing/jaeger-ui/blob/feeaccd1502b2748b96ca3619e9c3dcb02bc0ed3/packages/jaeger-ui/src/components/App/TopNav.js#L62)

Add `openInSameTab` boolean field to [ConfigMenuItem type](https://github.com/jaegertracing/jaeger-ui/blob/feeaccd1502b2748b96ca3619e9c3dcb02bc0ed3/packages/jaeger-ui/src/types/config.js#L17) which, if set to true will set link target to ___self__

zablvit:
Signed-off-by: Vitaliy Zabolotskyy <zablvit@gmail.com>

Resolves #275 
Adds option to ui-config, so added link can be opened in the same tab 

Added corresponding boolean field to ConfigMenuItem and setting link target accordingly.


packages/jaeger-ui/src/components/App/TopNav.test.js
```diff
@@ -30,6 +30,7 @@ describe('<TopNav>', () => {
     {
       label: 'Twitter',
       url: 'https://twitter.com/JaegerTracing',
+      openInSameTab: true,
```
yurishkuro: does this make a difference?

packages/jaeger-ui/src/types/config.js
```diff
@@ -17,6 +17,7 @@
 export type ConfigMenuItem = {
   label: string,
   url: string,
+  openInSameTab: boolean,
```
yurishkuro: is there a way to indicate this as optional field?

packages/jaeger-ui/src/components/App/TopNav.test.js
```diff
@@ -87,5 +89,11 @@ describe('<TopNav>', () => {
       expect(item.prop('label')).toBe(labelAbout);
       expect(item.prop('items')).toBe(dropdownItems);
     });
+
+    it('adds target=_self to top-level item', () => {
+        const item = wrapper.find(`[href="${githubUrl}"]`);
+        expect(item.length).toBe(1);
+        expect(item.find(`[target="_self"]`).length).toBe(1);
```
yurishkuro: could we also have a negative test, i.e. if `openInSameTab` not specified, then `target=_blank`

packages/jaeger-ui/src/components/App/TopNav.js
```diff
@@ -93,9 +94,10 @@ export function TopNavImpl(props: Props) {
...
           }
           const item = ((m: any): ConfigMenuItem);
+          const target = item.openInSameTab ? "_self" : "_blank";
           return (
             <Menu.Item key={item.label}>
-              <a href={item.url} target="_blank" rel="noopener noreferrer">
+              <a href={item.url} target={target} rel="noopener noreferrer">
                 {item.label}
               </a>
             </Menu.Item>
```
yurishkuro: can this be DRY-ed with the above code into a helper function?

packages/jaeger-ui/src/components/App/TopNav.test.js
zablvit: In `renders the nested menu items` test, there is assertion if this is passed further.
I tried to make a test for rendered dropdown, but I just don't know, how to do it.

packages/jaeger-ui/src/types/config.js
zablvit: I'm not sure. Currently all these fields are optional (e.g you can specify only `url`).
Maybe some additional checks, but I don't know...
Anyway it should be stated in documentation.

packages/jaeger-ui/src/components/App/TopNav.test.js
zablvit: You're right, will add this soon.

packages/jaeger-ui/src/components/App/TopNav.js
zablvit: Ok, makes sense.

packages/jaeger-ui/src/components/App/TopNav.test.js
zablvit: Done

packages/jaeger-ui/src/components/App/TopNav.js
zablvit: Done

packages/jaeger-ui/src/components/App/TopNav.js
```diff
@@ -52,16 +52,23 @@ if (getConfigValue('dependencies.menuEnabled')) {
   });
 }
 
+function getItemLink(item: ConfigMenuItem) {
```
tiffon: Seems like this can also add the wrapping `<Menu.Item>`?

packages/jaeger-ui/src/types/config.js
tiffon: I'd suggest renaming to:
```
openInSelf?: boolean,
```

What do you think?

FYI, the name's trailing `?` indicates the field is optional.

yurishkuro: Fwiw I think SameTab is more self describing then "self". Target may be even better, direct match to HTML attribute.

tiffon: I like `target` as well, and actually suggested that and then edited the comment. It would be more flexible and simplify the implementation, a bit, as well.

zablvit: I think openInSameTab is better than target because of more control over input - only true/false.
On the other hand target attribute would indeed add a bit of flexibility.
You decide :)

tiffon: Let's go with `target?: string,`

packages/jaeger-ui/src/components/App/TopNav.test.js
```diff
@@ -87,5 +96,32 @@ describe('<TopNav>', () => {
...
+      expect(item.length).toBe(1);
+      expect(item.find(`[target="_blank"]`).length).toBe(1);
+    });
+
+    describe('<CustomNavDropdown>', () => {
+      beforeEach(() => {
+        wrapper = shallow(<TopNav.CustomNavDropdown {...configMenuGroup} />);
+      });
+
+      it('renders sub-memu items', () => {
```
everett980: sub-menu *

Issue #282, opened 2018-11-28
[embed] Scatter plot links to trace lose embed mode when clicked


tiffon:
## Requirement - what kind of business use case are you trying to solve?

When in embed mode on the search page, clicking a dot from the scatter plot should retain embed mode.

## Problem - what in Jaeger blocks you from solving the requirement?

When in embed mode on the search page, clicking a dot from the scatter plot loses retain embed mode, currently.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

Retain embed mode when viewing the trace.

## Any open questions to address

n/a

tiffon:
cc @aljesusg

aljesusg:
We can close this issue ;) Thanks @tiffon

tiffon:
Right. :+1:

If you add `Fixes #N.` to the PR it will close automatically.

aljesusg:
I forgot I added #N to the commit but not in description my bad ^^. Thanks @tiffon

PR #283, opened 2018-11-28
Fix link in scatter plot when embed mode #282

aljesusg:
#282 

Fix link in the scatter plot to the trace with embed mode

fix_scatter_plot


aljesusg:
@tiffon it's ready the fix :)

PR #286, opened 2018-11-30
Revisions for search and trace detail embed mode

### LINKED ISSUE #279: [embed] Enable configuration of any view when in embed mode

Enhancement to embed mode, e.g. query param `embed=...`.

Currently (in #263), when embedding a search view, it's not possible to customize the embedded view of trace-details. For instance, if the query param `mapCollapsed` is added to the embedded search page URL it will not be carried over to the trace detail view when a trace is selected from the search results.

Propagate embed query params when changing views.

## Any open questions to address

Seems like this will be a breaking change so `embed=vN+1` should be used to scope the effects of this change, where `N` is the current embed version.

### LINKED ISSUE #288: [Search] Multiple searches can cause the wrong results to show

Show search results for the last executed search, only.

When multiple searches are executed, they are shown in the order they finish. The last search to finish is the result set that shows when all is said and done, regardless of whether or not it was the last requested search. This can not only show results from a former search but also results that don't match the search form (if it was modified for a subsequent search).

Only show the last request search results. E.g. if multiple searches are executed, do not show any results except the last.

tiffon:
Tweaks to the search page and trace timeline embedded view. Mostly from prior comments.

- Rename query parameter for embedding to start with `ui` and use the
  page as the first word, e.g. `uiSearchHideGraph`

- Change query parameters for the minimap and trace details from hiding
  to showing, e.g. `hidemap` -> `uiTimelineShowMap`

- Require query parameters to configure embedded to have value=1

- Save the embed query params in Redux state instead of passing them
  around. Fixes #279.

- Use a `Link` with an icon instead of text buttons for opening the
  standalone view of the page

- Propagate whether the trace detail page is from the search page or
  not via the `Location#state` member on the React Router` Location`

- When returning to the search page use the previous results instead
  of executing a new search. This is done by storing the query with
  the search results.

- Adjusted aesthetic of "Back to Search" button on trace detail page

- Sequester parsing and stripping query parameters for the embed mode
  to a util

- In various places switch to using the `component/*/url.js#getUrl`
  functions instead of `prefixUrl(...)`

- Rename various component props

- Keep redux search query synced with redux search result (and their
  processing). Also fixes #288.

### TODO

- [x] Bolster test coverage

### Screenshots

screen shot 2018-11-30 at 3 31 52 am
screen shot 2018-11-30 at 3 32 14 am
screen shot 2018-11-30 at 3 33 33 am
screen shot 2018-11-30 at 3 34 00 am
screen shot 2018-11-30 at 3 34 08 am
screen shot 2018-11-30 at 3 34 17 am
screen shot 2018-11-30 at 3 34 56 am

cc @aljesusg


packages/jaeger-ui/src/utils/embedded-url.js
```diff
@@ -0,0 +1,62 @@
...
+
+const getStrings = value => (typeof value === 'string' ? value : _flatMap(value, getStrings));
+
+const VALUE_ENABLED = '1';
+export const VERSION_0 = 'v0';
+
+// uiEmbed=v0
+// uiSearchHideGraph=1
+// uiTimelineHideMap
+// uiTimelineHideDetails
```
aljesusg: // uiTimelineShowMap=1 to enable
// uiTimelineShowDetails=1 to enable

packages/jaeger-ui/src/components/App/Page.js
```diff
@@ -52,26 +56,29 @@ export class PageImpl extends React.Component<Props> {
...
         </Layout>
       </div>
     );
   }
 }
 
 // export for tests
-export function mapStateToProps(state: { router: { location: Location } }) {
+export function mapStateToProps(state: ReduxState) {
+  const { embedded } = state;
```
everett980: couldn't this `embedded` flag come from router? or are there other ways of enabling beyond query params?

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.js
```diff
@@ -37,11 +38,11 @@ type SearchResultsProps = {
   cohortAddTrace: string => void,
   cohortRemoveTrace: string => void,
   diffCohort: FetchedTrace[],
+  disableComparisions: boolean,
```
everett980: typo? `disableComparisons` instead of `disableComparisions`

packages/jaeger-ui/src/components/App/Page.js
aljesusg: I tested this and you can pass uiEmbed=v0 in query parameters.

aljesusg:
Hi @tiffon , I am working in one of the use of case with this...I wonder if for example in this gif if user navigate from search results to the trave view (inside an iframe) he can't change preferences to show details or minimap. What do you think about to add some buttons in the UI to enable/disable this too? 

peek 2018-12-10 16-16

tiffon:
@aljesusg Thanks for the suggestion! How about we default the minimap and details to collapsed but the user can expand the trace name to show them. This would be the default. We can also have a mode where the embed params can either disable showing the minimap and details, entirely, (each independently) or have them showing, by default?

aljesusg:
Maybe we should show them by default, anyway from searchtraces page you can pass the query Params related with trace pages like uiShowDetails and the UI propagate this param to the view when the user click a trace in the result.

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.js
```diff
@@ -145,22 +146,20 @@ export default class SearchResults extends React.PureComponent<SearchResultsProp
...
-                    target="_blank"
-                    href={getSearchURL()}
-                  >
-                    View Results
-                  </Button>
-                </label>
-              )}
-              <h2 className="ub-m0">
+              <h2 className="ub-m0 u-flex-1">
                 {traces.length} Trace{traces.length > 1 && 's'}
```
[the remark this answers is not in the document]
tiffon: It's possible to have zero traces, but this line wouldn't be reached, in that case, because a "No results" type of message is shown, instead.

tiffon:
@aljesusg Thanks for the suggestions!

@everett980 Thanks for the review!

I've made the following changes to the embedded mode:

- The TracePageHeader is collapsible when embedded, and it defaults to open
  - `uiTimelineCollapseTitle=1` will default it to collapsed
- The summary data (number of services, etc.) and minimap are now shown, by default
  - `uiTimelineHideMinimap=1` will hide the minimap
  - `uiTimelineHideSummary=1` will hide the trace summary
  - If both are hidden, the title is no longer collapsible because there is nothing to show
- I've restyled the `TracePageHeader`; I thought it looked more organized than the current, non-embedded header, so I consolidated the two classes

Here are a few screenshots. The search page hasn't changed.

#### Normal, header is open
ss-00-normal-open


#### Normal, header is collapsed
ss-01-normal-collapsed


#### Normal with back button to return to search results, header is open
ss-02-normal-with-back-open


#### Normal with back button, header is collapsed
ss-03-normal-with-back-collapsed


#### Embedded with back button, header is open
ss-04-embedded-with-back-open


#### Embedded with back button, header is collapsed
ss-05-embedded-with-back-collapsed



#### Embedded with back button, summary is hidden
ss-06-embedded-with-back-hide-summary



#### Embedded with back button, minimap is hidden
ss-07-embedded-with-back-hide-map




#### Embedded with back button, both summary and minimap are hidden
ss-08-embedded-with-back-hide-both



#### Embedded without a back button, collapsed
ss-09-embedded-no-back-collapsed

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.js
tiffon: Thanks!

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.js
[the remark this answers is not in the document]
tiffon: Marking as resolved since this view will not be shown when there are 0 traces.

packages/jaeger-ui/src/components/TracePage/index.test.js
```diff
@@ -153,21 +152,38 @@ describe('<TracePage>', () => {
     expect(cancelScroll.mock.calls).toEqual([[]]);
   });
 
-  it('no render TracePageHeader if queryparam embed', () => {
-    wrapper.setProps({ embed: true });
-    expect(wrapper.find(TracePageHeader).length).toBe(0);
-  });
+  describe('embed mode', () => {
```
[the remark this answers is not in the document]
tiffon: Moved the test. :+1:

packages/jaeger-ui/src/components/TracePage/index.test.js
```diff
@@ -153,21 +152,38 @@ describe('<TracePage>', () => {
...
+        },
+      };
+      wrapper.setProps({ embedded });
+    });
 
-  it('collapse map if queryparam mapCollapsed', () => {
-    wrapper.setProps({ mapCollapsed: true, embed: true });
-    expect(wrapper.find(SpanGraph).length).toBe(0);
-  });
+    it('no render TracePageHeader if queryparam embed', () => {
```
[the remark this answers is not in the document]
tiffon: I cleaned up the tests, generally.

packages/jaeger-ui/src/components/TracePage/index.test.js
```diff
@@ -153,21 +152,38 @@ describe('<TracePage>', () => {
...
+    it('collapses the minimap by default', () => {
+      expect(wrapper.find(SpanGraph).length).toBe(0);
+    });
 
-  it('open a window when goFullView is called', () => {
-    wrapper.setProps({ id: '12345' });
-    global.open = jest.fn();
-    wrapper.instance().goFullView();
-    expect(global.open).toBeCalledWith(prefixUrl('/trace/12345'), '_blank');
+    it('shows the minimap when the embed parameter is true', () => {
```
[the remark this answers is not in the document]
tiffon: Moved the map into the header.

packages/jaeger-ui/src/components/TracePage/index.test.js
```diff
@@ -153,21 +152,38 @@ describe('<TracePage>', () => {
...
+      wrapper.setProps({
+        embedded: { timeline: { showMap: true, showDetails: false } },
+      });
+      expect(wrapper.find(SpanGraph).length).toBe(1);
+    });
+
+    it('open a window when goFullView is called', () => {
```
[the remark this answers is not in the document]
tiffon: Moved into the header and re-phrased.

aljesusg:
Hi @tiffon , I tested it, nice refactor and enhancement. Thanks !!!

packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageHeader.test.js
```diff
@@ -0,0 +1,123 @@
...
+  });
+
+  it('renders a <SpanGraph>', () => {
+    expect(wrapper.find(SpanGraph).length).toBe(1);
+  });
+
+  describe('observes the visibility toggles for various UX elements', () => {
+    it('hides the minimap when hideMap === true', () => {
+      wrapper.setProps({ hideMap: true });
+      expect(wrapper.find(SpanGraph).length).toBe(0);
```
everett980: this test and the following do not confirm that the prop in the test hides the UX element, just that it does not exist. before each `setProps` line, there should be a presence check (for this test `expect(wrapper.find(SpanGraph).length).toBe(1);`)

packages/jaeger-ui/src/components/TracePage/index.js
```diff
@@ -357,35 +337,32 @@ export class TracePageImpl extends React.PureComponent<TracePageProps, TracePage
...
-    };
-
-    const tracePageEmbedProps = {
-      searchParams,
-      fromSearch,
-      enableDetails,
-      onGoFullViewClicked: this.goFullView,
+      resultCount: findMatchesIDs ? findMatchesIDs.size : 0,
+      showArchiveButton: !isEmbedded && archiveEnabled,
+      showShortcutsHelp: !isEmbedded,
```
everett980: to follow the idea of versioned embedding, should these properties be part of the embedded state so that future versions of embedded could affect these separately? i.e.: `showShortCutsHelp = embedded.showShortCutsHelp,`

packages/jaeger-ui/src/reducers/trace.test.js
```diff
@@ -18,56 +18,194 @@ import traceGenerator from '../demo/trace-generators';
...
+        ...traces,
+        [traceID]: { id: traceID, error: expect.any(Error), state: fetchedState.ERROR },
+      };
+      expect(state.traces).toEqual(outcome);
+    });
+  });
+
+  it('handles a failed request', () => {
+    const error = 'error-info';
+    const traces = { preExisting: Math.random() };
```
everett980: here and other places:
why use `Math.random()` as "filler data"? It seems as though it'd be clearer to simply write `{ preExistingID: 'preExistingTraceDate' }`

packages/jaeger-ui/src/components/TracePage/index.test.js
```diff
@@ -369,10 +347,18 @@ describe('mapDispatchToProps()', () => {
 });
 
 describe('mapStateToProps()', () => {
-  it('maps state to props correctly', () => {
-    const id = 'abc';
-    const trace = {};
-    const state = {
+  const id = 'abc';
```
everett980: clearer variable names and values in such a large file could help. since this is a traceID (I believe), the variable name and value could be traceID. It interferes with object shorthand below but is clearer if there are errors.

packages/jaeger-ui/src/components/TracePage/index.test.js
```diff
@@ -369,10 +347,18 @@ describe('mapDispatchToProps()', () => {
...
+  const trace = {};
+  const embedded = Math.random();
```
everett980: I mentioned this below as well, why use `Math.random()`, an equally valid, but easier to debug test value could be `const embedded = 'truthy embedded value';`

packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageHeader.test.js
tiffon: Thanks

Issue #215, opened 2018-06-04
Reset Viewing Layer zoom 


yuribit:
It would be nice to have a way to reset the selection on the Viewing Layer in the Trace Page Header. 
The easiest way to do that could be a "Reset Selection" button right above the Viewing Layer. 

Something like this:
![example](https://user-images.githubusercontent.com/1260900/41007755-22fa9000-68dc-11e8-8894-305525fc1ca1.gif)

What do you think?

tiffon:
@yuribit Thanks for creating this ticket!

Great suggestion. Maybe it can show on mouse-over of the minimap, just below the bottom of the map. That way it won't obscure image or other zoom controls.

tiffon:
This was fixed in #290 (although there is a regression where the minimap dragging is not working, at the moment).

PR #290, opened 2018-12-11
Add a Button to Reset Viewing Layer Zoom (#215)

everett980:
- Issue #215 - Reset Viewing Layer Zoom

- Renders a button when either scrubber is not at the appropriate extreme.
- Button is hidden unless user mouses over the timeline.
- On click, resets viewing layer zoom back to fully zoomed out, both scrubbers back to appropriate extreme.

reset selection webm


packages/jaeger-ui/src/components/TracePage/SpanGraph/ViewingLayer.css
```diff
@@ -62,3 +62,15 @@ limitations under the License.
   top: 0;
   user-select: none;
 }
+
+.viewRangeTimeResetButton {
```
tiffon: Please follow the `ViewingLayer--subComponent` naming pattern.

packages/jaeger-ui/src/components/TracePage/SpanGraph/ViewingLayer.js
```diff
@@ -16,6 +16,7 @@
 
 import * as React from 'react';
 import cx from 'classnames';
+import { Button } from 'antd';
```
tiffon: Please sort the imports on the string portion.

packages/jaeger-ui/src/components/TracePage/SpanGraph/ViewingLayer.js
```diff
@@ -224,7 +225,18 @@ export default class ViewingLayer extends React.PureComponent<ViewingLayerProps,
   };
 
   /**
-   * Randers the difference between where the drag started and the current
+   * Resets the zoom to fully zoomed out.
+   *
+   * @param {Syntheticevent<HTMLButtonElement>} event - Click event created by clicking on button.
+   */
+  _resetTimeZoomClickHandler = (event: SyntheticEvent<HTMLButtonElement>) => {
+    event.stopPropagation();
```
tiffon: Are these (stopProp, preventDefault) still necessary if the button is on top of the SVG instead of under it?

packages/jaeger-ui/src/components/TracePage/SpanGraph/ViewingLayer.test.js
```diff
@@ -255,6 +255,37 @@ describe('<SpanGraph>', () => {
         });
       });
     });
+
+    describe('.viewRangeTimeResetButton', () => {
+      it('should render viewRangeResetButton if props.viewRange.time.current[0] !== 0', () => {
```
tiffon: Can you add a test to verify it's not rendered when current = [0, 1] ?

packages/jaeger-ui/src/components/TracePage/SpanGraph/ViewingLayer.js
everett980: The handlers do not seem necessary. I am accustomed to being fairly restrictive about preventing event bubbling which may not necessary here. I have removed these two function calls since it behaves the same without them.

packages/jaeger-ui/src/components/TracePage/SpanGraph/ViewingLayer.css
everett980: will do. I also added a note for the future style guide.

packages/jaeger-ui/src/components/TracePage/SpanGraph/ViewingLayer.js
everett980: sorted. I also added a note for this for the future style guide. I'm guessing our sort is case-insensitive and `.` comes before `[a-zA-Z]`?

packages/jaeger-ui/src/components/TracePage/SpanGraph/ViewingLayer.css
everett980: renamed the class and a related variable

packages/jaeger-ui/src/components/TracePage/SpanGraph/ViewingLayer.js
tiffon: More info about the sort of imports: https://github.com/jaegertracing/jaeger-ui/blob/style-guide-wip/CONTRIBUTING.md#imports

And, I forgot to mention (doh!) please have `import * as React from ...` as the first import.

Issue #244, opened 2018-09-04
Validate duration fields in UI


objectiser:
## Requirement - what kind of business use case are you trying to solve?

Validate the duration fields before allowing the search

## Problem - what in Jaeger blocks you from solving the requirement?

Currently an invalid duration field can be specified and sent to the backend, resulting in an error in the main area.

![jaeger-invalid-min-duration](https://user-images.githubusercontent.com/164562/45026859-90925880-b037-11e8-9bb2-68a7737b58ce.png)

## Proposal - what do you suggest to solve the problem or improve the existing situation?

`Find Traces` button only enabled if all of the fields are valid - so when user enters content into the min/max duration fields it must be valid before the search button is re-enabled.

everett980:
I didn't use the correct syntax in the PR that addressed this, closing now. Please re-open if the issue persists.

PR #291, opened 2018-12-12
Add popover and prevent submit if duration params are invalid (#244)

everett980:
- Duration fields without units result in an error, this PR adds popovers when these fields lack units and prevents submit (#244 )

If the user times in an invalid duration, a popover is displayed to indicate that units are necessary. The submit button is disabled if any fields are invalid. To avoid jumping on the user as soon as the start typing, the popover is not visible until the user has blurred. After the user has blurred, the popover is visible as long as the field is invalid, even if the field becomes focused again.

244 invalid popover

### DIFF
--- a/packages/jaeger-ui/src/components/SearchTracePage/SearchForm.js
+++ b/packages/jaeger-ui/src/components/SearchTracePage/SearchForm.js
@@ -38,9 +38,12 @@ import './SearchForm.css';
 const FormItem = Form.Item;
 const Option = Select.Option;
 
-const AdaptedInput = reduxFormFieldAdapter(Input);
-const AdaptedSelect = reduxFormFieldAdapter(Select);
-const AdaptedVirtualSelect = reduxFormFieldAdapter(VirtSelect, option => (option ? option.value : null));
+const AdaptedInput = reduxFormFieldAdapter({ AntInputComponent: Input, isValidatedInput: true });
+const AdaptedSelect = reduxFormFieldAdapter({ AntInputComponent: Select });
+const AdaptedVirtualSelect = reduxFormFieldAdapter({
+  AntInputComponent: VirtSelect,
+  onChangeAdapter: option => (option ? option.value : null),
+});
 
 export function getUnixTimeStampInMSFromForm({ startDate, startDateTime, endDate, endDateTime }) {
   const start = `${startDate} ${startDateTime}`;
--- a/packages/jaeger-ui/src/components/SearchTracePage/SearchForm.js
+++ b/packages/jaeger-ui/src/components/SearchTracePage/SearchForm.js
@@ -74,6 +74,17 @@ export function traceIDsToQuery(traceIDs) {
   return traceIDs.split(',');
 }
 
+export const placeholderDurationFields = 'e.g. 1.2s, 100ms, 500us';
+export function validateDurationFields(value) {
+  if (!value) return undefined;
+  return /\d[\d\\.]*(us|ms|s|m|h)$/.test(value)
+    ? undefined
+    : {
+        content: `Please enter a number followed by a duration unit, ${placeholderDurationFields}`,
+        title: 'Please match the requested format.',
+      };
+}
+
 export function convertQueryParamsToFormDates({ start, end }) {
   let queryStartDate;
   let queryStartDateTime;
@@ -155,6 +166,7 @@ export class SearchFormImpl extends React.PureComponent {
   render() {
     const {
       handleSubmit,
+      invalid,
       selectedLookback,
       selectedService = '-',
       services,
@@ -323,13 +335,20 @@ export class SearchFormImpl extends React.PureComponent {
           <Field
             name="minDuration"
             component={AdaptedInput}
-            placeholder="e.g. 1.2s, 100ms, 500us"
+            placeholder={placeholderDurationFields}
             props={{ disabled }}
+            validate={validateDurationFields}
           />
         </FormItem>
 
         <FormItem label="Max Duration">
-          <Field name="maxDuration" component={AdaptedInput} placeholder="e.g. 1.1s" props={{ disabled }} />
+          <Field
+            name="maxDuration"
+            component={AdaptedInput}
+            placeholder={placeholderDurationFields}
+            props={{ disabled }}
+            validate={validateDurationFields}
+          />
         </FormItem>
 
         <FormItem label="Limit Results">
@@ -342,7 +361,11 @@ export class SearchFormImpl extends React.PureComponent {
           />
         </FormItem>
 
-        <Button htmlType="submit" disabled={disabled || noSelectedService} data-test={markers.SUBMIT_BTN}>
+        <Button
+          htmlType="submit"
+          disabled={disabled || noSelectedService || invalid}
+          data-test={markers.SUBMIT_BTN}
+        >
           Find Traces
         </Button>
       </Form>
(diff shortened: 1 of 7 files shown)


yurishkuro:
I would rather display an error in the UI than try to infer which units the user had in mind by typing just a number. E.g. `please specify units, e.g. "ms"`.

tiffon:
Per @yurishkuro comment, we can add a `pattern` attribute and use the `title` attribute to indicate a message.

```
pattern="\d[\d\.]*(us|ms|s|m|h)"
title="Enter a number followed by a duration unit, e.g. 20us, 30ms, 1.1s"
```

screen shot 2018-12-17 at 1 34 59 pm

The message only shows up if the user tries to submit the form while the field is invalid.

To take it a bit further, the `:invalid` pseudo-class can be used to style the input as red-ish.

screen shot 2018-12-17 at 1 36 31 pm

packages/jaeger-ui/src/components/SearchTracePage/SearchForm.js
```diff
@@ -98,6 +99,11 @@ export function convertQueryParamsToFormDates({ start, end }) {
   };
 }
 
+function addDefaultUnitToDuration(duration) {
+  if (!duration) return null;
+  return /[0-9]$/.test(duration) ? `${duration}ms` : duration;
```
tiffon: Side note: `1.1` would fail this regex.

packages/jaeger-ui/src/utils/redux-form-field-adapter.css
```diff
@@ -0,0 +1,20 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+input.ant-input.is-invalid {
+  border: rgba(255, 0, 0, 0.5) 1px solid;
+  background-color: rgba(255, 0, 0, 0.1);
+}
```
tiffon: Would suggest creating a `.SearchForm--validatedInput.is-invalid` style for this instead of overriding `.ant-input`, directly.

Also, this has a very strong aesthetic. What do you think of the following?

```
background: #fff9f8;
border: 1px solid #c00;
```

packages/jaeger-ui/src/utils/redux-form-field-adapter.js
```diff
@@ -14,22 +14,31 @@
...
+import './redux-form-field-adapter.css';
+
 export default function reduxFormFieldAdapter(
   AntInputComponent: Class<React.Component<*, *>>,
   onChangeAdapter: () => void
 ) {
   return function _reduxFormFieldAdapter(props: any) {
-    const { input: { value, onChange }, children, ...rest } = props;
+    // inputRest includes necessary props such as onBlur and value
+    const { input: { onChange, ...inputRest }, children, ...rest } = props;
```
tiffon: If you destruct `className` from `props`, you can add it to the `cx(...)` call. This will allow for the `.SearchForm--validatedInput` class to be added and avoid overriding `.ant-input` styles.

Also, for accessibility to folks who don't work primarily in JS, we've been using `Boolean(value)` instead of `!!value` to cast to a boolean.

Seems like it might be useful to have `isInvalid` only `true` when [`!rest.meta.active`](https://redux-form.com/7.0.3/docs/api/field.md/#-code-meta-active-boolean-code-) ?

packages/jaeger-ui/src/utils/redux-form-field-adapter.js
```diff
@@ -14,22 +14,31 @@
...
+    const isInvalid = rest.meta.touched && !!rest.meta.error;
     return (
-      <AntInputComponent
-        value={value}
-        onChange={onChangeAdapter ? (...args) => onChange(onChangeAdapter(...args)) : onChange}
-        {...rest}
-      >
-        {children}
-      </AntInputComponent>
+      <Popover placement={'bottomLeft'} visible={isInvalid} {...rest.meta.error}>
```
tiffon: Maybe a parameter can be added to `reduxFormFieldAdapter` for whether or not the `<Popover />` should be used? When custom time-range is selected, six input fields are not using it but it's still be rendered.

```js
export default function reduxFormFieldAdapter(
   AntInputComponent: Class<React.Component<*, *>>,
   onChangeAdapter: () => void,
   isValidatedInput?: boolean,
 ) {
```

packages/jaeger-ui/src/utils/redux-form-field-adapter.js
everett980: I'm fine with using `Boolean(value)` instead of `!!value`, but I was basing it off of [airbnb's stye guide regarding 'no new'](https://github.com/airbnb/javascript#coercion--booleans) we should make a note in our style guide if we want to deviate from airbnb.

tiffon: Agreed. [This is what I have](https://github.com/jaegertracing/jaeger-ui/blob/style-guide-wip/CONTRIBUTING.md#code-style) so far for style-guide content. We should merge it with what you're working on.

packages/jaeger-ui/src/utils/redux-form-field-adapter.css
everett980: Updated, this is far less aggressive.

packages/jaeger-ui/src/utils/redux-form-field-adapter.js
everett980: added a parameter which also fixed an issue with the dropdown

packages/jaeger-ui/src/components/SearchTracePage/SearchForm.test.js
```diff
@@ -275,6 +276,36 @@ describe('<SearchForm>', () => {
...
+  it('should return `undefined` if the value is falsy', () => {
+    expect(validateDurationFields('')).toBeUndefined();
+    expect(validateDurationFields(null)).toBeUndefined();
+    expect(validateDurationFields(undefined)).toBeUndefined();
+  });
+
+  it('should return Popover-compliant error object if the value is a populated string that does not adhere to expected format', () => {
+    expect(validateDurationFields('100')).toEqual({
+      content: 'Please enter a number followed by a duration unit, e.g. 1.2s, 100ms, 500us',
+      title: 'Please match the requested format.',
```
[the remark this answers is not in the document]
tiffon: Just a thought. I would probably prefer looser, but I think it's a personal preference.

packages/jaeger-ui/src/components/SearchTracePage/SearchForm.js
```diff
@@ -38,9 +38,12 @@ import './SearchForm.css';
 const FormItem = Form.Item;
 const Option = Select.Option;
 
-const AdaptedInput = reduxFormFieldAdapter(Input);
-const AdaptedSelect = reduxFormFieldAdapter(Select);
-const AdaptedVirtualSelect = reduxFormFieldAdapter(VirtSelect, option => (option ? option.value : null));
+const AdaptedInput = reduxFormFieldAdapter({ AntInputComponent: Input, isValidatedInput: true });
```
tiffon: Would suggest breaking this out into two wrapped components, one with `isValidatedInput: true` and one without.

everett980: yup! meant to do that all along. updated!

Issue #204, opened 2018-04-10
A way to copy json text of json-formatted tag values


mabn:
The way json values are displayed is pretty cool and readable (yay), but sometimes I want to copy such value and use it somewhere else. E.g. in case of request payloads or message content to replay them in some other tool.

The formatting removes quotes from field names so I have to either add them manually or find the relevant value in trace json and "unescape" it.

It would be cool if there was some way to either disable the formatting temporarily for a given field or to somehow copy the raw value to the clipboard.

yurishkuro:
I've think a copy icon would be useful here

tiffon:
@mabn Thanks for creating this ticket!

SGTM. The call-to-action can appear on hover of the table row. IMO the text "Copy" is more clear than an icon.

PR #292, opened 2018-12-13
Add a copy icon to entries in KeyValuesTable (#204)

everett980:
- Tags are human-readable ( 🎉), but harder to paste into other tools.

- Add icon to rows in KeyValueTable to copy that row's data.

204 copy tags with tooltip
(again, my converted gifs have a weird background color issue, looks fine in the browser and .webm recording)

Copied JSON:
```
{
  "key": "sql.query",
  "type": "string",
  "value": "SELECT * FROM customer WHERE customer_id=731"
}
```

I went with the copy icon over the word copy since I believe it is fairly ubiquitous, but open to changing.


yurishkuro:
> I don't love that there is no "copy confirmation" so I welcome advice on how to indicate that to the user.

the copy icon can be animated to transform to a green checkbox.

Q: in what format is the tag being copied? I.e. how does the above example look after pasting it as text?

everett980:
> Q: in what format is the tag being copied? I.e. how does the above example look after pasting it as text?

@yurishkuro the `JSON.stringify(row, null, 2)` looks like the following:
```
{
  "key": "hostname",
  "type": "string",
  "value": "6376c38cb700"
}
```
The `null, 2` arguments add new lines with two spaces for indentation. It helps considerably with legibility and just about any tool that accepts JSON input can handle the white-space characters.

everett980:
> the copy icon can be animated to transform to a green checkbox.

> I would suggest using the tooltip on the icon to state / indicate:
> 
> * "Copy JSON" on mouseover
> * "Copied to clipboard" on click
> * * Set a mouseLeaveDelay to something like 500ms (or so) to let the user more easily read the message
> What do you think?

@yurishkuro @tiffon I went with the tooltip over the icon change as the tooltip will help with clarity for both before and after the copy event.
antd's tooltip struggles with a left-positioned tooltip's title increasing in length, so I used the title "Copy JSON" before the row has been copied, and "Copied" for after copy. I would hope that a user wouldn't be reliant on the "to clipboard" detail to know where copied data goes. I will update the gif in the description to include the tooltips.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.css
```diff
@@ -40,6 +40,18 @@ limitations under the License.
...
 
+.KeyValueTable--copyColumn {
+  text-align: right;
+}
+
+.KeyValueTable--copyIcon {
+  visibility: hidden;
+}
+
+tr:hover > .KeyValueTable--copyColumn > .KeyValueTable--copyIcon {
```
tiffon: Maybe a class should be added for the row? That will shorten the other selectors and avoid use of the element selector. Maybe the `KeyValueTable--body` class can be removed.

yarn.lock
```diff
@@ -2701,6 +2702,13 @@ copy-template-dir@1.3.0:
     readdirp "^2.0.0"
     run-parallel "^1.1.4"
 
+copy-to-clipboard@^3:
+  version "3.0.8"
+  resolved "https://unpm.uberinternal.com/copy-to-clipboard/-/copy-to-clipboard-3.0.8.tgz#f4e82f4a8830dce4666b7eb8ded0c9bcc313aba9"
```
tiffon: This needs to be hitting the yarn registry.

everett980: updated

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.css
everett980: added `.KeyValueTable--row` and remove references to `tr` from .css file. still need `KeyValueTable--body` for vertical alignment of table itself.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.js
```diff
@@ -55,54 +56,104 @@ type KeyValuesTableProps = {
...
+                // eslint-disable-next-line react/no-array-index-key
+                <tr key={`${row.key}-${i}`}>
+                  <td className="KeyValueTable--keyColumn">{row.key}</td>
+                  <td>{valueMarkup}</td>
+                  <td className="KeyValueTable--copyColumn">
+                    <Tooltip
+                      arrowPointAtCenter
+                      mouseLeaveDelay={0.5}
+                      onVisibleChange={visible => this.handleTooltipVisibilityChange(row, visible)}
+                      placement={'left'}
```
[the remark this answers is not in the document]
everett980: I added the rule and ran `yarn eslint --fix`

PR #309, opened 2019-01-08
Log Markers on Spans

sfriberg:
Signed-off-by: Staffan Friberg <sfriberg@kth.se>

Currently to see log entries as part of the span you need to expand the details. This PR adds markers in the main Span UI for log entries with a popover showing the log details. Tightly packed log entries are grouped as a single marker until zoomed enough to be rendered separately.

The SpanBar class will now create markers for log entries.

Thanks to @tiffon for cleaning up the CSS and switching to Antd Popover instead of Tooltip.

One caveat with the current implementation is that it will on a click in the popover expand the Span details in the main UI instead of anything in table inside the Popover. Any ideas how to get the more expected behavior of expanding the table in the popover instead?

screenshot from 2019-01-07 21-35-40


sfriberg:
> Maybe instead of showing all logs in the popup it should only show the log for the current tick, but with all log fields expanded (must be careful with max popup size in this case). It might get tricky / sensitive when there are many ticks next to each other.

It only shows for the "current tick", but I merge ticks close together. 
The span in question has 18 ticks, and only 3 show for this marker, if you would zoom in further on it it would split in to 3 separate ticks provided they have different timestamps.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/utils.test.js
```diff
@@ -27,29 +27,29 @@ import traceGenerator from '../../../demo/trace-generators';
...
-      const args = { min: 1, max: 2, start: 1.25, end: 1.75, viewStart: 0.25, viewEnd: 0.75 };
-      const { start, end } = getViewedBounds(args);
+      const args = { min: 1, max: 2, viewStart: 0.25, viewEnd: 0.75 };
+      const { start, end } = createViewedBoundsFunc(args)(1.25, 1.75);
       expect(start).toBe(0);
       expect(end).toBe(1);
     });
 
     it('works for a sub-range that within a sub-view', () => {
       const args = { min: 100, max: 200, start: 130, end: 170, viewStart: 0.1, viewEnd: 0.9 };
```
everett980: you can remove the `start` and `end` args

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/utils.js
```diff
@@ -13,29 +13,35 @@
...
- * @param  {number} end       The end of the sub-range.
  * @param  {number} viewStart The start of the zoom, on a range of [0, 1],
  *                            relative to the `min`, `max`.
  * @param  {number} viewEnd   The end of the zoom, on a range of [0, 1],
  *                            relative to the `min`, `max`.
- * @return {Object}           The resultant range.
+ * @returns {(number, number) => Object} Created view bounds function
  */
-export function getViewedBounds({ min, max, start, end, viewStart, viewEnd }) {
+export function createViewedBoundsFunc({ min, max, viewStart, viewEnd }) {
```
everett980: You could define and export  type for the output of this function, then in `VirtualizedTraceView` and `SpanBarRow` you can say that `this.getViewedBounds` is of type `viewedBoundsFunctionType` instead of declaring. not crucial, but somewhat DRYer and cleaner.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBar.js
```diff
@@ -26,21 +32,37 @@ type SpanBarProps = {
...
   viewStart: number,
+  getViewedBounds: (number, number) => { start: number, end: number },
   rpc: {
     viewStart: number,
     viewEnd: number,
     color: string,
   },
   setLongLabel: () => void,
   setShortLabel: () => void,
+  trace: Trace,
```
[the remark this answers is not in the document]
sfriberg: Will change to push only the time stamp down, was debating back and forth as well on this and decided on sending the whole trace. Don't know if there is any perf difference with complex objects vs a simple primitive?

Perhaps something to consider for later is if the trace->span->log should backpointers so you could get the span from a log and then the trace from a span.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/utils.test.js
sfriberg: good catch, though I had caught them all.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/utils.js
sfriberg: done

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBar.js
```diff
@@ -15,7 +15,13 @@
 // limitations under the License.
 
 import React from 'react';
+import { Popover } from 'antd';
+import _groupBy from 'lodash/groupBy';
+import _values from 'lodash/values';
+
 import { onlyUpdateForKeys, compose, withState, withProps } from 'recompose';
+import type { Span, Trace } from '../../../types/trace';
+import AccordianLogs from './SpanDetail/AccordianLogs';
```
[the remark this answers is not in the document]
sfriberg: should be the right order with the next push

sfriberg:
> I think it makes sense to keep the popover non-interactive. If the user is interacting with the popover to expand logs, etc., and then accidentally moves the mouse outside the popover, they will lose that state. If the user is interested in details, seems like expanding the row is a nice way to both present the details and keep the log-hints / popover simple.
> 
> What do you think?
> 
> If we end up going down the road of keeping the popover as-is, so clicking it expands the row instead of interactively expands items in the popover, then what do you think of remove the chevrons (the expand / collapse ">") from the popover, so the log items no longer look interactive?

The main reason to allow expanding is to allow reading longs lines which otherwise would be truncated. I think most people would be fine with losing the state in the popup and common practice in other applications as well. It is just intended for some quick interaction.

I'm fine with either solution, and leave the decision to you. :)

Agree about removing the chevrons if we make it completely non-interactive.

Back to the original question. I can't seem to figure out how to disable that a click in the popover expands the span details. Any clue what needs to change to disable that behavior? (that behavior is intuitive and unexpected :) )

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBar.js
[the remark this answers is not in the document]
sfriberg: done

tiffon:
I found an issue where log markers that are on the far left, right at time=0 (for the current zoom) couldn't be triggered. It was a CSS issue; I pushed an update. Hope you don't mind.

>I can't seem to figure out how to disable that a click in the popover expands the span details. Any clue what needs to change to disable that behavior? (that behavior is non-intuitive and unexpected :) )

I think adding event listeners for `onClick` to the popover and `<AccordianLogs>` and then calling `event.stopPropagation()` should do the trick.

>The main reason to allow expanding is to allow reading longs lines which otherwise would be truncated. I think most people would be fine with losing the state in the popup and common practice in other applications as well. It is just intended for some quick interaction.

I'm hesitant to have such full-fledged functionality in the logs popovers. Seems like it will get wierd when the height of the popover expands, might be a pain keep the popover positioned, correctly. 

I think it might be a nice UX to keep the span detail opening when you click in the popover but also expand the appropriate log entries. I tested it out, [here](https://github.com/tiffon/jaeger-ui/commit/0137034c6351301624cbb87adcbb2ec84022fd71). What do you think?

sfriberg:
Was able to disable the clicking in the div. Trying to do it with the AccordionLogs would require several changes to expose the event to the handler methods give to AccordionLogs.

I like the idea, my main concern is that it expands something that easily gets in the way of understanding the flow of the span in the UI. Think there are two ways to improve on that side. Easy one would be to make a second click collapse the details again. A more longterm/experimental way would be to consider if the span details should be shown on the side (static sidebar on the right or left) instead of inline with the spans which breaks visibility of spans as you want to look at details of one of them.

tiffon:
I'll look into the onClick issue.

I see what you mean about the span detail row interrupting the visual flow of span bars. I kind of wonder what the popover might look like if the logs are expanded in it? 

Re switching to a details panel on the side, I'm absolutely a fan of this. This will allow us to use the same functionality to show span details in the trace graph and trace diffs views. (I thought we had a ticket for moving to a side panel, but I guess not.)

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/AccordianLogs.js
```diff
@@ -57,13 +69,13 @@ export default function AccordianLogs(props: AccordianLogsProps) {
...
-              isOpen={openedItems.has(log)}
-              linksGetter={linksGetter}
               data={log.fields || []}
+              highContrast
+              interactive={interactive}
+              isOpen={openedItems ? openedItems.has(log) : false}
               label={`${formatDuration(log.timestamp - timestamp)}`}
-              onToggle={() => onItemToggle(log)}
+              linksGetter={linksGetter}
+              onToggle={interactive && onItemToggle ? () => onItemToggle(log) : null}
```
sfriberg: No sure what the preference/recommendation is.

Would it make sense to use undefined instead of null as it is defined as "onToggle?" in the properties?

More of a general question about using null vs undefined than anything.

CHANGELOG.md
```diff
@@ -21,6 +21,8 @@
 
 * **Trace detail:** Add indent guides to trace timeline view ([#172](https://github.com/jaegertracing/jaeger-ui/issues/172)) ([@everett980](https://github.com/everett980) in [#297](https://github.com/jaegertracing/jaeger-ui/pull/297))
 
+* **Trace detail:** Log Markers on Spans ([Fix #119](https://github.com/jaegertracing/jaeger-ui/issues/119)) ([@sfriberg](https://github.com/sfriberg) in [#309](https://github.com/jaegertracing/jaeger-ui/pull/309))
```
yurishkuro: this shouldn't go into the past release

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBarRow.js
```diff
@@ -86,12 +87,15 @@ export default class SpanBarRow extends React.PureComponent<SpanBarRowProps> {
...
+      traceStartTime,
       span,
-      viewEnd,
-      viewStart,
     } = this.props;
     const { duration, hasChildren: isParent, operationName, process: { serviceName } } = span;
     const label = formatDuration(duration);
+    const viewBounds = getViewedBounds(span.startTime, span.startTime + span.duration);
+    const viewStart = viewBounds.start;
+    const viewEnd = viewBounds.end;
```
yurishkuro: does this account for a possible zoom on the minimap? I.e. do we test with less than 100% minimap selection?

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/AccordianLogs.js
```diff
@@ -74,3 +86,11 @@ export default function AccordianLogs(props: AccordianLogsProps) {
     </div>
   );
 }
+
+AccordianLogs.defaultProps = {
+  interactive: true,
```
yurishkuro: what is the meaning of the "interactive" flag?

CHANGELOG.md
tiffon: @yurishkuro Great catch, thanks!

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBarRow.js
tiffon: Yes, it accounts for a zoom.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/AccordianLogs.js
tiffon: As a practice, I generally use `null`, which differentiates something set to `null` from something that's not there. E.g.

```js
const o = {};
console.log(o.notThere === null); // false
// for convenience, `==` is easier than `value === null || value === undefined`
console.log(o.notThere == null);  // true
```

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/AccordianLogs.js
tiffon: When set to `false` the component is not interactive.

tiffon:
@everett980 Thanks for the catch re the flow types. Turns out `@ flow` was missing form the file!

tiffon:
## @sfriberg The log markers are merged! 🎉 🎉 🎉 

Thanks so much for putting this together!

PR #312, opened 2019-01-11
Add ability to copy node data from graph

### LINKED ISSUE #304: Ability to copy node data from graph
It would be nice to be able to easily copy information from nodes to share, add to tickets, search on timeline, etc.

Nodes are not copyable because click + drag to select causes the graph to pan.

Small copy icon similar to #204

everett980:
- Resolves #304 
- Adds ability to copy operation and service name from nodes which wasn't possible due to an onDrag handler.

- Move KeyValueTable's copy icon into a component (<CopyIcon />) and consume that component in KeyValueTable, OpNode, and drawNode.

image


yurishkuro:
@everett980 please follow the same guidelines for PR description as for commit messages (described in CONTRIBUTING)

packages/jaeger-ui/src/components/TraceDiff/TraceDiffGraph/drawNode.css
```diff
@@ -74,6 +74,16 @@ limitations under the License.
   padding: 0.3rem 0.5rem 0.3rem 0.75rem;
 }
 
+.DiffNode--popover .DiffNode--copyIcon,
+.DiffNode:not(:hover) .DiffNode--copyIcon {
+  display: none;
+}
```
tiffon: Is the use of `:not(...)` necessary? Would something like the following work?

```css
.DiffNode .DiffNode--copyIcon { display: none; }
.DiffNode:hover .DiffNode--copyIcon { display: unset; }
```

packages/jaeger-ui/src/components/TraceDiff/TraceDiffGraph/drawNode.css
```diff
@@ -74,6 +74,16 @@ limitations under the License.
...
+
+.DiffNode--labelCell {
+  display: flex;
+  justify-content: space-between;
+}
```
tiffon: Can you merge this with the CSS class on [LOC 73](https://github.com/jaegertracing/jaeger-ui/pull/312/files#diff-d6dac11b88f602aa428194a0a0179eb3R73)?

packages/jaeger-ui/src/components/common/copy-icon.js
```diff
@@ -0,0 +1,69 @@
+// @flow
```
tiffon: Because this is has a component as the default export, can you name the file `CopyIcon.js`?

packages/jaeger-ui/src/components/TraceDiff/TraceDiffGraph/drawNode.js
```diff
@@ -57,6 +59,9 @@ class DiffNode extends React.PureComponent<Props> {
             </td>
             <td className={`DiffNode--labelCell ${className}`}>
               <strong>{service}</strong>
+              <span className="DiffNode--copyIcon">
```
tiffon: You migth be able to skip having a wrapping node if you add a [`className` prop](https://github.com/jaegertracing/jaeger-ui/blob/046de99a5d384cf62f29eb27db1f0c49351c55c9/packages/jaeger-ui/src/components/common/NewWindowIcon.js#L27-L28) to the `CopyIcon` component?

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.css
```diff
@@ -44,12 +44,8 @@ limitations under the License.
   text-align: right;
 }
 
-.KeyValueTable--copyIcon {
-  visibility: hidden;
-}
-
-.KeyValueTable--row:hover > .KeyValueTable--copyColumn > .KeyValueTable--copyIcon {
-  visibility: unset;
+.KeyValueTable--row:not(:hover) > .KeyValueTable--copyColumn > .KeyValueTable--copyIcon {
```
tiffon: Is the use of `:not(:hover)` to avoid defining two styles? :+1:

packages/jaeger-ui/src/components/common/copy-icon.js
```diff
@@ -0,0 +1,69 @@
...
+
+import { Icon, Tooltip } from 'antd';
+import { CopyToClipboard } from 'react-copy-to-clipboard';
+
+type propsType = {
+  copyText: string,
+  tooltipTitle: string,
+};
+
+type stateType = {
```
tiffon: Can you use initial caps on type names? E.g. `StateType`

packages/jaeger-ui/src/components/common/copy-icon.js
```diff
@@ -0,0 +1,69 @@
...
+};
+
+export default class CopyIcon extends React.PureComponent<propsType, stateType> {
+  constructor(props: propsType) {
+    super(props);
+
+    this.state = {
+      hasCopied: false,
+    };
+  }
```
tiffon: I think the constructor is not necessary, here. I believe the following will work:

```js
export default class CopyIcon extends React.PureComponent<propsType, stateType> {
   state = {
    hasCopied: false,
  };
  // ...
```

packages/jaeger-ui/src/components/common/copy-icon.js
```diff
@@ -0,0 +1,69 @@
...
+
+  handleClick = () => {
+    this.setState({
+      hasCopied: true,
+    });
+  };
+
+  handleTooltipVisibilityChange = (visible: boolean) => {
+    if (!visible) {
```
tiffon: This might also check for the current state:

```js
if (!visible && this.state.hasCopied) {
  // ...
}
```

packages/jaeger-ui/src/components/TraceDiff/TraceDiffGraph/drawNode.css
everett980: is there a reason to avoid `:not`?
I started using to avoid defining multiple styles to achieve one goal (hiding something).
I only discovered when working on #310 when I was styling based on both hover state and focus state in: `/packages/jaeger-ui/src/components/TraceDiff/TraceDiffGraph/TraceDiffGraph.css`
That seemed easier than defining multiple styles so I started applying to more places.

packages/jaeger-ui/src/components/common/copy-icon.js
everett980: renamed

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.css
everett980: yup! seemed simpler

packages/jaeger-ui/src/components/common/copy-icon.js
everett980: capitalized

packages/jaeger-ui/src/components/common/copy-icon.js
everett980: yeah this change works, simplified

packages/jaeger-ui/src/components/common/copy-icon.js
everett980: good call, potentially saves a render (or at least a check to determine if it should render)

packages/jaeger-ui/src/components/TraceDiff/TraceDiffGraph/drawNode.js
everett980: that is cleaner, added

packages/jaeger-ui/src/components/TraceDiff/TraceDiffGraph/drawNode.css
tiffon: Word.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.test.js
```diff
@@ -14,11 +14,34 @@
...
+  const wrapper = shallow(
+    <LinkValue href={href} title={title}>
+      {childrenText}
+    </LinkValue>
+  );
+
+  it('renders as expected', () => {
+    expect(wrapper.find('a').prop('href')).toBe(href);
+    expect(wrapper.find('a').prop('title')).toBe(title);
+    expect(wrapper.find('a').text()).toMatch(/childrenText/);
```
tiffon: Nit: I would typically recommend saving to a const, `const anchor = wrapper.find(...)`

Issue #306, opened 2019-01-03
Shift from Flow to TypeScript


tiffon:
In the Flow vs TypeScript comparison, TypeScript has emerged as more robust and popular than Flow.

From the [State of JavaScript 2018 Survey](https://2018.stateofjs.com/javascript-flavors/overview/):

| Language   | Would like to learn | Would use again | Total | Source                                                            |
|------------|---------------------|-----------------|-------|-------------------------------------------------------------------|
| Flow       | 24.2%               | 10.2%           | 34.6% | [link](https://2018.stateofjs.com/javascript-flavors/flow/)       |
| TypeScript | 33.7%               | 46.7%           | 80.4% | [link](https://2018.stateofjs.com/javascript-flavors/typescript/) |

[DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) has more type definitions than [flow-typed](https://github.com/flow-typed/flow-typed).

There are also many bugs or shortcomings in Flow which do not appear to be issues in TypeScript. For example, in Flow:

```
class Thing<T, U> {
  static newThing<V>(factory: () => V) {
    const thing: Thing<string, V> = new Thing(factory);
    return thing;
  }

  items: T[];
  factory: () => U;
  constructor(factory: () => U) {
    this.factory = factory;
    this.items = [];
  }
}

const fn: () => number = () => 0;
const t: Thing<string, number> = Thing.newThing<number>(fn);
const t2: Thing<string, number> = Thing.newThing<number>(() => 0);
```

[Results in the following error](https://flow.org/try/#0MYGwhgzhAEAqAWBLAdgcwDywDTQKoD5oBvAKGmggBcxLFhpkBTAdwRQwDV8AKAMzGCUA9gCcAngC5o3AJTQAvIQ5zS5csCHIq0SkjRS2adFRHscXBQxZw9qPgOHiZAbjJqRjSgFcRyHbddyAF8SN0RKRgBbCAMAbQBdQOh+QVFJaTlFPCSNLUoRL1SReyL02QVCXBU3cl1ECAA6FMcxS2a0pNqkRvComHloBKSQkJJc7V5kKXKs5C9IgCNGEUsZwgAGV3HKHQNbY3yzBnmlkUIBw1QGplZ9ucXlnkmXMc1tSgAmPfYD0zQce6nc42djXFiXdCAx7cNbQdYuIA):

```
16: const t: Thing<string, number> = Thing.newThing<number>(fn);
[...]

PR #331, opened 2019-02-16
Shift plexus to TypeScript (from flowtypes)

tiffon:
Partially addresses #306.

Outstanding items for this PR (before feedback):
- [x] Linting TypeScript in plexus
- [x] Document the build related changes
- [x] Misc cleanup
- [x] Resurrect the demo


packages/plexus/src/DirectedGraph/builtins/PureNodes.tsx
```diff
@@ -18,15 +16,16 @@ import * as React from 'react';
...
 type Props = {
-  classNamePrefix: string,
-  getNodeLabel: ?(Vertex) => string | React.Node,
-  layoutVertices: ?(LayoutVertex[]),
-  setOnNode: ?(Vertex) => ?Object,
-  vertexRefs: { current: HTMLElement | null }[],
-  vertices: Vertex[],
+  classNamePrefix: string;
+  getNodeLabel: ((vertex: TVertex) => React.ReactNode) | void;
+  layoutVertices: TLayoutVertex[] | void;
```
everett980: more of a question than a change suggestion:
when would we want to have this instead of `layoutVertices?: TLayoutVertex[]`?

packages/plexus/src/DirectedGraph/types.tsx
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
everett980: does the year here matter? github is of the impression that this is a new file (not that types.js was renamed types.tsx)

packages/plexus/src/DirectedGraph/types.tsx
```diff
@@ -0,0 +1,57 @@
...
+export type TPropsFactoryFn<TInput> = (value: TInput) => TObjectOfAny | void;
+
+export type TDirectedGraphState = {
+  edges: TEdge[];
+  layoutEdges: TLayoutEdge[] | void;
+  layoutGraph: TLayoutGraph | void;
+  layoutPhase: number;
+  layoutVertices: TLayoutVertex[] | void;
+  sizeVertices: TSizeVertex[] | void;
+  // vertexRefs: { current: HTMLElement | null }[],
```
everett980: remove?

packages/plexus/src/DirectedGraph/types.tsx
```diff
@@ -0,0 +1,57 @@
...
+  zoomEnabled: boolean;
+  zoomTransform: ZoomTransform;
+};
+
+export type TDirectedGraphProps = {
+  arrowScaleDampener: number;
+  className: string;
+  classNamePrefix: string;
+  edges: TEdge[];
+  // getEdgeLabel: ?(TEdge) => React.Node,
```
everett980: why is this commented instead of `getEdgeLabel?:` ? was edge label support remove / not implemented?

packages/plexus/src/LayoutManager/Coordinator.tsx
```diff
@@ -96,10 +102,10 @@ export default class Coordinator {
       id,
       options,
       input: { edges, vertices },
-      status: { phase: 'not-started' },
+      status: { phase: ECoordinatorPhase.NotStarted },
     };
     const isDotOnly = Boolean(options && options.useDotEdges);
-    const phase = isDotOnly ? 'dot-only' : 'positions';
+    const phase = isDotOnly ? EWorkerPhase.DotOnly : EWorkerPhase.Positions;
```
everett980: `EWorkerPhase` was imported from a `./types` file but is being used to define values. Should `EWorkerPhase` exist as `WorkerPhases` in a constants file? I think mixing constants and types could get messy. I don't expect `EWorkerPhase` to have usable values, just an enum definition.

packages/plexus/src/LayoutManager/Coordinator.tsx
```diff
@@ -17,33 +15,41 @@
 /* eslint-disable no-console */
 
 import * as convCoord from './dot/conv-coord';
-import workerFactory from './layout.worker';
+import LayoutWorker from 'worker/LayoutManager/layout.worker';
```
everett980: is this now from an npm package? if so I believe it should be the first import with a new line after.

packages/plexus/src/LayoutManager/Coordinator.tsx
```diff
@@ -176,31 +181,37 @@ export default class Coordinator {
...
+      Boolean(workerMessage.meta && workerMessage.meta.layoutId !== this.currentLayout.id);
+    if (type === EWorkerErrorType.Error) {
+      console.error(`${isStale ? '(stale) ' : ''}Error from viz.worker`, event);
       return;
     }
     if (isStale) {
       return;
     }
-    if (type === 'layout-error') {
+    if (type === EWorkerErrorType.LayoutError) {
```
everett980: this if statement, the following and the proceeding `type ===` are also comparing values to types. Maybe that's just a standard for `Enum`s that I'm not used to but it seems as though they shouldn't be used in that way.

packages/plexus/src/LayoutManager/getLayout.tsx
```diff
@@ -19,43 +17,69 @@ import viz from 'viz.js';
...
+  message: string;
+};
+
+type TVerticesValidity = TValidityError | TValidityOk | TValidityWarn;
 
 const SHIFT_THRESHOLD = 0.015;
 const VALIDITY_OK = 'VALIDITY_OK';
-const VALIDITY_ERROR = 'VALIDITY_ERROR';
 const VALIDITY_WARN = 'VALIDITY_WARN';
+const VALIDITY_ERROR = 'VALIDITY_ERROR';
```
everett980: not _that_ important but seems like this should've moved up a line, not down

packages/plexus/src/LayoutManager/layout.worker.tsx
```diff
@@ -0,0 +1,63 @@
...
+  currentMeta = meta;
+  const { layoutError, ...result } = getLayout(meta.phase, edges, vertices, options);
+  const type = layoutError ? EWorkerErrorType.LayoutError : meta.phase;
+  const message: TWorkerOutputMessage = { meta, type, ...result };
+  ctx.postMessage(message);
+  currentMeta = null;
+}
+
+function handleError(errorType: string, event: ErrorEvent) {
+  const { colno, error, filename, lineno, message } = event;
```
everett980: previously these variables came from `event.data`

packages/plexus/src/LayoutManager/layout.worker.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const payload: TWorkerErrorMessage = {
+    type: EWorkerErrorType.Error,
+    meta: currentMeta,
+    errorMessage: {
+      colno,
+      error,
```
everett980: previously `errorMessage.error` was guaranteed to have the exactly the keys `code, message, name, stack` even if some may be not be present on `event.data.error` or if more are present on `event.data.error`
Not sure if this matters, but it may change the shape of this object.

tiffon:
@everett980 Thanks for reviewing! Will check out your comments.

> ```
> key: Type | void;
> or
> key?: Type;
> or
> key?: Type | void;
> ```

In the above, [`key: Type | void`](https://www.typescriptlang.org/docs/handbook/basic-types.html#void) means the value will either be an instance of `Type` or one of the values `null` or `undefined`. It's the equivalent of [`key: ?Type` in flow](https://flow.org/en/docs/types/maybe/).

`key?: Type` means the presence of `key` is optional. 

`key?: Type | void` is a combo of the two: the presence of key is optional and if present it can be either a `Type` instance or `null` or `undefined`.

> Thirdly, if we want to be super thorough in type names, we may want to regex fix the type names that don't start with a `T`.

Agreed. Turns out I missed a few!

> Then one other point is that we don't need to use `;` instead of `,` in type definitions. I'm not overly attached to `,` over `;` but do have a slight preference. If you do as well we may want to change but if not it's fine either way.

I think we should use `;` because that's the style used by the [TS team](https://www.typescriptlang.org/docs/handbook/interfaces.html#optional-properties).

packages/jaeger-ui/src/utils/DraggableManager/README.md
```diff
@@ -45,7 +45,7 @@ To use a DraggableManager instance, relevant mouse events should be piped to the
 
 ```html
 <div className="DividerDemo--realm">
-  <div className="DividerDemo--divider" onMouseDown={this._dragManager.handleMouseDown} />
+  <div className="DividerDemo--divider" onMouseDown="{this._dragManager.handleMouseDown}" />
```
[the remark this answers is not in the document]
tiffon: Great catch! Probably the type should be `jsx`.

packages/plexus/.neutrinorc.js
```diff
@@ -0,0 +1,73 @@
+const { join } = require('path');
```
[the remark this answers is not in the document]
tiffon: Yeah, this is one of the main files I had in mind for bolstering the comments around the build related changes.

packages/jaeger-ui/config-overrides.js
```diff
@@ -41,6 +41,9 @@ const modifyVars = lessToJs(loadedVarOverrides);
 
 function webpack(_config) {
   let config = _config;
+  // const resolvedPath = require.resolve('react-scripts/config/paths');
```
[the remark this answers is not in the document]
tiffon: Thanks.

packages/plexus/src/DirectedGraph/DirectedGraph.tsx
```diff
@@ -131,14 +130,14 @@ export default class DirectedGraph extends React.PureComponent<DirectedGraphProp
...
-  constructor(props: DirectedGraphProps) {
+  constructor(props: TDirectedGraphProps) {
     super(props);
     const { edges, vertices, zoom: zoomEnabled } = props;
     if (Array.isArray(edges) && edges.length && Array.isArray(vertices) && vertices.length) {
       this.state.layoutPhase = PHASE_CALC_SIZES;
       this.state.edges = edges;
       this.state.vertices = vertices;
-      this.state.vertexRefs = vertices.map(React.createRef);
+      this.state.vertexRefs = vertices.map(() => React.createRef<HTMLElement>());
```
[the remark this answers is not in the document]
tiffon: Thanks.

packages/plexus/src/DirectedGraph/DirectedGraph.tsx
```diff
@@ -166,16 +165,18 @@ export default class DirectedGraph extends React.PureComponent<DirectedGraphProp
     }
   }
 
-  _onPositionsDone = (result: Cancelled | PositionsDone) => {
-    if (!result.isCancelled) {
+  _onPositionsDone = (result: TCancelled | TPositionsDone) => {
+    // typescript requires the `=== false` check for correct type inference
```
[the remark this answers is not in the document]
tiffon: Took a bit, but I got this working. Great catch!

packages/plexus/src/LayoutManager/Coordinator.tsx
tiffon: The enums are types, and they define a set of values of that type. For instance, `EWorkerPhase.DotOnly` is a value of type `EWorkerPhase`.

Here are [examples of the upsides](https://www.typescriptlang.org/play/index.html#src=enum%20ThingAB%20%7B%0D%0A%20%20%20%20A%20%3D%20'A'%2C%0D%0A%20%20%20%20B%20%3D%20'B'%2C%0D%0A%7D%0D%0A%0D%0Aenum%20ThingAB2%20%7B%0D%0A%20%20%20%20A%20%3D%20'A'%2C%0D%0A%20%20%20%20B%20%3D%20'B'%2C%0D%0A%7D%0D%0A%0D%0Aconst%20one%3A%20ThingAB%20%3D%20ThingAB.A%3B%0D%0Aconst%20two%3A%20ThingAB2%20%3D%20ThingAB2.A%3B%0D%0Aconsole.log(one%20%3D%3D%3D%20two)%3B%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%2F%2F%20always%20false%20error%20%20%0D%0Aconst%20three%3A%20ThingAB2%20%3D%20ThingAB.A%3B%20%20%20%20%20%20%2F%2F%20not%20assignable%20error%0D%0A%0D%0A):

```typescript
enum ThingAB {
    A = 'A',
    B = 'B',
}

enum ThingAB2 {
    A = 'A',
    B = 'B',
}

const one: ThingAB = ThingAB.A;
const two: ThingAB2 = ThingAB2.A;
console.log(one === two);               // always false error  
const three: ThingAB2 = ThingAB.A;      // not assignable error
```

I would say enums are pretty nice, and I've missed them in JS and flow. 

Unfortunately, unions of enums (or something akin to unions) are not possible in TypeScript (not sure about other languages), so I did some strange stuff with `EWorkerPhase` deriving values from `ECoordinatorPhase`, but otherwise I believe their use in plexus is fairly idiomatic.

The [official rationale](https://en.wikipedia.org/wiki/Enumerated_type#Rationale) for them.

Enums in [C#](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/enum#example-2), [Swift](https://docs.swift.org/swift-book/LanguageGuide/Enumerations.html), [Java](https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html), and [others](https://rosettacode.org/wiki/Enumerations).

packages/plexus/src/LayoutManager/Coordinator.tsx
tiffon: Well, I needed to create a type definition for the layout.worker that reflected it being used by `Coordinator.tsx` as a `WebWorker`. In TypeScript, I don't think you can create auxiliary type defs for local files. So, I made an alias in webpack and then added a type def for the alias. 

So, it's a local file loaded via an alias resolver.  

Seems like it should be kept in the local files section of imports, but moved to the bottom.

packages/plexus/src/DirectedGraph/DirectedGraph.tsx
```diff
@@ -202,9 +203,9 @@ export default class DirectedGraph extends React.PureComponent<DirectedGraphProp
     this.setState({ zoomTransform });
   };
 
-  _constrainZoom = (transform: D3Transform, extent: [[number, number], [number, number]]) => {
+  _constrainZoom = (transform: ZoomTransform, extent: [[number, number], [number, number]]) => {
     const [, [vw, vh]] = extent;
-    const { height: h, width: w } = this.state.layoutGraph || {};
+    const { height: h = null, width: w = null } = this.state.layoutGraph || {};
     if (h == null || w == null) {
       // for flow
```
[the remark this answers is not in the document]
tiffon: `state.layoutGraph` can be `null`... so I think you're right.

packages/plexus/src/DirectedGraph/builtins/PureNodes.tsx
tiffon: If the prop is provided in all cases, then it should be `layoutVertices:`. If the prop is sometimes omitted, then it should be `layoutVertices?:`

packages/plexus/src/DirectedGraph/types.tsx
tiffon: @yurishkuro How do you feel about omitting the year from copyright notices? I think it's pretty common to leave it out?

https://github.com/facebook/jest/blob/51817fd79a3744078a82861f3763dc346ca0cbd6/packages/babel-jest/src/index.ts#L1-L10

packages/plexus/src/DirectedGraph/types.tsx
tiffon: :+1:

packages/plexus/src/DirectedGraph/types.tsx
tiffon: It was not included. But, could be. So... I'll remove it. :+1:

packages/plexus/src/LayoutManager/Coordinator.tsx
tiffon: These are comparing values to values. The type is `EWorkerErrorType` and `EWorkerErrorType.LayoutError` is a value with that type. So, the code above is comparing a value with a value.

packages/plexus/src/LayoutManager/getLayout.tsx
tiffon: These were actually unused, so I remove them.

packages/plexus/src/LayoutManager/layout.worker.tsx
tiffon: Yes, and that was in error.

packages/plexus/src/LayoutManager/layout.worker.tsx
tiffon: The current code is more aligned with the spec. Not sure why I had the previous code, might have been from experimentation.

typings/index.d.ts
```diff
@@ -14,10 +12,4 @@
...
 // limitations under the License.
 
-export type ErrorEvent = {
-  message: string,
-  filename: string,
-  lineno: number,
-  colno?: number,
-  error?: any,
-};
+/// <reference path="custom.d.ts" />
```
everett980: Does this file serve a purpose if it is just comments?

tiffon: Hey, thought I responded to this. Yes, it does. Added docs for this, here:

https://github.com/jaegertracing/jaeger-ui/pull/331/files#diff-c21202b241d0cd40822db520c8b00770R103

packages/plexus/demo/src/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { render } from 'react-dom';
+
+import largeDg, { getNodeLabel as getLargeNodeLabel } from './data-large';
+import { edges as dagEdges, vertices as dagVertices } from './data-dag';
+import { colored as colorData, getColorNodeLabel, setOnColorEdge, setOnColorNode } from './data-small';
+import { DirectedGraph, LayoutManager } from '../../src';
+import { TVertex } from '../../src/types';
+
+// const { TVertex } = types;
```
everett980: can be removed

packages/plexus/demo/src/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+import './index.css';
+
+const { classNameIsSmall } = DirectedGraph.propsFactories;
+
+const addAnAttr = () => ({ 'data-rando': Math.random() });
+const setNodeClassName = (vertex: TVertex) => ({
```
everett980: might as well call it `setOnNode` since that's how it's used and it does more than define the class name (unless the `onClick: () => console.log(...)` is meant to be removed)

packages/plexus/BUILD.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Script `prepublishOnly`
+
+Executed after `yarn install` is run in the project root.
+
+### Dependencies (dev and otherwise)
+
+#### `viz.js@1.8.1`
+
+This specific version of [viz.js](https://github.com/mdaines/viz.js) is used to avoid a regression. Meanwhile, [looks like `2.x.x`](https://github.com/mdaines/viz.js/issues/120#issuecomment-389281407) has recovered a lot of ground.
```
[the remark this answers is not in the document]
tiffon: I already created a ticket, will add it to the comment.

packages/plexus/demo/src/index.tsx
tiffon: :+1:

BUILD.md
```diff
@@ -0,0 +1,111 @@
...
+
+Note: This configuration is extended by `./packages/plexus/.eslintrc.js`.
+
+## `.flowconfig`
+
+Being phased out.
+
+## `.travis.yml`
+
+Currently `./packages/plexus` doesn't have any tests... But, when it does, `.travis.yml` needs to be updated to send coverage info for all `./packages/*` to codecov.io.
```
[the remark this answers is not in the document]
tiffon: https://github.com/jaegertracing/jaeger-ui/issues/340

BUILD.md
```diff
@@ -0,0 +1,111 @@
...
+
+`tsc-lint-debug` is for diagnosing problems with linking, resolving files, or aliases in TypeScript code. It lists the files involved in the compilation.
+
+### `husky` . `hooks` . `pre-commit`
+
+Runs the `lint` and `test` scripts.
+
+## `.eslintrc.js`
+
+Pretty basic. Needs to be cleaned up. The `airbnb` configuration needs to be updated.
```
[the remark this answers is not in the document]
tiffon: Not seeing the warning I usually, see... I might have resolved it. I'll create a ticket when I see it again.

BUILD.md
```diff
@@ -0,0 +1,111 @@
...
+
+## `.travis.yml`
+
+Currently `./packages/plexus` doesn't have any tests... But, when it does, `.travis.yml` needs to be updated to send coverage info for all `./packages/*` to codecov.io.
+
+[`yarn install --frozen-lockfile`](https://yarnpkg.com/lang/en/docs/cli/install/#toc-yarn-install-frozen-lockfile) ensures installs in CI fail if they would typically mutate the lockfile.
+
+## `lerna.json`
+
+We should probably audit our use of `lerna` to make sure a) it's necessary and b) it's idiomatic if it is necessary. We have ended up relying quite a bit on `yarn` workspaces, which has reduced the relevance of `lerna`.
```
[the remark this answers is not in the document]
tiffon: https://github.com/jaegertracing/jaeger-ui/issues/341

Issue #333, opened 2019-02-21
Timeline visualization regression


ledor473:
## Requirement - what kind of business use case are you trying to solve?
Visualize traces

## Problem - what in Jaeger blocks you from solving the requirement?
I'm not totally sure how but seems like the pull request https://github.com/jaegertracing/jaeger-ui/pull/297 broke the indentation in the timeline view in some cases.

If I checkout the previous commit in master (ac4f7a7f5a66d8eea5a20f45bb3b86ac65ebd9e7) I see this:
<img width="420" alt="pre" src="https://user-images.githubusercontent.com/1359779/53206430-f337da80-35fe-11e9-9832-a233cddd6cf5.png">

But when I checkout this commit from master (f23eb487572a67047869dc2ea70098d943091146) I see this:
<img width="420" alt="post" src="https://user-images.githubusercontent.com/1359779/53206437-f763f800-35fe-11e9-9b0c-876c78dca77f.png">

Here's the trace that is in the screenshot above: https://gist.github.com/ledor473/a51e66aee1f14285eef4953ae7ee6d59

## Proposal - what do you suggest to solve the problem or improve the existing situation?
I'm not sure what broke it exactly, so its hard to propose a fix.

yurishkuro:
@tiffon @everett980 please take a look

everett980:
Thanks for raising!
I have a change to ancestor-ids in [this pr](https://github.com/jaegertracing/jaeger-ui/pull/310)
I will verify that it fixes the issue for this trace (it should).

PR #335, opened 2019-02-22
Process FOLLOWS_FROM spans in TraceView

rubenvp8510:
- This fixes https://github.com/jaegertracing/jaeger-ui/issues/333

- I noticed that there is FOLLOWS_FROM attributes on some spans that wasn't take into account when the indentation of the spans is computes on the TreeView.

Not sure if this is the proper way of handling this, I left two comments for two assumptions that I did. if some of the maintainers can review please. @tiffon @everett980 

Thanks!

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanTreeOffset.js
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanTreeOffset.js
@@ -16,7 +16,6 @@
 
 import cx from 'classnames';
 import _get from 'lodash/get';
-import _find from 'lodash/find';
 import React from 'react';
 import IoChevronRight from 'react-icons/lib/io/chevron-right';
 import IoIosArrowDown from 'react-icons/lib/io/ios-arrow-down';
@@ -54,14 +53,7 @@ export class UnconnectedSpanTreeOffset extends React.PureComponent<SpanTreeOffse
     super(props);
 
     this.ancestorIds = [];
-    let currentSpan: Span = props.span;
-    while (currentSpan) {
-      currentSpan = _get(_find(currentSpan.references, { refType: 'CHILD_OF' }), 'span');
-      if (currentSpan) {
-        this.ancestorIds.push(currentSpan.spanID);
-      }
-    }
-
+    this.computeAncestors(props.span);
     // Some traces have multiple root-level spans, this connects them all under one guideline and adds the
     // necessary padding for the collapse icon on root-level spans.
     this.ancestorIds.push('root');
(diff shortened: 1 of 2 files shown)


packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanTreeOffset.js
```diff
@@ -52,14 +52,7 @@ export class UnconnectedSpanTreeOffset extends React.PureComponent<SpanTreeOffse
...
 
     this.ancestorIds = [];
-    let currentSpan: Span = props.span;
-    while (currentSpan) {
-      currentSpan = _get(_find(currentSpan.references, { refType: 'CHILD_OF' }), 'span');
-      if (currentSpan) {
-        this.ancestorIds.push(currentSpan.spanID);
-      }
-    }
-
```
tiffon: Looking at the code that builds the trace tree, which is then turned into the array of spans with parents preceding children (and children are sorted by startTime):

https://github.com/jaegertracing/jaeger-ui/blob/38efa49e96d34cb666fa232da14346ceccd896d4/packages/jaeger-ui/src/selectors/trace.js#L72-L75

Looks like the algorithm (which could probably be better) is to use the span ID of the first reference, as long as that reference is FOLLOWS_FROM or CHILD_OF.

In accumulating the ancestor IDs, we should aim to be consistent with that process so it accurately reflects the order the list.

Therefore, maybe the following would work?

```jsx
while (currentSpan && Array.isArray(currentSpan.references)) {
  const { (refType, span } = currentSpan.references[0] || {};
  if (span && (refType === 'FOLLOWS_FROM' || refType === 'CHILD_OF')) {
    this.ancestorIds.push(span.spanID);
    currentSpan = span;
  }
}
```

What do you think? 

@everett980 Can you weigh in?

everett980: I think the precedent for handling reference is not particularly strong. AFAIK there is nothing preventing a span from being `FOLLOWS_FROM` multiple other spans. With regards to indentation it should only indent once (based on depth) but I do not know how the UI handles this case in other ways. Does the span with two `FOLLOWS_FROM` show up twice in the list, once per reference?

Purely from a UI layout perspective I like the implementation in your comment. However, I am uncertain if we should have better support for multiple references.

yurishkuro: Multiple references are rarely used within the same trace (i.e. to represent a JOIN), the more realistic use case is something like multiple traces coalescing into a single batch write operation, which causally depends on all of the previous traces. However, the _first_ reference in the list is always used to pick the trace ID for the new span (actually, different Jaeger clients might be different in that regard, e.g.. they may give priority to the _first child-of reference_).

rubenvp8510:
I like the algorithm you put in the comments, simpler that mine, at the beginning I thought we will give an special treatment to the FOLLOWS_FROM, that is why I split it in two different things, but after an analysis of my implementation and  your code it seems like from the UI perspective (or at least for this view) we won't do a distinction. Correct me if this is wrong.

I'll simplify my code.

rubenvp8510:
I also think that we need to put some tests for this, to avoid regressions.

rubenvp8510:
@tiffon @everett980 I added one test to take FOLLOWS_FROM into account. I think this is ready for another review, if there is no more comments and you decide it's ok, then it is ready for merge.

Issue #247, opened 2018-09-12
Performance for large number of traces


pavolloffay:
## Requirement - what kind of business use case are you trying to solve?

At the moment showing a query for a large number of results e.g. limit=1500 fetches a lot of data to UI which makes it not responsible enough. 

Q: Not sure if the issue is caused by a large amount of data or just displaying points in the top graph.

It's related to https://github.com/jaegertracing/jaeger/issues/954 which asks for adding sorting logic on the server side. If the backed can fetch and sort more traces in memory than UI this can be a workaround.

Also somehow indirectly related to  https://github.com/jaegertracing/jaeger/issues/1051 and https://github.com/jaegertracing/jaeger/issues/960

## Proposal - what do you suggest to solve the problem or improve the existing situation?

Instead of returning all data to UI e.g. traces with tags, logs operation names the backend could return summary with only necessary data. Then when on clicking on a trace UI would query backed for full trace instance.

tiffon:
@pavolloffay, thanks for creating this ticket.

The current search implementation has some severe issues, returning full traces being one of the issues, along with 

- [#892 - Search returns results in different order every time](https://github.com/jaegertracing/jaeger/issues/892)
- [#690 - cannot query for minduration and tags simultaneously](https://github.com/jaegertracing/jaeger/issues/690)
- [#166 - jaeger-query might not return search results due to `LIMIT` parameter](https://github.com/jaegertracing/jaeger/issues/166)

The issue of loading full traces for search results can be addressed with the current implementation of search. But, we've primarily been focused on addressing the issues mentioned in jaeger#166 and related tickets. As part of that work effort we intend to lighten the payload for search results.

yurishkuro:
Also this [#169 - Provide GraphQL query service](https://github.com/jaegertracing/jaeger/issues/169)

Justin-YimYim:
Hi guys, has this issue solved? If I understand it rightly, the performance issue after loading 1500+ traces is because of the DOM generation on the Search Trace Page. I have made a paging button to split the DOM generation and I have used 10K traces to test it and could be rendered. Should I make a pull request or should I create a new issue branch?

yurishkuro:
The issue is that if your search results contain 1000+ traces, the results themselves are only showing a limited set of information about each trace that can be represented by a handful of data elements, but in order to display them the UI nonetheless needs to load complete traces and pre-process them (like finding the root span to display name & latency).

Justin-YimYim:
[...]

PR #337, opened 2019-02-26
Virtualized traced results list

rubenvp8510:
- This PR is part of the solution of the Issue #247, I still think we need to implement a new endpoint, but this definitely makes the UI more responsive.  **This is still work in progress.**

- When the trace search page has too many traces to show, it makes the UI less responsive, almost unusable in some cases. This is because the amount of DOM objects that the browser needs to handle. With virtualization we only render and maintain objects that are visible to the user.


objectiser:
@rubenvp8510 @tiffon Realise this is marked WIP, but just wanted to check if there is any chance this could make it into 1.11, or would it be best to leave for a patch release?

rubenvp8510:
@objectiser This is almost ready, I need to make sure that one of the test pass, then I can remove the WIP. I think I can have this by tomorrow end of the day, but I need @tiffon feedback.

pavolloffay:
That would be great, if you also could cut 1.1.1 release of UI it would be awesome.

tiffon:
Also, seems like [a test is broken](https://travis-ci.org/jaegertracing/jaeger-ui/builds/499083711#L1258).

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ResultItem.js
```diff
@@ -60,55 +60,57 @@ export default class ResultItem extends React.PureComponent<Props> {
...
+                    const { name, numberOfSpans: count } = service;
+                    return (
+                      <li key={name} className="ub-inline-block ub-m1">
+                        <Tag
+                          className="ResultItem--serviceTag"
+                          style={{ borderLeftColor: colorGenerator.getColorByKey(name) }}
+                        >
+                          {name} ({count})
+                        </Tag>
+                      </li>
```
tiffon: For this, the `<Tag...>` seems pretty heavy. What about using something home-grown?

```jsx
<li
  key={name}
  className="ResultItem--serviceTag"
  style={{ borderLeftColor: colorGenerator.getColorByKey(name) }}
>
  {name} <span className="ResultItem--serviceCount">{count}</span>
</li>
```

```css
.ResultItem--serviceTag {
  background: #fff;
  border-left: 15px solid transparent;
  border-radius: 4px;
  box-shadow: 0 0 1px rgba(0, 0, 0, 0.5);
  color: rgba(0, 0, 0, 0.65);
  display: inline-block;
  font-size: 0.9em;
  line-height: 1;
  margin: 0.3em 0.35em;
  padding: 0.3em;
}

.ResultItem--serviceCount {
  font-size: 0.9em;
  opacity: 0.5;
  padding-left: 0.2em;
}
```

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.js
```diff
@@ -173,20 +210,17 @@ export default class SearchResults extends React.PureComponent<SearchResultsProp
...
-                  toggleComparison={this.toggleComparison}
-                  trace={trace}
-                  disableComparision={disableComparisons}
-                />
-              </li>
-            ))}
-          </ul>
+          <ListView
+            dataLength={this.props.traces.length}
+            itemHeightGetter={() => 155}
```
tiffon: I think this needs to have the `initialDraw` prop set, otherwise it will render 300 items, from the start.

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.js
```diff
@@ -83,9 +84,16 @@ export const sortFormSelector = formValueSelector('traceResultsSort');
...
   props: SearchResultsProps;
-
   static defaultProps = { skipMessage: false };
 
+  constructor(props) {
+    super(props);
+    const { queryOfResults, diffCohort } = this.props;
+    const searchUrl = getUrl(stripEmbeddedState(queryOfResults));
+    const cohortIds = new Set(diffCohort.map(datum => datum.id));
+    this.state = { searchUrl, cohortIds };
```
tiffon: Currently, this state is not being updated when new props come in, so selecting result items for comparison is not working.

Seems like using [memoize-one](https://github.com/alexreardon/memoize-one), instead, would be less work than maintaining the state.

PR #359, opened 2019-03-25
Convert from Flow to Typescript for Jaeger-UI

everett980:
- Typescript has proven to be superior than Flow, so this PR converts from Flow to Typescript.

- Convert all files with `@flow`, as well as some of the simple non-flow files.
- Add script to trace typescript coverage.


tiffon:
Fixes #306.

Issue #361, opened 2019-03-27
[trace timeline] Add a "collapse to matches" button next to the Find input


tiffon:
Add a button on the Find input, near the up and down chevrons, that triggers the collapsing of parent spans that aren't ancestors of matching nodes.

This would match the behaviour of loading the page when there is a `uiFind` value that has at least one match.

The button should be an icon, not the literal text "collapse to matches". Maybe the up and down buttons can be changed to arrows (instead of chevrons) and the icon for the new button can be the double-right chevron.

everett980:
Having some issues with Icons atm. Icons that aren't used elsewhere in the project are blank. I think it is caused by using a slightly older version of antd that needs to load them all. If we update antd we'll get svg icons which would be nice.

Aside from that, the change is nearly ready. I still need to fix the Icons and add new tests, but there should be a diff for this sometime tomorrow 🤞

everett980:
While working on this I noticed that expanding / collapsing many rows at once is a bit jarring. Based on the current implementation of detail rows and hidden children IDs I think it would be non-trivial to add some sort of transition to hiding/showing children and expanding/collapsing rows, but would make the experience a bit smoother.

tiffon:
Fixed by #367.

Issue #360, opened 2019-03-26
Unable to copy & paste span ID


yurishkuro:
## Requirement - what kind of business use case are you trying to solve?

I can't to highlight span IU to copy & paste it elsewhere.

## Problem - what in Jaeger blocks you from solving the requirement?

The popup shows saying  "click to add to filter", the span ID cannot be highlighted either by double-clicking or by dragging the mouse.

![image](https://user-images.githubusercontent.com/3523016/55026289-188e7e80-4fd9-11e9-8662-7c726a4eb0bc.png)

## Proposal - what do you suggest to solve the problem or improve the existing situation?

Add a magnifying glass for "click to add to filter" next to the span ID, leave span ID as plain text.

everett980:
What are your thoughts on making the tooltip the clickable element instead?
Or instead of an add to filter icon, add a copy icon next to the ID which is a pattern present on KV tables and graph nodes?

Temporary work around: Right click the button (spanID) and select "copy"

yurishkuro:
Copy button might work. What is the use case for adding span ID to the filter anyway? I am already looking at it.

everett980:
The filter is shareable, so a user can add the ID then copy the url to send a link to a specific span within the trace. Or multiple spans if they select multiple IDs.

tiffon:
@yurishkuro, @everett980 

I think it makes sense to have an icon for both filtering and copying. Currently, the filter functionality has a low level of discoverability.

OTOH, it's reasonable for the filter icon to actually be a share icon, since that's when adding it to the filter really makes sense.

Yuri, not sure if you're aware, but whatever is in the find input is added to the URL. And, if a URL with a `uiFind` value is viewed, then the parent spans that aren't ancestors of matching spans are collapsed and the spans that match have their details expanded. For instance, here is an entry with three matching spans:

![image](https://user-images.githubusercontent.com/3471184/51396551-47860100-1b0d-11e9-9891-4d9d74bdd1ad.png)

I think it makes to add a button on the Find input, near the up and down arrows, that triggers the collapsing of parents that aren't ancestors of matching nodes. Created #361 to capture this.

tiffon:
Fixed in #367.

PR #367, opened 2019-04-02
add uiFind icons, scroll to first match

everett980:
- Solves #361 
- Solves #360 

- Moves the click handler to add a spanID to uiFind to a button next to the spanID
- Adds a collapse and scroll to match button to TracePageSearchBar
- Scrolls to first uiFind match when the TimelineView is mounted

### Unfocused Find
Aside from the icon button changes, the uiFind input looks the same before focusing
image
### Focused Find
Focusing the input results in it widening
image

### Icon close up
New Icon is currently hovered to show distinction. Vertical borders have been removed.
image

### Deep link icon
New deep link icon, with tooltip (and with safari's scrollbar rearing its head)
image

### Focused button
Slight highlight indicates focus.
image


packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/index.tsx
```diff
@@ -113,13 +113,11 @@ export default function SpanDetail(props: SpanDetailProps) {
...
         <small className="SpanDetail--debugInfo">
-          <Tooltip title="Click ID to add to filter">
-            <span className="SpanDetail--debugLabel" data-label="SpanID:" />{' '}
-            <button className="SpanDetail--debugValue" type="button" onClick={() => addToUiFind(spanID)}>
-              {spanID}
-            </button>
+          <span className="SpanDetail--debugLabel" data-label="SpanID:" />{' '}
+          {spanID}
+          <Tooltip placement="topRight" title="Click to add to filter for deep linking">
+            <Button className="SpanDetail--debugValue" htmlType="button" icon="down" onClick={() => addToUiFind(spanID)}/>
```
everett980: arbitrary icon because many antd icons are showing up as blank... TODO

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
```diff
@@ -158,6 +159,10 @@ export class VirtualizedTraceViewImpl extends React.PureComponent<VirtualizedTra
     setTrace(trace, uiFind);
   }
 
+  componentDidMount() {
+    setTimeout(this.props.scrollToFirstVisibleSpan);
```
everett980: Without the `setTimeout` it scrolls to the bottom. I think the rows may not have height in the initial render.
@tiffon do you have any insight on why this would scroll too far without the `setTimeout`

packages/jaeger-ui/src/components/TracePage/index.tsx
```diff
@@ -330,10 +332,15 @@ export class TracePageImpl extends React.PureComponent<TProps, TState> {
...
     }
-    //  =  ?  : null;
-    // const spanFindMatches = !traceGraphView ? this._filterSpans(uiFind || '', _get(trace, 'data.spans')) : null;
+
     const isEmbedded = Boolean(embedded);
     const headerProps = {
+      focusUiFind: () => {
+        if (trace.data) {
+          focusUiFind(trace.data, uiFind);
+          setTimeout(this._scrollManager.scrollToFirstVisibleSpan);
```
everett980: The scroll destination changes as a result of dispatching `focusUiFind`. This `setTimeout` achieves the desired result but definitely seems like the wrong way to achieve this.
@tiffon how would you recommend scrolling to the first match only after the rows have been expanded/collapsed and had their children shown/hidden?

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
tiffon: Looks like the constructor calls `setTrace(...)`. When visiting the URL, directly, the `setTrace(...)` applies the find matching.

When coming from the search page the `setTimeout` has no effect; the find is already applied to the redux state via the reducer that changed the state to be viewing the trace.

So, seems like the rows are rendered but they're not collapsed, yet, when visiting a trace directly from a URL. Probably it's scrolling to the position of the row, but then the non-matching rows are collapsed and the target scroll position ends up getting truncated with the result of being scrolled to the bottom.

packages/jaeger-ui/src/components/TracePage/ScrollManager.tsx
```diff
@@ -255,6 +260,10 @@ export default class ScrollManager {
     this._scrollToVisibleSpan(-1);
   };
 
+  scrollToFirstVisibleSpan = () => {
+    this._scrollToVisibleSpan(1, 0);
```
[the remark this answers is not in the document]
everett980: when `startRow` is not provided, it starts at the top of the visible spans, not from the very first span. `_scrollToVisibleSpan(1, 0)` means from the first span in the current trace, scroll down (1) to the first span that exists in `uiFindMatches`

packages/jaeger-ui/src/components/TracePage/index.tsx
everett980: Unfortunately an instance property wouldn't work because the component that updates as a result of `focusUiFind` is not `TracePage/index.tsx`. I got around this by adding a property to the `detailStates` Map and `childrenHiddenIDs` Set made by `calculateHiddenIdsAndDetailStates` in `TracePage/duck.tsx` and reading that property in `componentDidUpdate` in `TraceTimelineViewer/VirtualizedTraceView.tsx`
It's slightly unorthodox in principle, but not so bad in practice.
[Here is the commit that solely addresses setTimeout](https://github.com/jaegertracing/jaeger-ui/pull/367/commits/ebf9473b9613815e861723b77e69ad7eb7c577e2)
TODO: more tests and icons

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/duck.tsx
```diff
@@ -121,14 +112,38 @@ function setTrace(state: TTraceTimeline, { uiFind, trace }: TTraceUiFindValue) {
...
+}
+
+
+function setTrace(state: TTraceTimeline, { uiFind, trace }: TTraceUiFindValue) {
+  const { traceID, spans } = trace;
+  if (traceID === state.traceID) {
+    return state;
+  }
+  const { spanNameColumnWidth } = state;
+  const newState = { ...newInitialState(), spanNameColumnWidth, traceID };
```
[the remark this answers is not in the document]
everett980: Nice, updated 👍

packages/jaeger-ui/src/components/TracePage/ScrollManager.tsx
```diff
@@ -131,7 +135,8 @@ export default class ScrollManager {
     }
     const { duration, spans, startTime: traceStartTime } = this._trace;
     const isUp = direction < 0;
-    const boundaryRow = isUp ? xrs.getTopRowIndexVisible() : xrs.getBottomRowIndexVisible();
+    const boundaryRow = _isUndefined(startRow) ? (isUp ? xrs.getTopRowIndexVisible() : xrs.getBottomRowIndexVisible()): startRow;
```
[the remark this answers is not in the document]
everett980: replaced and removed all lodash imports that are no longer used in this file.
I agree about nested ternaries, was a quick and dirty WIP. It's now two separate ternaries.

packages/jaeger-ui/src/components/TracePage/ScrollManager.tsx
```diff
@@ -184,7 +189,7 @@ export default class ScrollManager {
 
       // If there are hidden children, scroll to the last visible span
       if (childrenAreHidden) {
-        let isFallbackHidden: boolean | TNil;
+        let isFallbackHidden: boolean;
```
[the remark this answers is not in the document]
everett980: I introduced it as `?boolean` which was necessary according to flow. Else it says:
`  ^ possibly uninitialized variable [1] is incompatible with boolean [2].`
I imagine we'll continue to find some `TNil` over-usage since I did `s/?\([a-zA-Z]\{1,\}\)/\1 | TNil/` pretty liberally.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/duck.tsx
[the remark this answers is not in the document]
everett980: nice, shortened 👍

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/index.tsx
tiffon: Yes, replace instead of append.

packages/jaeger-ui/src/components/TracePage/ScrollManager.tsx
```diff
@@ -132,7 +132,7 @@ export default class ScrollManager {
     const { duration, spans, startTime: traceStartTime } = this._trace;
     const isUp = direction < 0;
     const boundaryRow = isUp ? xrs.getTopRowIndexVisible() : xrs.getBottomRowIndexVisible();
-    const spanIndex = xrs.mapRowIndexToSpanIndex(boundaryRow);
+    const spanIndex = xrs.mapRowIndexToSpanIndex(startRow != null ? startRow : boundaryRow);
```
tiffon: Is calculating `boundaryRow` even when it is not used intentional?

packages/jaeger-ui/src/components/TracePage/ScrollManager.tsx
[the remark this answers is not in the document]
tiffon: :+1:

packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageSearchBar.css
```diff
@@ -15,11 +15,19 @@ limitations under the License.
...
+  justify-content: flex-end;
 }
 
 .TracePageSearchBar--bar {
-  display: flex;
+  max-width: 100%;
+  transition: max-width 1.5s;
+}
+
+.TracePageSearchBar--bar:not(:focus-within) {
```
tiffon: Why prefer the use of `:not(...)` over:

```css
.TracePageSearchBar--bar {
  max-width: 20rem;
  transition: max-width 0.5s;
}

.TracePageSearchBar--bar:focus-within {
  max-width: 100%;
}
```

Also, the `1.5s` seems a bit slow?

packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageSearchBar.tsx
```diff
@@ -48,28 +57,40 @@ export function TracePageSearchBarFn(props: TracePageSearchBarProps & { forwarde
...
-        />
+        {navigable && (
+          <>
+            <Button
+              className={cx(btnClass, 'ant-btn-icon-only')}
+              disabled={!textFilter}
+              htmlType="button"
+              onClick={focusUiFindMatches}
+            >
+              <IoAndroidLocate />
```
tiffon: Does the icon look a bit low?

```css
margin-top: -4px;
```

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/index.tsx
```diff
@@ -113,13 +114,16 @@ export default function SpanDetail(props: SpanDetailProps) {
...
-              {spanID}
-            </button>
+          <span className="SpanDetail--debugLabel" data-label="SpanID:" /> {spanID}
+          <Tooltip placement="topRight" title="Click to add to filter for deep linking">
+            <Button
+              className="SpanDetail--debugValue ant-btn-icon-only"
+              htmlType="button"
+              onClick={() => addToUiFind(spanID)}
+            >
+              <ZoomIn size="1.4em" />
```
tiffon: I really think it makes sense to treat this as an anchor, which is pretty traditional, instead of appending it to the search field. 

What do you think?

tiffon: Also, the ti icons have a pretty different aesthetic than the others used in the UI. I generally avoid them.

tiffon: Re `ant-btn-icon-only`, I recommend not using Ant's CSS classes, directly, as that will expose the styling to what amounts to being a private API. This will increase the risk of unintended consequences when upgrading.

packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageSearchBar.css
everett980: the `:not(...)` was a relic of an earlier approached, updated :+1:
Changed to 0.5s and seems fine. Initially the width delta was larger and when it was too fast it felt instantaneous rather than smooth.

packages/jaeger-ui/src/components/TracePage/ScrollManager.tsx
everett980: I see we have `boundary` which is the index of a row and `boundaryRow` which is also the index of a row...
changed it to a `let boundaryRow` with an `if ... else if ... else` so that `xrs.get${Top/Bottom}RowIndexVisible` isn't called unnecessarily.

packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageSearchBar.tsx
everett980: I couldn't tell if it was actually lower or if the up/down arrows are high. when compared to the close icon it seemed closer. I believe it's location is fixed now.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/index.tsx
everett980: > I really think it makes sense to treat this as an anchor, which is pretty traditional, instead of appending it to the search field.

If the functionality for clicking the icon is to deep link to a `spanID`, not add to the `uiFind`, do you think it'd make sense for this button to instead be "Copy deep link to this span"? On click it'd copy the url but with urlParam for `uiFind` to instead only be the `spanID` instead of whatever is currently present. Replacing the current `uiFind` seems odd, but copying a link to a single id does seem like a useful use case.

everett980: > Re `ant-btn-icon-only`, I recommend not using Ant's CSS classes, directly, as that will expose the styling to what amounts to being a private API. This will increase the risk of unintended consequences when upgrading.

Generally I agree, but if antd's styling for `antd-btn-icon-only` changes, wouldn't we want this `<Button><SomeIcon /></Button>` to update accordingly? By creating a class that mimic's `antd-btn-icon-only` then on upgrading antd we risk having inconsistent icon buttons. Seems like either way we have a risk when updating the library.

tiffon: #### Which icon(s) to use and append vs replace

> do you think it'd make sense for this button to instead be "Copy deep link to this span"? On click it'd copy the url but with urlParam for `uiFind` to instead only be the `spanID` instead of whatever is currently present.

That sounds like it would work. I would suggest the icon for this be a link icon, maybe the `IoIosLink` in react-icons?

For the case of appending to `uiFind`, the current approach of a magnifying glass with a `+` inside works really well, IMO (not the ti style, but the general idea). Seems like this will be a rare use-case, though. What do you think?

Since these icons will start to look kind of noisy, maybe they can only show up on hover of the span detail panel?

@yurishkuro You had an interest in this functionality, what do you think of having an icon for copying a deep-link to this specific span and one for adding this span to the find (via append)?

#### Using `.antd-btn-icon-only`

One of the main issues with using `.antd-btn-icon-only` (by being renamed, etc) is that it could go away in a future update. In that case, we would miss the updated styling and have broken styling on our `<Button>` that we might miss in inspecting the impact of updating.

By creating a class that mimics the `.antd-btn-icon-only` styling, we would have some risk of inconsistent styling, but I would say that's preferable over risk of broken styling, but still not ideal.

I would think the ideal scenario is to use the React components, as intended. This will generally have the lowest risk of the above scenarios.

Why not go with something like the following?

```jsx
<Button className="AddCustomStyling--etc" ghost size="small">
  <SomeIconHere />
</Button>
```

everett980: > #### Which icon(s) to use and append vs replace
> > do you think it'd make sense for this button to instead be "Copy deep link to this span"? On click it'd copy the url but with urlParam for `uiFind` to instead only be the `spanID` instead of whatever is currently present.
> 
> That sounds like it would work. I would suggest the icon for this be a link icon, maybe the `IoIosLink` in react-icons?
> 
> For the case of appending to `uiFind`, the current approach of a magnifying glass with a `+` inside works really well, IMO (not the ti style, but the general idea). Seems like this will be a rare use-case, though. What do you think?

I think the append spanID to `uiFind` use-case is really rare, and can be easily done by simply copying and pasting the spanID. I would opt for just one `IoIosLink` icon to copy a deep link url.

> Since these icons will start to look kind of noisy, maybe they can only show up on hover of the span detail panel?

With just one icon that is only visible when the user has expanded the detail row I don't think hiding it until hovering is necessary.

> @yurishkuro You had an interest in this functionality, what do you think of having an icon for copying a deep-link to this specific span and one for adding this span to the find (via append)?

I'm also interested in user's opinions, but as Yuri pointed out on the PR that added `uiFind`, it seems as though adding an already-visible spanID to the `uiFind` doesn't have a purpose outside of deep linking, which would be achieved with the link icon, not the append-to-search icon.

> #### Using `.antd-btn-icon-only`
> One of the main issues with using `.antd-btn-icon-only` (by being renamed, etc) is that it could go away in a future update. In that case, we would miss the updated styling and have broken styling on our `<Button>` that we might miss in inspecting the impact of updating.
> 
> By creating a class that mimics the `.antd-btn-icon-only` styling, we would have some risk of inconsistent styling, but I would say that's preferable over risk of broken styling, but still not ideal.
> 
> I would think the ideal scenario is to use the React components, as intended. This will generally have the lowest risk of the above scenarios.
> 
> Why not go with something like the following?
> 
> ```js-jsx
> <Button className="AddCustomStyling--etc" ghost size="small">
>   <SomeIconHere />
> </Button>
> ```

I hadn't thought that they might remove/rename the class. I also realized that the class isn't intended for `<Button>`s that have children so it could have unintended changes. I added a class to achieve this functionally without using `.antd-btn-icon-only`

packages/jaeger-ui/src/components/TracePage/index.test.js
```diff
@@ -132,6 +135,32 @@ describe('<TracePage>', () => {
...
+  describe('focusUiFindMatches', () => {
+    let scrollToFirstVisibleSpanSpy;
+
+    beforeEach(() => {
+      scrollToFirstVisibleSpanSpy = jest.spyOn(wrapper.instance()._scrollManager, 'scrollToFirstVisibleSpan');
+      scrollToFirstVisibleSpanSpy.mockReset();
+      defaultProps.focusUiFindMatches.mockReset();
+    });
+
+    it('calls _scrollManager.scrollToFirstVisibleSpan and props.focusUiFindMatches with props.trace.data and uiFind when props.trace.data is present', () => {
```
tiffon: Is the call to `_scrollManager.scrollToFirstVisibleSpan` necessary?

packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageSearchBar.css
```diff
@@ -15,22 +15,35 @@ limitations under the License.
 */
 
 .TracePageSearchBar {
-  max-width: 20rem;
+  width: 40%;
+}
+
+.TracePageSearchBar--InputGroup {
```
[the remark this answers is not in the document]
everett980: Nice, updated
Where is the list of `.ub-` classes?

packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageSearchBar.tsx
```diff
@@ -48,28 +57,40 @@ export function TracePageSearchBarFn(props: TracePageSearchBarProps & { forwarde
   };
 
   return (
-    <div className="ub-flex-auto ub-mx2 TracePageSearchBar">
+    <div className="TracePageSearchBar">
       {/* style inline because compact overwrites the display */}
-      <Input.Group compact style={{ display: 'flex' }}>
+      <Input.Group className="TracePageSearchBar--InputGroup" compact style={{ display: 'flex' }}>
```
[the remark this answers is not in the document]
everett980: The comment above the inline `style` seems to be accurate. I'm going to leave it as is.

packages/jaeger-ui/src/components/TracePage/index.tsx
```diff
@@ -306,6 +305,14 @@ export class TracePageImpl extends React.PureComponent<TProps, TState> {
     }
   }
 
+  focusUiFindMatches = () => {
+    const { trace, focusUiFindMatches, uiFind } = this.props;
+    if (trace && trace.data) {
+      this._scrollManager.scrollToFirstVisibleSpan();
```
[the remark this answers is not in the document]
everett980: doesn't seem to be, removed :+1:

packages/jaeger-ui/src/components/TracePage/index.test.js
everett980: nope, removed :+1:

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/duck.tsx
```diff
@@ -68,67 +70,87 @@ export const actionTypes = generateActionTypes('@jaeger-ui/trace-timeline-viewer
...
-    return state;
-  }
-  const { spanNameColumnWidth } = state;
-
-  if (!uiFind) {
-    // No filter, so we're done
-    return { ...newInitialState(), spanNameColumnWidth, traceID };
-  }
-  // There is a filter, so collapse all rows except matches and their ancestors; show details for matches
+function calculateHiddenIdsAndDetailStates(uiFind: string, spans: Span[]) {
```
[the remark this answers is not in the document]
everett980: much better name, updated!

packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageSearchBar.css
```diff
@@ -15,22 +15,35 @@ limitations under the License.
...
 .TracePageSearchBar--btn {
+  border-left: none;
   transition: 0.2s;
 }
 
 .TracePageSearchBar--btn.is-disabled {
   opacity: 0.5;
 }
+
+.TracePageSearchBar--btn.locate-btn {
```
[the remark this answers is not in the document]
everett980: updated

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/index.tsx
```diff
@@ -69,6 +68,7 @@ export default function SpanDetail(props: SpanDetailProps) {
       value: formatDuration(relativeStartTime),
     },
   ];
+  const deepLinkCopyText = `${window.location.origin}${window.location.pathname}?uiFind=${spanID}`;
```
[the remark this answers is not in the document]
everett980: I guess I'm just used to seeing `window.location` instead of `location` on its own, but `location` on its own works fine and is shorter so I updated.

I agree about consolidating `uiFind` in its own PR. Should that just be when `uiFind` is in a string? If so, this seems to be the only instance based on a quick search (`ag --ts '['"'"'"`].+uiFind'`). But I think it would make sense to be a constant whenever it is interacting with the URL (i.e.: here and when interacting with `queryString.parse` / `queryString.stringify`)

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
```diff
@@ -158,6 +159,20 @@ export class VirtualizedTraceViewImpl extends React.PureComponent<VirtualizedTra
     setTrace(trace, uiFind);
   }
 
+  shouldComponentUpdate(nextProps: VirtualizedTraceViewProps) {
+    const nextPropKeys = Object.keys(nextProps) as (keyof VirtualizedTraceViewProps)[];
+    for (let i = 0; i < nextPropKeys.length; i += 1) {
+      if (nextProps[nextPropKeys[i]] !== this.props[nextPropKeys[i]]) {
+        if (nextPropKeys[i] === 'shouldScrollToFirstUiFindMatch') {
+          if (nextProps[nextPropKeys[i]]) return true;
```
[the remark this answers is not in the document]
everett980: Added two inline comments.

That said, I typically rely on unit tests to serve as documentation as they cannot be out of date. Good spec names identify the expected functionality.
Though that has the downside of not tying documentation to specific lines.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/index.tsx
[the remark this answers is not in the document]
everett980: Also, eslint dislikes `location` but doesn't mind `window.location`. What is your preference on using `window.location` vs `// eslint-disable-line no-restricted-globals`

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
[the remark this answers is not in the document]
tiffon: LGTM

packages/jaeger-ui/src/components/TracePage/index.tsx
```diff
@@ -280,6 +280,9 @@ export class TracePageImpl extends React.PureComponent<TProps, TState> {
     if (this.props.trace && this.props.trace.data) {
       this.traceDagEV = calculateTraceDagEV(this.props.trace.data);
     }
+    if (traceGraphView) {
+      this.focusUiFindMatches();
+    }
```
tiffon: This seems kind of strange. So, it will focus the UI find matches when you go back to the timeline view?

everett980: I thought I had removed this... It doesn't work without more changes and as you said it may not be the desired functionality.
Will really remove and push.

PR #372, opened 2019-04-17
Limit the thickness of spans in the minimap

### LINKED ISSUE #370: Limit the thickness of spans in the minimap
In the screenshot below, the spans in the minimap are about 4x thicker than the spans in the main timeline. I think it looks very weird aesthetically and might be confusing to new users. Maybe we can set a limit on the max thickness of the spans to be no thicker than in the main timeline view?

image

rubenvp8510:
- Fixes #370 

- This imposes a limit of the thickness of spans in the minimap. (not more than the span in the main view)

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TracePageHeader/SpanGraph/render-into-canvas.test.js
+++ b/packages/jaeger-ui/src/components/TracePage/TracePageHeader/SpanGraph/render-into-canvas.test.js
@@ -21,6 +21,7 @@ import renderIntoCanvas, {
   MAX_TOTAL_HEIGHT,
   MIN_ITEM_WIDTH,
   MIN_TOTAL_HEIGHT,
+  MAX_ITEM_HEIGHT,
 } from './render-into-canvas';
 
 const getCanvasWidth = () => window.innerWidth * 2;
@@ -124,16 +125,19 @@ describe('renderIntoCanvas()', () => {
         { input: items[1].serviceName, output: [1, 1, 1] },
         { input: items[2].serviceName, output: [2, 2, 2] },
       ];
+      const cHeight =
+        items.length < MIN_TOTAL_HEIGHT ? MIN_TOTAL_HEIGHT : Math.min(items.length, MAX_TOTAL_HEIGHT);
+
       const expectedDrawings = [
         getBgFillRect(),
         ...items.map((item, i) => {
           const { valueWidth, valueOffset } = item;
           const color = expectedColors[i].output;
           const fillStyle = `rgba(${color.concat(ITEM_ALPHA).join()})`;
-          const height = MIN_TOTAL_HEIGHT / items.length;
+          const height = Math.min(MAX_ITEM_HEIGHT, Math.max(MIN_ITEM_HEIGHT, cHeight / items.length));
           const width = valueWidth / totalValueWidth * getCanvasWidth();
           const x = valueOffset / totalValueWidth * getCanvasWidth();
-          const y = height * i;
+          const y = cHeight / items.length * i;
           return { fillStyle, height, width, x, y };
         }),
       ];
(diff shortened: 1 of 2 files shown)


packages/jaeger-ui/src/components/TracePage/TracePageHeader/SpanGraph/render-into-canvas.tsx
```diff
@@ -36,7 +37,9 @@ export default function renderIntoCanvas(
   canvas.width = cWidth;
   // eslint-disable-next-line no-param-reassign
   canvas.height = cHeight;
-  const itemHeight = Math.max(MIN_ITEM_HEIGHT, cHeight / items.length);
+  const computedItemHeight = Math.max(MIN_ITEM_HEIGHT, cHeight / items.length);
+  const itemHeight = computedItemHeight > MAX_ITEM_HEIGHT ? MAX_ITEM_HEIGHT : computedItemHeight;
```
yurishkuro: Why mix ?: and max?

w = Math.min(MAX, Math.max(MIN, expr))

rubenvp8510: I just wanted to be more explicit, but I think using math.min and math.max is more elegant. I'll change.

rubenvp8510: Done

packages/jaeger-ui/src/components/TracePage/TracePageHeader/SpanGraph/render-into-canvas.test.js
```diff
@@ -124,16 +125,19 @@ describe('renderIntoCanvas()', () => {
...
         ...items.map((item, i) => {
           const { valueWidth, valueOffset } = item;
           const color = expectedColors[i].output;
           const fillStyle = `rgba(${color.concat(ITEM_ALPHA).join()})`;
-          const height = MIN_TOTAL_HEIGHT / items.length;
+          const height = Math.min(MAX_ITEM_HEIGHT, Math.max(MIN_ITEM_HEIGHT, cHeight / items.length));
           const width = valueWidth / totalValueWidth * getCanvasWidth();
           const x = valueOffset / totalValueWidth * getCanvasWidth();
-          const y = height * i;
+          const y = cHeight / items.length * i;
```
yurishkuro: Can we have the same min/max logic in the test? It's easier to read.

rubenvp8510: Actually the test has the min/max logic,

rubenvp8510: I'll attach screenshots with small trace. But I think if we limit the height of the minimap for this case we will end with a really small minimap  (not too much aesthetic IMHO) but I'll attach both screenshots.

tiffon: Thanks

rubenvp8510:
I attached 3 possible implementations here.

1) This PR , limit the thickness of the span bar

Screenshot from 2019-04-23 10-23-19
 

2) Limit the canvas for render (but not the minimap height)

Screenshot from 2019-04-23 11-01-46

3) Also limit the heigh of the minimap according to the number of spans.


Screenshot from 2019-04-23 11-05-07

yurishkuro:
the last two screenshots look wrong - if we reduce the height of the minimap, we should still calculate it as MIN(max_allowed_height, default_span_height * span_count), where default_span_height is the same as in the main timeline view.

rubenvp8510:
Ha! I did the calculations using MIN(max_allowed_height, default_span_height * span_count) but I used the MIN_ITEM_HEIGHT as a default_span_height, that explains why it looks bad.

If I understand correctly default_span_height should be the height of the span in the main timeline. Ok I'll update it.

rubenvp8510:
Screenshot from 2019-04-23 13-01-20
@yurishkuro is this way correct?

Thanks for review!

yurishkuro:
that's what I was suggesting, but now that I see it, it doesn't look particularly great (probably because there's no whitespace between spans).

@tiffon what are your creative suggestions?

tiffon:
1 looks the great, IMO. Is that the one currently set to be merged?

rubenvp8510:
Yes, option 1 is implemented in this PR.

PR #373, opened 2019-04-25
Minor enhancement : Use Arrow Link for the Force Dependency graph

Etienne-Carriere:
Replace simple link by Arrow link in the Force Dependency graph so that the graph is more
understandable

Signed-off-by: Etienne Carriere <etienne.a.carriere@socgen.com>

Currently, in the Force Directed Graph, we don't know the direction of the dependency between 2 services.

Replacing "not arrow link" by "Arrow link" in the graph

### DIFF
--- a/packages/jaeger-ui/src/components/DependencyGraph/DependencyForceGraph.js
+++ b/packages/jaeger-ui/src/components/DependencyGraph/DependencyForceGraph.js
@@ -60,6 +60,7 @@ export default class DependencyForceGraph extends Component {
   render() {
     const { nodes, links } = this.props;
     const { width, height } = this.state;
+    const nodesMap = new Map(nodes.map(node => [node.id, node]));
 
     return (
       <div
@@ -103,7 +104,12 @@ export default class DependencyForceGraph extends Component {
             />
           ))}
           {links.map(({ opacity, ...link }) => (
-            <JaegerForceGraphArrowLink key={`${link.source}=>${link.target}`} opacity={opacity} link={link} />
+            <JaegerForceGraphArrowLink 
+                key={`${link.source}=>${link.target}`} 
+                opacity={opacity} 
+                link={link}
+                targetRadius={nodesMap.get(link.target).radius}
+            />
           ))}
         </InteractiveForceGraph>
       </div>
--- a/packages/jaeger-ui/src/components/DependencyGraph/DependencyForceGraph.js
+++ b/packages/jaeger-ui/src/components/DependencyGraph/DependencyForceGraph.js
@@ -13,9 +13,10 @@
 // limitations under the License.
 
 import React, { Component } from 'react';
-import { InteractiveForceGraph, ForceGraphNode, ForceGraphArrowLink } from 'react-vis-force';
+import { InteractiveForceGraph, ForceGraphNode}  from 'react-vis-force';
 import { window } from 'global';
 import { debounce } from 'lodash';
+import {default as JaegerForceGraphArrowLink} from './JaegerForceGraphArrowLink';
 
 import { nodesPropTypes, linksPropTypes } from '../../propTypes/dependencies';
 
@@ -102,7 +103,7 @@ export default class DependencyForceGraph extends Component {
             />
           ))}
           {links.map(({ opacity, ...link }) => (
-            <ForceGraphArrowLink key={`${link.source}=>${link.target}`} opacity={opacity} link={link} />
+            <JaegerForceGraphArrowLink key={`${link.source}=>${link.target}`} opacity={opacity} link={link} />
           ))}
         </InteractiveForceGraph>
       </div>
--- a/packages/jaeger-ui/src/components/DependencyGraph/DependencyForceGraph.js
+++ b/packages/jaeger-ui/src/components/DependencyGraph/DependencyForceGraph.js
@@ -13,10 +13,10 @@
 // limitations under the License.
 
 import React, { Component } from 'react';
-import { InteractiveForceGraph, ForceGraphNode}  from 'react-vis-force';
+import { InteractiveForceGraph, ForceGraphNode } from 'react-vis-force';
 import { window } from 'global';
 import { debounce } from 'lodash';
-import {default as JaegerForceGraphArrowLink} from './JaegerForceGraphArrowLink';
+import ForceGraphArrowLink from './ForceGraphArrowLink';
 
 import { nodesPropTypes, linksPropTypes } from '../../propTypes/dependencies';
 
@@ -104,11 +104,11 @@ export default class DependencyForceGraph extends Component {
             />
           ))}
           {links.map(({ opacity, ...link }) => (
-            <JaegerForceGraphArrowLink 
-                key={`${link.source}=>${link.target}`} 
-                opacity={opacity} 
-                link={link}
-                targetRadius={nodesMap.get(link.target).radius}
+            <ForceGraphArrowLink
+              key={`${link.source}=>${link.target}`}
+              opacity={opacity}
+              link={link}
+              targetRadius={nodesMap.get(link.target).radius}
             />
           ))}
         </InteractiveForceGraph>
--- a/packages/jaeger-ui/src/components/DependencyGraph/DependencyForceGraph.js
+++ b/packages/jaeger-ui/src/components/DependencyGraph/DependencyForceGraph.js
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 import React, { Component } from 'react';
-import { InteractiveForceGraph, ForceGraphNode, ForceGraphLink } from 'react-vis-force';
+import { InteractiveForceGraph, ForceGraphNode, ForceGraphArrowLink } from 'react-vis-force';
 import { window } from 'global';
 import { debounce } from 'lodash';
 
@@ -102,7 +102,7 @@ export default class DependencyForceGraph extends Component {
             />
           ))}
           {links.map(({ opacity, ...link }) => (
-            <ForceGraphLink key={`${link.source}=>${link.target}`} opacity={opacity} link={link} />
+            <ForceGraphArrowLink key={`${link.source}=>${link.target}`} opacity={opacity} link={link} />
           ))}
         </InteractiveForceGraph>
       </div>
--- a/packages/jaeger-ui/src/components/DependencyGraph/JaegerForceGraphArrowLink.js
+++ b/packages/jaeger-ui/src/components/DependencyGraph/JaegerForceGraphArrowLink.js
@@ -57,30 +57,25 @@ export default class JaegerForceGraphArrowLink extends PureComponent {
   }
 
   render() {
-    const { link, targetRadius, edgeOffset, ...spreadable } = this.props;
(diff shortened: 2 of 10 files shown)


Etienne-Carriere:
Example for the following dependency graph : 
`{"dependencies":[
        {"parent":"service1","child":"service2","callCount":1000},
        {"parent":"service2","child":"service3","callCount":100},
        {"parent":"service1","child":"service3","callCount":100},
        {"parent":"service3","child":"service4","callCount":10},
        {"parent":"service4","child":"service3","callCount":101}],
 "timestamp":"2019-05-02T00:00:00Z"
}
`

Currently : 

CURRENTLY_Not_Directed_Graph_with_label

Proposal : 
PROPOSAL_Directed_Graph_with_label

Etienne-Carriere:
@tiffon, thank you for the remark. 

I made some tests and the targetRadius is more for an homothetic ratio which is not great for "big nodes" so I try a translation of the marker on the link and I have something interesting . 

Unfortunately, I have to patch the  ForceGraphArrowLink (https://github.com/uber/react-vis-force/blob/master/src/components/ForceGraphArrowLink.js) . 2 possible solutions : 
* Patch the react-vis-force lib but seems to be not maintained (last patch in 2017)
* Have a monkey-patch (new composant JaegerForceGraphArrowLink) in Jaeger-ui 

What do you prefer ?

tiffon:
I see what you mean. It's very unfortunate, but yeah, we're probably not going to get much traction on updating the library, and I would generally prefer not to have a fork of it... 

@yurishkruo, can you weigh in on supporting arrows in the links? The main issue is that in cases where the nodes have a large radius, the node covers the arrows.

Screen Shot 2019-05-03 at 12 47 53 PM


Note: The purple arrows are my annotations.

In the image above, the opacity is reduced. This allows the arrows, which are behind the green nodes, to be visible. The dark, concentric effects in the large nodes are the visual effect of the arrows that are behind the nodes.


What do you think?

Etienne-Carriere:
@tiffon , for helping you in the decision, here is the solution with the "monkey_patch" : 
https://github.com/societe-generale/jaeger-ui/tree/directed_link_monkey_patch
https://github.com/societe-generale/jaeger-ui/commit/0b76ab5b47484f4b114e42bedd9440665dcb991b

tiffon:
@Etienne-Carriere Using a local arrow link works great!

In the layout of the edges, I noticed that sometimes the edges are off-center and / or the arrow is hidden behind the target node.

Screen Shot 2019-05-14 at 3 49 12 PM


The following changes seem to address these issues:


```diff
diff --git a/packages/jaeger-ui/src/components/DependencyGraph/DependencyForceGraph.js b/packages/jaeger-ui/src/components/DependencyGraph/DependencyForceGraph.js
index e27ab40..a7e856b 100644
--- a/packages/jaeger-ui/src/components/DependencyGraph/DependencyForceGraph.js
+++ b/packages/jaeger-ui/src/components/DependencyGraph/DependencyForceGraph.js
@@ -60,6 +60,7 @@ export default class DependencyForceGraph extends Component {
   render() {
     const { nodes, links } = this.props;
     const { width, height } = this.state;
+    const nodesMap = new Map(nodes.map(node => [node.id, node]));
 
     return (
       <div
@@ -103,7 +104,12 @@ export default class DependencyForceGraph extends Component {
             />
           ))}
           {links.map(({ opacity, ...link }) => (
-            <JaegerForceGraphArrowLink key={`${link.source}=>${link.target}`} opacity={opacity} link={link} />
+            <JaegerForceGraphArrowLink
+              key={`${link.source}=>${link.target}`}
+              opacity={opacity}
+              link={link}
+              targetRadius={nodesMap.get(link.target).radius}
+            />
           ))}
         </InteractiveForceGraph>
       </div>
diff --git a/packages/jaeger-ui/src/components/DependencyGraph/JaegerForceGraphArrowLink.js b/packages/jaeger-ui/src/components/DependencyGraph/JaegerForceGraphArrowLink.js
index 3160f56..aaf9516 100644
--- a/packages/jaeger-ui/src/components/DependencyGraph/JaegerForceGraphArrowLink.js
+++ b/packages/jaeger-ui/src/components/DependencyGraph/JaegerForceGraphArrowLink.js
@@ -28,7 +28,7 @@ import { ForceGraphLink} from 'react-vis-force';
 }
 
 
-export default class JaegerForceGraphArrowLink extends PureComponent { 
+export default class JaegerForceGraphArrowLink extends PureComponent {
   static get propTypes() {
     return {
       link: PropTypes.shape({
@@ -57,30 +57,25 @@ export default class JaegerForceGraphArrowLink extends PureComponent {
   }
 
   render() {
-    const { link, targetRadius, edgeOffset, ...spreadable } = this.props;
+    const { link, targetRadius, edgeOffset: _, ...spreadable } = this.props;
     const id = `arrow-${linkId(link)}`;
     return (
       <g>
         <defs>
           <marker
             id={id}
-            markerWidth={(targetRadius * 3) + 1}
-            markerHeight={(targetRadius * 3) + 1}
-            refX={ targetRadius + link.target_node_size }
-            refY={targetRadius}
+            markerWidth={6}
+            markerHeight={4}
+            refX={5 + targetRadius}
+            refY={2}
             orient="auto"
             markerUnits="strokeWidth"
           >
-            {targetRadius > 0 && (
-              <path
-                d={`M0,0 L0,${targetRadius * 2} L${targetRadius * 3},${targetRadius} z`}
-                fill={spreadable.stroke || spreadable.color}
-              />
-            )}
+            {targetRadius > 0 && <path d="M0,0 L0,4 L6,2 z" fill={spreadable.stroke || spreadable.color} />}
           </marker>
         </defs>
 
-        <ForceGraphLink {...spreadable} link={link} edgeOffset={targetRadius} markerEnd={`url(#${id})`} />
+        <ForceGraphLink {...spreadable} link={link} markerEnd={`url(#${id})`} />
       </g>
     );
   }
```



Screen Shot 2019-05-14 at 3 44 25 PM

These changes

- Uses the target node's radius instead of the default of `2`
- Use a static drawing of the arrow head and only adjust the refX to shift it based on the target node's radius
- Omits the `edgeOffset` prop from the `ForceGraphLink` because that doesn't seem to do anything constructive

What do you think?

Apologies for the slow reply! Thank you for the diff!

Etienne-Carriere:
@tiffon , I put the monkey_patch branch and then take in account your patch proposal. After you review, I propose to squash the commits before merge.

packages/jaeger-ui/src/components/DependencyGraph/JaegerForceGraphArrowLink.js
```diff
@@ -0,0 +1,88 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
tiffon: A couple of requests for this file.

Can you update the 2017 to 2019 in the copyright?

We're aiming to use TypeScript for all new files. Would you mind converting to *.tsx?

Also, I am thinking the `Jaeger` scoping of `JaegerForceGraphArrowLink` is kind of implied; what do you think of renaming to just `ForceGraphArrowLink`?

packages/jaeger-ui/src/components/DependencyGraph/JaegerForceGraphArrowLink.js
```diff
@@ -0,0 +1,88 @@
...
+  value: PropTypes.number,
+}).isRequired,
+      targetRadius: PropTypes.number,
+      edgeOffset: PropTypes.number,
+      className: PropTypes.string,
+      opacity: PropTypes.number,
+      stroke: PropTypes.string,
+      strokeWidth: PropTypes.number
+    };
+  }
```
tiffon: As we're using TypeScript, can you use a type instead of `prop-types`? E.g.

```tsx
type TProps = {
  link: {
    source: string;
    target: string;
    value: string;
  };
  targetRadius: number;
  // etc
};

export default class JaegerForceGraphArrowLink extends PureComponent<TProps> {
  // ...
```

packages/jaeger-ui/src/components/DependencyGraph/DependencyForceGraph.js
```diff
@@ -13,9 +13,10 @@
 // limitations under the License.
 
 import React, { Component } from 'react';
-import { InteractiveForceGraph, ForceGraphNode, ForceGraphLink } from 'react-vis-force';
+import { InteractiveForceGraph, ForceGraphNode}  from 'react-vis-force';
 import { window } from 'global';
 import { debounce } from 'lodash';
+import {default as JaegerForceGraphArrowLink} from './JaegerForceGraphArrowLink';
```
tiffon: The `{ default as ... }` is not necessary; it can simply be:

```js
import AnyName from './the-file';
// same as:
import { default as AnyName } from './the-file';
```

[Ref](http://exploringjs.com/es6/ch_modules.html#_default-exports-one-per-module)

packages/jaeger-ui/src/components/DependencyGraph/JaegerForceGraphArrowLink.js
```diff
@@ -0,0 +1,83 @@
...
+    const { link, targetRadius, edgeOffset: _, ...spreadable } = this.props;
+    const id = `arrow-${linkId(link)}`;
+    return (
+      <g>
+        <defs>
+          <marker
+            id={id}
+            markerWidth={6}
+            markerHeight={4}
+            refX={ 5 +  link.target_node_size }
```
tiffon: I think you can use `targetRadius` instead of `link.target_node_size`? The reason I figured using `targetRadius` is better is because `link.target_node_size` was not accurate. But, if your update fixes it, then we can get rid of `targetRadius`.

packages/jaeger-ui/src/components/DependencyGraph/DependencyForceGraph.js
Etienne-Carriere: agreed

packages/jaeger-ui/src/components/DependencyGraph/JaegerForceGraphArrowLink.js
Etienne-Carriere: agreed

Etienne-Carriere:
@tiffon , I take in account all your comments but the conversion from JS to TypeScript . Sorry but I an not an JS dev only a "patcher"

yurishkuro:
ping

Etienne-Carriere:
pong . Yes, I am late with writing the unit test . Will try this week

Etienne-Carriere:
The ts linter has some issues as it seems to be the first test in TypeScript. I will try to fix it.

tiffon:
Thanks for adding tests!

I think the CI failure is from adding types for jasmine and enzyme. These added types for node, which defines a `Global` type (apparently) and is in conflict with the `const global` defined in the custom typings.

We generally write tests in plain JavaScript, and it's preferable to be consistent. 

Can you change the tests to be JS? Or, do you mind if I make the edit?

Etienne-Carriere:
I put the test back in JS

Etienne-Carriere:
@tiffon , do you know why the codecov hook is hanged ?

yurishkuro:
I restarted the build

PR #375, opened 2019-05-03
Deduplicate tags for spans

rubenvp8510:

- This solves issue: https://github.com/jaegertracing/jaeger/issues/1470, at least from the UI perspective, not sure if this is the best way.

- I check for duplicated tags, and remove it (I left only one of them)
- Also has a boolean to indicate that span has duplicated tags so we can show a warning or something 

Screenshot of warning:

Screenshot from 2019-05-03 12-29-02

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/AccordianText.css
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/AccordianText.css
@@ -1,5 +1,5 @@
 /*
-Copyright (c) 2017 Uber Technologies, Inc.
+Copyright (c) 2019 Uber Technologies, Inc.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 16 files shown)


yurishkuro:
I like the warning icon, but we have a built-in place for warnings in the spans (below), which may contain more than one warning. So it would be better to display them as a list. Maybe we can add another expandable item below logs called `Warnings({count})`, which may be prefixed with the warning icon for better visuals, and not show up at all if span has no warnings.

https://github.com/jaegertracing/jaeger/blob/3fff10206109b54c9d8a42cb1347336836a9b0ea/model/model.pb.go#L359

rubenvp8510:
Ahh! I haven't noticed that field in span object, in that case it makes more sense to put warnings there. I'll do the modifications and update this PR, with new screenshots also.

rubenvp8510:
Screenshot from 2019-05-04 13-09-57

@yurishkuro WDYT?

yurishkuro:
@rubenvp8510 s/s looks good to me. I  assume the Warnings will not show up if there are no warnings?

@tiffon is this ready to merge?

rubenvp8510:
@yurishkuro yes, the warning accordion  does not show up if there are no warning on the span

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/AccordianText.css
```diff
@@ -0,0 +1,40 @@
...
+  cursor: initial;
+}
+
+.AccordianText--header.is-high-contrast:not(.is-empty):hover {
+  background: #ddd;
+}
+
+.AccordianText--emptyIcon {
+  color: #aaa;
+}
```
tiffon: I don't think these styles are used anywhere?

- `.AccordianText--header.is-empty`
- `.AccordianText--header.is-high-contrast:not(.is-empty):hover` (as a result of the above)
- `.AccordianText--emptyIcon`

rubenvp8510:
Hi @tiffon I did all requested modifications, also I modified the style according to your suggestion, I think looks better and more "inline" with the actual UI style.

I attached some screen-shots of the latest modifications.

Screenshot from 2019-05-15 18-23-30
Screenshot from 2019-05-15 18-23-19

rubenvp8510:
Looks very good! Thanks for the improvement!

tiffon:
Hey, awesome, done! :+1:

PR #378, opened 2019-05-10
Create parser function and test with mock data

everett980:
- Begin work on DDG, this PR parses expected network payload into _The Model_.

- Parsing function and tests


packages/jaeger-ui/src/utils/DeepDependencyGraph/parse-payload.test.js
```diff
@@ -0,0 +1,86 @@
...
+    const focalPathElemArgument = groupOperations ? { service: focalPathElem.service } : focalPathElem;
+    const { paths, services } = parsePayload(payload, focalPathElemArgument);
+    const serviceNames = Object.keys(services);
+
+    expect(serviceNames).toEqual(Array.from(new Set(_map(_flatten(payload), 'service'))));
+    serviceNames.forEach(serviceName => {
+      expect(Object.keys(services[serviceName].operations)).toEqual(
+        Array.from(new Set(_map(_filter(_flatten(payload), { service: serviceName }), 'operation')))
+      );
+    });
```
tiffon: Why not use native JS functionality for this type of thing?

```js
// if `services` and `service.operations` are changed to `Map`s
const inputFlat = [].concat(...payload);

const inputSvcs = new Set(inputFlat.map(item => item.service));
expect(new Set(services.keys())).toEqual(inputSvcs);

for (let [name, service] of services) {
  const ops = inputFlat.map(item => item.service === service ? item.operation : null).filter(Boolean);
  expect(new Set(service.operations.keys())).toEqual(new Set(ops));
}
```

packages/jaeger-ui/src/utils/DeepDependencyGraph/parse-payload.tsx
```diff
@@ -0,0 +1,112 @@
+// Copyright (c) 2019 Uber Technologies, Inc.
```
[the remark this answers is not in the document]
tiffon: `model` seems like a better home for this than `utils`...?

packages/jaeger-ui/src/utils/DeepDependencyGraph/parse-payload.tsx
```diff
@@ -0,0 +1,112 @@
...
+  members: TPathElem[];
+};
+
+type TServiceMap = Record<string, TService>;
+
+type TParsedPayload = {
+  paths: TPath[];
+  services: TServiceMap;
+};
+
```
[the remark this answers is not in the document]
everett980: SGTM, I will rename and move to a `types` file.

packages/jaeger-ui/src/utils/DeepDependencyGraph/parse-payload.tsx
```diff
@@ -0,0 +1,112 @@
...
+type TPath = {
+  focalIdx: number;
+  members: TPathElem[];
+};
+
+type TServiceMap = Record<string, TService>;
+
+type TParsedPayload = {
```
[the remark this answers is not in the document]
everett980: I had thought that the hash and request parameters would live in the URL? There isn't much need to have them in the redux state, unless I'm missing something.

packages/jaeger-ui/src/utils/DeepDependencyGraph/parse-payload.test.js
everett980: To me, the value of using lodash here is that it is very declarative. The five nested function calls is dense, but I can remove `Array.from` and add `new Set` to the `expect` part. Then it reads as `_flatten` the payload, `_filter` to those with the target `serviceName`, `_map` to get `operation`, then create the `new Set` for comparison.

The code you provided is good, but I have to read more carefully to validate it. IMO `const ops = inputFlat.map(item => item.service === service ? item.operation : null).filter(Boolean);` is as dense as `_map(_filter(_flatten(payload), { service: serviceName }), 'operation')` and requires returning `null` from the `map` in order to do the `.filter(Boolean)` which isn't bad but is more to remember while checking the code.

packages/jaeger-ui/src/utils/DeepDependencyGraph/parse-payload.test.js
```diff
@@ -0,0 +1,86 @@
...
+  });
+
+  it('checks both operation and service when calculating focalIdx when both are provided', () => {
+    const { almostDoubleFocalPath } = testResources;
+    parsedOutputValidator({ paths: [almostDoubleFocalPath], focalIndices: [4] });
+  });
+
+  it('checks only service when calculating focalIdx when only service is provided', () => {
+    const { almostDoubleFocalPath } = testResources;
+    parsedOutputValidator({ paths: [almostDoubleFocalPath], focalIndices: [2], groupOperations: true });
```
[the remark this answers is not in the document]
everett980: I was tying it to the queryParam that would enable this feature, but `ignoreFocalOperations` is more relevant to this test file.

packages/jaeger-ui/src/utils/DeepDependencyGraph/parse-payload.test.resources.js
```diff
@@ -0,0 +1,77 @@
...
+};
+
+const firstPathElem = simplePathElemMaker('first');
+const beforePathElem = simplePathElemMaker('before');
+const midPathElem = simplePathElemMaker('mid');
+const afterPathElem = simplePathElemMaker('after');
+const lastPathElem = simplePathElemMaker('last');
+
+export const simplePath = [firstPathElem, beforePathElem, focalPathElem, afterPathElem, lastPathElem];
+export const longSimplePath = pathLengthener([
```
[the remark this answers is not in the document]
everett980: that is what I meant to do, will update :+1:

packages/jaeger-ui/src/utils/DeepDependencyGraph/parse-payload.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+export default function parsePayload(
+  payload: TPayload,
+  { service: focalService, operation: focalOperation }: { service: string; operation?: string }
+): TParsedPayload {
+  const serviceMap: TServiceMap = {};
+  let visibilityIdx = 0;
+
+  const paths = payload.map(payloadPath => {
+    // path with stand-in values in order to have obj to which to assign memberOf for each pathElem.
```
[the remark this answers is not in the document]
everett980: `const path = { /* ... */ }` needs to be declared so that `const pathElem = { memberOf: path, /* ... */ }` works below.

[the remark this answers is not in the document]
everett980: changed to `const path = {} as TPath;` instead of having magic `-1` to remember, and avoids stand-in empty members array that is never used.

packages/jaeger-ui/src/utils/DeepDependencyGraph/parse-payload.tsx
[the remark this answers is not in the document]
tiffon: I think it also needs to be in the redux store to match with the URL.

packages/jaeger-ui/src/utils/DeepDependencyGraph/parse-payload.tsx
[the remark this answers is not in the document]
tiffon: :+1:

packages/jaeger-ui/src/utils/DeepDependencyGraph/parse-payload.test.js
tiffon: I think it's about preference. Personally, I lean toward native. But, to each their own.

packages/jaeger-ui/src/utils/DeepDependencyGraph/parse-payload.tsx
```diff
@@ -0,0 +1,112 @@
...
+  pathIdx: number;
+  visibilityIdx: number;
+};
+
+type TPath = {
+  focalIdx: number;
+  members: TPathElem[];
+};
+
+type TServiceMap = Record<string, TService>;
```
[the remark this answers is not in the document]
everett980: Good call, updated. Just a minor issue with typescript.
```
if (!serviceMap.has(serviceName)) {
  serviceMap.set(serviceName, { /* ... */ });
}
if (!serviceMap.get(serviceName).operations.has(operationName)) {
  /* ... */
}
```
TypeScript complains that `serviceMap.get(serviceName)` could be undefined, but of course it can't be.

packages/jaeger-ui/src/utils/DeepDependencyGraph/parse-payload.tsx
```diff
@@ -0,0 +1,112 @@
...
+// limitations under the License.
+
+type TPayload = {
+  operation: string;
+  service: string;
+}[][];
+
+type TService = {
+  name: string;
+  operations: Record<string, TOperation>;
```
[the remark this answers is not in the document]
everett980: Updated, but has same issue as previous comment.

packages/jaeger-ui/src/utils/DeepDependencyGraph/parse-payload.tsx
```diff
@@ -0,0 +1,112 @@
...
+          pathElems: [],
+        };
+      }
+      const pathElem = {
+        memberOf: path,
+        operation: serviceMap[service].operations[operation],
+        pathIdx: i,
+        visibilityIdx: visibilityIdx++,
+      };
+      Object.defineProperty(pathElem, 'distance', {
```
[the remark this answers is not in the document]
everett980: Updated to a class. Definitely cleaner, but the class definition is quite verbose. Not necessarily a con, but does warrant further thought into file structure.

packages/jaeger-ui/src/utils/DeepDependencyGraph/parse-payload.tsx
```diff
@@ -0,0 +1,112 @@
...
+          name: operation,
+          service: serviceMap[service],
+          pathElems: [],
+        };
+      }
+      const pathElem = {
+        memberOf: path,
+        operation: serviceMap[service].operations[operation],
+        pathIdx: i,
+        visibilityIdx: visibilityIdx++,
```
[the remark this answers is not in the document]
everett980: no it's not. Initially when padding out `0`s it didn't offer much value. But with empty values in the CSV, it does have value. Also it will help with `show/hide` (an) entire hop(s). I will revisit this.

packages/jaeger-ui/src/utils/DeepDependencyGraph/parse-payload.tsx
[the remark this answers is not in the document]
everett980: how come? i'm not sure what it is used for aside from validating that the BE response when sharing a link is the same response that the shared DDG was constructed from.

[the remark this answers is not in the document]
everett980: Also, seems like link sharing validation through the data hash is part of the Beta enhancement of shareable links and not part of the core MVP.

packages/jaeger-ui/src/utils/DeepDependencyGraph/parse-payload.tsx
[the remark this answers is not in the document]
everett980: visibilityIdx is now concentric!

packages/jaeger-ui/src/model/ddg/transform-ddg-data.tsx
```diff
@@ -0,0 +1,105 @@
...
+    const members = payloadPath.map(({ operation: operationName, service: serviceName }, i) => {
+      if (!serviceMap.has(serviceName)) {
+        serviceMap.set(serviceName, {
+          name: serviceName,
+          operations: new Map(),
+        });
+      }
+      const service = serviceMap.get(serviceName) as TDdgService;
+
+      if (!service.operations.has(operationName)) {
```
tiffon: A way to avoid extra checks and `as`:

```tsx
let service = serviceMap.get(serviceName);
if (!service) {
    service = {
        name: serviceName,
        operations: new Map(),
    };
    serviceMap.set(serviceName, service);
}
console.log(service.name); // no error - service is known to be non-null
```

The same approach can be used with `operation`.

packages/jaeger-ui/src/model/ddg/transform-ddg-data.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+export default function transformDdgData(
+  payload: TDdgPayload,
+  { service: focalService, operation: focalOperation }: { service: string; operation?: string }
+): TDdgTransformedDdgData {
+  const serviceMap: TDdgServiceMap = new Map();
+  const pathElemsByDistance: TDdgPathElemsByDistance = new Map();
+
+  const paths = payload.map(payloadPath => {
+    const path = {} as TDdgPath;
```
tiffon: `as` circumvents type-checking. So, I would suggest sticking with `-1`...?

packages/jaeger-ui/src/model/ddg/transform-ddg-data.tsx
```diff
@@ -0,0 +1,105 @@
...
+      throw new Error('A payload path lacked the focalNode');
+    }
+
+    path.members = members;
+    members.forEach(member => {
+      if (pathElemsByDistance.has(member.distance)) {
+        (pathElemsByDistance.get(member.distance) as PathElem[]).push(member);
+      } else {
+        pathElemsByDistance.set(member.distance, [member]);
+      }
```
tiffon: Same as above, the use of a local var can help typescript along:

```tsx
let members = pathElemsByDistance.get(member.distance);
if (members) {
  members.push(member);
} else {
  pathElemsByDistance.set(member.distance, [member]);
}
```

I think typescript just doesn't have a full appreciation for what `Map#has()` offers...

packages/jaeger-ui/src/model/ddg/transform-ddg-data.tsx
```diff
@@ -0,0 +1,105 @@
...
+    if (
+      (Math.abs(downstream) < upstream && pathElemsByDistance.has(downstream)) ||
+      !pathElemsByDistance.has(upstream)
+    ) {
+      nextArrayToIndex = pathElemsByDistance.get(downstream--) as PathElem[];
+    } else {
+      nextArrayToIndex = pathElemsByDistance.get(upstream++) as PathElem[];
+    }
+    nextArrayToIndex.forEach(setPathElemVisibilityIdx);
+  }
```
tiffon: Some comments on what's going on here would be helpful to our future selves...

packages/jaeger-ui/src/model/ddg/types.tsx
```diff
@@ -0,0 +1,78 @@
...
+export type TDdgTransformedDdgData = {
+  pathElemsByDistance: TDdgPathElemsByDistance;
+  paths: TDdgPath[];
+  services: TDdgServiceMap;
+};
+
+export class PathElem {
+  memberOf: TDdgPath;
+  operation: TDdgOperation;
+  pathIdx: number;
```
tiffon: At first I thought this field referred to the index of the parent path, `memberOf`, within the `TDdgTransformedDdgData`.

Maybe `memberIdx` would indicate otherwise?

packages/jaeger-ui/src/model/ddg/types.tsx
```diff
@@ -0,0 +1,78 @@
...
+  private _visibilityIdx?: number;
+
+  constructor({ path, operation, pathIdx }: { path: TDdgPath; operation: TDdgOperation; pathIdx: number }) {
+    this.memberOf = path;
+    this.operation = operation;
+    this.pathIdx = pathIdx;
+
+    operation.pathElems.push(this);
```
tiffon: Kind of strange for `PathElem` to add itself to `operation`. Currently, it needs to know `operation` is an object with a `pathElems` array property.

Moving the `.push()` in here instead of in the transform function doesn't seem worth the coupling.

packages/jaeger-ui/src/model/ddg/types.tsx
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  get distance() {
+    return this.pathIdx - this.memberOf.focalIdx;
```
tiffon: I like this, but on second thought, the `PathElem` knowing details about `TDdgPath` like `focalIdx`, and being dependent on how focal the index is represented, is unfortunate.

tiffon:
I'm curious about merging this directly into master. We can merge to master, incrementally, and hide behind a feature flag. What do you think?

packages/jaeger-ui/src/model/ddg/types.tsx
everett980: sure, will rename to `memberIdx`

Is there value in a `TDdgPath` knowing it's index in the `TDdgModel`? I've been building off of this PR for jag-500 (create edges and vertices for PPE with distinguished operations) and it seems like data structures that contain `PathElem`s has been more useful than data structures containing `Path`s.

packages/jaeger-ui/src/model/ddg/types.tsx
everett980: I was thinking of making `TDdgPath` a class as well, so that `focalIdx` can be a getter. The benefit of that would be the ability to enable/disable distinguishing operations without needing to recalculate all `focalIdx`s.
It is somewhat coupled, but the value is high. Through TypeScript (and good tests) it's pretty clear if/when this coupling is broken.

everett980:
> I'm curious about merging this directly into master. We can merge to master, incrementally, and hide behind a feature flag. What do you think?

That sounds fine to me, would we omit DDG PRs from the `CHANGELOG.md` until the feature flag is removed? If not, there should be some indication that the changes are experimental.

packages/jaeger-ui/src/model/ddg/types.tsx
everett980: I will move the `push` to `transform-ddg-data`

packages/jaeger-ui/src/model/ddg/transform-ddg-data.tsx
everett980: Okay, I'll revert to object with dummy-values.

packages/jaeger-ui/src/model/ddg/transform-ddg-data.tsx
everett980: Nice, I will out this approach.

packages/jaeger-ui/src/model/ddg/types.tsx
```diff
@@ -0,0 +1,78 @@
...
+    if (this._visibilityIdx == null) {
+      this._visibilityIdx = visibiliityIdx;
+    } else {
+      throw new Error('Visibility Index cannot be changed once set');
+    }
+  }
+
+  get visibilityIdx(): number {
+    if (this._visibilityIdx == null) {
+      throw new Error('Visibility Index was never set for this PathElem');
```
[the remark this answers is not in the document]
everett980: If it is a warning, it needs either a default return value, or to sometimes not return a number. Either option would result in some odd "edge-case" handling. A `PathElem` that lacks a `visibilityIdx` is an indication that an error occurred when creating `PathElem`s from the payload.

packages/jaeger-ui/src/model/ddg/types.tsx
```diff
@@ -0,0 +1,78 @@
...
+    this.pathIdx = pathIdx;
+
+    operation.pathElems.push(this);
+  }
+
+  get distance() {
+    return this.pathIdx - this.memberOf.focalIdx;
+  }
+
+  set visibilityIdx(visibiliityIdx: number) {
```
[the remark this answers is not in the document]
everett980: good catch, ran into that on my other branch, will fix.

packages/jaeger-ui/src/model/ddg/types.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+export type TDdgPathElemsByDistance = Map<number, PathElem[]>;
+
+export type TDdgTransformedDdgData = {
+  pathElemsByDistance: TDdgPathElemsByDistance;
+  paths: TDdgPath[];
+  services: TDdgServiceMap;
+};
+
+export class PathElem {
```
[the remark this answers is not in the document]
everett980: I was on the fence initially, SGTM.

packages/jaeger-ui/src/model/ddg/types.tsx
```diff
@@ -0,0 +1,78 @@
...
+export type TDdgPath = {
+  focalIdx: number;
+  members: PathElem[];
+};
+
+export type TDdgServiceMap = Map<string, TDdgService>;
+
+export type TDdgPathElemsByDistance = Map<number, PathElem[]>;
+
+export type TDdgTransformedDdgData = {
```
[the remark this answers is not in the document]
everett980: `Model` is fairly generic, but definitely better than `DdgTransformedDdgData`. Will update to `TDdgModel` but still remain open to a better name.

packages/jaeger-ui/src/model/ddg/types.tsx
```diff
@@ -0,0 +1,78 @@
...
+};
+
+export type TDdgPath = {
+  focalIdx: number;
+  members: PathElem[];
+};
+
+export type TDdgServiceMap = Map<string, TDdgService>;
+
+export type TDdgPathElemsByDistance = Map<number, PathElem[]>;
```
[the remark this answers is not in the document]
everett980: It lets the `Map` types be reused across components, it's been worth it on when consuming this output in the next story.

packages/jaeger-ui/src/model/ddg/transform-ddg-data.tsx
everett980: I think it's an issue with our TS config. This TS playground has no issue with the `if (!map.has(key)) map.set(key, value); map.get(key);` pattern:

https://www.typescriptlang.org/play/#src=const%20functionIndexer%3A%20Map%3C(any)%20%3D%3E%20any%2C%20number%20%3E%20%3D%20new%20Map()%3B%0D%0A%0D%0Afunction%20getFnIdx(fn%3A%20(any)%20%3D%3E%20any)%3A%20number%20%7B%0D%0A%20%20%20%20if%20(!functionIndexer.has(fn))%20%7B%0D%0A%20%20%20%20%20%20%20%20functionIndexer.set(fn%2C%20functionIndexer.size%3B%0D%0A%20%20%20%20%7D%0D%0A%20%20%20%20return%20functionIndexer.get(fn)%3B%0D%0A%7D

(Not sure why Github won't let me link to that using `[label](link)` even when using `ctrl + k`...)

everett980: Though I updated to the pattern you suggested and it works well. But then shortly below this in the `while` loop it is more convoluted than this change.

packages/jaeger-ui/src/model/ddg/transform-ddg-data.tsx
everett980: agreed, added comments

packages/jaeger-ui/src/model/ddg/types.tsx
everett980: Scratch what I said about `focalIdx` as a getter, since if the `focalIdx` changes as a result of the `focalNode` changing, all `visibilityIdx`s would be invalid.

packages/jaeger-ui/src/model/ddg/transform-ddg-data.tsx
```diff
@@ -12,64 +12,65 @@
...
-  TDdgOperation,
-  TDdgPath,
-  TDdgServiceMap,
-  TDdgPathElemsByDistance,
-  TDdgTransformedDdgData,
-  PathElem,
-} from './types';
+import { PathElem, TDdgModel, TDdgPayload, TDdgPath, TDdgPathElemsByDistance, TDdgServiceMap } from './types';
 
 export default function transformDdgData(
```
tiffon: Since we moved to the naming convention of filenames reflecting the default export, the file should be named `transformDdgData.tsx`. Unfortunately, that was not the initial approach taken, so it's not reflected, everywhere.

everett980: I see, I will rename. The usage of `kebab-case` vs `PascalCase` vs `camelCase` file names definitely seem inconsistent. It might be worth mentioning the style guide portion of `CONTRIBUTING.md`.

packages/jaeger-ui/src/model/ddg/transform-ddg-data.tsx
everett980: I agree that we should avoid `as` as much as possible as it generally indicates something is mistyped. Fortunately it isn't a [total sidestep](https://www.typescriptlang.org/play/#src=type%20TFoo%20%3D%20%7B%0D%0A%20%20%20%20foo%3A%20string%3B%0D%0A%7D%0D%0A%0D%0Atype%20TBar%20%3D%20%7B%0D%0A%20%20%20%20bar%3A%20string%3B%0D%0A%7D%0D%0A%0D%0Afunction%20barToFoo(bar%3A%20TBar)%3A%20TFoo%20%7B%0D%0A%20%20%20%20return%20bar%20as%20TFoo%3B%0D%0A%7D) but still should be avoided.

If we could somehow "teach" TS `Map#has` and any future related shortcomings of TS that'd be ideal, but failing that the code blocks in this thread serve as good examples for helping TS.

I updated the related code, but with a `do { /* ... */ } while (/* ... */);` loop instead and it works well.

packages/jaeger-ui/src/model/ddg/transform-ddg-data.tsx
```diff
@@ -78,6 +79,7 @@ export default function transformDdgData(
     return path;
   });
 
+  // Assign visibility indices such there is a positive, dependent correlation between visibilityIdx and distance
```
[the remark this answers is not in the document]
everett980: That sounds reasonable. I implemented the sort as:
```
const stringifyPayloadEntry = ({ service, operation }: TDdgPayloadEntry) => `${service}::${operation}`;

/* ... */
  const paths = payload
    .slice()
    .sort((a, b) =>
      a
        .map(stringifyPayloadEntry)
        .join()
        .localeCompare(b.map(stringifyPayloadEntry).join())
    )
```
It seems to work well and should work at scale. Did you have a different sort in mind?

packages/jaeger-ui/src/model/ddg/transform-ddg-data.tsx
tiffon: It probably is inconsistent due to not renaming everything when adopting the approach... The approach is to name based on default export, if there is one. If not, then a `descriptive-name` kebab-case name.

packages/jaeger-ui/src/model/ddg/transform-ddg-data.tsx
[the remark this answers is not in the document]
tiffon: Not so sure I'd say this will scale, localeCompare is dog slow:

Screen Shot 2019-05-22 at 1 55 11 PM

https://jsperf.com/localecompare/38

packages/jaeger-ui/src/model/ddg/transform-ddg-data.tsx
tiffon: Yeah, we need to put a bit of time into a style-guide. I started some notes, here:

https://github.com/jaegertracing/jaeger-ui/blob/style-guide-wip/CONTRIBUTING.md

tiffon: Created a ticket for it: https://github.com/jaegertracing/jaeger-ui/issues/385

packages/jaeger-ui/src/model/ddg/transformDdgData.tsx
```diff
@@ -0,0 +1,125 @@
...
+  TDdgModel,
+  TDdgPayload,
+  TDdgPayloadEntry,
+  TDdgPath,
+  TDdgPathElemsByDistance,
+  TDdgServiceMap,
+  TDdgVisibilityIdxToPathElem,
+} from './types';
+
+const stringifyPayloadEntry = ({ service, operation }: TDdgPayloadEntry) => `${service}::${operation}`;
```
tiffon: For lexicographical comparison, `:` is before some alphanumerics and after others:

- `a:op` is after `a0::op`
- `a:op` is before `aa::op`

I suggest using `\v` as the delimiter as it's very unlikely to be present in a service or operation name and is consistent in its lexicographical sorting for alphanumerics.

- `a\vop` is before `a0::op`
- `a\vop` is before `aa::op`

packages/jaeger-ui/src/model/ddg/transformDdgData.tsx
```diff
@@ -0,0 +1,125 @@
...
+  const serviceMap: TDdgServiceMap = new Map();
+  const pathElemsByDistance: TDdgPathElemsByDistance = new Map();
+
+  const paths = payload
+    .slice()
+    .sort((a, b) =>
+      a
+        .map(stringifyPayloadEntry)
+        .join()
+        .localeCompare(b.map(stringifyPayloadEntry).join())
```
tiffon: I added a note in another thread, but `localeCompare` has very poor performance: https://jsperf.com/localecompare/38

tiffon: Also, you might consider doing the stringify and join once, up-front. For large sets of paths, this might be quite a hit on perf.

packages/jaeger-ui/src/model/ddg/transform-ddg-data.tsx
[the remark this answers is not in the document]
everett980: Wow that is really slow! I had no idea. I will update it.
I think for our use of static ordering, not necessarily the most logical ordering (e.g.: `aZ` before `aa` is fine as long as it is consistent), `(a, b) => a > b ? 1 : -1` should be a sufficient comparison to avoid needing to convert to lowercase unnecessarily.

packages/jaeger-ui/src/model/ddg/transformDdgData.tsx
everett980: changed to `\v` so it's easier to reason about (which is important) but ultimately it doesn't affect the outcome since order rationale doesn't matter, just order consistency.

packages/jaeger-ui/src/model/ddg/transformDdgData.tsx
everett980: I put in a map to avoid `path.map(stringifyPayloadEntry).join()` on the same path twice. And changed from `a.localeCompare(b)` to `a > b ? 1 : -1`

PR #384, opened 2019-05-21
Issue 356 maximum lookback should be retention period

### LINKED ISSUE #356: The maximum search lookback should be the trace retention period
The lookback option in the search form, on the search page, should have the full retention period of traces as the max option. Currently, it is hardcoded to two days, but the retention period is configurable.

@vprithvi, @jaegertracing/elasticsearch is it possible for the query service to know this retention so it can be passed into the UI, dynamically? The alternative is to make a field in the UI config, but it would then need to be maintained as part of a deployment rather than adjusting, automatically.

everett980:
- Closes #356 

- Adds the ability to configure the lookback dropdown on the search page through `get-config.tsx`

### DIFF
--- a/packages/jaeger-ui/src/components/SearchTracePage/SearchForm.js
+++ b/packages/jaeger-ui/src/components/SearchTracePage/SearchForm.js
@@ -96,10 +96,6 @@ export const lookbackOptions = [
     value: '3h',
   },
   {
-    label: '4 Hours',
-    value: '4h',
-  },
-  {
     label: '6 Hours',
     value: '6h',
   },
@@ -144,12 +140,22 @@ export const lookbackOptions = [
 export const optionsWithinMaxLookback = _memoize(
   () => {
     const maxLookback = getConfigValue('search.maxLookback');
-    const { value: maxLookbackValue } = maxLookback;
     const now = new Date();
-    const minTimestamp = lookbackToTimestamp(maxLookbackValue, now);
-    const options = lookbackOptions.filter(({ value }) => lookbackToTimestamp(value, now) >= minTimestamp);
-    if (options[options.length - 1].value !== maxLookbackValue) {
-      options.push(maxLookback);
+    const minTimestamp = lookbackToTimestamp(maxLookback.value, now);
+    const lookbackToTimestampMap = new Map();
+    const options = lookbackOptions.filter(({ value }) => {
+      const lookbackTimestamp = lookbackToTimestamp(value, now);
+      lookbackToTimestampMap.set(value, lookbackTimestamp);
+      return lookbackTimestamp >= minTimestamp;
+    });
+    const lastInRangeIndex = options.length - 1;
+    const lastInRangeOption = options[lastInRangeIndex];
+    if (lastInRangeOption.label !== maxLookback.label) {
+      if (lookbackToTimestampMap.get(lastInRangeOption.value) !== minTimestamp) {
+        options.push(maxLookback);
+      } else {
+        options.splice(lastInRangeIndex, 1, maxLookback);
+      }
     }
     return options.map(({ label, value }) => (
       <Option key={value} value={value}>
@@ -224,7 +230,7 @@ export function submitForm(fields, searchTraces) {
   if (lookback !== 'custom') {
     const now = new Date();
     start = lookbackToTimestamp(lookback, now);
-    end = moment(now).valueOf() * 1000;
+    end = now * 1000;
   } else {
     const times = getUnixTimeStampInMSFromForm({
       startDate,
--- a/packages/jaeger-ui/src/components/SearchTracePage/SearchForm.js
+++ b/packages/jaeger-ui/src/components/SearchTracePage/SearchForm.js
@@ -14,10 +14,11 @@
 
 import * as React from 'react';
 import { Form, Input, Button, Popover, Select } from 'antd';
-import _memoize from 'lodash/memoize';
+import _get from 'lodash/get';
 import logfmtParser from 'logfmt/lib/logfmt_parser';
 import { stringify as logfmtStringify } from 'logfmt/lib/stringify';
 import moment from 'moment';
+import memoizeOne from 'memoize-one';
 import PropTypes from 'prop-types';
 import queryString from 'query-string';
 import IoHelp from 'react-icons/lib/io/help';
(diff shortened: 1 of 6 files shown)


packages/jaeger-ui/src/setupTests.js
```diff
@@ -25,3 +25,6 @@ const createSerializer = require('enzyme-to-json').createSerializer;
 
 Enzyme.configure({ adapter: new EnzymeAdapter() });
 expect.addSnapshotSerializer(createSerializer({ mode: 'deep' }));
+
+// Calls to get-config.tsx warn if this global is not a function
+window.getJaegerUiConfig = () => ({});
```
everett980: This removes 21 warnings when running tests on master, and this branch introduced many more before this change.

yurishkuro:
image

Use "resolves" or "closes" or "fixes", otherwise it's not being recognized by GitHub automatically.

everett980:
> Use "resolves" or "closes" or "fixes", otherwise it's not being recognized by GitHub automatically.
That explains why my previous PRs merely "mentioned" issues without closing them and I had to do it manually. Thanks!

packages/jaeger-ui/src/utils/config/get-config.test.js
```diff
@@ -22,19 +22,30 @@ import defaultConfig, { deprecations } from '../../constants/default-config';
 
 describe('getConfig()', () => {
   let oldWarn;
+  let oldWindowGetJaegerUiConfig;
   let warnFn;
 
-  beforeEach(() => {
+  beforeAll(() => {
     oldWarn = console.warn;
+    oldWindowGetJaegerUiConfig = window.getJaegerUiConfig;
```
tiffon: Seems like the change in `setupTests` will cause the old function to be the no op?

packages/jaeger-ui/src/components/SearchTracePage/SearchForm.js
```diff
@@ -71,6 +73,93 @@ export function convTagsLogfmt(tags) {
...
+  {
+    label: '2 Hours',
+    value: '2h',
+  },
+  {
+    label: '3 Hours',
+    value: '3h',
+  },
+  {
+    label: '4 Hours',
```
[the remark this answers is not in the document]
everett980: For some reason I thought it was pre-existing... Will remove :+1:

packages/jaeger-ui/src/components/SearchTracePage/SearchForm.js
```diff
@@ -71,6 +73,93 @@ export function convTagsLogfmt(tags) {
...
+    label: '3 Weeks',
+    value: '3w',
+  },
+  {
+    label: '4 Weeks',
+    value: '4w',
+  },
+];
+
+export const optionsWithinMaxLookback = _memoize(
```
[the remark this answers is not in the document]
everett980: `_memoize`'s comparison function returns a string, while `memoizeOne` expects a `boolean` which would necessitate keeping the comparison value on hand ourselves.

If the config values cannot change without a refresh (as opposed to just unlikely to change) then memoizeOne without any comparison function would work.

packages/jaeger-ui/src/components/SearchTracePage/SearchForm.js
```diff
@@ -71,6 +73,93 @@ export function convTagsLogfmt(tags) {
...
+  () => {
+    const maxLookback = getConfigValue('search.maxLookback');
+    const { value: maxLookbackValue } = maxLookback;
+    const now = new Date();
+    const minTimestamp = lookbackToTimestamp(maxLookbackValue, now);
+    const options = lookbackOptions.filter(({ value }) => lookbackToTimestamp(value, now) >= minTimestamp);
+    if (options[options.length - 1].value !== maxLookbackValue) {
```
everett980: These needs to check the timestamp, not the value.
E.g.: If the user configures `'1w'` it shouldn't be added just because `'1w' !== '7d'`

packages/jaeger-ui/src/components/SearchTracePage/SearchForm.test.js
```diff
@@ -126,6 +130,85 @@ describe('conversion utils', () => {
   });
 });
 
+describe('lookback utils', () => {
+  describe('lookbackToTimestamp', () => {
+    const hourInMicroseconds = 60 * 60 * 1000 * 1000;
+    const now = new Date();
+    const nowInMicroseconds = moment(now).valueOf() * 1000;
```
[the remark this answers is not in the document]
everett980: that works! will also update the component file as that's where I copied this line from.

packages/jaeger-ui/src/utils/config/get-config.test.js
everett980: yeah it does, but `get-config.test.js` doesn't need to worry about the global defined in `setupTests`. Should that change for whatever reason, we wouldn't want to have to change it here too.

packages/jaeger-ui/src/components/SearchTracePage/SearchForm.test.js
```diff
@@ -126,6 +130,85 @@ describe('conversion utils', () => {
...
+        expect(props.children[1]).toBe(expectedOptions[i].label);
+      });
+    });
+
+    it("includes config.search.maxLookback if it's not part of standard options", () => {
+      const configValue = {
+        label: '4 Days - configValue',
+        value: '4d',
+      };
+      const expectedOptions = [...lookbackOptions.slice(0, 9), configValue];
```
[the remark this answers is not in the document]
everett980: What would be the benefit of redefining the entries here? Though I think moving it to a constants file with an all-caps name may help.

packages/jaeger-ui/src/components/SearchTracePage/SearchForm.js
everett980: ended up slightly different, but handles cases of `'1w'` vs `'7d'`

packages/jaeger-ui/src/utils/config/get-config.test.js
everett980: Seems as those this store-n-restore is wholly unnecessary as `setupTests.js` is run before each file!
Just learned that from trial and error and [here](https://jestjs.io/docs/en/configuration#setupfilesafterenv-array)!

PR #386, opened 2019-05-24
create graph edges and vertices from parsed payload

everett980:
- Consumes `visibilityKey` and `ddgModel` to create edges and vertices for plexus.

- Create `visibilityKey` management util for creating, updating, and comparing `visibilityKey`s.
- Create class that adds and removes edges and vertices based on changing `visibilityKey`s

## To Discuss
- `DdgEVManager` errors can probably be warnings except for adding "disjoint" `pathElem`, or possibly ignored outright (no warning) in non-problematic cases (e.g.: trying to add `pathElem` that is already in EV)
- Memoization can probably be removed
- Should all class methods that aren't public be private? I think it would be worth adding a very restrictive configuration of [this eslint rule](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/explicit-member-accessibility.md), though I must admit I am biased towards verbose code when it adds clarity.
  - Configuration I would specify:
```
{
  accessibility: 'explicit',
  overrides: {
    accessors: 'explicit',
    constructors: 'no-public',
    methods: 'explicit',
    properties: 'explicit',
    parameterProperties: 'no-public'
  }
}
```


packages/jaeger-ui/src/model/ddg/ddgEdgesAndVertices.test.js
```diff
@@ -0,0 +1,329 @@
...
+    const simpleTestfocalPathElem = simpleDdgModel.paths[0].members[2];
+    const expectedKeyEntry = pathElem => `${pathElem.operation.service.name}::${pathElem.operation.name}`;
+    const expectedFocalPathElemKey = expectedKeyEntry(simpleTestfocalPathElem);
+    const emptyDdgEdgesAndVertices = new DdgEdgesAndVertices({ ddgModel: {} });
+
+    it('creates key for focal pathElem', () => {
+      expect(emptyDdgEdgesAndVertices.getVertexKey(simpleTestfocalPathElem)).toBe(expectedFocalPathElemKey);
+    });
+
+    it('creates key for downstream pathElem', () => {
```
everett980: this should be `'creates key for upstream pathElem'`

packages/jaeger-ui/src/model/ddg/ddgEdgesAndVertices.test.js
```diff
@@ -0,0 +1,329 @@
...
+    });
+  });
+
+  describe('convergent paths', () => {
+    const oneHopVisibilityKey = createVisibilityKey([0, 1, 2, 3, 4, 5]);
+    const fullVisibilityKey = changeVisibility({
+      visibilityKey: oneHopVisibilityKey,
+      showIndices: [6, 7, 8, 9],
+    });
+    const downstreamAndFocalValidationParams = [
```
everett980: I think this is `upstreamAndFocalValidationParams`...

packages/jaeger-ui/src/model/ddg/ddgEdgesAndVertices.test.js
```diff
@@ -0,0 +1,329 @@
...
+          ingressNeighborVisibilityIndices: [3],
+        },
+        {
+          visibilityIndices: [9],
+          ingressNeighborVisibilityIndices: [7],
+        },
+      ]);
+    });
+
+    it('removes pathElems without removing other pahtElems with same service & operation', () => {
```
everett980: typo `pahtElems`

packages/jaeger-ui/src/model/ddg/types.tsx
```diff
@@ -50,3 +51,24 @@ export type TDdgModel = {
   services: TDdgServiceMap;
   visibilityIdxToPathElem: TDdgVisibilityIdxToPathElem;
 };
+
+export type TDdgEdgeIdentifiers = 'egressEdges' | 'ingressEdges';
+
+export class DdgVertex {
```
everett980: This type should be in own file with test, and should be compatible with plexus vertex type

packages/jaeger-ui/src/model/ddg/types.tsx
```diff
@@ -50,3 +51,24 @@ export type TDdgModel = {
...
+
+  constructor({ key }: { key: string }) {
+    this.egressEdges = new Map();
+    this.key = key;
+    this.ingressEdges = new Map();
+    this.pathElems = new Set();
+  }
+}
+
+export type TDdgEdge = {
```
everett980: This type needs to be compatible with plexus edge type

yurishkuro:
@everett980 references to internal tickets (Jag 500) are confusing in oss, please link in the other direction (internal jira to github)

packages/jaeger-ui/src/model/ddg/DdgEdge.tsx
```diff
@@ -0,0 +1,26 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { TEdge } from '@jaegertracing/plexus/lib/types';
+import { DdgVertex } from './types';
+
+export default class DdgEdge implements TEdge<{}> {
```
everett980: I have waffled on if this should be a class that `implements TEdge<{ from: DdgVertex, to: DdgVertex }>` so that edges would have an easier and direct link to its endpoints.
I have decided against it because the data structures are pretty bi-directionally entangled and this extra bi-directional connection did not seem to have as much value as the other connections do.

packages/jaeger-ui/src/model/ddg/DdgVertex.tsx
```diff
@@ -0,0 +1,63 @@
...
+    this.ingressEdges = new Map();
+    this.pathElems = new Set();
+  }
+
+  toJSON() {
+    type TDigestibleEdges = Record<string, DdgEdge>;
+    const digestibleEgressEdges: TDigestibleEdges = {};
+    const digestibleIngressEdges: TDigestibleEdges = {};
+    this.egressEdges.forEach((edge, egreesNeighbor) => {
+      digestibleEgressEdges[egreesNeighbor.key] = edge;
```
everett980: this should probably just be `digestibleEgressEdges[egressNeighbor]` to use the `get [Symbol.toStringTag]` method below which better indicates that this is a JSON representation of a map of vertices to edges, not keys to edges.
Also, `egrees` and `ingrees` are typos

packages/jaeger-ui/src/model/ddg/DdgVertex.test.js
```diff
@@ -0,0 +1,81 @@
...
+    return vertex;
+  }
+
+  function makeTestEdge(from, to) {
+    const edge = new DdgEdge({ from, to });
+    from.egressEdges.set(to, edge);
+    to.ingressEdges.set(from, edge);
+  }
+
+  const twoHopUpstreamVertexKey = 'twoHopUpstreamVertexKey';
```
everett980: these keys don't need to be `const`s they can just be the argument to `makeTestVertex` directly...

packages/jaeger-ui/src/model/ddg/DdgEVManager.test.js
```diff
@@ -0,0 +1,353 @@
...
+   * @param {Object[]} expectedVertices - The vertices that the DdgEvManager should have.
+   * @param {number[]} expectedVertices[].visibilityIndices - The visibility indices that should all share one
+   *     DdgVertex.
+   * @param {number[]} expectedVertices[].egressNeighborVisibilityIndices - A single visibilityIdx is
+   *     sufficient to define a neighboring vertex. For each egressNeighborVisibilityIdx, the expectedVertex
+   *     should have an egressEdge from the expectedVertex to the vertex represented by the
+   *     egressNeighborVisibilityIdx.
+   * @param {number[]} expectedVertices[].ingressNeighborVisibilityIndices - A single visibilityIdx is
+   *     sufficient to define a neighboring vertex. For each ingressNeighborVisibilityIdx, the expectedVertex
+   *     should have an ingressEdge from the expectedVertex to the vertex represented by the
```
everett980: `egressEdge`s are `from the vertex represented by the ingressNeighborVisibilityIdx to the expectedVertex`, this was copied and not flipped...

packages/jaeger-ui/src/model/ddg/DdgEVManager.tsx
```diff
@@ -0,0 +1,152 @@
...
+  TDdgPathElemsByDistance,
+  TDdgPath,
+  TDdgServiceMap,
+} from './types';
+
+export default class DdgEVManager {
+  lastVisibilityKey: string;
+  pathElemsByDistance: TDdgPathElemsByDistance;
+  pathElemToVertex: Map<PathElem, DdgVertex>;
+  paths: TDdgPath[];
```
everett980: This is unused, should be removed until a use case arises for needing paths (which probably won't be in this component)

packages/jaeger-ui/src/model/ddg/DdgEVManager.tsx
```diff
@@ -0,0 +1,152 @@
...
+  // inbound/outbound edges are visible for a vertex, but maybe not as vertices could be densitiy-aware and
+  // provide that to this fn. could also be property on pathElem that gets set by addPathElems
+  // tl;dr may move in late-alpha
+  private getVertexKey = (pathElem: PathElem): string => {
+    const { distance, memberOf } = pathElem;
+    const { focalIdx, members } = memberOf;
+    const startIdx = Math.min(focalIdx, focalIdx + distance);
+
+    return members
+      .slice(startIdx, startIdx + Math.abs(distance) + 1)
```
everett980: I think this could  just be `members.slice(Math.min(pathElem.memberOf.focalIdx, pathElem.memberIdx), Math.max(pathElem.memberOf.focalIdx, pathElem.memberIdx) + 1).map(/* ... */).join('|');`
a bit clearer than `startIdx` and `Math.abs(distance) + 1` IMO

packages/jaeger-ui/src/model/ddg/PathElem.test.js
```diff
@@ -52,4 +54,58 @@ describe('PathElem', () => {
...
+    const operations = simplePath.map(({ operation, service }) => ({
+      name: operation,
+      service: {
+        name: service,
+      },
+    }));
+    const path = {
+      focalIdx: 2,
+    };
+    const members = operations.map((operation, i) => new PathElem({ memberIdx: i, operation, path }));
```
everett980: `members` can be made directly from `simplePath.map` rather than needing to `map` twice.

packages/jaeger-ui/src/model/ddg/DdgEVManager.tsx
```diff
@@ -0,0 +1,141 @@
...
+import { PathElem, DdgEdge, DdgVertex, TDdgModel, TDdgPathElemsByDistance, TDdgServiceMap } from './types';
+
+export default class DdgEVManager {
+  lastVisibilityKey: string;
+  pathElemsByDistance: TDdgPathElemsByDistance;
+  pathElemToVertex: Map<PathElem, DdgVertex>;
+  services: TDdgServiceMap;
+  edges: Set<DdgEdge>;
+  vertices: Map<string, DdgVertex>;
+  visibilityIdxToPathElem: Map<number, PathElem>;
```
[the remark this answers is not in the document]
everett980: I was just thinking that. I will convert to an array tomorrow and see how that feels (should be good though).

packages/jaeger-ui/src/model/ddg/DdgEdge.tsx
everett980: I figured `implements` added clarity that this class was intended to work with a plexus type.
I am planning on converting the `DdgEdge` class to just an object.
At that point I don't need `TDdgEdge` at all, I can just use `TEdge` from plexus.
Thoughts?

packages/jaeger-ui/src/model/ddg/DdgVertex.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+import { DdgEdge, PathElem } from './types';
+
+type TDdgVertex = {
+  egressEdges: Map<DdgVertex, DdgEdge>;
+  ingressEdges: Map<DdgVertex, DdgEdge>;
+  pathElems: Set<PathElem>;
+};
+
+export default class DdgVertex implements TVertex<TDdgVertex> {
```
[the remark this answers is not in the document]
everett980: It works implicitly without `implements`, but I am wondering if `implements` adds some explicit clarity that this class will be used/usable where plexus takes vertices.
Based on [this article you sent me earlier](https://medium.com/@martin_hotell/interface-vs-type-alias-in-typescript-2-7-2a8f1777af4c) it seems as though implements works just as well for types and interfaces.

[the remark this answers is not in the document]
tiffon: In this case it seems okay because the `TDdgVertex` is not exported. But, if it were, it would likely cause some confusion.

It's also somewhat confusing in that `TDdgVertex` is a partial of `DdgVertex`. So, maybe the following would be more clear?

```js
type TDdgVertex = TVertex<{
  egressEdges: Map<DdgVertex, DdgEdge>;
  ingressEdges: Map<DdgVertex, DdgEdge>;
  pathElems: Set<PathElem>;
}>;

export default class DdgVertex implements TDdgVertex {
//...
}
```

packages/jaeger-ui/src/model/ddg/DdgEdge.tsx
tiffon: Are you going to remove this class? I think just using `TEdge` from plexus makes sense.

packages/jaeger-ui/src/model/ddg/DdgVertex.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  constructor({ key }: { key: string }) {
+    this.egressEdges = new Map();
+    this.key = key;
+    this.ingressEdges = new Map();
+    this.pathElems = new Set();
+  }
+
+  toJSON() {
+    type TDigestibleEdges = Record<string, DdgEdge>;
```
tiffon: Why is this type defined in this method?

packages/jaeger-ui/src/model/ddg/DdgVertex.tsx
```diff
@@ -0,0 +1,63 @@
...
+  pathElems: Set<PathElem>;
+
+  constructor({ key }: { key: string }) {
+    this.egressEdges = new Map();
+    this.key = key;
+    this.ingressEdges = new Map();
+    this.pathElems = new Set();
+  }
+
+  toJSON() {
```
tiffon: This doesn't return JSON so the name `toJSON` is a misnomer.

packages/jaeger-ui/src/model/ddg/DdgEVManager.tsx
```diff
@@ -0,0 +1,137 @@
...
+              ? {
+                  from: connectedVertex,
+                  to: vertex,
+                }
+              : {
+                  from: vertex,
+                  to: connectedVertex,
+                }
+          );
+          vertex[pathElem.focalSideEdgesKey].set(connectedVertex, newEdge);
```
tiffon: If the connected vertex is on the edge, why use it as a key to the edge?

packages/jaeger-ui/src/model/ddg/DdgEVManager.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+  constructor({ ddgModel }: { ddgModel: TDdgModel }) {
+    this.edges = new Set();
+    this.lastVisibilityKey = '';
+    this.pathElemToVertex = new Map();
+    this.vertices = new Map();
+    this.visibilityIdxToPathElem = ddgModel.visibilityIdxToPathElem;
+  }
+
+  private addElems = (newIndices: number[]) => {
```
tiffon: Comments on addElems / removeElems would be useful.

packages/jaeger-ui/src/model/ddg/DdgEVManager.tsx
```diff
@@ -0,0 +1,137 @@
...
+  // provide that to this fn. could also be property on pathElem that gets set by addPathElems
+  // tl;dr may move in late-alpha
+  private getVertexKey = (pathElem: PathElem): string => {
+    const { memberIdx, memberOf } = pathElem;
+    const { focalIdx, members } = memberOf;
+
+    return members
+      .slice(Math.min(focalIdx, memberIdx), Math.max(focalIdx, memberIdx) + 1)
+      .map(({ operation }) => `${operation.service.name}::${operation.name}`)
+      .join('|');
```
tiffon: I recommend using `"\t"` for the service -> operation delimiter and `"\n"` for the join operator.

packages/jaeger-ui/src/model/ddg/types.tsx
```diff
@@ -41,12 +43,15 @@ export type TDdgPath = {
...
 
 export type TDdgModel = {
-  pathElemsByDistance: TDdgPathElemsByDistance;
+  distanceToPathElems: TDdgDistanceToPathElems;
   paths: TDdgPath[];
   services: TDdgServiceMap;
   visibilityIdxToPathElem: TDdgVisibilityIdxToPathElem;
 };
+
+export type TDdgEdgeIdentifier = 'egressEdges' | 'ingressEdges';
```
tiffon: Why not use an enum, here?

packages/jaeger-ui/src/model/ddg/DdgVertex.tsx
```diff
@@ -0,0 +1,63 @@
...
+    return {
+      egressEdges: digestibleEgressEdges,
+      key: this.key,
+      ingressEdges: digestibleIngressEdges,
+      pathElems: this.pathElems,
+    };
+  }
+
+  toString() {
+    return JSON.stringify(this.toJSON(), null, 2);
```
tiffon: Seems like this is what `toJSON` should return?

packages/jaeger-ui/src/model/ddg/DdgVertex.tsx
```diff
@@ -0,0 +1,63 @@
...
+  toJSON() {
+    type TDigestibleEdges = Record<string, DdgEdge>;
+    const digestibleEgressEdges: TDigestibleEdges = {};
+    const digestibleIngressEdges: TDigestibleEdges = {};
+    this.egressEdges.forEach((edge, egressNeighbor) => {
+      digestibleEgressEdges[egressNeighbor[Symbol.toStringTag]] = edge;
+    });
+    this.ingressEdges.forEach((edge, ingressNeighbor) => {
+      digestibleIngressEdges[ingressNeighbor[Symbol.toStringTag]] = edge;
+    });
```
tiffon: Can you add comments to the above?

packages/jaeger-ui/src/model/ddg/PathElem.tsx
```diff
@@ -52,4 +52,42 @@ export default class PathElem {
...
+      ...this.toJSONHelper(),
+      memberOf: {
+        focalIdx: this.memberOf.focalIdx,
+        members: this.memberOf.members.map(member => member.toJSONHelper()),
+      },
+    };
+  }
+
+  toString() {
+    return JSON.stringify(this.toJSON(), null, 2);
```
tiffon: Why override `toString` instead of returning this from `toJSON`?

packages/jaeger-ui/src/model/ddg/transformDdgData.tsx
```diff
@@ -18,36 +18,36 @@ import {
...
+  const pathsComparators: Map<TDdgPayloadEntry[], string> = new Map();
 
   const paths = payload
     .slice()
     .sort((a, b) => {
-      let aCompareValue = pathsComparisonMap.get(a);
-      if (!aCompareValue) {
-        aCompareValue = a.map(stringifyPayloadEntry).join();
-        pathsComparisonMap.set(a, aCompareValue);
+      let aComparator = pathsComparators.get(a);
```
tiffon: What was wrong with `aCompareValue` (aside from length)?

IIRC, a comparator is a thing that compares, e.g. a function or something with a `comparator.compare(a, b)` method?

packages/jaeger-ui/src/model/ddg/DdgEVManager.tsx
everett980: SGTM

packages/jaeger-ui/src/model/ddg/DdgEVManager.tsx
```diff
@@ -0,0 +1,137 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { compareKeys } from './visibility-key';
+
+import { PathElem, DdgEdge, DdgVertex, TDdgModel } from './types';
+
+export default class DdgEVManager {
+  private edges: Set<DdgEdge>;
+  private lastVisibilityKey: string;
```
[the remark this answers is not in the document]
everett980: `lastVisibilityKey` is used to reuse edges and vertices over the lifespan of the class, as a new visibilityKey is compared to this value. `prevVisKey` is probably a better name.

packages/jaeger-ui/src/model/ddg/DdgEdge.tsx
everett980: SGTM I will use `TEdge`. I think that change will also simplify something else as well.

packages/jaeger-ui/src/model/ddg/DdgEVManager.tsx
everett980: Using the vertices as keys makes it very easy to check if vertex `ƒ` has another vertex `A` as an ingress or egress edge (`ƒ.egressEdges.has(A)` or `ƒ.ingressEdges.has(A)`, respectively). It also makes it easy to remove edges when hiding `PathElem`s.
The other simple alternatives would be to either have:
1) a `Set<TEdge>` that would require checking if a specific vertexKey exists as a `to`/`from` on a set entry
or
2) a `Set<string>` of vertex keys to check if the vertexKey exists and then look it up in `DdgEVManager.edges`

packages/jaeger-ui/src/model/ddg/types.tsx
everett980: tbh I forget `enum` exists! we never used before because this "works"
but since this case is exactly what enums are for, updated.

packages/jaeger-ui/src/model/ddg/DdgVertex.tsx
everett980: moved to top of file

packages/jaeger-ui/src/model/ddg/transformDdgData.tsx
everett980: reverted to `aCompareValue`
I thought comparator was akin to measure as both a verb and a noun and was looking to shorten...

packages/jaeger-ui/src/model/ddg/DdgVertex.tsx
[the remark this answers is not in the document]
everett980: that is clearer.

packages/jaeger-ui/src/model/ddg/DdgEVManager.tsx
everett980: added

packages/jaeger-ui/src/model/ddg/DdgVertex.tsx
everett980: I added a comment block above the `toJSON` method. lmk if there are still points of ambiguity

packages/jaeger-ui/src/model/ddg/DdgVertex.tsx
everett980: let's consolidate the toJSON / toString talk in [one discussion](https://github.com/jaegertracing/jaeger-ui/pull/386#discussion_r292510033)

packages/jaeger-ui/src/model/ddg/PathElem.tsx
everett980: let's consolidate the toJSON / toString talk in [one discussion](https://github.com/jaegertracing/jaeger-ui/pull/386#discussion_r292510033)

packages/jaeger-ui/src/model/ddg/DdgEVManager.tsx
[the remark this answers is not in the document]
everett980: key might not be the best term for this concept

packages/jaeger-ui/src/model/ddg/DdgEVManager.tsx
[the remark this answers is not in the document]
everett980: converted to array and it works swimmingly

packages/jaeger-ui/src/model/ddg/PathElem.tsx
tiffon: Doh! [toJSON](https://tc39.es/ecma262/#sec-serializejsonproperty)

packages/jaeger-ui/src/model/ddg/DdgEVManager.tsx
```diff
@@ -0,0 +1,231 @@
...
+
+export default class DdgEVManager {
+  private edges: Set<TEdge>;
+  private edgeToFarSideOfEdgePathElems: Map<TEdge, Set<PathElem>>;
+  private farSideOfEdgePathElemsToEdge: Map<Set<PathElem>, TEdge>;
+  private prevVisKey: string;
+  private pathElemToFarSideOfEdgePathElems: Map<PathElem, Set<PathElem>>;
+  private pathElemToVertex: Map<PathElem, DdgVertex>;
+  private vertices: Map<string, DdgVertex>;
+  private visIdxToPathElem: PathElem[];
```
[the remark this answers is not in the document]
tiffon: Also, there is a lot of redundancy in the data. I think this adds to the complexity. It's an optimization that facilitates quick access but at the cost of maintaining more state, which needs to be aligned during updates, etc. As opposed to having a simpler representation of the state, which would make updates easier to implement and follow (albeit with more traversing instead of accessing via an index or key).

packages/jaeger-ui/src/model/ddg/DdgEVManager.tsx
```diff
@@ -0,0 +1,231 @@
...
+          }
+
+          // The pathElem needs to be associated with the pre-existing edge so that edges can be removed when
+          // hiding pathElems even if no vertices are removed, which is possible with certain density settings
+          elems.add(pathElem);
+          this.pathElemToFarSideOfEdgePathElems.set(pathElem, elems);
+        }
+      }
+    });
+  };
```
tiffon: The comments to `showElems` and `hideElems` are very useful! :+1:

packages/jaeger-ui/src/model/ddg/types.tsx
```diff
@@ -41,12 +42,16 @@ export type TDdgPath = {
 
 export type TDdgServiceMap = Map<string, TDdgService>;
 
-export type TDdgPathElemsByDistance = Map<number, PathElem[]>;
-export type TDdgVisibilityIdxToPathElem = Map<number, PathElem>;
+export type TDdgDistanceToPathElems = Map<number, PathElem[]>;
```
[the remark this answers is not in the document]
everett980: distance can be negative

packages/jaeger-ui/src/model/ddg/DdgEVManager.tsx
```diff
@@ -0,0 +1,231 @@
...
+  private vertices: Map<string, DdgVertex>;
+  private visIdxToPathElem: PathElem[];
+
+  constructor({ ddgModel }: { ddgModel: TDdgModel }) {
+    this.edges = new Set();
+    this.edgeToFarSideOfEdgePathElems = new Map();
+    this.farSideOfEdgePathElemsToEdge = new Map();
+    this.pathElemToFarSideOfEdgePathElems = new Map();
+    this.pathElemToVertex = new Map();
+    this.prevVisKey = '';
```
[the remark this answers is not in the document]
everett980: it's used by `getEdgesAndVertices` which in turn calls `compareKeys` to determine which edges and vertices need to be added/removed

packages/jaeger-ui/src/model/ddg/DdgEVManager.tsx
```diff
@@ -0,0 +1,231 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { TEdge } from '@jaegertracing/plexus/lib/types';
+
+import { compareKeys } from './visibility-key';
+
+import { EDdgEdgeKeys, PathElem, DdgVertex, TDdgModel } from './types';
+
+export default class DdgEVManager {
```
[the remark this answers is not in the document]
everett980: `TDdgModel` is simply the parsed payload that lives in state, while `DdgEVManager` uses a fixed model and changing `visibilityKey`s to create/remove edges and vertices. not sure how these two could be combined without having instances of a self-mutating class in the redux state?

packages/jaeger-ui/src/model/ddg/DdgEVManager.tsx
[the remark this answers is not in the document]
everett980: yeah it's definitely heavy in the amount of maps and bi-directional mappings. it adds complexity, but I think it will be required for performance when changing the visibility of many `pathElem`s at once, such as when hiding children of `pathElem` that has had its descendants fully expanded.

packages/jaeger-ui/src/model/ddg/DdgEVManager.tsx
```diff
@@ -0,0 +1,231 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { TEdge } from '@jaegertracing/plexus/lib/types';
+
+import { compareKeys } from './visibility-key';
```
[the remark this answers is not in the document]
everett980: so we'd replace `foo-key/fooKey` with `foo-codec/fooCodec`, `getBuckets` with `decode`, and `createKey` with `encode`?
SGTM and helps avoid the use of `key` which definitely has other meanings than "encoding"

packages/jaeger-ui/src/model/ddg/DdgEVManager.tsx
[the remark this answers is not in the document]
everett980: While falling asleep last night I realized that I can remove the interim step of `pathElemToFarSideOfEdgeElems: Map<pathElem, Set<PathElem>>`
edges can be associated with pathElems exactly like vertices are. `pathElemToEdge: Map<PathElem, TEdge>`, and back: `edgeToFarSideElems: Map<TEdge, Set<PathElem>>`

Furthermore, I realized that `this.edges` can be supplanted by `this.edgeToFarSideElems.keys()`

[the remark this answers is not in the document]
everett980: currently, these `private` properties are simply alphabetical. I think with the pruning in the previous comment and some grouping with comments it becomes fairly manageable:

```
// PathElems given to constructor that serve as underlying data of a graph
private readyonly visIdxToPathElem: PathElem[];

// over the lifespan of a manager, it is provided changing keys/codecs. The last processed key/codec is stored in order to avoid creating every collection of edges & vertices from scratch when large parts may be reused / unchanged.
private prevVisKey: string;

// Non-focal pathElems need to be connected to the focalNode. If non-focal pathElems are hidden, we need to know which edges may be removed.
private pathElemToEdge: Map<PathElem, TEdge>;
// Since a single edge may be shared by multiple pathElems, we need to track them all so that hiding one pathElem doesn't result in an edge being removed if other pathElems for that same edge are still visible
private edgeToFarSideElems: Map<TEdge, Set<PathElem>>;

// Visible pathElems need a vertex to be rendered
private pathElemToVertex: Map<PathElem, DdgVertex>;
// Since pathElems can share a vertex, vertices need to be tracked by their key
private vertices: Map<string, DdgVertex>;
```

I'm working on a separate branch three degrees removed so I didn't just make these changes and push, but I didn't want to lose these notes before our meeting tomorrow

packages/jaeger-ui/src/model/ddg/DdgEVManager.tsx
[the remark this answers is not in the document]
everett980: thinking about how to avoid storing this. would the alternative approach in `getEdgesAndVertices` be:
```
getEdgesAndVertices = (codec: string): { edges: TEdge[], vertices: TDdgVertex[] } => {
  const indices = _map(Array.from(this.pathElemToVertex.keys()), 'visibilityIdx');
  const existingCodec = encode(indices); // encode is new name for createKey
  const { shown, hidden } = compareCodecs({ new: codec, old: existingCodec });
  /* ... unchanged ... */
}
```
?

This seems inefficient compared to storing a variable. Another option instead of `compareCodecs` would be:
```
getEdgesAndVertices = (codec: string): { edges: TEdge[], vertices: TDdgVertex[] } => {
  const currIndices = _map(Array.from(this.pathElemToVertex.keys()), 'visibilityIdx');
  const newIndices = decode(codec);
  const { shown, hidden } = compareIndices({ new: newIndices, old: currIndices });
  /* ... unchanged ... */
}
```

packages/jaeger-ui/src/model/ddg/DdgEVManager.tsx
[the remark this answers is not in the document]
tiffon: The instance in redux state doesn't need to be mutable. It can return a new instance when anything is changed.

Currently, the instance retains a lot of state, but that state is mostly used internally. So, what's the advantage of it? Further, the state that is returned, the edges and vertices, is mutated, e.g. a `DdgVertex.pathElems` can have values added and removed from it. This breaks strict equality checking to determine if there have changes, which will be useful when we get to managing plexus graphs.

packages/jaeger-ui/src/model/ddg/DdgEVManager.tsx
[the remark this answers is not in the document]
tiffon: `codec` has the `encode` and `decode` functions, the resultant object of `decode` is not the `codec`.

packages/jaeger-ui/src/model/ddg/Graph.test.js
```diff
@@ -0,0 +1,234 @@
...
+        {
+          visIndices: [4],
+          focalSideNeighbors: [2],
+        },
+      ]);
+    });
+  });
+
+  describe('convergent paths', () => {
+    it('adds separate vertices for equal PathElems that have different focalPaths, even those with equal focalSideNeighbors', () => {
```
tiffon: What is a `focalPaths`?

packages/jaeger-ui/src/model/ddg/Graph.test.js
```diff
@@ -0,0 +1,234 @@
...
+        },
+      ]);
+    });
+
+    it('reuses edge when possible', () => {
+      const convergentGraph = new Graph({
+        ddgModel: convergentModel,
+      });
+      const sharedEdgeElem0 = convergentGraph.visIdxToPathElem[5];
+      const sharedEdgeElem1 = convergentGraph.visIdxToPathElem[4];
```
tiffon: Why not name these consts `elemA` and `elemB`?

packages/jaeger-ui/src/model/ddg/PathElem.tsx
```diff
@@ -52,4 +52,43 @@ export default class PathElem {
     }
     return this._visibilityIdx;
   }
+
+  get focalSideNeighbor(): PathElem | null {
+    if (!this.distance) return null;
+    return this.memberOf.members[this.memberIdx - this.distance / Math.abs(this.distance)];
```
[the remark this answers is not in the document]
everett980: nice, updated

packages/jaeger-ui/src/model/ddg/Graph.tsx
```diff
@@ -0,0 +1,136 @@
...
+  };
+
+  public getVisible = (visKey: string): { edges: TEdge[]; vertices: TDdgVertex[] } => {
+    const edges: Set<TEdge> = new Set();
+    const vertices: Set<TDdgVertex> = new Set();
+    decode(visKey).forEach(visIdx => {
+      const pathElem = this.visIdxToPathElem[visIdx];
+      if (pathElem) {
+        const edge = this.pathElemToEdge.get(pathElem);
+        if (edge) edges.add(edge);
```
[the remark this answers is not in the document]
everett980: The order only matters in the constructor, and since `visIdxToPathElem` is an array, it is already known that the order is pertinent.

packages/jaeger-ui/src/model/ddg/Graph.test.js
everett980: discussed IRL, but for the future, `focalPath` is the DDG equivalent to `ancestralPath`. Ancestral Path is based off of the trace's root span, while `focalPath` is based off of the focalNode

PR #388, opened 2019-06-04
Wrap span tag values

epkugelmass:
Long tag values are difficult to read as they are rendered in a scroll box.

Wrap the tag value instead.

From

image

to

image

Signed-off-by: Elan Kugelmass <epkugelmass@gmail.com>

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/index.css
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/index.css
@@ -22,7 +22,7 @@ limitations under the License.
   line-height: 17px;
   font-size: 13px;
   font-family: monospace;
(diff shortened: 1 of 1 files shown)


yurishkuro:
I can see arguments for either type of display, e.g. a huge value that's not relevant to you at the moment will mess up the display. Maybe this should be toggled somehow.

epkugelmass:
> I can see arguments for either type of display, e.g. a huge value that's not relevant to you at the moment will mess up the display. Maybe this should be toggled somehow.

I agree with you. This measly PR represents the limit of what I'm able to contribute to this project right now. Maybe we should close this in favor of a more detailed issue.

tiffon:
@yurishkuro values don't get very large due to truncation.

PR #393, opened 2019-06-13
Upgrade a few packages

tiffon:

A few out of date packages are being flagged.

Update:

- js-yaml
- sinon (latest doesn't use js-yaml)
- fstream
- tar
- handlebars

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/Tween.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/Tween.tsx
@@ -57,7 +57,9 @@ export default class Tween {
       this.callbackComplete = onComplete;
       this.callbackUpdate = onUpdate;
       if (delay) {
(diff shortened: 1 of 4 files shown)


packages/jaeger-ui/src/components/TracePage/Tween.tsx
```diff
@@ -57,7 +57,9 @@ export default class Tween {
       this.callbackComplete = onComplete;
       this.callbackUpdate = onUpdate;
       if (delay) {
-        this.timeoutID = setTimeout(this._frameCallback, delay);
+        // @types/node sets the return type of setTieout to NodeJS.Timeout, so prefix
```
everett980: Nice work around. Some thoughts
1) typo in `setTieout`, should be `setTimeout`
2) is it worth documenting this somewhere
3) minor but I believe this can be a one line comment

PR #403, opened 2019-07-02
Adds a configuration option to customize the standard datetime format

leogomes:
Signed-off-by: Leonardo FREITAS GOMES <leonardo.f.gomes@gmail.com>

- Current datetime format used is not precise. It shows the trace start time with a minute precision.

- Adds a new configuration option that allows users to specify the datetime format they want to use, so that users can display time using format options like fractional second and timezone.


yurishkuro:
Can you provide a screenshot? Which timestamps is this affecting?

Also, there should  be some documentation about the config file format, it needs to be updated to describe the new property.

leogomes:
Here's a possible new date format set:
image

and here's the original one:
image

This only affects this timestamp now. With my change, the STANDARD_DATETIME_FORMAT is customized, so it could affect other dates in the future. So far, it seems to be only used by the "Trace Start" header field.

leogomes:
For documentation update, it would be a separate PR here: https://github.com/jaegertracing/documentation/blob/master/content/docs/next-release/frontend-ui.md

right?

packages/jaeger-ui/src/constants/default-config.tsx
```diff
@@ -66,6 +66,7 @@ export default deepFreeze(
         gaID: null,
         trackErrors: true,
       },
+      standardDatetimeFormat: FALLBACK_STANDARD_DATETIME_FORMAT,
```
tiffon: Since this is used in only one place, I don't think it's the "standard" date time format.

leogomes: I called it like that, because it overrides the format used by the STANDARD_DATETIME_FORMAT constant. Today, it's only used in one place, but if other bits of the UI start using https://github.com/jaegertracing/jaeger-ui/blob/master/packages/jaeger-ui/src/utils/date.tsx#L64 they will also be affected. I'm fine with calling it "traceStartDateFormat" too :)

leogomes:
> Instead of adding a config value, maybe it makes sense to change the trace page header to show a more precise datetime, instead? What is your preferred level of precision?

Up to fractional seconds would be a nice format: 'MMMM Do YYYY, HH:mm:ss.SSS' 
image

packages/jaeger-ui/src/constants/default-config.tsx
leogomes: @tiffon What do you suggest? Shall I remove the config option and add the fractional seconds to the format as shown below? Or should I keep the config and just rename the variable to remove the standard_ prefix?

tiffon:
@yurishkuro The fractional seconds work for you? I think it looks good and is preferable over another config option.

yurishkuro:
It's fine. Using smaller font for decimals is another option (but harder to implement).

tiffon:
@leogomes 	Thanks for your help with this.

Reducing the contrast might be preferable over making the text smaller; it's already pretty small.

Screen Shot 2019-07-10 at 11 41 16 PM

That was done using the `u-tx-muted` util CSS class:

```tsx
  {
    key: 'timestamp',
    label: 'Trace Start',
    renderer: (trace: Trace) => {
      const dateStr = formatDatetime(trace.startTime);
      const match = dateStr.match(/^(.+)(\.\d+)$/);
      return match ? <>{match[1]}<span className="u-tx-muted">{match[2]}</span></> : dateStr;
    }
  },
```

leogomes:
Hello @tiffon, @yurishkuro, 

What's the status on this one? Have @tiffon changes been merged somewhere and we can close this one?

Thanks,
Leo

yurishkuro:
@tiffon any more feedback?

tiffon:
@leogomes 	Thanks. I put together a PR for it: 

https://github.com/jaegertracing/jaeger-ui/pull/430

LMK if you have any feedback.

I think this PR is good to close.

PR #405, opened 2019-07-03
Ddg graph node

tiffon:
Add graph node rendering for the DDG graph.

Screen Shot 2019-06-24 at 2 08 00 PM

Screen Shot 2019-06-24 at 2 08 12 PM


packages/jaeger-ui/src/components/DeepDependencyGraph/getNodeLabel/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+const noops = {
+  setViewModifier(vertexKey: string, viewModifier: EViewModifier, enabled: boolean) {
+    // eslint-disable-next-line no-console
+    console.log(`set view modifier: ${enabled ? 'on' : 'OFF'} ${viewModifier} -- ${vertexKey}`);
+  },
+};
+
+export default function getNodeLabel(vertex: TDdgVertex) {
+  const { key, operation, service } = vertex;
+  const isFocalNode = service.length % 3 === 0;
```
everett980: since `model/ddg/Graph.tsx` was updated, this value should be destructured from the vertex.

packages/jaeger-ui/src/components/DeepDependencyGraph/getNodeLabel/index.test.js
```diff
@@ -0,0 +1,35 @@
...
+describe('getNodeLabel()', () => {
+  it('returns a <DdgNode/>', () => {
+    const operation = 'the-operation';
+    const service = 'the-service';
+    const ddgNode = getNodeLabel({ operation, service });
+    expect(ddgNode).toBeDefined();
+    const wrapper = shallow(ddgNode);
+    expect(wrapper).toMatchSnapshot();
+    // set as the focal node
+    // TODO(joe): update test after vertex indicates focal node or not
```
everett980: I think this test can be enabled now

packages/jaeger-ui/src/components/DeepDependencyGraph/getNodeLabel/DdgNode.tsx
```diff
@@ -0,0 +1,68 @@
...
+import cx from 'classnames';
+
+import focalNodeIcon from './focalNodeIcon';
+import setFocusButtonIcon from './setFocusButtonIcon';
+import NewWindowIcon from '../../common/NewWindowIcon';
+import { EViewModifier } from '../../../model/ddg/types';
+
+import './DdgNode.css';
+
+type TProps = {
```
everett980: would it make sense for this to be `type TProps = TDdgVertex & { focalNodeUrl..., viewModifierStuff }`?

packages/jaeger-ui/src/components/DeepDependencyGraph/getNodeLabel/index.tsx
tiffon: Great catch, this is a strange one.

packages/jaeger-ui/src/components/DeepDependencyGraph/getNodeLabel/DdgNode.tsx
tiffon: The props don't quite line up:

- `vertexKey` vs `key` (which can't be named `key` as a prop)
- `label` is not present

tiffon:
@everett980 This is ready for another review. Thanks.

PR #406, opened 2019-07-03
DDG service operation selectors

tiffon:

Allow users to select a service and operation for the DDG.

Adds service and operation selector components.

Previously, fetching the data for the DDG was implemented in the header. This has been removed from `Header` and placeholder data is being supplied via `DeepDependencies/index`. The main implementation of this is not provided in this PR.

-----

### ↓ Initial state

The operation selector is omitted because a service is not selected.

Screen Shot 2019-07-03 at 4 22 16 PM

-----

### ↓ Dropdown is open

Screen Shot 2019-07-03 at 4 22 24 PM

-----

### ↓ A search term is entered into the dropdown

Screen Shot 2019-07-03 at 4 22 50 PM

-----

### ↓ Acronym based search term

Screen Shot 2019-07-03 at 4 22 55 PM

-----

### ↓ A service is selected, operation selector is shown

Screen Shot 2019-07-03 at 4 23 09 PM

-----

### ↓ Search term for the operation is entered

Screen Shot 2019-07-03 at 4 23 22 PM

-----

### ↓ Both service and operation are selected

Screen Shot 2019-07-03 at 4 23 29 PM


packages/jaeger-ui/src/components/common/FilteredList/index.test.js
```diff
@@ -0,0 +1,153 @@
...
+      jest.useFakeTimers();
+      indices = {
+        visibleStartIndex: 1,
+        visibleStopIndex: props.options.length - 1,
+      };
+      wrapper.instance().onListItemsRendered(indices);
+      jest.runAllTimers();
+    });
+
+    afterEach(jest.useRealTimers);
```
everett980: should this be in `afterAll` (and the `useFakeTimers` can be in `beforeAll`)?

packages/jaeger-ui/src/components/common/FilteredList/highlightMatches.tsx
```diff
@@ -0,0 +1,131 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import * as React from 'react';
+import memoize from 'lru-memoize';
+
+enum EHighlightCadence {
```
everett980: I still think `StartsWith` is clearer than `Cadence`. I see now how there are more cases than `[match, notMatch]` and `[notMatch, match, otherNotMatch]`, but it still boils down to `StartsWith` match, and then alternates from there.

Which I suppose is the same as even/odd, but somehow `Cadence` implies that it's possible there could be some option like `EveryThird = 3,` or some such. So I guess it's just semantics...

tiffon: I did away with the `EHighlightCadence` and instead added a function that determines if a capture group should be highlighted or not:

```ts
type TIsHighlightGroupFn = (index: number) => boolean;
```

This is now a param to `toHighlights`:

```ts
function toHighlights(isHighlightGroup: TIsHighlightGroupFn, match: RegExpMatchArray) { ... }
```

Which uses it to determine if a group should be highlighted (skipping group 0, which is the full match):

```ts
      if (isHighlightGroup(i)) {
        return <mark key={`${tx + i}`}>{tx}</mark>;
      }
      return tx;
```

Then, in place of the enum there are:

```ts
// In the matches from the regular expressions created in `MatchHighlighter`,
// text that should be highlighted is interleaved with text that shouldn't be
// highlighted. E.g. alternating capturing groups should be highlighted,
// regardless of how many capturing groups there are in the match. Therefore, we
// need to establish if the first capturing group (i.e. match[1]) should be
// highlighted or if it should start at match[2]. Then, we alternate, from there.
// This results in either highlighting even groups (starting at match[2]) or odd
// groups (starting at match[1]). However, we don't actually need to highlight
// odd groups, but just the first group (match[1]), so we can simplify that test.
const highlightEvenGroups = toHighlights.bind(null, (index: number) => index % 2 === 0);
const highlightFirstGroup = toHighlights.bind(null, (index: number) => index === 1);
```

Which are used directly, in the `MatchHighlighter#matchers` array.

I think this is an improvement, but LMK if you disagree.

PR #411, opened 2019-07-10
Add an ESLint rule requiring the names of interfaces to be prefixed with "I"

tiffon:

Make it an error to have an interface without `I` instead of the current functionality which is to error if it is prefixed with an `I`.

Add an eslint rule and rename existing interfaces.

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineRow.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineRow.tsx
@@ -16,19 +16,19 @@ import * as React from 'react';
 
 import './TimelineRow.css';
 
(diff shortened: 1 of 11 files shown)


packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineRow.tsx
```diff
@@ -16,19 +16,19 @@ import * as React from 'react';
 
 import './TimelineRow.css';
 
-type TimelineRowProps = {
+type TTimelineRowProps = {
```
everett980: Are leading `T`s for types being enforced along with `I`s for interfaces or was this just because you caught it?
If not, we should find a rule for that as well.

PR #415, opened 2019-07-16
Remove patch check from master branch

tiffon:

Screen Shot 2019-07-16 at 5 24 22 PM

Currently if a PR that's merged to master has less than 90% coverage the Jaeger UI repo shows up as failing some checks even if the coverage for the project is above 90%.

This PR endeavor to remove the `patch` check from the master branch.

Add a `.codecov.yml` config file and configure to only apply the patch check to pull requests.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,5 @@
+coverage:
+  status:
+    project:
+      default:
+        only_pulls: true


yurishkuro:
Patch check is not required for merge
image

tiffon:
@yurishkuro The intent of this PR is to remove it from the overall repo status check, which drives the green checkbox or red x next to the repo name in the header, by limiting the `patch` check to pull requests.

Screen Shot 2019-07-16 at 5 24 22 PM

yurishkuro:
I still don't follow. Patch == PR, is it not? Or do you mean we're getting this as red on master builds?

tiffon:
@yurishkuro I believe the `patch` status check runs on master based on the most recent merge. That's my theory, anyway. 

The field `only_pulls` is from the docs on [commit status](https://docs.codecov.io/docs/commit-status#section-patch-status), which show the default config:

```yaml
coverage:
  status:
    patch:
      default:
        # basic
        target: auto
        threshold: null
        base: auto 
        # advanced
        branches: null
        if_no_uploads: error
        if_not_found: success
        if_ci_failed: error
        only_pulls: false
        flags: null
        paths: null
```

tiffon:
@yurishkuro I'm not so sure it worked. Will verify with the next PR merged. The config might replace the default for `project` instead of merging with it.

PR #420, opened 2019-07-22
Implement non-measured layers of nodes to plexus

tiffon:

Add support for rendering supplemental layers of nodes in plexus graphs.

Nodes can now also be rendered as layers of non-measured nodes. Non-measured means they take the position and size of nodes in as parameters to the render function. This is handy for decorating the main nodes with supplemental visual elements, such as UX based highlights or borders that scale well when zoom is being used.

Other changes:

- Add a comparison of using SVG vs HTML for borders and emphasis on nodes to the demo; show as both medium and small sized graphs

- Consolidate common code to the `HtmlLayer` and `SvgLayer` components

- Switch to the more succinct `getClassName` prop instead of `classNamePrefix` for internal components

- Allow `SvgEdgesLayer` to have `defs` prop when it's a standalone layer

## Screenshots

For the screenshots, we're adding a 1 pixel border to the nodes and highlighting nodes with labels that start with a vowel.

The HTML version of the border and emphasis is rendered with a 1 pixel border on the node and the vowel emphasis is via a padded, striped background with a one pixel dashed outline.

For the version which uses SVG to look better when zoomed out:

- The border is a rect with a 2 pixel stroke behind the main node's HTML content (causing 1 pixel to be viewable)
- The vowel emphasis has three layers:
  - Top layer: A SVG rect with a 10 pixel stroke (5 pixels are viewable)
  - Middle layer: The same HTML padded background as the HTML version
  - Bottom layer: A SVG rect with a 12 pixel stroke (causing 1 pixel to be viewable)

The rationale for the SVG + HTML version is the HTML background will be visible when the node is not zoomed out. When it is zoomed out, the HTML padded background will become small and the strokes will stay viewable. This allows the detailed styling of the HTML version to be seen while near normal scale but for the emphasis, in general, to still be visible when the scale is small.

### The HTML version

Normal zoom ↓

Screen Shot 2019-07-22 at 12 21 45 PM

Zoomed out to a medium size ↓

Screen Shot 2019-07-22 at 12 36 35 PM

Zoomed out to a small size ↓

Screen Shot 2019-07-22 at 12 48 52 PM

----

### The SVG + HTML version

Normal zoom ↓

Screen Shot 2019-07-22 at 12 35 54 PM

Zoomed out to a medium size ↓

Screen Shot 2019-07-22 at 12 36 44 PM

Zoomed out to a small size ↓

Screen Shot 2019-07-22 at 12 49 03 PM


packages/plexus/src/LayeredDigraph/SvgLayer.tsx
```diff
@@ -20,40 +20,69 @@ import { assignMergeCss, getProps } from './utils';
...
                 graphState={graphState}
               />
             ))}
           </defs>
         )}
+        {children}
+      </g>
+    );
+    if (extraWrapper) {
+      content = <g {...extraWrapper}>{content}</g>;
```
everett980: since the consumer can add props to the "standard" `g`, what is the use case for the extra `g`?

everett980: Ultimately, this results in:
```
<svg>
  <g className={ consumerDefinedClassName } style={ CONSTANT_STYLE }>
    <g className={ anotherConsumerDefinedClassName } transform={ basedOnZoomManager }>
      <g { ...extraWrapperPropsDefinedByConsumer }>
        <g { ...containerPropsAlsoDefinedByConsumer }>
          {defs /* ... */}
          {children}
        </g>
      </g>
    </g>
  </g>
</svg>
```
I don't know if there is a downside to having multiple nested `<g>`s, but it seems like this is more complicated than necessary. It leaves me wondering when a consumer would define something through `setOnContainer` vs `extraWrapper`, and which of the two `classNames` that I defined I should style.

tiffon: `extraWrapper` is for internal use only; it's not part of the public API.. So, the public API is not ambiguous, the user would use `setOnContainer`. 

Mainly, `extraWrapper` so the edges can have the `stroke` defaulted to `#000` without it inadvertently overriding the CSS styling set through the public API. E.g. the container the user applies `setOnContainer` to inherits the `stroke="#000"` but can override it via CSS styling. 

Right now there is a `<g>` for each thing that needs to be accomplished. To reduce the nesting would entail combining some `<g>` elements, which will be a bit more complex to map the code to the DOM elements.

I agree the levels of nesting could be reduced. But I'm pretty sure it's still very light as this happens on the scale of the number of layers not the number of vertices or edges in the graph. I would think the perf impact is negligible.

packages/plexus/src/LayeredDigraph/types.tsx
```diff
@@ -143,8 +146,8 @@ export type TLayer<T = {}, U = {}> = TOneOfFour<
 >;
 
 export type TMeasurableNodeProps<T = {}> = Omit<TMeasurableNodeRenderer<T>, 'measurable'> & {
-  classNamePrefix?: string;
-  hidden?: boolean;
+  getClassName: (name: string) => string;
```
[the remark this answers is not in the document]
tiffon: (Damn, my first comment was lost due to flaky internet)

The public API is `clasNamePrefix` which defaults to `plexus`. 

Only internal components use `getClassName` which is basically just:

```tsx
const { classNamePrefix: prefix = 'plexus' } = this.props;
const getClassName = (name: string) = `${prefix} ${prefix}--LayeredDigraph--${name}`;

// e.g.
getClassName(`SvgEdge`) === 'plexus plexus--LayeredDigraph--SvgEdge';
```

It just serves to shorten and the class name generation and ensure consistency.

PR #421, opened 2019-07-22
Raise real ddg request

everett980:
This PR removes the consumption of `sample-paths.test.resources.js` and instead raises the request to jaeger-query to get a DDG.

The structure of the DDG payload is slightly different now, so tests were updated to provide this new structure and a super simple `map` was added to `transformDdgData`.

### DIFF
--- a/packages/jaeger-ui/src/model/ddg/visibility-codec.test.js
+++ b/packages/jaeger-ui/src/model/ddg/visibility-codec.test.js
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 import { encode, decode, encodeDistance } from './visibility-codec';
-import { focalPayloadElem, longSimplePath, shortPath, simplePath } from './sample-paths.test.resources';
+import { focalPayloadElem, longSimplePath, shortPath, simplePath, wrap } from './sample-paths.test.resources';
 import transformDdgData from './transformDdgData';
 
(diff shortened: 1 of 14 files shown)


packages/jaeger-ui/src/model/ddg/visibility-codec.test.js
```diff
@@ -57,8 +57,8 @@ describe('visibility-codec', () => {
   });
 
   describe('encodeDistance', () => {
-    const ddgModel = transformDdgData([longSimplePath, simplePath], focalPayloadElem);
-    const shortModel = transformDdgData([shortPath], focalPayloadElem);
+    const ddgModel = transformDdgData([longSimplePath, simplePath].map(wrap), focalPayloadElem);
+    const shortModel = transformDdgData([shortPath].map(wrap), focalPayloadElem);
```
tiffon: Why not just wrap in the test resources file?

everett980: the paths are used when validating as well, so the options are either:
* export wrapped paths, and need to unwrap in more places than it's used wrap
* export both unwrapped and wrapped paths, but that becomes more variables to keep track of
* export unwrapped paths, and wrap right before feeding to transform, which feels the cleanest due to the other uses of unwrapped paths

PR #424, opened 2019-07-23
Wire service & operation selectors to redux & url

everett980:
### Changes:
Service & Operation selectors now get their options from redux, and get their values from the url.


packages/jaeger-ui/src/components/DeepDependencies/Header/index.tsx
```diff
@@ -13,70 +13,91 @@
...
+    }
+  }
 
-  setOperation = (operation: string) => {
-    // istanbul ignore next
-    this.setState({ operation });
+  setService = (service: string) => {
+    const { fetchServiceOperations, setService, operationsForService } = this.props;
+    if (!Reflect.has(operationsForService, service)) {
+      fetchServiceOperations(service);
```
tiffon: Having the `Header` fetch the services or operations, if they're needed, is inconsistent with `index.tsx` fetching the deps-graph data instead `Graph` taking care of that.

What do you think?

tiffon: Ping...

everett980: just pushed the updated version

everett980:
Good point, it could possibly save on renders (e.g.: when changing service before `operationsForService` updates)
I'll update on the UiFind branch.

PR #430, opened 2019-08-04
Show seconds in trace start time on the trace page

tiffon:

Trace start time, on the trace page, is not granularity enough.

Show the seconds and sub-seconds of the trace start time in the trace page header.

The seconds and sub-seconds are presented with reduced contrast:

Screen Shot 2019-08-03 at 11 02 48 PM

The seconds and sub-seconds are shown at full contrast when the datetime value is hovered:

Screen Shot 2019-08-03 at 11 02 55 PM

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageHeader.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageHeader.tsx
@@ -68,7 +68,18 @@ export const HEADER_ITEMS = [
   {
     key: 'timestamp',
     label: 'Trace Start',
-    renderer: (trace: Trace) => formatDatetime(trace.startTime),
+    renderer: (trace: Trace) => {
+      const dateStr = formatDatetime(trace.startTime);
+      const match = dateStr.match(/^(.+)(:\d\d\.\d+)$/);
+      return match ? (
+        <span className="TracePageHeader--overviewItem--value">
+          {match[1]}
+          <span className="TracePageHeader--overviewItem--valueDetail">{match[2]}</span>
+        </span>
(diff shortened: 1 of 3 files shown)


packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageHeader.tsx
```diff
@@ -68,7 +68,18 @@ export const HEADER_ITEMS = [
   {
     key: 'timestamp',
     label: 'Trace Start',
-    renderer: (trace: Trace) => formatDatetime(trace.startTime),
+    renderer: (trace: Trace) => {
+      const dateStr = formatDatetime(trace.startTime);
+      const match = dateStr.match(/^(.+)(:\d\d\.\d+)$/);
```
yurishkuro: I would only lower the contrast for sub-seconds, not for seconds

packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageHeader.tsx
```diff
@@ -68,7 +68,18 @@ export const HEADER_ITEMS = [
...
+      return match ? (
+        <span className="TracePageHeader--overviewItem--value">
+          {match[1]}
+          <span className="TracePageHeader--overviewItem--valueDetail">{match[2]}</span>
+        </span>
+      ) : (
+        dateStr
```
yurishkuro: should this also use `<span>`  for consistency?

tiffon: No, it's better to leave it off if it's not serving a purpose.

PR #436, opened 2019-08-08
Fixes handling of numeric strings in span tag values

### LINKED ISSUE #396: String tags are being interpreted as numbers
hello,friend
two questions:

1. 

i write a tag like this
image

but jaeger-ui show error,like this
image

1. 
i write a tag like this
image

image

image

yntelectual:
Signed-off-by: Matus Majchrak <matus.majchrak@gmail.com>
Fixes the rendering of span tag values that contain numeric strings

Hi this is my first PR to this project so bear with me :) 

Resolves #396 

As stated in bug #396, span tags containing numeric strings will be converted to actual numbers and rendered in scientific notation if they are large enough. 

As hinted by @tiffon, this commit limits the json conversion of tag values only to cases, when the value is an actual string and is not a primitive type(i.e. only if it is object or array) 

The one open point from my side, is what would be a meaningful test for the rendering of tag value containing real json object(if it is required) as I had some trouble creating an elegant test case that would verify the `dangerouslySetInnerHTML` contents(right now I verify it by call to `html()` and then string matching...) and what role should tag `type` attribute play in the rendering: 
e.g.
given span tag payload:
```
{
  "key": "numberTag",
 "type": "int",
  "value": 123
},
{
  "key": "numberAsString",
 "type": "string",
  "value": "123"
}
```
Not sure if this is a valid case. 
In any case, looking forward to your feedback.

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.tsx
@@ -23,13 +23,14 @@ import { KeyValuePair, Link } from '../../../../types/trace';
 
 import './KeyValuesTable.css';
 
-const jsonObjectOrArrayStartRegex = '/^({|[)/';
+const jsonObjectOrArrayStartRegex = /^(\[|\{)/;
 
 function parseIfComplexJson(value: any) {
   // if the value is a string representing actual json object or array, then use json-markup
-  if (typeof value === 'string' && jsonObjectOrArrayStartRegex.match(value)) {
+  if (typeof value === 'string' && value.match(jsonObjectOrArrayStartRegex)) {
     // otherwise just return as is
     try {
+      console.log('parsing JSON');
       return JSON.parse(value);
       // eslint-disable-next-line no-empty
     } catch (_) {}
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.tsx
@@ -27,10 +27,9 @@ const jsonObjectOrArrayStartRegex = /^(\[|\{)/;
 
 function parseIfComplexJson(value: any) {
   // if the value is a string representing actual json object or array, then use json-markup
-  if (typeof value === 'string' && value.match(jsonObjectOrArrayStartRegex)) {
+  if (typeof value === 'string' && jsonObjectOrArrayStartRegex.test(value)) {
     // otherwise just return as is
     try {
-      console.log('parsing JSON');
       return JSON.parse(value);
       // eslint-disable-next-line no-empty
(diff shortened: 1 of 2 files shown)


packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.tsx
```diff
@@ -23,6 +23,8 @@ import { KeyValuePair, Link } from '../../../../types/trace';
 
 import './KeyValuesTable.css';
 
+const jsonObjectOrArrayStartRegex = new RegExp('^[{[]+');
```
tiffon: Why not use a regular expression literal? `/^(\{|\[)/`

yntelectual: Makes sense, will update it later today(seems like the project can not be built on windows...) and push the changes. Thanks

yntelectual:
alright, could you, please, give one more try?

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.tsx
```diff
@@ -23,11 +23,18 @@ import { KeyValuePair, Link } from '../../../../types/trace';
...
-function parseIfJson(value: any) {
-  try {
-    return JSON.parse(value);
-    // eslint-disable-next-line no-empty
-  } catch (_) {}
+const jsonObjectOrArrayStartRegex = /^(\[|\{)/;
+
+function parseIfComplexJson(value: any) {
+  // if the value is a string representing actual json object or array, then use json-markup
+  if (typeof value === 'string' && value.match(jsonObjectOrArrayStartRegex)) {
```
tiffon: I think this can be `jsonObjectOrArrayStartRegex.test(value)` to avoid generating match data?

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.tsx
```diff
@@ -23,11 +23,18 @@ import { KeyValuePair, Link } from '../../../../types/trace';
...
+    // otherwise just return as is
+    try {
+      console.log('parsing JSON');
```
tiffon: Is this log statement a left-over?

yntelectual: oops, fixed now

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.tsx
yntelectual: right

tiffon:
# :tada: Thanks for contributing!! :tada:

yntelectual:
Pleasure is all mine, well setup project and great feedback.

PR #442, opened 2019-08-19
[DDG stacked - 6] DDG styling and performance improvements 

tiffon:

Styling and performance issues.

Styling adjustments and various performance improvements.

### DIFF
--- a/packages/jaeger-ui/src/components/DeepDependencies/Graph/getNodeRenderers.tsx
+++ b/packages/jaeger-ui/src/components/DeepDependencies/Graph/getNodeRenderers.tsx
@@ -21,7 +21,7 @@ import { TDdgVertex, EViewModifier } from '../../../model/ddg/types';
 import './getNodeRenderers.css';
 
 export default function getNodeRenderers(findMatches: Set<TDdgVertex>, viewModifiers: Map<string, number>) {
-  function renderVectorBorder(lv: TLayoutVertex<TDdgVertex>) {
+  function vectorBorder(lv: TLayoutVertex<TDdgVertex>) {
     // eslint-disable-next-line no-bitwise
     const isHovered = (viewModifiers.get(lv.vertex.key) || 0) & EViewModifier.Hovered;
     // eslint-disable-next-line no-bitwise
@@ -44,7 +44,16 @@ export default function getNodeRenderers(findMatches: Set<TDdgVertex>, viewModif
     );
   }
 
-  function renderVectorFindEmphasisOutline(lv: TLayoutVertex<any>) {
+  if (!findMatches.size) {
+    return {
+      vectorBorder,
+      htmlFindEmphasis: null,
+      vectorFindColorBand: null,
+      vectorFindOutline: null,
+    };
+  }
+
+  function vectorFindOutline(lv: TLayoutVertex<any>) {
     if (!findMatches.has(lv.vertex)) {
       return null;
     }
@@ -60,14 +69,14 @@ export default function getNodeRenderers(findMatches: Set<TDdgVertex>, viewModif
     );
   }
 
-  function renderHtmlFindEmphasis(lv: TLayoutVertex<any>) {
+  function htmlFindEmphasis(lv: TLayoutVertex<any>) {
     if (!findMatches.has(lv.vertex)) {
       return null;
     }
     return <div className="DdgNode--HtmlFindEmphasis" />;
   }
 
-  function renderVectorFindEmphasisColorBand(lv: TLayoutVertex<any>) {
+  function vectorFindColorBand(lv: TLayoutVertex<any>) {
     if (!findMatches.has(lv.vertex)) {
       return null;
     }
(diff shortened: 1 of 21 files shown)


packages/jaeger-ui/src/components/DeepDependencies/Graph/getNodeRenderers.tsx
```diff
@@ -84,9 +93,9 @@ export default function getNodeRenderers(findMatches: Set<TDdgVertex>, viewModif
   }
 
   return {
-    htmlFindEmphasis: renderHtmlFindEmphasis,
-    vectorBorder: renderVectorBorder,
-    vectorFindColorBand: renderVectorFindEmphasisColorBand,
-    vectorFindOutline: renderVectorFindEmphasisOutline,
+    htmlFindEmphasis,
```
everett980: :+1:

packages/jaeger-ui/src/components/DeepDependencies/Graph/index.css
```diff
@@ -14,16 +14,22 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-.DdgArrow.is-pathHovered {
+.Ddg--Edges {
+  stroke: #444;
+  stroke-width: 0.7;
```
everett980: is this overridden by `zoom-transform` based styling?

packages/jaeger-ui/src/components/DeepDependencies/Graph/index.css
```diff
@@ -14,16 +14,22 @@ See the License for the specific language governing permissions and
...
+}
+
+.Ddg--Edges.is-withViewModifiers {
```
everett980: this seems to accomplish the `is-notMatched` but now seems a little less clear IMO

everett980: I see, it's to apply one class to all `Edges` not a class to every `Edge`. Missed the `s` at first

packages/jaeger-ui/src/components/DeepDependencies/Header/NameSelector.tsx
```diff
@@ -42,22 +42,37 @@ export default class NameSelector extends React.PureComponent<TProps, TState> {
...
+  }
+
   setValue = (value: string) => {
     this.props.setValue(value);
-    this.setState({ popoverVisible: false });
+    this.changeVisible(false);
+  };
+
+  private onBodyClicked = () => {
+    if (this.listRef.current && !this.listRef.current.isMouseWithin()) {
```
everett980: nice!

packages/jaeger-ui/src/components/DeepDependencies/index.css
```diff
@@ -14,11 +14,16 @@ See the License for the specific language governing permissions and
...
+  height: 100%;
   left: 0;
-  position: fixed;
+  position: absolute;
   right: 0;
-  top: 0;
+}
+
+.Ddg--graphWrapper {
+  flex: 1;
```
everett980: this + `flex-direction: column` is certainly clearer than before with `TOP_NAV_HEIGHT`
but what would cause there to be overflow?

everett980: probably want to avoid the `TOP_NAV_HEIGHT` elsewhere as well, if possible.

packages/plexus/src/Digraph/Node.tsx
```diff
@@ -38,6 +40,10 @@ function getHtmlStyle(lv: TLayoutVertex<any>) {
 export default class Node<T = {}> extends React.PureComponent<TProps<T>> {
   render() {
     const { getClassName, layerType, renderNode, renderUtils, setOnNode, layoutVertex } = this.props;
+    const nodeContent = renderNode(layoutVertex, renderUtils);
```
everett980: this `nodeContent` can be used in the `Wrapper` below

packages/plexus/src/Digraph/types.tsx
```diff
@@ -108,7 +108,7 @@ export type TMeasurableNodeRenderer<T = Record<string, unknown>> = {
 };
 
 export type TNodeRenderer<T = Record<string, unknown>> = {
-  renderNode: TRenderNodeFn<T>;
+  renderNode: TRenderNodeFn<T> | null;
```
everett980: I see why this should be nullable, but it should be possible to still use this def in `NodesLayer.tsx` and `Nodes.tsx`, just making this one value [`NonNullable`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html)

packages/jaeger-ui/src/components/DeepDependencies/Graph/index.css
everett980: I would call this `.Ddg--Edges.contains-emphasis` because view modifiers other than emphasis shouldn't cause all edges to have their stroke opacity lowered.

packages/jaeger-ui/src/components/DeepDependencies/Graph/index.css
everett980: my mistake, zoom-transform influences `stroke-opacity` not `stroke-width`

packages/jaeger-ui/src/components/DeepDependencies/Graph/index.css
everett980: not worth the performance hit to distinguish between VMs that necessitate de-emphasizing other edges and (non-existent) VMs that don't. worth revisiting when other edge VMs are introduced.

packages/plexus/src/Digraph/types.tsx
everett980: simplified `NodesLayer.tsx` and `Nodes.tsx` types

PR #445, opened 2019-08-29
Derive DDG from search results

### LINKED ISSUE #399: Service dependency view based on trace search criteria

The current service dependency tab presents service dependency information related to all services being traced based on information derived from a batch job that generally will only run once per day (Apache Spark job). User's have no ability to customise the view based on a subset of services related to specific trace paths, time range, tags, etc. As the information is only calculated at the end of day, it also means users have to wait until the following day to see the results - which means a significant delay in getting feedback on changes being deployed to their services.

The requirement is to provide a "service dependency view" as an alternative visualisation of the set of trace instances retrieved from the current trace overview page.

Currently users can only view service dependency information (in a production environment) if using the Apache Spark job - which generally will only run at the end of day, so there is a delay in seeing the results. The view also shows a graph of all services, so in an environment with a large number of microservices, this can be overwhelming.

Provide an alternative view from the trace overview page. So users can enter their search criteria and initially see the list of trace instances as now, and then switch view to see the same list of trace instances but now presented in a service dependency graph.

The graph can be constructed by analysing the client/server span pairs in each trace instance to extract the source and destination service names. Where a client span has no associated server span, it could use the `peer.service` tag or host from the `http.url` to provide a representation of the external service.

If we want to go beyond the current service dependency graph representation to provide more elaborate information:

* Relationships could be directional on the graph
* Latency information could be included, as well as number of requests
* Links could be per operation rather than just at the service level
* If a standard way of defining service version was supported, e.g. a `version` tag, then would be good to differentiate requests going to different versions of the service - as a way to identify when traffic has migrated away from an older version
* Auto refresh cycle

Some of these features could be enabled via checkboxes, to allow the initial view to be simple.

rubenvp8510:
- Derived a DDG form the search results.

Resolves #399 

- Used trace search results for derive a DDG. this is still work in progress.. 

Screenshot from 2019-08-30 00-02-50
Screenshot from 2019-08-30 00-02-24

Added a button on the top right of the trace results (it does not have a button style yet)


yurishkuro:
Do you have some ideas how that would be useful?

objectiser:
@yurishkuro Its the approach that was discussed on the bi-weekly call a couple of months back - for users who want a minimal backend installation, but still benefit from a service dependency graph showing the services involved in the traces they are interested in.

objectiser:
@rubenvp8510 On the trace instance view page it is possible to switch between views by using a dropdown. Wondering if the same would work on the trace search page for the dependency view - so the region below the scatter plot could be switched between a trace instance list and the dependency view? Not sure if the region would be too small - but the benefit is that the search criteria would still be visible allowing the user to tweak the criteria and press the search button again and see the update in the dependency view.

rubenvp8510:
@objectiser yes it could be, my concern was the size of the view, but I can modify this PR and attach an screenshot so we can evaluate it.

everett980:
@rubenvp8510 I think the button should exist in the "Compare Traces" section.

Something like this (except without hiding the trace name and details):
image

This way you can select a subset of search results. It would also support making selections, then searching again, and adding selections from the second search.

The DDG view would need a button to go back to search when generated in this manner, which would preserve the selected traceIDs (there is a url param that supports this).

rubenvp8510:
@everett980 That sounds good, the only thing is that in that way there is no way to use all traces, unless I selected all traces, or am i missing something here?, I don''t see a button "select all" ,but of course we can add it.

everett980:
@rubenvp8510 yeah we would need to add a select all (or "add all" to current selection if any traces are already selected).
as well as maybe a "clear selection" button.

everett980:
@objectiser 
> I like the simplicity of being able to switch between the list and dependency graph view with a single button
I agree that the simplicity of this approach is very nice.

> so possibly if the "select individual traces" option is also required, that could be in addition?
The ability to include selected traces from previous searches would be beneficial but can be tackled later on.

everett980:
@rubenvp8510 It seems as though `DeepDependencies/traces.tsx` is a fork of `DeepDependencies/index.tsx`. This is going to result in a lot of maintenance effort as the feature is rapidly evolving.

I'm wondering if it would make sense for `DeepDependencies/traces.tsx` to `import { DeepDependencyGraphPageImpl } from './';`
That way `DeepDependencies/traces.tsx` and `DeepDependencies/index.tsx` can have different `mapStateToProps` functions, but otherwise have less repetition and less to maintain.

I believe the only change that would be necessary for `DeepDependencyGraphPageImpl` would be to hide the select service & operation drop downs:
image
(Which would require a small change to `DeepDependencies/Header.tsx` as well)

rubenvp8510:
@everett980 I totally agree with your comments. there are a couple of more changes that we need to do it other than just hidden a menu, but I don't think it will be hard to do it now that I understand better how the graph works.

I'll update the PR with this, I think it will also help with the coverage, as we are reusing more parts.

objectiser:
@everett980 Are there any other issues that need to be addressed with this one, or can it be merged?

everett980:
@objectiser I gave it a quick look at the changes last week and it looked good to me. haven't had the chance to give it a final pass over yet but I should have time Monday.

objectiser:
@everett980 Great, thanks!

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.js
```diff
@@ -95,6 +102,19 @@ export default class SearchResults extends React.PureComponent<SearchResultsProp
     }
   };
 
+  updateUrlState = () => {
```
everett980: This doesn't seem to be in use

packages/jaeger-ui/src/components/SearchTracePage/index.js
```diff
@@ -248,9 +249,11 @@ export function mapStateToProps(state) {
   if (serviceError) {
     errors.push(serviceError);
   }
+  const location = router.location;
   const sortBy = sortFormSelector(state, 'sortBy');
   const traceResults = sortedTracesXformer(traces, sortBy);
   return {
+    location,
```
everett980: I don't believe this prop is being used

packages/jaeger-ui/src/components/SearchTracePage/index.test.js
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import { MemoryRouter } from 'react-router-dom';
```
everett980: I don't think the changes to this file are still necessary.

packages/jaeger-ui/src/model/ddg/transformTracesToPaths.tsx
```diff
@@ -0,0 +1,74 @@
...
+        .forEach(leaf => {
+          const path = spanAncestorIds(leaf)
+            // @ts-ignore
+            .map(id => convertSpan(map.get(id), trace.data));
+          // @ts-ignore
+          path.push(convertSpan(map.get(leaf.spanID), trace.data));
+          if (hasFocal(path, focalService, focalOperation)) {
+            paths.push({
+              path,
+              trace_id: traceId,
```
everett980: note to self: this will need to be updated after my in-progress work

packages/jaeger-ui/src/model/ddg/transformTracesToPaths.tsx
```diff
@@ -0,0 +1,74 @@
...
+    const map: Map<string, Span> = new Map();
+    if (trace.data) {
+      const traceId = trace.data.traceID;
+      trace.data.spans
+        .filter(span => {
+          map.set(span.spanID, span);
+          return !span.hasChildren;
+        })
+        .forEach(leaf => {
+          const path = spanAncestorIds(leaf)
```
everett980: This can be cleaned up slightly to avoid the `@ts-ignore`s

packages/jaeger-ui/src/model/ddg/transformTracesToPaths.test.js
```diff
@@ -0,0 +1,66 @@
...
+        },
+      ],
+      processes: {
+        p1: {
+          serviceName: 'customer',
+        },
+      },
+    },
+  ];
+  it('transforms trace results payload', () => {
```
everett980: these tests could be a bit more rigorous to check multiple traces and to check that the `hasFocal` filter works.

packages/jaeger-ui/src/components/SearchTracePage/index.test.js
everett980: oh I see, it's because a child of this component changed and mount makes that relevant... I'll see if I can get rid of mount.

everett980: I see, without this the `withRouter` in `SearchResults` returns something that enzyme dislikes.

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.css
```diff
@@ -35,3 +35,7 @@ limitations under the License.
 .SearchResults--resultLink:hover {
   color: inherit;
 }
+
+.SearchResults--ddg-container {
+  min-height: 600px;
```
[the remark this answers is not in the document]
everett980: update to use flex

PR #449, opened 2019-09-16
Make svc&op slugs instead of query params for GA

everett980:
- Analyzing query parameters in Google Analytics is not supported, while it is easy to track segments of the url path.

- change `/deep-dependencies?service=myServiceName&operation=myOperationName` to `/deep-dependencies/myServiceName/myOperationName`
- service and operation remain optional


everett980:
@saminzadeh
Thanks for the info about GA support. I dug into it some and it seems as though it's actually easier to get the insights if they are parameters.
Additionally the slug vs param is an interesting distinction. That stack overflow answer is insightful for APIs, but leaves some ambiguity for URLs.
I think according to that, service would be a slug while operation would be a parameter.

Since the view is the same for all (DDG), and the service & operation are simply specifying which DDG to view, I think it is good as is.

PR #452, opened 2019-09-19
Ddg hidden ui matches count

everett980:
- Currently, the DDG UI will report 0 matches if all matches are hidden, or under represent the number of matches. This would be counter productive to certain use cases.

- Show `visible / total` uiFind matches:
image

## Future Work
- Change the indicator into a button that will expand the graph to show the hidden matches.


everett980:
Test coverage is added in [this PR](https://github.com/jaegertracing/jaeger-ui/pull/453) which extends the functionality.

PR #455, opened 2019-09-24
Add script tags from config to body

everett980:
- We need the ability to run non-OSS code on the UI without treating Jaeger UI as a component library.

- Adds scripts field to config, which are added to the document body.
- Initially only inline scripts are supported, as the config is a JSON and we need the ability to define an onLoad function if we were to use a `src`


everett980:
@yurishkuro the only documentation for the config json in this repo is a link to jaegertracing.io, otherwise we rely on type definitions.
I'll update the linked page tomorrow.

Issue #117, opened 2017-11-02
traceName relies on traceID to equal spanID


terinjokes:
**Description**

The search result model creates a name for each trace, used by the search results page. The name is set while iterating over each result's spans, when the trace ID equals the span ID[^0].

[^0]: https://github.com/jaegertracing/jaeger-ui/blob/79d262dd12e2cd64679b5d5a14aa0c7060fa85e8/src/model/search.js#L65-L68

This assumption works for native Jaeger libraries. Unfortunately, if you're using Jaeger alongside Zipkin, you'll find this doesn't hold true for some Zipkin implementations, where these can differ.[^1]

> When an incoming request has no trace information attached, we generate a random trace ID and span ID. The span ID can be reused as the lower 64-bits of the trace ID, but it can also be completely different.

[^1]: http://zipkin.io/pages/instrumenting.html

Most of our services using Zipkin are fronted by a service that generates (64-bit) trace IDs that match root span IDs, but there's a small handful that aren't. As a result our scatterplot occasionally shrugs a bit.

¯\\\_(ツ)\_/¯

**Steps to reproduce the issue:**
1. Submit spans to a jaeger collector where the root span is started in a Zipkin implementation that generates separate random numbers for a root trace id and span id (eg, opentracing-zipkin-go) or generates 128-bit trace IDs.
2. Search for the trace in jaeger-ui.
3. Hover over a point of the scatterplot.

**Describe the results you received:**

¯\\\_(ツ)\_/¯

**Describe the results you expected:**

A trace name made up of the root's service and operation names.

**Additional information you deem important (e.g. issue happens only occasionally):**

Based on my grok of the code, the traces returned in the search results are tree sorted, just like on timeline pages. Thus the first row is either the actual root, or a reasonable alternative.

If this is acceptable, I'm happy to submit a pull request changing the implementation.

¯\\\_(ツ)\_/¯

yurishkuro:
This is a bug, imo. There has never been a guarantee stated that root span id == trace id. The UI should simply look for a span with parentID=0/null.

tiffon:
@terinjokes Thanks for reporting this and for the information about spanID vs traceID of root spans.

From what I can tell, the root span is generally the first span in the data payload. But, that's reliant on the sort of the spans which is not assumed to be part of the API contract with the query service. I think @yurishkuro's suggestion of using the first span without a parent or follows-from reference as the root span will be more robust to changes in the datastore or query service.

A PR is definitely welcomed!

PR #461, opened 2019-10-07
Fix trace name resolution for spans with remote parents

tmthrgd:
When encountering a trace whose root span contains a reference to a remote parent span in a different trace, the UI would fail to determine the root span and instead display the fallback `<trace-without-root-span>` trace name.

This change makes it so that the first span that has no parents in the current trace is correctly considered the root span. Previously only a span with no references/parents was considered
the root span.

To simplify the logic and ensure consistency, `transformTraceData` was changed to use the fixed `getTraceName`.

Traces with remote parents are commonly used when handling HTTP or RPC requests from untrusted clients. See the [opencensus-specs util/HandleUntrustedRequests.md](https://github.com/census-instrumentation/opencensus-specs/blob/8600b7497457a277e681552005556407e9666997/utils/HandleUntrustedRequests.md) for an example of this approach.

I understand that I should add a test for this, but I'm unsure where would be appropriate. If someone can point me to an appropriate file, I'll add that to this PR.

<details>
<summary>An example of a trace that contains a remote parent.</summary>
<pre>
{
    "data": [
        {
            "traceID": "46412397683b6cee4f02fa1ac7e4d0db",
            "spans": [
                {
                    "traceID": "46412397683b6cee4f02fa1ac7e4d0db",
                    "spanID": "7ef1d2e931ada231",
                    "operationName": "operation",
                    "references": [
                        {
                            "refType": "CHILD_OF",
                            "traceID": "79b24a4a9fb5d4e46895ab64f55dfc90",
                            "spanID": "9111732227172a77"
                        }
                    ],
                    "startTime": 1257894000000000,
                    "duration": 1000,
                    "tags": [],
                    "logs": [],
                    "processID": "p1",
                    "warnings": null
                }
            ],
            "processes": {
                "p1": {
                    "serviceName": "service",
                    "tags": []
                }
            },
            "warnings": null
        }
    ],
    "total": 0,
    "limit": 0,
    "offset": 0,
    "errors": null
}
</pre>
</details>

Updates #117
Updates #134

### DIFF
--- a/packages/jaeger-ui/src/model/trace-viewer.js
+++ b/packages/jaeger-ui/src/model/trace-viewer.js
@@ -14,6 +14,6 @@
 
 // eslint-disable-next-line import/prefer-default-export
 export function getTraceName(spans) {
-  const span = spans.filter(sp => !sp.references || !sp.references.length)[0];
+  const span = spans.find(sp => !sp.references || !sp.references.some(ref => ref.traceID === sp.traceID));
   return span ? `${span.process.serviceName}: ${span.operationName}` : '';
 }
--- a/packages/jaeger-ui/src/model/transform-trace-data.tsx
+++ b/packages/jaeger-ui/src/model/transform-trace-data.tsx
@@ -17,6 +17,7 @@ import _isEqual from 'lodash/isEqual';
 import { getTraceSpanIdsAsTree } from '../selectors/trace';
 import { KeyValuePair, Span, SpanData, Trace, TraceData } from '../types/trace';
 import TreeNode from '../utils/TreeNode';
+import { getTraceName } from './trace-viewer';
 
 function deduplicateTags(spanTags: Array<KeyValuePair>) {
   const warningsHash: Map<string, string> = new Map<string, string>();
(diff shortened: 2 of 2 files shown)


everett980:
@tmthrgd Thanks for submitting this PR.
The best place to test `/src/model/trace-viewer` would be a new file: `/src/model/trace-viewer.test.js`
`/src/model/transform-trace-data` unfortunately does not have a dedicated test file yet, and I think that would be beyond the scope of this PR.

Additionally, if you could convert `trace-viewer` to typescript (and update the suffix to `.tsx`) that would be great. I believe the only change necessary would be:
```
import { SpanData } from '../types/trace';

// eslint-disable-next-line import/prefer-default-export
export function getTraceName(spans: SpanData[]): string {
```

packages/jaeger-ui/src/model/trace-viewer.js
```diff
@@ -14,6 +14,6 @@
 
 // eslint-disable-next-line import/prefer-default-export
 export function getTraceName(spans) {
-  const span = spans.filter(sp => !sp.references || !sp.references.length)[0];
+  const span = spans.find(sp => !sp.references || !sp.references.some(ref => ref.traceID === sp.traceID));
```
yurishkuro: This doesn't seem like an equivalent or a superset of the previous logic.

I would rather fix this like this:
  * find all spans that either have no parent reference, or a parent reference that is not found in the trace
  * if only one such span is found, it's the root (i.e. source of "trace name")
  * if more than one such span is found, then
    * if exactly one of them has no parent refs, it's the root
    * otherwise give up (or, could still do something like pick the one with the earliest start time)

Clearly, we need a set of unit tests for this logic.

tmthrgd:
Looks like this was superseded and fixed by #541.

PR #463, opened 2019-10-10
Make ddg nodes circular and closer together

everett980:
- DDG Nodes are fairly spread out, making it impossible to view many nodes at once while keeping text size legible.
- Additionally, rectangular nodes do not compliment future plans for node decoration.

- Make node spacing denser.
- Make nodes the smallest possible enclosing circle, instead of rectangles of fixed height.

### Before:
image
### After
image

Unfortunately blanking out the text fails to capture how much easier it is to read the nodes, and since the font isn't mono spaced it's hard to replace with stand in text.


yurishkuro:
can you post screenshots?

everett980:
Good point, updated with sanitized screenshots and sent you two with text.

yurishkuro:
yeah, maybe replace service names with xxxx of appropriate length.

Also, it seems like the node radius changes based on the length of the service name. I think this will be pretty confusing, as it is pretty common visualization technique to vary node side based on some _metric_ rather than the name length. I would rather let the long names bleed over, or move them below the circles.

everett980:
Overflowing text will probably be problematic if we want to decorate nodes with a partially full band on the outside of the node.
Alternatively, we could size down the text (or truncate it with an ellipsis and a tool tip) if it makes the node too large.

everett980:
scaling the nodes to the same size should be pretty straight forward, and will allow typical length strings to be represented legibly.

PR #466, opened 2019-10-14
Consume new API format

everett980:
The payload format has changed slightly, update `transformDdgData` to handle this change, and update view traces to consume multiple traceIDs per path elem.

### DIFF
--- a/packages/jaeger-ui/src/model/ddg/transformDdgData.test.js
+++ b/packages/jaeger-ui/src/model/ddg/transformDdgData.test.js
@@ -26,7 +26,7 @@ describe('transform ddg data', () => {
       ? { service: focalPayloadElem.service }
       : focalPayloadElem;
     const { paths, services, visIdxToPathElem } = transformDdgData(
-      payload.map(testResources.wrap),
+      testResources.wrap(payload),
       focalPayloadElemArgument
     );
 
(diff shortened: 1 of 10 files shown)


packages/jaeger-ui/src/model/ddg/transformDdgData.test.js
```diff
@@ -166,21 +166,43 @@ describe('transform ddg data', () => {
...
+  it('adds traceIDs to paths from attributes', () => {
+    const { focalPayloadElem, doubleFocalPath, simplePath, wrap } = testResources;
+    const payload = wrap([simplePath, doubleFocalPath]);
+    payload.dependencies.forEach((dependency, i) => {
+      // eslint-disable-next-line no-param-reassign
+      dependency.attributes = [
+        {
+          key: 'exemplar_trace_id',
+          value: `trace ${i} a`,
+        },
```
vprithvi: @yurishkuro I don't recall our previous discussion - but is there a particular reason to prefer this to combining multivalued keys in the backend as so? 
```
{ 
key: 'exemplar_trace_id',
values: {'traceId1', 'traceId2`, ...}
}
```

everett980: Additionally, is there a reason to prefer the above implementation or the live implementation over:
```
attributes: {
  exemplar_trace_id: ['traceID1', 'traceID2', ...],
  [someOtherKey]: ['value0', 'value1'],
},
```

Issue #467, opened 2019-10-17
Issue with Jaeger UI visualizing span with multiple parents


ppatierno:
Hi,
I am working on adding Jaeger tracing support to the [Strimzi](http://strimzi.io) HTTP bridge component which exposes the Kafka cluster through an HTTP RESTful API.
An HTTP client (producer/consumer) can connect to the bridge via HTTP and sending/receiving Kafka messages through HTTP requests on specific endpoints.
I am having a use case where a new span created by the bridge could have multiple parents; let me explain better.

An HTTP consumer sends a GET request to a specific endpoint for getting Kafka messages, so for doing an actual "poll" operation (in Kafka terms).
With just one HTTP GET it's able to get more Kafka messages (in one shot) encoded in JSON format into the HTTP response from the bridge.
Now, if these Kafka messages have already tracing information brought via headers (i.e. uber-tracer-id), the new "poll" span which the bridge creates, after getting messages, should be somehow "follows-from" all of them.
The things become more interesting saying that the Kafka messages could have different uber-tracer-id(s) because they could be sent by different producer applications (or just one producer) in different times or; it's also true the opposite, that one HTTP producer could send more Kafka messages encoded as JSON in the HTTP request body in one shot, so the Kafka messages spans will have the same uber-tracer-id having as parent one "send" operation span.

Coming back at the consumer side, my question is about having a span with multiple parents.
I noticed that using a child-of or a follows-from relationship, puts the "poll" span just under one of the spans related to the received message and not under both.
First use case.

- an HTTP producer sends two messages in one shot (so just one "send" span with two children "To_test") ... which is fine
- the HTTP consumer polls for messages, the bridge reads them from Kafka so the two "From_test" children of related "To_test" which is fine, but the newly created "poll" has as a parent just one of them ... not both.

![scenario_1](https://user-images.githubusercontent.com/5842311/67001736-af886880-f0da-11e9-8553-695b6137d625.png)

Other than the picture, the scenario_1.txt shows that "poll" has actually two references to same trace-id (about the "send") and different "span-id" (related to the two messages read) but it's visualized under just one of them.

Second use case:

- an HTTP producer sends two messages with two different HTTP requests, so two different "send" spans with related "To_test" ... which is fine
- the HTTP consumer polls for messages,  the bridge reads them from Kafka so the two "From_test" children of related "To_test" which is fine (they are part of the two different traces), but the newly created "poll" has as a parent just one of them ... not both.

[...]

PR #477, opened 2019-10-29
Jaeger UI visualizing span with multiple parents 

rubenvp8510:
-  See description on Issue #467 

- Virtualize of the spaces in multiple parents (repeating the span in each parent), using an icon to indicate which spans has more than one parent, also were able to navigate to those parents (or highlight it)

Screenshot from 2019-11-07 07-15-35
Screenshot from 2019-11-07 07-15-16
Screenshot from 2019-11-07 07-15-01


rubenvp8510:
@everett980 I attached some screenshots of this, Could you please provide me some feedback? Thanks!

yurishkuro:
Are there changes to the main timeline, or is the addition only the new icon on the expanded spans? 

When the span is expanded I would rather show an additional section References, after Tags and Logs, where refs can be shown in a table, with links. That seems more usable than a mouse over pop up.

rubenvp8510:
> Are there changes to the main timeline, or is the addition only the new icon on the expanded spans?
There are one change to the timeline, before this PR the "poll" span was only displayed on the first parent, now is showing twice (per each parent)

Also an icon to indicate that span has more parents 
> 
> When the span is expanded I would rather show an additional section References, after Tags and Logs, where refs can be shown in a table, with links. That seems more usable than a mouse over pop up.

That was my second choice, and something I was about to ask, if the accordion section is better I can do the change.

yurishkuro:
> There are one change to the timeline, before this PR the "poll" span was only displayed on the first parent, now is showing twice (per each parent)

Unless there are some explicit visual indications of a duplicate, I think showing the span twice is very misleading. A span represents some work, this work happened once.

I think accordion fits more consistently with the existing design of span details.

rubenvp8510:
> > There are one change to the timeline, before this PR the "poll" span was only displayed on the first parent, now is showing twice (per each parent)
> 
> Unless there are some explicit visual indications of a duplicate, I think showing the span twice is very misleading. A span represents some work, this work happened once.

Exactly, that is why I added an icon, any other suggestion on how to handle this?

> 
> I think accordion fits more consistently with the existing design of span details.

yurishkuro:
But the icon is only visible when you expand the span. The visual difference needs to exist in the gantt chart itself. Maybe showing the span bar hollow, I don't know, I am not a UX designer. Even hollow bar is not enough imo (also, when the span is visually very short, hollow wouldn't even work).

everett980:
Should descendants of spans with multiple references also be duplicated?
As is, it looks like the diff makes descendants of spans with multiple references descendants of the root span instead.

rubenvp8510:
I agree we should show something in the Gantt char (still not sure how ) , also in the left view of the spam, May be something after the operation name?

But I think showing the "span" per parent is not too bad, the span preserves his ID, so it is clear that is the same span. If we don't show it duplicated ,  showing only on the first parent,  we are missing the fact that there are other references for that span.

yurishkuro:
I definitely think the children should not be repeated. Maybe the multi-parent span should only have full details in one place in the chart, and all other copies of that span (a) shown visually different, e.g. a hollow box, and (b) when expanded just display a link to the first occurrence where full details can be seen.

objectiser:
> (b) when expanded just display a link to the first occurrence where full details can be seen

Instead of two steps (i.e. displaying a link) - when the duplicate is expanded, it just actually expands the first occurrence and focus on that span?

yurishkuro:
> Instead of two steps (i.e. displaying a link) - when the duplicate is expanded, it just actually expands the first occurrence and focus on that span?

it might lead to poor user experience, as the first instance of the span could be far away, so the screen would need to scroll, etc., making the user lose the context. An explicit link avoids that.

rubenvp8510:
> Should descendants of spans with multiple references also be duplicated?
Well good question,  this PR does that, but reading the discussion I'm not sure that is a correct way to proceed. The thing is, an span with multiple parents represent more than one path in the trace.

> As is, it looks like the diff makes descendants of spans with multiple references descendants of the root span instead.

rubenvp8510:
> I definitely think the children should not be repeated. Maybe the multi-parent span should only have full details in one place in the chart, and all other copies of that span (a) shown visually different, e.g. a hollow box, and (b) when expanded just display a link to the first occurrence where full details can be seen.


I agree with a), for b) Is not less usable  need to go to another place to see the full details?

yurishkuro:
> Is not less usable need to go to another place to see the full details?

if you follow that argument, then why not show all children of that repeated span as well. I don't think it plays well. For all practical purposes that span is _not really there_, as each span has only one true parent (e.g. in OpenTelemetry the parent is enforced, whereas in Jaeger we infer the parent from one or more refs).

everett980:
To avoid duplicating spans, could we add two interact-able icons?

The first next to the name of the span with multiple references. If there is one other reference, clicking the icon would navigate to its other reference. If there are more than two references, all references but the first would be in a dropdown. Clicking a ref in the dropdown would similarly navigate to the that reference.

The second icon would be on any span that is referenced but is rendered with a different parent span. Clicking this icon would navigate to that reference. Likewise this would need to have a dropdown view if it has been referenced by multiple other spans rendered elsewhere.

rubenvp8510:
> To avoid duplicating spans, could we add two interact-able icons?
> 
> The first next to the name of the span with multiple references. If there is one other reference, clicking the icon would navigate to its other reference. If there are more than two references, all references but the first would be in a dropdown. Clicking a ref in the dropdown would similarly navigate to the that reference.
> 
> The second icon would be on any span that is referenced but is rendered with a different parent span. Clicking this icon would navigate to that reference. Likewise this would need to have a dropdown view if it has been referenced by multiple other spans rendered elsewhere.


I like this idea, I can do the modifications to this PR.

rubenvp8510:
I attached some screenshots with the suggested changes, I can change the icons if you have a better idea of what icons I should use, I can also add a references accordion in the details if that helps.
 
Screenshot from 2019-11-17 10-56-51

Screenshot from 2019-11-17 10-58-02

Screenshot from 2019-11-17 10-57-16

yurishkuro:
I like the icon!

On the first screenshot, the popup shows two lines, one with name and ID, another just the ID, why is that? Is it because the second reference is to another trace? Would ie make sense to annotation them with `(this trace) {service}:{operartion} - {id}` and `(another trace) - {id}`? (Ideally we'd retrieve the name of the other trace, but that can be done later.

Are we going to list references in the span details in accordion?

rubenvp8510:
> I like the icon!
> 
> On the first screenshot, the popup shows two lines, one with name and ID, another just the ID, why is that? Is it because the second reference is to another trace? Would ie make sense to annotation them with `(this trace) {service}:{operartion} - {id}` and `(another trace) - {id}`? (Ideally we'd retrieve the name of the other trace, but that can be done later.
> 

Yes, the reason was because the span is on another trace and, under the actual circumstances I don't have the operation name. I'll change the text to the text you suggested, only have the ID there does not look good.




> Are we going to list references in the span details in accordion?

Yes I will add the references accordion on details.

rubenvp8510:
Icons on the span row.
Screenshot from 2019-11-22 11-37-00


Here, the menu with references to other traces.
Screenshot from 2019-11-22 11-37-24

References on span details.
Screenshot from 2019-11-22 09-31-36


@yurishkuro  @everett980  @objectiser WDYT?

yurishkuro:
in the span details I think we want to display reference type (child-of, follows-from)

rubenvp8510:
> in the span details I think we want to display reference type (child-of, follows-from)

Ahh, true, I going to add it.

rubenvp8510:
Added ref type.

Screenshot from 2019-11-22 12-26-13

rubenvp8510:
This is ready for another review.

yurishkuro:
@everett980 is this good to merge?

rubenvp8510:
@everett980  @yurishkuro  What is the status of this? Any other comment that needs to be addressed? 

Thanks!

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/ReferencesButton.test.js
```diff
@@ -0,0 +1,73 @@
...
+  const focusMock = jest.fn();
+  const oneReference = trace.spans[1].references;
+
+  const moreReferences = oneReference.slice();
+  moreReferences.push({
+    refType: 'CHILD_OF',
+    traceID: trace.traceID,
+    spanID: trace.spans[2].spanID,
+  });
+  const baseProps = {
```
everett980: `uiFind`, `history`, `location`, and `trace` are not part of TReferencesButtonProps`

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/ReferencesButton.test.js
```diff
@@ -0,0 +1,73 @@
...
+      const props = { ...baseProps, references: oneReference };
+      const wrapper = shallow(<ReferencesButton {...props} />);
+      const dropdown = wrapper.find(Dropdown);
+      const refLink = wrapper.find(ReferenceLink);
+      const tooltip = wrapper.find(Tooltip);
+
+      expect(dropdown.length).toBe(0);
+      expect(refLink.length).toBe(1);
+      expect(refLink.first().props().className).toBe('multi-parent-button');
+      expect(tooltip.length).toBe(1);
```
everett980: might as well provide a `tooltipText` since it's a mandatory prop then check that `tooltip.prop('title')` is `props.tooltipText`

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/ReferencesButton.test.js
```diff
@@ -0,0 +1,73 @@
...
+    location: {
+      search: null,
+    },
+    traceID: trace.traceID,
+    trace: {
+      data: trace,
+    },
+    focusSpan: focusMock,
+  };
+  describe('rendering', () => {
```
everett980: don't need this `describe` as there are no tests outside of this `describe`

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/ReferencesButton.test.js
```diff
@@ -0,0 +1,73 @@
...
+    it('render with dropdown', () => {
```
everett980: `it('renders single reference', () => {`

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/ReferencesButton.test.js
```diff
@@ -0,0 +1,73 @@
...
+      const refLink = wrapper.find(ReferenceLink);
+      const tooltip = wrapper.find(Tooltip);
+
+      expect(dropdown.length).toBe(0);
+      expect(refLink.length).toBe(1);
+      expect(refLink.first().props().className).toBe('multi-parent-button');
+      expect(tooltip.length).toBe(1);
+    });
+
+    it('render with dropdown', () => {
```
everett980: `it('renders multiple references', () => {`

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/ReferencesButton.test.js
```diff
@@ -0,0 +1,73 @@
...
+  describe('rendering', () => {
+    it('render with dropdown', () => {
+      const props = { ...baseProps, references: oneReference };
+      const wrapper = shallow(<ReferencesButton {...props} />);
+      const dropdown = wrapper.find(Dropdown);
+      const refLink = wrapper.find(ReferenceLink);
+      const tooltip = wrapper.find(Tooltip);
+
+      expect(dropdown.length).toBe(0);
+      expect(refLink.length).toBe(1);
```
everett980: should add: `expect(refLink.prop('reference')).toBe(oneReference[0]);`

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/ReferencesButton.test.js
```diff
@@ -0,0 +1,73 @@
...
+    });
+
+    it('render with dropdown', () => {
+      const props = { ...baseProps, references: moreReferences };
+      const wrapper = shallow(<ReferencesButton {...props} />);
+      const dropdown = wrapper.find(Dropdown);
+      expect(dropdown.length).toBe(1);
+      const menuInstance = shallow(dropdown.first().props().overlay);
+      const submenuItems = menuInstance.find(Menu.Item);
+      expect(submenuItems.length).toBe(2);
```
everett980: should add:
```
submenuItems.forEach((submenuItem, i) => {
  expect(submenuItem.find(ReferenceLink).prop('reference')).toBe(moreReferences[i]);
});
```

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/AccordianReferences.test.js
```diff
@@ -0,0 +1,124 @@
...
+      process: {
+        serviceName: 'service2',
+      },
+    },
+    spanID: 'span4',
+    traceID: 'trace1',
+  },
+  {
+    refType: 'CHILD_OF',
+    span: {
```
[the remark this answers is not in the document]
everett980: at least `ReferenceButton` uses its truthiness as a signal for internal v external so it should be falsy here.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/ReferencesButton.css
```diff
@@ -0,0 +1,29 @@
...
+limitations under the License.
+*/
+
+.multi-parent-button {
+  padding: 0 5px;
+  color: #000;
+}
+
+.external-trace-ref {
+  float: right;
```
everett980: `float` should be avoided. Once I submit I'll link how I recommended getting around `float` in another file in this PR.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/ReferencesButton.css
```diff
@@ -0,0 +1,29 @@
...
+  margin-left: 5px;
```
everett980: is this left side margin only for when there are both references and referrals? if so the selector for this styling should be `.multi-parent-button ~ .multi-parent-button`

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/ReferenceLink.test.js
```diff
@@ -0,0 +1,65 @@
...
+  };
+
+  const externalRef = {
+    refType: 'CHILD_OF',
+    traceID: 'trace2',
+    spanID: 'span2',
+  };
+
+  describe('rendering', () => {
+    it('render for external trace', () => {
```
everett980: i think this spec name and the following were swapped

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
```diff
@@ -220,6 +223,18 @@ export class VirtualizedTraceViewImpl extends React.Component<VirtualizedTraceVi
     }
   }
 
+  focusSpan = (uiFind: string) => {
```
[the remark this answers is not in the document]
everett980: actually I don't think this will work due to these calls synchronously updating the redux state but the render that would consume the change caused by `focusUiFindMatches` is async... so I think this is fine for now

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/AccordianReferences.test.js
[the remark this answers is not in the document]
rubenvp8510: Not necessary, I validated both cases, when `null` or `span.traceID!=traceID` . Can we assume that if there is no span (or is null) this is a reference to another trace? if so, I think we can use only that fact for validation on `ReferenceButton`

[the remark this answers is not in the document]
rubenvp8510: Sorry not on ReferenceButton but on Accordian https://github.com/jaegertracing/jaeger-ui/pull/477/files#diff-0900665c662dc3c2086584f6d2fd05abR53

I would like to unify the criteria. so my question is still the same, is enough to validate only for null span? or we should consider both? (if span is not null validate the traceID).

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/AccordianReferences.tsx
```diff
@@ -0,0 +1,139 @@
...
+              </ReferenceLink>
+            </li>
+          );
+        })}
+      </ul>
+    </div>
+  );
+}
+
+export default class AccordianReferences extends React.PureComponent<AccordianReferencesProps> {
```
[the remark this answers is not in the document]
rubenvp8510: I totally agree with this, and I think this could be a separate PR, as this is getting too big.

I'm interested on working on it, so I can create the issue or if you create it can assign it to me if you think is appropriate.

rubenvp8510:
@everett980 Thanks for the review!

I think I covered almost all (if not all) the comments/requested changes, except the accordion stuff which I think would be in another PR.

everett980:
Thanks for the quick turn around! Tomorrow morning I'll re-review (I started to and what I saw so far looks good) and will create the follow up issue for Accordian consolidation.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/AccordianReferences.test.js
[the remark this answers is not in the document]
everett980: Currently it seems safe to treat `Boolean(ref.span)` as an indication of a local span reference. transform-trace-data only populates `ref.span` if it is in `data.spans` ([source]()https://github.com/jaegertracing/jaeger-ui/blob/master/packages/jaeger-ui/src/model/transform-trace-data.tsx#L151)

everett980:
I don't recall hitting submit feedback (hence the lack of any top-level message to go with it)... but anyway.
@rubenvp8510 The PR looks good. I left four little things that would be good to tweak but if you don't have time before a target UI release of tomorrow I can probably just make the changes myself.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/AccordianReferences.tsx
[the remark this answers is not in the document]
everett980: I cannot seem to assign it to you, so here is the issue: https://github.com/jaegertracing/jaeger-ui/issues/498

PR #488, opened 2019-11-22
Ddg optional operation

everett980:
- Current DDG is a service & operation oriented dependency graph. We want to be able to have a service dependency graph.

- Allow operation field on focal vertex to be an array (filtered based on visibility)
- Raise request when service is specified
- Allow operation to be cleared
- Allow operation to be specified on focal node with multiple operations
- Improve DDG node hovered CSS


everett980:
Build passed but never updated on this view...

PR #489, opened 2019-11-28
Sort span tags in alphabetical order

### LINKED ISSUE #218: Sort tags in expanded span view
**Description**

Expanding a span on the UI shows collapsed list of tags which can be expanded again. When there are many tags it's hard to find the one a person is looking for 

E.g. when all headers are added as tags and I'm looking for a particular one it's usually easier to use ctrl+f and type part of its name.

An example:
Alt text

Additionally some tags might be more significant than others so it might make sense to have a way to prioritize them (e.g. user_id, or http.url for http client/server spans). But it's something extra.

Expand any span

Tags which appear to be in random order. It's hard to find the important ones with a first glance.

Tags ordered in some way which makes it easy to "understand" what's on the list

It's not bug. I'm also not sure if sorting is the best solution but the current display is not ideal.

nabam:
Resolves #218 
Expanding a span on the UI shows collapsed list of tags which can be expanded again. When there are many tags it's hard to find the one a person is looking for

- Trace span tags are getting sorted after deduplication in reducer
- New option `topTagPrefixes?: string[]` introduced in UI configuration file. Tags with keys matching this prefixes will be displayed at the top preserving order of prefixes.

### DIFF
--- a/packages/jaeger-ui/src/constants/default-config.tsx
+++ b/packages/jaeger-ui/src/constants/default-config.tsx
@@ -66,7 +66,6 @@ export default deepFreeze(
         gaID: null,
         trackErrors: true,
       },
-      topTagPrefixes: ['http.'],
     },
     // fields that should be individually merged vs wholesale replaced
     '__mergeFields',
--- a/packages/jaeger-ui/src/constants/default-config.tsx
+++ b/packages/jaeger-ui/src/constants/default-config.tsx
@@ -66,6 +66,7 @@ export default deepFreeze(
         gaID: null,
         trackErrors: true,
       },
+      topTagPrefixes: ['http.'],
     },
     // fields that should be individually merged vs wholesale replaced
     '__mergeFields',
--- a/packages/jaeger-ui/src/model/transform-trace-data.test.js
+++ b/packages/jaeger-ui/src/model/transform-trace-data.test.js
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { orderTags } from './transform-trace-data';
+import { orderTags, deduplicateTags } from './transform-trace-data';
 
 describe('orderTags()', () => {
   it('correctly orders tags', () => {
@@ -29,5 +29,23 @@ describe('orderTags()', () => {
       { key: 'a.ip', value: '8.8.8.8' },
       { key: 'b.ip', value: '8.8.4.4' },
     ]);
+  });
+});
+
+describe('deduplicateTags()', () => {
+  it('deduplicates tags', () => {
+    const tagsInfo = deduplicateTags([
+      { key: 'b.ip', value: '8.8.4.4' },
+      { key: 'b.ip', value: '8.8.8.8' },
+      { key: 'b.ip', value: '8.8.4.4' },
+      { key: 'a.ip', value: '8.8.8.8' },
+    ]);
+
+    expect(tagsInfo.tags).toEqual([
+      { key: 'b.ip', value: '8.8.4.4' },
+      { key: 'b.ip', value: '8.8.8.8' },
+      { key: 'a.ip', value: '8.8.8.8' },
+    ]);
+    expect(tagsInfo.warnings).toEqual(['Duplicate tag "b.ip:8.8.4.4"']);
   });
 });
--- a/packages/jaeger-ui/src/model/transform-trace-data.test.js
+++ b/packages/jaeger-ui/src/model/transform-trace-data.test.js
@@ -1,0 +1,33 @@
+// Copyright (c) 2019 The Jaeger Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { orderTags } from './transform-trace-data';
+
+describe('orderTags()', () => {
+  it('correctly orders tags', () => {
+    const orderedTags = orderTags(
+      [
+        { key: 'b.ip', value: '8.8.4.4' },
+        { key: 'http.status_code', value: '200' },
+        { key: 'a.ip', value: '8.8.8.8' },
+      ],
+      ['http.']
+    );
+    expect(orderedTags).toEqual([
+      { key: 'http.status_code', value: '200' },
+      { key: 'a.ip', value: '8.8.8.8' },
+      { key: 'b.ip', value: '8.8.4.4' },
+    ]);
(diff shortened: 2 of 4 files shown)


nabam:
I'm not very familiar with javascript, react and redux. Please let me know if I'm doing something bizarre.

packages/jaeger-ui/src/constants/default-config.tsx
```diff
@@ -66,6 +66,7 @@ export default deepFreeze(
         gaID: null,
         trackErrors: true,
       },
+      topTagPrefixes: ['http.'],
```
yurishkuro: the priority sorting feels contrary to the intention of this PR. Is it really necessary?

nabam: Sorting by itself makes it much easier to find relevant information in a trace compare to experience with tags in the random order. At least the tag one is trying to follow is at the same position in every span. 

I was trying to find a way to access tags that that are more relevant than others faster. Issue #218 mentions that and in my installation I have a similar problem
Screenshot 2019-12-02 at 10 01 14
Without priority sorting I would have to name tags the way they are alphabetically in order of importance to get most relevant ones in that view.

Though I bet there are more elegant solutions for that than priority sorting. I think it's quite a standard approach to have many tags some of which are used rarely for things like troubleshooting corner cases and others are crucial for most of the usage scenarios.

packages/jaeger-ui/src/model/transform-trace-data.tsx
```diff
@@ -32,6 +33,24 @@ function deduplicateTags(spanTags: Array<KeyValuePair>) {
   return { tags, warnings };
 }
 
+export function orderTags(spanTags: Array<KeyValuePair>, topPrefixes: Array<string>) {
+  const orderedTags: Array<KeyValuePair> = [...spanTags];
+  orderedTags.sort((a, b) => {
+    for (let i = 0; i < topPrefixes.length; i++) {
```
everett980: I would make `topPrefixes` optional, then wrap this `for` loop in an `if` statement and drop the default value on line 131

packages/jaeger-ui/src/model/transform-trace-data.tsx
```diff
@@ -32,6 +33,24 @@ function deduplicateTags(spanTags: Array<KeyValuePair>) {
   return { tags, warnings };
 }
 
+export function orderTags(spanTags: Array<KeyValuePair>, topPrefixes: Array<string>) {
+  const orderedTags: Array<KeyValuePair> = [...spanTags];
+  orderedTags.sort((a, b) => {
```
everett980: I think this should be case insensitive, so below this I would add:
```
const aKey = a.key.toLowerCase();
const bKey = b.key.toLowerCase();
```
and then use `aKey` and `bKey` in place of `a.key` and `b.key`

and above `tree.walk` on line 111 I would add:
```
const prefixes = getConfigValue('topTagPrefixes');
const topTagPrefixes = prefixes && prefixes.map(p => p.toLowerCase())`;
```

packages/jaeger-ui/src/constants/default-config.tsx
everett980: I like the priority sorting as a config option, but I wouldn't add a default value.

packages/jaeger-ui/src/model/transform-trace-data.tsx
```diff
@@ -32,6 +33,24 @@ function deduplicateTags(spanTags: Array<KeyValuePair>) {
   return { tags, warnings };
 }
 
+export function orderTags(spanTags: Array<KeyValuePair>, topPrefixes: Array<string>) {
+  const orderedTags: Array<KeyValuePair> = [...spanTags];
```
everett980: It seems as though `spanTags.slice()` is faster than `[...spanTags]`
[source](https://measurethat.net/Benchmarks/ShowResult/84928)

packages/jaeger-ui/src/model/transform-trace-data.test.js
```diff
@@ -0,0 +1,51 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { orderTags, deduplicateTags } from './transform-trace-data';
+
+describe('orderTags()', () => {
+  it('correctly orders tags', () => {
+    const orderedTags = orderTags(
+      [
```
everett980: If you could expand the test data and expected output a bit to cover that it sorts within a tag prefix (i.e.: `'http.a'` is before `'http.z'`) and that it respects the order of `topTagPrefixes` (i.e.: if `topTagPrefixes` is `['z.', 'a.', 'http.']` then the result will have this order), that would be great.


I think this would do it:
```
const orderedTags = orderTags(
  [
    { key: 'b.ip', value: '8.8.4.4' },
    { key: 'http.status_code', value: '200' },
    { key: 'a.ip', value: '8.8.8.8' },
    { key: 'z.ip', value: '8.8.8.16' },
    { key: 'http.message', value: 'ok' },
  ],
  ['z.', 'a.', 'http.']
)
expect(orderedTags).toEqual([
  { key: 'z.ip', value: '8.8.8.16' },
  { key: 'a.ip', value: '8.8.8.8' },
  { key: 'http.message', value: 'ok' },
  { key: 'http.status_code', value: '200' },
  { key: 'b.ip', value: '8.8.4.4' },
]);
```

packages/jaeger-ui/src/model/transform-trace-data.tsx
```diff
@@ -15,10 +15,11 @@
 import _isEqual from 'lodash/isEqual';
 
 import { getTraceSpanIdsAsTree } from '../selectors/trace';
+import { getConfigValue } from '../utils/config/get-config';
 import { KeyValuePair, Span, SpanData, Trace, TraceData } from '../types/trace';
 import TreeNode from '../utils/TreeNode';
 
-function deduplicateTags(spanTags: Array<KeyValuePair>) {
+export function deduplicateTags(spanTags: Array<KeyValuePair>) {
```
everett980: I believe this is only exported for tests, so please add `// exported for tests` above this line.

packages/jaeger-ui/src/model/transform-trace-data.tsx
```diff
@@ -32,6 +33,24 @@ function deduplicateTags(spanTags: Array<KeyValuePair>) {
...
+      const p = topPrefixes[i];
+      if (a.key.startsWith(p) && !b.key.startsWith(p)) {
+        return -1;
+      }
+
+      if (!a.key.startsWith(p) && b.key.startsWith(p)) {
+        return 1;
+      }
+    }
+    return a.key.localeCompare(b.key);
```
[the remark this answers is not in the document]
everett980: `aKey` and `bKey` are defined [here](https://github.com/jaegertracing/jaeger-ui/pull/489#discussion_r354033729), sorry that got out of order

packages/jaeger-ui/src/model/transform-trace-data.tsx
nabam: I think that `toLowerCase` mapping can be done inside `orderTags`. It's a bit of performance loss but makes code more testable.

PR #490, opened 2019-12-01
Show combined trace

lookfwd:
- Adds the ability to combine and show many traces in a single trace view. This is useful in many cases, including when searching for a specific tag (e.g. for a session ID) and then showing all those traces in a single view. It also reveals concurrent interactions e.g. between transactions and disk flushes, in the case of a database.

jaeger-feature-preview

Here's for example a case of 3 traces combined in a single screen:

three-traces

- It informally extends the definition of trace id to a set of `|` separated ids in the form `traceid1|traceid2...`. For example the url `/trace/traceid1|traceid2` will use a trace with id `traceid1|traceid2`. Composite ids are handled as usual everywhere, but at the very low level (`jaeger-api`) where instead of fetching a single trace, we now delegate to `JaegerAPI.searchTraces({ traceID: ids }` and then merge the results to a single composite trace with `mergeAll()`. A button is added that allows a user to select many traces by reusing the diff selection mechanism.

Many things can be improved but this is a starting point for discussion.

### DIFF
--- a/packages/jaeger-ui/src/actions/jaeger-api.js
+++ b/packages/jaeger-ui/src/actions/jaeger-api.js
@@ -15,11 +15,48 @@
 import { createAction } from 'redux-actions';
 import JaegerAPI from '../api/jaeger';
 
-export const fetchTrace = createAction(
-  '@JAEGER_API/FETCH_TRACE',
-  id => JaegerAPI.fetchTrace(id),
-  id => ({ id })
-);
+export function mergeAll(id) {
+  return traces => {
+    if (traces.data.length < 2) {
+      return traces;
+    }
+    const ot = { traceID: id, spans: [], processes: {}, warnings: [] };
+
+    const processMap = {};
+
+    traces.data.forEach(trace => {
+      const tracePidMap = {};
+      Object.entries(trace.processes).forEach(([pid, process]) => {
+        const pJson = JSON.stringify(process);
+        let targetPid = processMap[pJson];
+        if (!targetPid) {
+          targetPid = `p${Object.keys(processMap).length + 1}`;
+          processMap[pJson] = targetPid;
+          ot.processes[targetPid] = process;
+        }
+        tracePidMap[pid] = targetPid;
+      });
+      trace.spans.forEach(span => {
+        const ns = span;
+        ns.processID = tracePidMap[ns.processID];
+        ot.spans.push(ns);
+      });
+      if (trace.warnings) {
+        trace.warnings.forEach(j => ot.warnings.push(j));
+      }
+    });
+
+    return { data: [ot] };
+  };
+}
+
+function reduceFetch(id) {
+  const ids = id.split('|');
+  const promise = ids.length < 2 ? JaegerAPI.fetchTrace(id) : JaegerAPI.searchTraces({ traceID: ids });
+  return promise && promise.then(mergeAll(id));
+}
+
+export const fetchTrace = createAction('@JAEGER_API/FETCH_TRACE', id => reduceFetch(id), id => ({ id }));
 
 export const fetchMultipleTraces = createAction(
   '@JAEGER_API/FETCH_MULTIPLE_TRACES',
--- a/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/DiffSelection.css
+++ b/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/DiffSelection.css
@@ -32,3 +32,7 @@ limitations under the License.
   border: 1px solid #ccc;
   border-bottom: none;
 }
+
+.DiffSelection--message button {
+  margin-left: 7px;
+}
--- a/packages/jaeger-ui/src/components/TracePage/url.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/url.tsx
@@ -22,6 +22,10 @@ export function getUrl(id: string) {
   return prefixUrl(`/trace/${id}`);
 }
 
+export function getCombineUrl(id: string[]) {
+  return prefixUrl(`/trace/${id.join('|')}`);
+}
+
 export function getLocation(id: string, state: Record<string, any> | TNil) {
   return {
     state,
--- a/packages/jaeger-ui/src/actions/jaeger-api.test.js
+++ b/packages/jaeger-ui/src/actions/jaeger-api.test.js
@@ -142,3 +142,88 @@ describe('actions/jaeger-api', () => {
     expect(called.verify()).toBeTruthy();
   });
 });
+
+describe('mergeAll()', () => {
+  const trace1 = {
+    traceID: 'trace1',
+    spans: [
+      {
+        traceID: 'trace1',
+        spanID: 'span-1',
+        processID: 'p1',
+      },
+    ],
+    processes: {
+      p1: {
+        serviceName: 'frontend',
+      },
+    },
+    warnings: ['warning 1'],
+  };
+  const trace2 = {
+    traceID: 'trace2',
+    spans: [
+      {
+        traceID: 'trace2',
+        spanID: 'span-a',
+        processID: 'p1',
+      },
+      {
+        traceID: 'trace2',
+        spanID: 'span-b',
+        processID: 'p2',
+      },
+    ],
+    processes: {
+      p1: {
+        serviceName: 'database',
+      },
+      p2: {
+        serviceName: 'frontend',
+      },
+    },
+    warnings: ['warning 2'],
+  };
+  it('mergeAll() of less than two traces returns identity', () => {
+    const merge = jaegerApiActions.mergeAll('id');
+    const traces = { data: [trace1] };
+    expect(merge(traces)).toEqual(traces);
+  });
+  it('mergeAll() of many traces returns merged trace', () => {
+    const merge = jaegerApiActions.mergeAll('id');
+    const traces = { data: [trace1, trace2] };
+    expect(merge(traces)).toEqual({
+      data: [
+        {
+          processes: {
+            p1: {
+              serviceName: 'frontend',
+            },
+            p2: {
+              serviceName: 'database',
+            },
+          },
+          spans: [
+            {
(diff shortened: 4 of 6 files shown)


yurishkuro:
Could you elaborate on how multiple traces are displayed?

lookfwd:
Yes, here's for example a case of 3 traces combined in a single screen:

three-traces

objectiser:
Just wondering whether there needs to be some way to visually know which lines are part of which trace instance. Currently I assume the traces are not interleaved? But at some point we may want that, in which case possibly having different background colours may be helpful.

Even if the traces are kept separate, it might be good to have some form of separator, either a line or perhaps alternating background colour.

lookfwd:
Yes, I was thinking about that... It becomes worse with traces that have missing top-level Spans. So I was thinking to add an artificial top-level span with the trace-ID as name. Then it should all become more clear. Do you think this would help?

yurishkuro:
That might work, but I don't know if you can do it purely in the UI model. If you have to do it in the trace data, then you would have to erase some data to make otherwise root spans children of the fake span. Maybe it's possible to only insert the row in the UI model.
 
Perhaps use Trace 1, Trace 2, etc. as "service name", and trace ID as "operation name".

lookfwd:
Makes sense and it’s more clean. In other words, in the UI... for every Span, if (multi_trace and my.traceId!=previous_span.traceId) { inject_a_fake_super_span_row(); }

packages/jaeger-ui/src/actions/jaeger-api.js
```diff
@@ -15,11 +15,48 @@
...
-  '@JAEGER_API/FETCH_TRACE',
-  id => JaegerAPI.fetchTrace(id),
-  id => ({ id })
-);
+export function mergeAll(id) {
+  return traces => {
+    if (traces.data.length < 2) {
+      return traces;
+    }
+    const ot = { traceID: id, spans: [], processes: {}, warnings: [] };
```
everett980: `ot` isn't the clearest name, maybe `multiTrace` would be clearer? I'd also move the following newline to precede this line. Also, I'm not sure what implications there may be from having an invalid `traceID` 🤔

packages/jaeger-ui/src/actions/jaeger-api.js
```diff
@@ -15,11 +15,48 @@
...
+        let targetPid = processMap[pJson];
+        if (!targetPid) {
+          targetPid = `p${Object.keys(processMap).length + 1}`;
+          processMap[pJson] = targetPid;
+          ot.processes[targetPid] = process;
+        }
+        tracePidMap[pid] = targetPid;
+      });
+      trace.spans.forEach(span => {
+        const ns = span;
```
everett980: I'm guessing this line is to avoid the eslint error `no-param-reassign`, but this will still result in the original span being updated because this is pass by reference.

Regardless, I'm not sure if there is any impact of changing the process IDs from what was originally on the span to a counter. @yurishkuro can you weigh in?

everett980: If we do stick with the pid -> counter, I think we can improve the performance compared to stringifying every process in every trace.

packages/jaeger-ui/src/actions/jaeger-api.js
```diff
@@ -15,11 +15,48 @@
...
+    });
+
+    return { data: [ot] };
+  };
+}
+
+function reduceFetch(id) {
+  const ids = id.split('|');
+  const promise = ids.length < 2 ? JaegerAPI.fetchTrace(id) : JaegerAPI.searchTraces({ traceID: ids });
+  return promise && promise.then(mergeAll(id));
```
everett980: the `promise &&` shouldn't be necessary. Does `fetchTrace` or `searchTraces` ever return something other than a promise (or throw an error, which would interrupt before this line anyway).

packages/jaeger-ui/src/actions/jaeger-api.js
```diff
@@ -15,11 +15,48 @@
...
+      if (trace.warnings) {
+        trace.warnings.forEach(j => ot.warnings.push(j));
+      }
+    });
+
+    return { data: [ot] };
+  };
+}
+
+function reduceFetch(id) {
```
everett980: `reduce` is a confusing prefix in an actions file, since there is a separate corresponding file called `reducers/jaeger-api`

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/DiffSelection.css
```diff
@@ -32,3 +32,7 @@ limitations under the License.
   border: 1px solid #ccc;
   border-bottom: none;
 }
+
+.DiffSelection--message button {
```
everett980: It would be better to add a class name to these buttons and then you could select them with:
`.DiffSelection--cohortBtn ~ .DiffSelection--cohortBtn {`

the `~` should mean that any `cohortBtn` following a `cohortBtn` would have the `margin-left` styling. if `~` doesn't work, you can replace the selector with `.DiffSelection--cohortBtn:not(:first-of-type)` should also work but has slightly worse browser compatibility (though we usually aren't incredibly concerned with supporting super old browsers).

packages/jaeger-ui/src/actions/jaeger-api.js
```diff
@@ -15,11 +15,48 @@
 import { createAction } from 'redux-actions';
 import JaegerAPI from '../api/jaeger';
 
-export const fetchTrace = createAction(
-  '@JAEGER_API/FETCH_TRACE',
-  id => JaegerAPI.fetchTrace(id),
-  id => ({ id })
-);
+export function mergeAll(id) {
```
everett980: if this is only exported for tests, please add `// exported for tests` above this line

packages/jaeger-ui/src/actions/jaeger-api.js
lookfwd: The issue was that some unit tests returned null.

packages/jaeger-ui/src/actions/jaeger-api.js
lookfwd: > I'm not sure if there is any impact of changing the process IDs from what was originally on the span to a counter.

I think it's not a counter, but a `"p?"` string where `?` in this case is a counter.

As far as I see, it's to correlate `span` data with the `process` data from a single trace. At that point in the pipeline there haven't been any computations or mappings done to it yet. So they're treated as random string identifiers.

> improve the performance

Typically there's very little code that you can optimize and improve performance in a system. I doubt I've touched any performance critical code. :)

packages/jaeger-ui/src/components/TracePage/url.tsx
```diff
@@ -22,6 +22,10 @@ export function getUrl(id: string) {
   return prefixUrl(`/trace/${id}`);
 }
 
+export function getCombineUrl(id: string[]) {
```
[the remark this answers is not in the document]
lookfwd: Great idea!

lookfwd:
All the comments are very good and need to be addressed.

> What does the pid change achieve? There might be a safer/simpler way to accomplish this.

I can see what you mean. Maybe we can prefix the process-IDs with the first part of the trace-id or something.
We can keep this super-clean, but we would have to do all the merging much later, in the UI and it would be more messy. That said, it might be a good idea to do it later.

packages/jaeger-ui/src/actions/jaeger-api.js
everett980: in the unit tests you can mock the return values for JaegerAPI
the code shouldn't need an extra check due to a test case

packages/jaeger-ui/src/actions/jaeger-api.js
everett980: The performance concern here comes from `JSON.stringfy`ing every process on all selected traces. I'm not sure how many processes are typical nor their expected sizes, but stringification can be pretty slow so it would be best to avoid.

lookfwd:
Thanks for all the review effort. This feature doesn't seem to be worth the effort at this point in time. Will see if it makes more sense in the future.

everett980:
> I can see what you mean. Maybe we can prefix the process-IDs with the first part of the trace-id or something.
This sounds like a good approach to me. Replace `span.processID` with``` `t-{traceID}:p-${processID}` ``` would be clear and hopefully still accomplish the same goals.

> We can keep this super-clean, but we would have to do all the merging much later, in the UI and it would be more messy. That said, it might be a good idea to do it later.
The advantage of doing it here is that the multi-trace accumulation will be used by multiple views (gantt chart/timeline and the trace graph)

_If_ we don't want to interlace the different traces, then this is also a good place to add dummy-spans to distinguish the traces. Very psuedo-codey, but in [this `for` loop](https://github.com/jaegertracing/jaeger-ui/pull/490/files#diff-9bcb45f0296bfa94fdd9e368a357d50dR27) something like:
```
const traceSeparatorSpan = {
  traceID: trace.ID,
  process: { serviceName: SOME_LABEL },
  operationName: SOME_LABEL,
};
multiTrace.spans.push(traceSeparatorSpan);
```
then in [this `for` loop](https://github.com/jaegertracing/jaeger-ui/pull/490/files#diff-9bcb45f0296bfa94fdd9e368a357d50dR39):
```
if (/* span.ref does not contain a local reference */) {
  /* add reference to traceSeparatorSpan */
}
```

everett980:
> Thanks for all the review effort. This feature doesn't seem to be worth the effort at this point in time. Will see if it makes more sense in the future.

I'm happy to review! I will see if we need this feature on this side (it's come up before) and possibly take this PR across the finish line. If not, it at least serves as a good basis for revisiting later.

lookfwd:
Great, yes. I'm happy to help but I don't know Jaeger UI enough right now. I have to go study a bit react/redux and the rendering/UI code. With my current level of understanding, I don't believe I will be able to finish this PR in a timely manner. :)

Issue #491, opened 2019-12-02
Jeager UI broken when accessing via IPv6 address


MaheshGPai:
## Requirement - what kind of business use case are you trying to solve?
Deploying jaeger in a pure IPv6 setup

## Problem - what in Jaeger blocks you from solving the requirement?
Logical issue in Jaeger UI module. The logic to determine **pathPrefix** does not consider that the **origin** may contain RegEx characters like "[" & "]"

![image](https://user-images.githubusercontent.com/14892092/69967465-78f99680-153e-11ea-8c4e-1c1dbad6cefe.png)

## Proposal - what do you suggest to solve the problem or improve the existing situation?
Correct the logic to determine pathPrefix

MaheshGPai:
PR https://github.com/jaegertracing/jaeger-ui/pull/494

PR #494, opened 2019-12-03
Fixes Jeager UI broken when accessing via IPv6 address (#491)

MaheshGPai:
Jeager UI broken when accessing via IPv6 address (#491)

Corrected the logic to determine pathPrefix by escaping RegEx characters

### DIFF
--- a/packages/jaeger-ui/src/utils/prefix-url.test.js
+++ b/packages/jaeger-ui/src/utils/prefix-url.test.js
@@ -16,7 +16,7 @@
 
 jest.mock('../site-prefix', () => `${global.location.origin}/a/site/prefix/`);
 
-import prefixUrl from './prefix-url';
+import prefixUrl, { getPathPrefix } from './prefix-url';
 
 const PATH_PREFIX = '/a/site/prefix';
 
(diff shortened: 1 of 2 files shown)


yurishkuro:
could you add unit tests?

packages/jaeger-ui/src/utils/prefix-url.test.js
```diff
@@ -37,3 +38,19 @@ describe('prefixUrl()', () => {
...
+    { source: undefined, target: '' },
+    { source: null, target: '' },
+    { source: `${IPV6_HOST_ADDRESS}`, target: `${IPV6_HOST_ADDRESS}` },
+    { source: `[${IPV6_HOST_ADDRESS}]`, target: `\\[${IPV6_HOST_ADDRESS}\\]` },
+    { source: `[${IPV6_HOST_ADDRESS}]:8080`, target: `\\[${IPV6_HOST_ADDRESS}\\]:8080` },
+  ];
+
+  tests.forEach(({ source, target }) => {
+    it(`regexEscape "${source}" correctly`, () => {
+      expect(regExEscape(source)).toBe(target);
```
yurishkuro: I think a better test would be to move the calculation of pathPrefix into a function and test that function with the permutation of addresses. Testing regExEscape is too low level and unnecessarily exposes implementation detail from the util file.

MaheshGPai: I agree. Have modified the tests.

yurishkuro:
@everett980 please review

packages/jaeger-ui/src/utils/prefix-url.tsx
```diff
@@ -15,6 +15,18 @@
...
 
 const origin = process.env.NODE_ENV === 'test' ? global.location.origin : window.location.origin;
+
+/**
+ * Escape any RegEx characters in a given string
+ *
+ * @param  {string} value The string which needs to be escaped
+ * @return {string} Escaped RegEx string
+ */
+function regExEscape(value?: string) {
```
everett980: this function will always be called with a `string`, so the `?` and the cast on the next line are unnecessary.
But since this is only called once, you could simply take the `replace` on line 27 and use it in the falsy branch of the ternary on line 41, as such:
```
const s = sitePref == null ? '' : sitePref.replace(/[-/\\^$*+?.()|[\]{}]/g, '\\$&');
```
with a comment above saying the `replace` will escape any RegEx characters in the `sitePref`

packages/jaeger-ui/src/utils/prefix-url.test.js
```diff
@@ -37,3 +37,25 @@ describe('prefixUrl()', () => {
...
+    { address: '[2a00:8a00:4000:751e:0000:0000:0000:001c]:8443' },
+    { address: '[::]:443' },
+    { address: '127.0.0.1' },
+    { address: '127.0.0.1:8443' },
+    { address: 'jaeger.mydomain.com:8443' },
+    { address: 'jaeger.mydomain.com' },
+  ];
+
+  tests.forEach(({ address }) => {
+    it(`getPathPrefix "${address}" correctly`, () => {
```
everett980: the test name should read as a sentence, so replace `getPathPrefix` with `handles` and it will read:
```
it(`handles "${address}" correctly`, () => {
```

MaheshGPai:
Have incorporated the requested changes.

Issue #496, opened 2019-12-04
Request list of services if service list is empty


objectiser:
## Requirement - what kind of business use case are you trying to solve?
See services in the dropdown that have reported traces after the UI had already started.

## Problem - what in Jaeger blocks you from solving the requirement?
When starting up a Jaeger UI in dev/test environment, if no traces have been reported the service dropdown is empty. When traces are reported, the user is required to refresh the browser to see the actual list.

## Proposal - what do you suggest to solve the problem or improve the existing situation?
If the service list is empty, then actively query the backend to retrieve the initial list.

## Any open questions to address
We may also want to consider a configurable polling period to automatically update the cached service list.

yurishkuro:
Good idea. I would use the count in parentheses as a progress indicator, ie replace 0 with "loading..." while loading.

What did you mean by cached service list? Afaik UI does not cache it; it's loaded every time you visit the Search page.

objectiser:
> What did you mean by cached service list?

My assumption - I normally just refresh the browser to get an uptodate list - but I guess changing between tabs might work. The main thing is that (ideally) it would be nice if the user does not need to necessarily force that update to see the latest list.

yurishkuro:
right, so `state` rather than `cache` of the UI.

PR #497, opened 2019-12-07
Request list of services if service list is empty

rubenvp8510:

- This address the issue mentioned on https://github.com/jaegertracing/jaeger-ui/issues/496 

- Added a polling timer for fetch services from the backend each N seconds (this should be configurable).

This is work in progress as I need to fix some details and polish the work.

### DIFF
--- a/packages/jaeger-ui/src/components/SearchTracePage/SearchForm.js
+++ b/packages/jaeger-ui/src/components/SearchTracePage/SearchForm.js
@@ -264,16 +264,19 @@ export class SearchFormImpl extends React.PureComponent {
       services,
       submitting: disabled,
     } = this.props;
-    const selectedServicePayload = services.find(s => s.name === selectedService);
+    const servicesLoaded = services != null;
+    const servicesList = (servicesLoaded && services.map(v => ({ label: v.name, value: v.name }))) || [];
+    const selectedServicePayload = (servicesLoaded && services.find(s => s.name === selectedService)) || [];
     const opsForSvc = (selectedServicePayload && selectedServicePayload.operations) || [];
     const noSelectedService = selectedService === '-' || !selectedService;
     const tz = selectedLookback === 'custom' ? new Date().toTimeString().replace(/^.*?GMT/, 'UTC') : null;
+    const servicesCount = (servicesLoaded && services.length) || 'Loading...';
     return (
       <Form layout="vertical" onSubmit={handleSubmit}>
         <FormItem
           label={
             <span>
-              Service <span className="SearchForm--labelCount">({services.length})</span>
+              Service <span className="SearchForm--labelCount">({servicesCount})</span>
             </span>
           }
         >
@@ -284,7 +287,7 @@ export class SearchFormImpl extends React.PureComponent {
             props={{
               disabled,
               clearable: false,
-              options: services.map(v => ({ label: v.name, value: v.name })),
+              options: servicesList,
               required: true,
             }}
           />
@@ -513,7 +516,7 @@ export function mapStateToProps(state) {
     // last search is only valid if the service is in the list of services
     const { operation: lastOp, service: lastSvc } = lastSearch;
     if (lastSvc && lastSvc !== '-') {
-      if (state.services.services.includes(lastSvc)) {
+      if (state.services && state.services.services && state.services.services.includes(lastSvc)) {
         lastSearchService = lastSvc;
         if (lastOp && lastOp !== '-') {
           const ops = state.services.operationsForService[lastSvc];
--- a/packages/jaeger-ui/src/components/SearchTracePage/index.js
+++ b/packages/jaeger-ui/src/components/SearchTracePage/index.js
@@ -22,6 +22,7 @@ import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
 import store from 'store';
 import memoizeOne from 'memoize-one';
+import _get from 'lodash/get';
 
 import SearchForm from './SearchForm';
 import SearchResults, { sortFormSelector } from './SearchResults';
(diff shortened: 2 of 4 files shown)


packages/jaeger-ui/src/components/SearchTracePage/index.js
```diff
@@ -44,6 +43,8 @@ const TabPane = Tabs.TabPane;
 
 // export for tests
 export class SearchTracePageImpl extends Component {
+  pollingTimer = null;
```
yurishkuro: name is too general, this component deals with other things as well

packages/jaeger-ui/src/components/SearchTracePage/index.js
```diff
@@ -63,12 +64,24 @@ export class SearchTracePageImpl extends Component {
       fetchMultipleTraces(needForDiffs);
     }
     fetchServices();
+    this.pollingForServices();
```
yurishkuro: `this.` is not used with other functions

packages/jaeger-ui/src/components/SearchTracePage/index.js
```diff
@@ -63,12 +64,24 @@ export class SearchTracePageImpl extends Component {
...
     if (service && service !== '-') {
       fetchServiceOperations(service);
     }
   }
 
+  componentWillUnmount() {
+    clearTimeout(this.pollingTimer);
+  }
+
+  pollingForServices() {
```
yurishkuro: use imperative verb: pollForServices

packages/jaeger-ui/src/reducers/services.js
```diff
@@ -19,7 +19,7 @@ import { localeStringComparator } from '../utils/sort';
 
 const initialState = {
   // `services` initial value of `null` indicates they haven't yet been loaded
-  services: null,
+  services: [],
```
yurishkuro: how will we distinguish not loaded vs. empty?

packages/jaeger-ui/src/components/SearchTracePage/index.js
rubenvp8510: I'll change it

packages/jaeger-ui/src/components/SearchTracePage/index.js
rubenvp8510: +1

packages/jaeger-ui/src/reducers/services.js
rubenvp8510: I was thinking on add a new flag on the state. This is some of the things I still need to add to this PR.

packages/jaeger-ui/src/components/SearchTracePage/index.js
```diff
@@ -63,12 +67,23 @@ export class SearchTracePageImpl extends Component {
       fetchMultipleTraces(needForDiffs);
     }
     fetchServices();
+    if (servicePollingEnabled === true) {
+      this.pollServicesTimer = setInterval(() => {
+        fetchServices();
```
everett980: Unfortunately this change will cause the UI to re-render twice per interval. [`fetchStarted` will set `state.services.loading` to `true`](https://github.com/jaegertracing/jaeger-ui/blob/master/packages/jaeger-ui/src/reducers/services.js#L29) when the long polling request is raised, then [`fetchDone` will set `state.services.loading` to `false`](https://github.com/jaegertracing/jaeger-ui/blob/master/packages/jaeger-ui/src/reducers/services.js#L35) even if nothing changed between the requests.

`fetchStarted` should probably only set `loading` to `true` if `state.services` is `null` or `[]` else return the existing state:
```
if (state.services && state.services.length) return state;
return { ...state, loading: true };
```

and `fetchDone` should only change the state if it was previously loading or if the services have changed (fortunately `services` are sorted anyway so this check won't increase time complexity):
```
const services = payload.data || [];
services.sort(localeStringComparator);
if (state.loading || !state.services || services.length !== state.services.length || services.some((service, i) => service !== state.services[i]) {
  return { ...state, services, error: null, loading: false };
}
return state;
```

packages/jaeger-ui/src/constants/default-config.tsx
```diff
@@ -57,6 +57,8 @@ export default deepFreeze(
         },
       ],
       search: {
+        servicePolling: true,
+        servicePollingInterval: 5000,
```
everett980: should this be defaulted to enabled? I would think that this should be opt-in.

packages/jaeger-ui/src/components/SearchTracePage/SearchForm.js
```diff
@@ -264,16 +264,19 @@ export class SearchFormImpl extends React.PureComponent {
...
       submitting: disabled,
     } = this.props;
-    const selectedServicePayload = services.find(s => s.name === selectedService);
+    const servicesLoaded = services != null;
+    const servicesList = (servicesLoaded && services.map(v => ({ label: v.name, value: v.name }))) || [];
+    const selectedServicePayload = (servicesLoaded && services.find(s => s.name === selectedService)) || [];
     const opsForSvc = (selectedServicePayload && selectedServicePayload.operations) || [];
     const noSelectedService = selectedService === '-' || !selectedService;
     const tz = selectedLookback === 'custom' ? new Date().toTimeString().replace(/^.*?GMT/, 'UTC') : null;
+    const servicesCount = (servicesLoaded && services.length) || 'Loading...';
```
everett980: this will result in `'Loading...'` if `services.length` is `0` regardless of if long polling services is enabled.

packages/jaeger-ui/src/constants/default-config.tsx
everett980: although I suppose the main use case would be when developing, where a config is not available 🤔

nadilas:
Any updates on this?

PR #502, opened 2019-12-18
Add adblocker and 0/single node disclaimers

everett980:
- DDG request is blocked by some adblockers
- Plexus does not render single vertex

- Add disclaimer to whitelist Jaeger on DDG request failure details view.
image
("New!" banner not in diff)
- Add disclaimer when there are zero or fewer vertices to view.
No dependencies found:
image
Dependencies found but hidden:
image

This PR is branched off of [another PR](https://github.com/jaegertracing/jaeger-ui/pull/501), [this commit contains the diff for this PR](https://github.com/jaegertracing/jaeger-ui/pull/502/commits/00a51f62bbd88e175edca6fa5285223f3de6112c)

### DIFF
--- a/packages/jaeger-ui/src/constants/default-config.tsx
+++ b/packages/jaeger-ui/src/constants/default-config.tsx
@@ -20,10 +20,6 @@ export default deepFreeze(
   Object.defineProperty(
     {
       archiveEnabled: false,
-      // TODO: remove default
-      deepDependencies: {
-        menuEnabled: true,
-      },
       dependencies: {
         dagMaxNumServices: FALLBACK_DAG_MAX_NUM_SERVICES,
         menuEnabled: true,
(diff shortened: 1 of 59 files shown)


yurishkuro:
there seems to be many changes not related to just disclaimers

everett980:
@yurishkuro the PR is branched off of another PR. There is a link in the PR description to the commit that contains the disclaimer changes.

everett980:
The middle branch was merged, so the diff is good now (except for two files with TODO delete changes comments that I definitely deleted on the middle branch.....)

packages/jaeger-ui/src/constants/default-config.tsx
```diff
@@ -20,6 +20,10 @@ export default deepFreeze(
   Object.defineProperty(
     {
       archiveEnabled: false,
+      // TODO: remove default
+      deepDependencies: {
+        menuEnabled: true,
```
yurishkuro: Do we really want that in oss?  There's no backend to generate DDGs.

everett980: yeah hence the TODO and comment above that I thought TODOs were dealt with...
definitely won't have this on by default. same with the `New!` css

everett980: Fixed

PR #504, opened 2019-12-19
Fix search results DDG path ordering

### LINKED ISSUE #485: Search results DDG looks incorrect for hotrod data
@everett980 a bunch of edges in this graph are incorrect

Screen Shot 2019-11-18 at 5 40 47 PM

everett980:
Signed-off-by: Everett Ross <reverett@uber.com>

- Closes #485

- Paths were reversed, except last element of path was still last (i.e.: 12345 was rendered as 43215)
- Paths were not filtered by kind, resulting in client span noise

### DIFF
--- a/packages/jaeger-ui/src/model/ddg/transformTracesToPaths.test.js
+++ b/packages/jaeger-ui/src/model/ddg/transformTracesToPaths.test.js
@@ -15,7 +15,7 @@
 import transformTracesToPaths from './transformTracesToPaths';
 
 describe('transform traces to ddg paths', () => {
-  const makeSpan = (spanName, childOf) => ({
+  const makeSpan = (spanName, childOf, kind) => ({
     hasChildren: true,
     operationName: `${spanName} operation`,
     processID: `${spanName} processID`,
@@ -28,6 +28,15 @@ describe('transform traces to ddg paths', () => {
           },
         ]
       : [],
+    tags:
+      kind === false
+        ? []
+        : [
+            {
+              key: 'span.kind',
+              value: kind === undefined ? 'server' : kind,
+            },
+          ],
     spanID: `${spanName} spanID`,
   });
   const makeTrace = (spans, traceID) => ({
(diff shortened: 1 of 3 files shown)


everett980:
> 1. Big nit for test logic: it would be a LOT easier to read if the test used letters to denote spans, rather than spelled out names, e.g. `makeTrace([A, B, C])` results in `path(A, B, C)`.

IME with tree-structured data, names that include relationships help with debugging. Granted the test trees here are small so they're not that hard to follow.
Also, test names like `kindlessSpan` and `clientSpan` are more clear about their distinguishing detail than `C` or `D`. If the spans are `A`, `B`, `C`, and `D` it's less clear why `transformTracesToPaths({ c: makeTrace([A, B, C]), d: makeTrace([A, B, D]) })` results in two paths of length two.

> 2. I am generally not a fan of service graph logic depending on span.kind tags. An edge between two services can be established purely from the transition between two services, regardless of the span annotations. So I am not clear as to the intention of the added test. May want to check the logic in the memory storage (although this one is for p2p graphs, but the detection is similar: https://github.com/jaegertracing/jaeger/blob/8b1ebe6be8583b86968e28fd9727c53f02fcd62b/plugin/storage/memory/memory.go#L61).

I thought that @vprithvi said we changed the definition of an edge from "change in serviceName" to "`span.kind===server` spans". There are pros and cons for each approach, but I'd like search results DDGs to match the flink job DDGs. @vprithvi can you confirm which definition we are using? If both definitions are valid, we can have a UI toggle or a config key to determine which rule we use for search results DDGs.

> 3. Could we introduce some simple JSON-based DSL for describing the traces, to make the data much more readable and the tests data driven? Such as:
> 
> ```json
> {
>   "test 1": {
>      "traces": [ 
>         [
>             { "service": "X", "name": "A" },
>             { "service": "Y", "name": "B", "parent": "A" },
>             { "service": "Z", "name": "C", "parent": "A" },
>         ],
>      ],
>      "expectedPaths": [
>         [
>             { "service": "X", "name": "A" },
>             { "service": "Y", "name": "B" },
>         ],
>         [
>             { "service": "X", "name": "A" },
>             { "service": "Z", "name": "C" },
>         ],
>      ],
>    }
> ```
> 
> Note that trace IDs are largely irrelevant, so can represent them just as arrays of spans, and span names can double as IDs in the DSL.

I agree that the `traceGenerator` used for tests should be beefed up, but I think that exceeds the scope of this fix.

packages/jaeger-ui/src/model/ddg/transformTracesToPaths.test.js
```diff
@@ -15,7 +15,14 @@
...
 describe('transform traces to ddg paths', () => {
-  const makeSpan = (spanName, childOf) => ({
+  const makePath = (pathSpans, trace) => ({
+    path: pathSpans.map(({ processID, operationName: operation }) => ({
+      service: trace.data.processes[processID].serviceName,
+      operation,
+    })),
+    attributes: [{ key: 'exemplar_trace_id', value: trace.data.traceID }],
+  });
+  const makeSpan = (spanName, childOf, kind) => ({
```
yurishkuro: `childOf` is a poor name. What is passed is a `parent`.

packages/jaeger-ui/src/model/ddg/transformTracesToPaths.test.js
```diff
@@ -45,109 +61,111 @@ describe('transform traces to ddg paths', () => {
       traceID,
```
yurishkuro: I'm a bit confused by how service names are assigned in L55. Looks like a string `"{first word of span id} service"`, but the first word of span is, per L47, is span name. So it seems _every_ span in these tests will have a different service name.

packages/jaeger-ui/src/model/ddg/transformTracesToPaths.test.js
everett980: good point, updated

packages/jaeger-ui/src/model/ddg/transformTracesToPaths.test.js
everett980: Yes every span gets a unique service and operation name. currently service name is not used as part of the trace to path calculation. once we settle on a specification for the flink job we will apply the same rule here and likely need to tweak the test.

PR #511, opened 2020-01-14
Configure search.maxLimit

### LINKED ISSUE #478: Configure max limit result in SearchForm.js

Hello!
I want to configure searchform "Limit Results" max property. 

I navigated the code and saw that is hardcoded 1500 as max limit in SearchForm.js

I think we can configure like environmet value the max limit, and if isn't configured, we can set 1500 as default value.

What do you think? 

Thank you !

GabrielDyck:
Signed-off-by: Gabriel Fernando Dyck <gabrielfernandodyck@gmail.com>

- Resolves #478

### DIFF
--- a/packages/jaeger-ui/src/components/SearchTracePage/SearchForm.js
+++ b/packages/jaeger-ui/src/components/SearchTracePage/SearchForm.js
@@ -34,7 +34,7 @@ import * as jaegerApiActions from '../../actions/jaeger-api';
 import { formatDate, formatTime } from '../../utils/date';
 import reduxFormFieldAdapter from '../../utils/redux-form-field-adapter';
 import { DEFAULT_OPERATION, DEFAULT_LIMIT, DEFAULT_LOOKBACK } from '../../constants/search-form';
-
+import getConfigValue from '../../utils/config/get-config';
 import './SearchForm.css';
 
 const FormItem = Form.Item;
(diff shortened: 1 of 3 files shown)


yurishkuro:
it should go into this file: https://github.com/jaegertracing/documentation/blob/master/content/docs/next-release/frontend-ui.md

GabrielDyck:
Hello!

I've added documentation for this in the PR https://github.com/jaegertracing/documentation/pull/348

Let me know if it's ok.
Thank you!

packages/jaeger-ui/src/components/SearchTracePage/SearchForm.js
```diff
@@ -443,7 +443,7 @@ export class SearchFormImpl extends React.PureComponent {
             type="number"
             component={AdaptedInput}
             placeholder="Limit Results"
-            props={{ disabled, min: 1, max: 1500 }}
```
yurishkuro: does the default-config below provide defaults per value, or only as a whole configuration?

GabrielDyck: It takes default per value.
If we configure maxLimit as 500, It will take 500.  If the value "maxLimit" isn't configure, it will take the default value, in this case, 1500.
@everett980 could you tell me if I am wrong?
Thank you!

everett980: @gabrieldyck you are correct. `default-config` is consumed per value.

everett980:
Documentation was merged, so this LGTM. Thanks for your patience!

PR #515, opened 2020-02-04
Add ga dimension for config cookie

everett980:
- Ability to specify custom dimensions.

- Read cookie and dimension names from config, if cookie is present, set dimension to cookie value.

### DIFF
--- a/packages/jaeger-ui/src/types/config.tsx
+++ b/packages/jaeger-ui/src/types/config.tsx
@@ -49,7 +49,7 @@ export type Config = {
     cookieToDimension?: {
       cookie: string;
       dimension: string;
-    };
+    }[];
     gaID: string | TNil;
     trackErrors: boolean | TNil;
   };
--- a/packages/jaeger-ui/src/types/config.tsx
+++ b/packages/jaeger-ui/src/types/config.tsx
@@ -46,6 +46,10 @@ export type Config = {
   scripts?: TScript[];
   topTagPrefixes?: string[];
   tracking?: {
+    cookieToDimension?: {
+      cookie: string;
+      dimension: string;
+    };
     gaID: string | TNil;
     trackErrors: boolean | TNil;
   };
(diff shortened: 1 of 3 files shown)


packages/jaeger-ui/src/types/config.tsx
```diff
@@ -46,6 +46,10 @@ export type Config = {
   scripts?: TScript[];
   topTagPrefixes?: string[];
   tracking?: {
+    cookieToDimension?: {
```
yurishkuro: perhaps this should be an array of mappings, so that it can be easily extended to other cookies (e.g. user's org, if we were able to get it).

packages/jaeger-ui/src/utils/tracking/index.tsx
```diff
@@ -45,6 +45,13 @@ const gaID = _get(config, 'tracking.gaID');
 export const isGaEnabled = isTest || isDebugMode || (isProd && Boolean(gaID));
 const isErrorsEnabled = isDebugMode || (isGaEnabled && Boolean(_get(config, 'tracking.trackErrors')));
 
+const cookieToDimension = _get(config, 'tracking.cookieToDimension');
+if (cookieToDimension) {
+  const match = ` ${document.cookie}`.match(new RegExp(`[; ]${cookieToDimension.cookie}=([^\\s;]*)`));
```
yurishkuro: - what is the meaning of the leading space?
- are you forced to use regexp because there is no structured API to access cookies?

packages/jaeger-ui/src/types/config.tsx
everett980: I was on the fence about it but just updated it.

packages/jaeger-ui/src/utils/tracking/index.tsx
everett980: 1) Leading space because the regex looks for `;` and/or ` ` as delimiter characters between cookies. and/or because the ` ` is optional in the spec. The leading space is necessary if the target cookie is the first cookie in the `document.cookie` string.
2) Correct. There are npm libraries for this (of course) but a library for a simple RegEx is overkill, and libraries may make certain assumptions about the cookies that aren't relevant here, meanwhile it's easy to know what our own RegEx is precisely doing.

everett980:
@yurishkuro 
> Any tests we should add?
The other code in the file run on JS load is also untested. I'm not sure of a good approach for it, so I'd leave it for now.

I'll dogfood it in staging w/ the config change before landing.

everett980:
Dogfooded successfully.

PR #516, opened 2020-02-05
add iconfont to remove the dependency on the external cdn

### LINKED ISSUE #187: Remove the dependency on `alicdn.com`
Jaeger might be deployed on a restricted network, so, depending on resources from `alicdn.com` should be avoided.

image

rgl:

This closes #187 

It would be nice not using the absolute address in `@icon-url: "/iconfont/iconfont"` but I didn't find a way.

This also adds a GitHub Actions Workflow to build and archive the result in the workflow result.


yurishkuro:
Why do we need to commit the font files to this repo? Where are they from? Which license? Couldn't we download them from an official location during build time?

packages/jaeger-ui/config-overrides-antd-vars.less
```diff
@@ -22,3 +22,7 @@
 
 // Table
 @table-row-hover-bg:#e5f2f2;
+
+// Font
+// see https://github.com/ant-design/antd-init/tree/master/examples/local-iconfont
+@icon-url: "/iconfont/iconfont"
```
yurishkuro: wouldn't this break people who're using `--query.base-path` option?

packages/jaeger-ui/config-overrides-antd-vars.less
```diff
@@ -22,3 +22,7 @@
 
 // Table
 @table-row-hover-bg:#e5f2f2;
+
+// Font
+// see https://github.com/ant-design/antd-init/tree/master/examples/local-iconfont
```
yurishkuro: That repo appears to be archived. What is the risk of using it?

rgl: There is no risk because the fonts were included in this MR. I've only placed that comment to hint from where the fonts came from.

And from what I understood, it was only archived because newer versions of antd library no longer use the cdn.

packages/jaeger-ui/config-overrides-antd-vars.less
rgl: I think it will. But I don't known enough about this antd library to known how to make that configurable. I think its worse to use the cdn than working around that at the http server level.

rgl:
> Why do we need to commit the font files to this repo?

It seemed the simpler solution.

> Where are they from?

They come from https://github.com/ant-design/antd-init/tree/master/examples/local-iconfont/iconfont.

> Which license?

I guess its the same as the antd library itself.

> Couldn't we download them from an official location during build time?

I guess we could download them at runtime from https://github.com/ant-design/antd-init/tree/master/examples/local-iconfont/iconfont, or even fork that repo and make a proper release of it.

everett980:
> or upgrade to a newer version of the antd library (which no longer depends on the cdn)

@rgl it seems to me that this could be the simplest option. though it will require manually checking and potentially fixing any styles that override antd

Issue #508, opened 2020-01-13
Extract TracePage components [Subtask]


aocenas:
## Requirement - what kind of business use case are you trying to solve?

Subtask from https://github.com/jaegertracing/jaeger-ui/issues/507

## Proposal - what do you suggest to solve the problem or improve the existing situation?

Way to extract this depends on what is required to be importable. For Grafana use case we only need part of the Trace page seen in red rectangle:
<img width="1416" alt="embed-trace-view-with-back-button" src="https://user-images.githubusercontent.com/1014802/72264412-0559a780-361b-11ea-8d4d-23e3d16b6a58.png">

This means toolbar and archiving components can be either hide-able by props or we could go the route of providing separate more low level components directly like Toolbar, Minimap, TimelineView, GraphView etc to be composed into an application as needed.

(@everett980): I expect this to be the comparatively straightforward part of this effort, however we will need to be mindful of the contract that gets created.
Global dependencies like redux and url state can be worked around by exporting both "connected" and "Impl" (implementation) forms of components, as done/consumed for Deep Dependency Graphs, and then adding a build process to create a standalone file for each desired "Impl" component. Some cleanup will be necessary for TracePage and its subcomponents, and the types should be significantly tightened for these use cases.
Fortunately, tracking is already configurable and off by default.

I think having a mixed connected and unconnected components can be avoided especially if lower level components are exposed. Redux (or other state management) can be left as a concern for the final application. If there is need to sidestep deep props passing, react context api could be used instead to have a localized and separated state management.

Another consideration is Antd component library dependency which may seem as too big dependency for such a package and also contains some global style overrides. This overlaps a bit with theme-ability.

## Any open questions to address

- What needs to be exported? One complete page component or a set of lower level components.
- Do we want/need to export connected components in the library?
- Can we remove AntD dependency? If so what to replace it with?

tiffon:
You might also consider creating a new package / yarn workspace which holds the components that are to be broken out. A few potential upsides:

- Decisive separation will likely make it much easier to determine what the boundaries between components should be and to understand those boundaries later and for new contributors
- It will likely be much easier to prevent coupling from creeping in
- Bugs that only manifest when the components are imported will be much more likely to be detected since Jaeger UI, itself, would be exercising that scenario (and presumably Uber would dog-food releases, too)
- The UI would serve as documentation for how to import and use the separated components

[...]

PR #517, opened 2020-02-06
Move redux up from VirtualizedTraceView

aocenas:

- Part of: https://github.com/jaegertracing/jaeger-ui/issues/508

- Pushing redux and url creation (focusSpan) dependencies up from VirtualizedTraceView and TraceTimelineViewer to TracePage so it will be easier to move them to separate package.

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
@@ -14,13 +14,7 @@
 
 import * as React from 'react';
 import cx from 'classnames';
-import { connect } from 'react-redux';
-import { bindActionCreators, Dispatch } from 'redux';
-import { withRouter, RouteComponentProps } from 'react-router-dom';
 
-// import { History as RouterHistory, Location } from 'history';
-
-import { actions } from './duck';
 import ListView from './ListView';
 import SpanBarRow from './SpanBarRow';
 import DetailState from './SpanDetail/DetailState';
@@ -33,15 +27,14 @@ import {
   ViewedBoundsFunctionType,
 } from './utils';
 import { Accessors } from '../ScrollManager';
-import { extractUiFindFromState, TExtractUiFindFromStateReturn } from '../../common/UiFindInput';
+import { TExtractUiFindFromStateReturn } from '../../common/UiFindInput';
 import getLinks from '../../../model/link-patterns';
 import colorGenerator from '../../../utils/color-generator';
-import { TNil, ReduxState } from '../../../types';
+import { TNil } from '../../../types';
 import { Log, Span, Trace, KeyValuePair } from '../../../types/trace';
 import TTraceTimeline from '../../../types/TTraceTimeline';
 
 import './VirtualizedTraceView.css';
-import updateUiFind from '../../../utils/update-ui-find';
 
 type RowState = {
   isDetail: boolean;
@@ -55,9 +48,9 @@ type TVirtualizedTraceViewOwnProps = {
   scrollToFirstVisibleSpan: () => void;
   registerAccessors: (accesors: Accessors) => void;
   trace: Trace;
-};
+  focusSpan: (uiFind: string) => void;
 
-type TDispatchProps = {
+  // was from redux
   childrenToggle: (spanID: string) => void;
   clearShouldScrollToFirstUiFindMatch: () => void;
   detailLogItemToggle: (spanID: string, log: Log) => void;
@@ -69,14 +62,11 @@ type TDispatchProps = {
   detailToggle: (spanID: string) => void;
   setSpanNameColumnWidth: (width: number) => void;
   setTrace: (trace: Trace | TNil, uiFind: string | TNil) => void;
-  focusUiFindMatches: (trace: Trace, uiFind: string | TNil, allowHide?: boolean) => void;
 };
 
 type VirtualizedTraceViewProps = TVirtualizedTraceViewOwnProps &
-  TDispatchProps &
   TExtractUiFindFromStateReturn &
-  TTraceTimeline &
-  RouteComponentProps;
+  TTraceTimeline;
 
 // export for tests
 export const DEFAULT_HEIGHTS = {
@@ -139,7 +129,7 @@ function getCssClasses(currentViewRange: [number, number]) {
 }
 
 // export from tests
-export class VirtualizedTraceViewImpl extends React.Component<VirtualizedTraceViewProps> {
+export default class VirtualizedTraceView extends React.Component<VirtualizedTraceViewProps> {
   clippingCssClasses: string;
   listView: ListView | TNil;
   rowStates: RowState[];
@@ -222,18 +212,6 @@ export class VirtualizedTraceViewImpl extends React.Component<VirtualizedTraceVi
       clearShouldScrollToFirstUiFindMatch();
     }
   }
-
-  focusSpan = (uiFind: string) => {
-    const { trace, focusUiFindMatches, location, history } = this.props;
-    if (trace) {
-      updateUiFind({
-        location,
-        history,
-        uiFind,
-      });
-      focusUiFindMatches(trace, uiFind, false);
-    }
-  };
 
   getAccessors() {
     const lv = this.listView;
@@ -332,6 +310,7 @@ export class VirtualizedTraceViewImpl extends React.Component<VirtualizedTraceVi
       findMatchesIDs,
       spanNameColumnWidth,
       trace,
+      focusSpan,
     } = this.props;
     // to avert flow error
     if (!trace) {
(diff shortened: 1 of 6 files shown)


packages/jaeger-ui/src/components/TracePage/index.test.js
```diff
@@ -699,15 +700,43 @@ describe('<TracePage>', () => {
...
+      });
+      expect(focusUiFindMatchesMock).toHaveBeenLastCalledWith(trace, spanName, false);
+    });
+  });
 });
 
 describe('mapDispatchToProps()', () => {
   it('creates the actions correctly', () => {
-    expect(mapDispatchToProps(() => {})).toEqual({
+    expect(mapDispatchToProps(() => {})).toMatchObject({
```
aocenas: There are few more functions because of the usage of:
```
...((bindActionCreators(timelineActions, dispatch) as unknown) as TDispatchPropsTimelineViewer),
```
In the map dispatch see a comment there.

packages/jaeger-ui/src/components/TracePage/index.tsx
```diff
@@ -430,15 +475,20 @@ export function mapStateToProps(state: ReduxState, ownProps: TOwnProps): TReduxP
...
 export function mapDispatchToProps(dispatch: Dispatch<ReduxState>): TDispatchProps {
   const { fetchTrace } = bindActionCreators(jaegerApiActions, dispatch);
   const { archiveTrace, acknowledge: acknowledgeArchive } = bindActionCreators(archiveActions, dispatch);
-  const { focusUiFindMatches } = bindActionCreators(timelineActions, dispatch);
-  return { acknowledgeArchive, archiveTrace, fetchTrace, focusUiFindMatches };
+  return {
+    acknowledgeArchive,
+    archiveTrace,
+    fetchTrace,
+    ...((bindActionCreators(timelineActions, dispatch) as unknown) as TDispatchPropsTimelineViewer),
```
aocenas: Not sure if the preferred way to do this. Seen it at few places and is definitely shorter than destructuring all the functions and then assign them again there is more stuff passed to the component than necessary needed.

aocenas:
Moving the redux up and passing an all the handlers makes for a pretty big props list but also I realised that all of the redux actions are simple UI toggles like expanding span. I wonder if this is so that navigating to search and back to the same trace will preserve the UI state or if that is just unneeded side-effect. Because I think lots of those could be hidden in a local component state making the props list shorter and probably making the components a bit easier to use if do not need to preserve the UI state on navigation.

Another question is whether to keep the `/TraceTimelineViewer/duck.tsx` where it is. I think if we keep the redux state it makes sense to still have it split like this ie having `TracePage/traceTimelineViewer.duck.tsx` and  `TracePage/archiveNotifier.duck.tsx` as separate files but can also be moved and merged into `TracePage/duck.tsx` later on.

packages/jaeger-ui/src/components/TracePage/index.test.js
```diff
@@ -80,13 +80,14 @@ describe('makeShortcutCallbacks()', () => {
 });
 
 describe('<TracePage>', () => {
-  TraceTimelineViewer.prototype.shouldComponentUpdate.mockReturnValue(false);
+  TraceTimelineViewer.mockReturnValue('');
```
everett980: what does this do? I've never mocked the return value of a React component before.

packages/jaeger-ui/src/components/TracePage/index.test.js
```diff
@@ -699,15 +700,43 @@ describe('<TracePage>', () => {
       expect(track.trackRange.mock.calls).toEqual([[src, range, [0, 1]]]);
     });
   });
+
+  describe('focusSpan', () => {
+    it('calls updateUiFind and focusUiFindMatches', () => {
+      const spanName = 'span1';
+      wrapper.instance().focusSpan(spanName);
```
everett980: I believe to follow the pattern (generally) present in the code base, this should be: `wrapper.find(TraceTimelineView).prop('focusSpan')(spanName);`

reason being: testing `wrapper.instance().fnName(arg)` can lean towards testing implementation, while testing `someChildComponent.prop('fnName')(arg)` tests functionality (in this case child component is able to set and focus uiFind)

packages/jaeger-ui/src/components/TracePage/index.tsx
```diff
@@ -45,24 +45,43 @@ import { TUpdateViewRangeTimeFunction, IViewRange, ViewRangeTimeUpdate } from '.
...
+import { Log, Span, Trace } from '../../types/trace';
 import { TraceArchive } from '../../types/archive';
 import { EmbeddedState } from '../../types/embedded';
 import filterSpans from '../../utils/filter-spans';
 import updateUiFind from '../../utils/update-ui-find';
 
 import './index.css';
+import TTraceTimeline from '../../types/TTraceTimeline';
+
+type TDispatchPropsTimelineViewer = {
```
everett980: this type should exist somewhere that `VirtualizedTraceView` can import it, to be DRYer.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
```diff
@@ -55,9 +48,9 @@ type TVirtualizedTraceViewOwnProps = {
   scrollToFirstVisibleSpan: () => void;
   registerAccessors: (accesors: Accessors) => void;
   trace: Trace;
-};
+  focusSpan: (uiFind: string) => void;
 
-type TDispatchProps = {
+  // was from redux
```
everett980: rather than commenting this, a type should be imported and `TVirtualizedTraceViewOwnProps` can be called `TProps` (no separation of own vs dispatch props) which would look like:
```
TProps = {
  currentViewRangeTime: [number, number];
  findMatchesIDs: Set<string> | TNil;
  focusSpan: (uiFind: string) => void;
  scrollToFirstVisibleSpan: () => void;
  registerAccessors: (accesors: Accessors) => void;
  trace: Trace;
} & TImportedType;
```

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
```diff
@@ -332,6 +310,7 @@ export class VirtualizedTraceViewImpl extends React.Component<VirtualizedTraceVi
       findMatchesIDs,
       spanNameColumnWidth,
       trace,
+      focusSpan,
```
everett980: admittedly nit, but please destructure and pass `focusSpan` in alphabetical order.

packages/jaeger-ui/src/components/TracePage/index.tsx
everett980: perhaps `TTraceTimelineDispatchFns` in `../../types/TTraceTimeline`

packages/jaeger-ui/src/components/TracePage/index.tsx
```diff
@@ -71,17 +90,17 @@ type TOwnProps = {
   match: Match<{ id: string }>;
 };
 
-type TReduxProps = {
+type TReduxProps = TExtractUiFindFromStateReturn & {
```
everett980: :+1:

packages/jaeger-ui/src/components/TracePage/index.tsx
everett980: I admittedly have a bias towards verbosity, but I find it clearer and preferable to destructure everything that is necessary and pass it along, rather than `(bindActionCreators(actions, dispatch) as unknown) as TType`.

It's a little more verbose in this case, but I'd still prefer it over the cast.

everett980:
> I wonder if this is so that navigating to search and back to the same trace will preserve the UI state or if that is just unneeded side-effect.

While the case of going from a trace back to search back to that trace in a single session is probably pretty rare, I believe this implementation also ensures that going from the Trace Gantt Chart to the Trace Graph and back to the Gantt Chart preserves the visible spans / span details / KV tables within span details. For this reason I believe it visibility state should not be lost on unmount.

However, if we wanted to simplify the props list and keep this functionality, `TraceTimelineViewer` could tuck its visibility state away into session storage [on `componentWillUnmount`](https://reactjs.org/docs/react-component.html#componentwillunmount) then restored in the `constructor` (not, not [`componentWillMount` as its deprecated](https://reactjs.org/docs/react-component.html#unsafe_componentwillmount)).

Using session storage instead of a highly verbose props list is actually probably the best approach as it will lessen the burden to embed this component.

everett980:
> I think if we keep the redux state it makes sense to still have it split like this ie having `TracePage/traceTimelineViewer.duck.tsx` and `TracePage/archiveNotifier.duck.tsx` as separate files but can also be moved and merged into `TracePage/duck.tsx` later on.

Using session storage makes this a moot point, but in case a similar pattern comes up later:
I also like having the divide between these two `duck`s. I think the best approach would be to have a `TracePage/duck.tsx` file that simply imports the two `duck`s and exports the full redux functionality for `TracePage`

PR #518, opened 2020-02-06
Push external link creation up to TracePage

aocenas:

- Part of https://github.com/jaegertracing/jaeger-ui/issues/508

- Extracts `linkToExternalSpan` and make it into external dependency. It also passes it on in a context so it does not have to be passed on through all the layers which seemed beneficial here but can be changed to standard props passing.


packages/jaeger-ui/src/components/TracePage/url/externalLinkContext.tsx
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
everett980: please update to 2020

Issue #509, opened 2020-01-13
Make TracePage components themeable [Subtask]


aocenas:
## Requirement - what kind of business use case are you trying to solve?

Subtask from https://github.com/jaegertracing/jaeger-ui/issues/507

## Proposal - what do you suggest to solve the problem or improve the existing situation?

The resulting package should provide some theme-ability options. For the Grafana usecase there is concern of making the look and feel more aligned with other parts of Grafana and also that we provide dark/light theme.

There are 2 main ways to do this:
- css3/scss/less styles with overridable variables. Right now Jaeger UI uses plain css that is loaded by webpack so may be closer to what is used right now. It is also used in AntD that Jaeger UI already uses https://ant.design/docs/react/customize-theme. On the other hand, consumers need to wire up the styling solution in their webpack (or other build system) configuration.
- CSS in JS solutions. This allows for theming purely in JS/TS context by using ThemeProvider wrappers. This approach is used for example in MaterialUI https://material-ui.com/customization/theming/. This allows for less configuration and theming can be to some extent type safe if used with Typescript. Also Grafana uses CSS in JS solution heavily (EmotionJS).

Apart from changes in styles there may be a need for injection of custom components. For example using a custom button or icon component may be more reasonable than trying to manipulate just the styles through theming. This could be done either by having optional props or by provider pattern.

## Any open questions to address

- What theming solution to use?
- What to do with AntD dependency and theming (see also https://github.com/jaegertracing/jaeger-ui/issues/508)?
- What level or granularity of theming to provide?

everett980:
@aocenas I considered commenting on [PR#507](https://github.com/jaegertracing/jaeger-ui/pull/522) but I think it makes sense to consolidate this discussion here.

> Also Grafana uses CSS in JS solution heavily (EmotionJS)

What led you/others at Grafana to land on EmotionJS?
I haven't yet fully gone down the rabbit hole of css-in-js, but it does seem like Emotion is running away from the competition since ~September ([npmtrends](https://www.npmtrends.com/emotion-vs-styled-components-vs-jss-vs-@emotion/core-vs-@emotion/styled-vs-styled-jsx-vs-glamor-vs-reactcss))

dprokop:
@everett980 please take a look at our overview here: https://github.com/grafana/grafana/issues/15438

[...]

PR #520, opened 2020-02-12
Allow UI elements to be injected and remove Antd dependency

aocenas:
Part of both https://github.com/jaegertracing/jaeger-ui/issues/509 and https://github.com/jaegertracing/jaeger-ui/issues/508

This tries to remove dependency on Antd. Issue with Antd is first the size of the overall dependency, global styles and problematic theming that would have theme both the trace UI and the Antd elements separately.

This provides a react context that needs to be used for injecting required UI components. This would  allow injecting custom UI element to make theming better. For example in Grafana we use custom component library to share styling and behaviour.

Caveats:
- It is hard to provide any defaults as that would mean a default UI library needs to be included. Antd could be defined as a peer dependency, that though would mean a warning if you do not install it which is probably not ideal. That UI components are required will make this a bit harder to use. Long term defaults can be implemented directly in the jeager ui though.
- This is designed so that the context provider is public and users implementing this as a library would use it directly ala Redux provider. Seems better than having long list of props but that is subjective.

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/index.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/index.tsx
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 import * as React from 'react';
-import { Input } from 'antd';
+import { Input, Popover } from 'antd';
 import { Location, History as RouterHistory } from 'history';
 import _clamp from 'lodash/clamp';
 import _get from 'lodash/get';
@@ -55,6 +55,7 @@ import { TraceArchive } from '../../types/archive';
 import { EmbeddedState } from '../../types/embedded';
 import filterSpans from '../../utils/filter-spans';
 import updateUiFind from '../../utils/update-ui-find';
+import UIElementsContext from './uiElementsContext';
 
 import './index.css';
 
@@ -395,15 +396,17 @@ export class TracePageImpl extends React.PureComponent<TProps, TState> {
             </section>
           ) : (
             <section style={{ paddingTop: headerHeight }}>
-              <TraceTimelineViewer
-                registerAccessors={this._scrollManager.setAccessors}
-                scrollToFirstVisibleSpan={this._scrollManager.scrollToFirstVisibleSpan}
-                findMatchesIDs={spanFindMatches}
-                trace={data}
-                updateNextViewRangeTime={this.updateNextViewRangeTime}
-                updateViewRangeTime={this.updateViewRangeTime}
-                viewRange={viewRange}
-              />
+              <UIElementsContext.Provider value={{ Popover }}>
+                <TraceTimelineViewer
+                  registerAccessors={this._scrollManager.setAccessors}
+                  scrollToFirstVisibleSpan={this._scrollManager.scrollToFirstVisibleSpan}
+                  findMatchesIDs={spanFindMatches}
+                  trace={data}
+                  updateNextViewRangeTime={this.updateNextViewRangeTime}
+                  updateViewRangeTime={this.updateViewRangeTime}
+                  viewRange={viewRange}
+                />
+              </UIElementsContext.Provider>
             </section>
           ))}
       </div>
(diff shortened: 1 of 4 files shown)


aocenas:
This is just a draft to get feedback on this approach.

packages/jaeger-ui/src/components/TracePage/uiElementsContext.tsx
```diff
@@ -0,0 +1,82 @@
...
+  | 'rightBottom';
+type PopoverInterface = React.ComponentClass<
+  {
+    content?: React.ReactNode;
+    arrowPointAtCenter?: boolean;
+    overlayClassName?: string;
+    placement?: TooltipPlacement;
+    children?: React.ReactNode;
+  },
+  {}
```
everett980: this should be `unknown`

packages/jaeger-ui/src/components/TracePage/index.tsx
```diff
@@ -395,15 +396,17 @@ export class TracePageImpl extends React.PureComponent<TProps, TState> {
...
-              <TraceTimelineViewer
-                registerAccessors={this._scrollManager.setAccessors}
-                scrollToFirstVisibleSpan={this._scrollManager.scrollToFirstVisibleSpan}
-                findMatchesIDs={spanFindMatches}
-                trace={data}
-                updateNextViewRangeTime={this.updateNextViewRangeTime}
-                updateViewRangeTime={this.updateViewRangeTime}
-                viewRange={viewRange}
-              />
+              <UIElementsContext.Provider value={{ Popover }}>
```
everett980: The `value` here should be a `const` at the top of the file. Something like `const UI_ELEMENTS = { Popover };`. Without that, ever render of `TracePage/index` will cause all consumers to re-render [(source)](https://reactjs.org/docs/context.html#caveats).

packages/jaeger-ui/src/components/TracePage/uiElementsContext.tsx
```diff
@@ -0,0 +1,82 @@
...
+  children: (elements: Elements) => React.ReactNode;
+};
+
+/**
+ * Convenience render prop style component to handle error state when elements are not defined.
+ */
+export function GetElementsContext(props: GetElementsContextProps) {
+  return (
+    <UIElementsContext.Consumer>
+      {(value: Elements | undefined) => {
```
everett980: Similar to my feedback on [another PR](https://github.com/jaegertracing/jaeger-ui/pull/518#pullrequestreview-364438678), is it possible to remove the `| undefined` and rely on TypeScript to check that the value is provided?

everett980: Although I suppose that depends on a way to check that `UiElementsContext.Consumer` is always used within a `UiElementsContext.Provider` 🤔

packages/jaeger-ui/src/components/TracePage/uiElementsContext.tsx
```diff
@@ -0,0 +1,82 @@
...
+type GetElementsContextProps = {
+  children: (elements: Elements) => React.ReactNode;
+};
+
+/**
+ * Convenience render prop style component to handle error state when elements are not defined.
+ */
+export function GetElementsContext(props: GetElementsContextProps) {
+  return (
+    <UIElementsContext.Consumer>
```
everett980: `UI` should be `Ui`. For the sake of camel/pascal case, acronyms should be treated as a word. [`UiFind` as an example](https://github.com/jaegertracing/jaeger-ui/blob/master/packages/jaeger-ui/src/components/common/UiFindInput.tsx)

everett980:
Also, how will this impact our ability to add functionality to JaegerUI?

For instance, the `UiPopover` interface only has five props defined. If we want to use Antd's `Popover`'s `onVisibleChange`, do we just update the type, mention it in the change log for the yet-to-be created `jaeger-components` package, and leave it to consumers to make sure their provided components are compatible with the types in `uiElementsContext` when they update?

PR #522, opened 2020-02-17
POC of theming with emotion

aocenas:

- part of https://github.com/jaegertracing/jaeger-ui/issues/509

This adds emotion and uses react context to propagate theme overrides to components.


.eslintrc.js
```diff
@@ -36,9 +36,11 @@ module.exports = {
       },
       plugins: ['@typescript-eslint'],
       rules: {
+        'no-useless-constructor': 'off',
         'no-unused-vars': 0,
         '@typescript-eslint/interface-name-prefix': ['error', 'always'],
         '@typescript-eslint/no-unused-vars': 1,
+        '@typescript-eslint/no-useless-constructor': 'error',
```
aocenas: After installing the emotion some packages were updated and an error in `no-useless-constructor` surfaced see: https://github.com/typescript-eslint/typescript-eslint/issues/420

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/index.css
```diff
@@ -14,10 +14,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-.TraceTimelineViewer {
-  border-bottom: 1px solid #bbb;
-}
-
 .json-markup {
```
aocenas: Other styles seem to be global overrides for external packages. These can be moved to global emotion styles too.

packages/jaeger-ui/src/components/TracePage/index.tsx
```diff
@@ -376,37 +377,45 @@ export class TracePageImpl extends React.PureComponent<TProps, TState> {
     };
 
     return (
-      <div>
-        {archiveEnabled && (
-          <ArchiveNotifier acknowledge={this.acknowledgeArchive} archivedState={archiveTraceState} />
-        )}
-        <div className="Tracepage--headerSection" ref={this.setHeaderHeight}>
-          <TracePageHeader {...headerProps} />
+      <ThemeProvider
```
aocenas: Just the provider is added here.

packages/jaeger-ui/src/site-prefix.tsx
```diff
@@ -20,7 +20,7 @@ if (!baseNode && process.env.NODE_ENV !== 'test') {
   throw new Error('<base> element not found');
 }
 
-const sitePrefix = baseNode ? baseNode.href : `${global.location.origin}/`;
+const sitePrefix = baseNode ? baseNode.href : `${((global as unknown) as TestGlobal).location.origin}/`;
```
aocenas: After the packages update the global custom declaration clashed with NodeJs types.

packages/jaeger-ui/typings/custom.d.ts
```diff
@@ -13,11 +13,12 @@
...
 
 // For jest
-declare const global: {
+declare type TestGlobal = {
   location: Location;
 }
+declare const global: TestGlobal;
 
-declare interface Window {
+declare interface Window { // eslint-disable-line @typescript-eslint/interface-name-prefix
```
aocenas: Another issue after the package update

packages/jaeger-ui/src/site-prefix.tsx
everett980: NodeJS type clashes are unfortunately the result of some changes to the yarn.lock that happen during adding packages:
https://github.com/jaegertracing/jaeger-ui/pull/522/files#diff-8ee2343978836a779dc9f8d6b794c3b2R1614
remove:
```
dependencies
  "@types/node" "*"
```
and `@types/node` itself from the `yarn.lock`
then `rm -rf node_modules && yarn install` _should_ fix these type issues.

packages/jaeger-ui/src/components/TracePage/index.tsx
```diff
@@ -376,37 +377,45 @@ export class TracePageImpl extends React.PureComponent<TProps, TState> {
...
+        value={
+          {
```
everett980: As mentioned [here](https://github.com/jaegertracing/jaeger-ui/pull/520#discussion_r387282602), this should be a `const` at the top of the file otherwise `ThemeProvider` consumers will re-render on every render of `TracePage/index`

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/index.tsx
```diff
@@ -26,6 +27,15 @@ import { TNil, ReduxState } from '../../../types';
 import { Span, Trace } from '../../../types/trace';
 
 import './index.css';
+import { createStyle, Theme, withTheme } from '../Theme';
+
+const getStyles = createStyle(theme => {
+  return {
+    TraceTimelineViewer: css`
+      border-bottom: ${theme.borderStyle};
```
everett980: Is the intent to provide all styles through `context`, or only configurable styles?

packages/jaeger-ui/src/components/TracePage/Theme.tsx
```diff
@@ -0,0 +1,78 @@
...
+) => {
+  let WithTheme: React.ComponentType<Omit<Props, 'theme'>> = props => {
+    return (
+      <ThemeConsumer>
+        {(theme: Theme) => (
+          <Component
+            {...({
+              ...props,
+              theme,
+            } as Props & { theme: Theme })}
```
everett980: this cast shouldn't be necessary.

packages/jaeger-ui/src/components/TracePage/Theme.tsx
```diff
@@ -0,0 +1,78 @@
...
+              ...value,
+            }
+          : defaultTheme;
+        return props.children(mergedTheme);
+      }}
+    </ThemeContext.Consumer>
+  );
+}
+
+export const withTheme = <Props extends { theme: Theme }, Statics extends {} = {}>(
```
everett980: is the second type used?

packages/jaeger-ui/src/components/TracePage/Theme.tsx
```diff
@@ -0,0 +1,78 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import React from 'react';
+import hoistNonReactStatics from 'hoist-non-react-statics';
+import memoizeOne from 'memoize-one';
+
+export type ThemeOptions = Partial<Theme>;
+
+export type Theme = {
```
everett980: I think this type would make more sense if the attributes were scoped to the component they were styling. Something like:
```
export type Theme = Record<string, Partial<emotion-style-object-type>>
```
(I assume emotion has a type for valid css objects)

then theme consumers can have different `borderStyle`s or what have you based on what is being styled, e.g.:
```
const theme: TTheme = {
  TraceTimelineViewer: {
    borderBottom: '1px solid #bbb',
  },
}
```

also types in this codebase should be prefixed by a `T` (enums get `E`, and interfaces get `I`)

yurishkuro:
don't think this will go further

PR #527, opened 2020-02-21
Allow to use non-root span as trace name

aliksend:
##  Which problem is this PR solving?
Allow to use name of span with tag trace_name == true as name of trace.
It is very useful when using with traefik to get clear trace names instead of "traefik: EntryPoint example.com"
It is not a significant change. Should I have been to open issue before?

- Change getTraceName to find spans with tag trace_name == true first in /jaeger-ui/packages/jaeger-ui/src/model/trace-viewer.js
- Use trace-viewer's  getTraceName in /packages/jaeger-ui/src/model/transform-trace-data.tsx
- Fix typo in /packages/jaeger-ui/src/utils/tracking/index.tsx

### DIFF
--- a/packages/jaeger-ui/src/model/trace-viewer.js
+++ b/packages/jaeger-ui/src/model/trace-viewer.js
@@ -14,6 +14,13 @@
 
 // eslint-disable-next-line import/prefer-default-export
 export function getTraceName(spans) {
-  const span = spans.filter(sp => !sp.references || !sp.references.length)[0];
+  // find span with "trace_name" tag
+  let span = spans.filter(
+    sp => sp.tags.filter(tg => tg.key === 'trace_name' && tg.value === true).length > 0
+  )[0];
+  if (span == null) {
+    // find span without references
+    span = spans.filter(sp => !sp.references || !sp.references.length)[0];
+  }
   return span ? `${span.process.serviceName}: ${span.operationName}` : '';
 }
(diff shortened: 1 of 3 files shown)


yurishkuro:
Conceptually, I don't like this change. The trace name comes from the data provided by instrumentation. This PR creates a mechanism in the UI that allows instrumentation to _override itself_. Why not fix the instrumentation in the first place?

aliksend:
> Why not fix the instrumentation in the first place?

I agree with you.
As I understand, conceptually correct fix will relate to opentracing project and will be harder to implement. Also it may drag out for a long time to spread to all tools including jaeger-ui. I'm not sure I can accompany this fix

May be you're right and I should close this PR and create Issue in opentracing project and wait for other contributors to implement it in opentracing and then in jaeger tools

yurishkuro:
which opentracing project are you referring to?

aliksend:
[Opentracing project](https://opentracing.io/) is a standard for different tracing tools like jaeger, zipkin etc. It defines basic concepts of tracing and conceptually correct way is to add new feature to base interface: [opentracing specification](https://github.com/opentracing/specification/blob/master/specification.md)

yurishkuro:
OpenTracing project itself is just an API. You specific issue is caused by some concrete instrumentation library.

yurishkuro:
I am going to close it, don't think it's the right direction.

PR #529, opened 2020-02-21
Move up redux from SpanTreeOffset

aocenas:
- Part of: #508

- Push redux parts from SpanTreeOffset up to TracePage

This based of https://github.com/jaegertracing/jaeger-ui/pull/517 but I cannot create a PR with that target as that branch is on my fork so this is draft until that is merged.

### DIFF
--- a/packages/jaeger-ui/src/utils/tracking/index.test.js
+++ b/packages/jaeger-ui/src/utils/tracking/index.test.js
@@ -19,11 +19,6 @@ jest.mock('./conv-raven-to-ga', () => () => ({
   message: 'jaeger/a',
 }));
 
-jest.mock('./index', () => {
-  process.env.REACT_APP_VSN_STATE = '{}';
-  return require.requireActual('./index');
-});
-
 import ReactGA from 'react-ga';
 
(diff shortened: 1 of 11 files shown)


packages/jaeger-ui/src/utils/tracking/index.test.js
```diff
@@ -19,11 +19,6 @@ jest.mock('./conv-raven-to-ga', () => () => ({
   message: 'jaeger/a',
 }));
 
-jest.mock('./index', () => {
```
aocenas: This is unrelated but did not find a way around it. The error this produces is https://github.com/facebook/jest/issues/2567 because of the `process` reference. I assume it could be caused be some dependency update but there is no lock file change here and I tried `rm -rf node_modules` and reinstalls multiple times without luck. Anybody knows why this could be happening?

everett980:
Regardless of the verdict on keeping span row visibility state in redux / some parent component, I think this state can be hidden away in `TraceTimelineViewer`'s `state`. There's no need to preserve which span rows are hovered on mount/unmount as the user's cursor should be controlling this behavior.

PR #530, opened 2020-02-23
Plexus: Maintain context by graphting new vertices w/o moving prior nodes

everett980:
- Adding or removing even a single vertex or edge caused all nodes and edges to be repositioned, making it impossible to follow changes in large graphs.

- Grapht new vertices onto existing vertices.
- Rebuild broken edges.
- Keep existing node and vertex edges when hiding nodes and/or vertices.
- Expand edges phase to handle adding edges when no vertices are added.
- Maintain zoom level and relative zoom location when graph size changes.

I still need to create some tickets for outstanding work, and add images, but otherwise this is ready.

### DIFF
--- a/packages/plexus/src/zoom/condenseIcon.tsx
+++ b/packages/plexus/src/zoom/condenseIcon.tsx
@@ -1,4 +1,4 @@
-// Copyright (c) 2018 Uber Technologies, Inc.
+// Copyright (c) 2020 Uber Technologies, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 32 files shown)


packages/plexus/src/zoom/condenseIcon.tsx
```diff
@@ -0,0 +1,25 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import * as React from 'react';
+
+// SVG is from Material Design Icons:
+// https://github.com/google/material-design-icons/blob/224895a86501195e7a7ff3dde18e39f00b8e3d5a/av/svg/production/ic_shuffle_48px.svg
+// LICENSE: Apache-2.0
```
everett980: @yurishkuro I followed the example in `src/zoom/resetIcon.tsx`, is this the correct way to note the LICENSE followed?

yurishkuro: it's good

PR #531, opened 2020-02-25
Lift antd dependencies from TraceTimelineViewer

aocenas:
Lifts antd dependencies from all descendants TraceTimelineViewer


aocenas:
Sorry this was created by accident

PR #536, opened 2020-03-03
Search results link spans

### LINKED ISSUE #526: Deep link spans from the trace search page
Similar to #255 in the way we only show partial (auto expand selected spans) trace when the page first load. 

We want a way to highlight spans in the trace search page. And when user click on given trace to open the detail page, automatically expand/highlight the specified spans while keep other spans collapsed.

We have analytics jobs to emit trace metrics for different services at trace/span level, for example if certain spans should have specific tag but not. The metrics have example traces. When user click on those examples, it will open the trace search page to show the list of traces. However it's very hard to quickly locate the problematic spans. So we updated the analytic jobs to include exemplar spanIds. However the jaeger-ui does not have a quick way to deep link the exemplar spans from the search page.

This will improve user experience, especially for large traces. It might be very hard to navigate through the a very long list of spans, when we know in advance the list of spanIds we are interested in, it will be very helpful for UI to just expand those spans for user.

- Search page to accept url pattern: 
https://{jaeger-ui-host}/search?span={spanId1},{spanId2}@{traceId1}&span={spanId3}@{traceId2}
and parse & save the exemplar spans.
- It will still use the list of traceIds to fetch the trace from backend.
- When user click on a trace, the trace page will open the details page using below format:
https://{jaeger-ui-host}/trace/{traceId1}?uiFind={spanId1}&uiFind={spanId2}

everett980:
- closes #526 
- Fixes `getExternalLink` for `ReferenceLink` used when a span has external references

- `SearchTracePage/url` parses `span` into `spanLinks` and `traceID` (unchanged: `traceID` is used to fetch traces instead of searching)
- `spanLinks` are given to `SearchResults` to pass a `uiFind` value to the `TracePage` (unchanged: this is how deep linking into `TracePage` already behaves)

### DIFF
--- a/packages/jaeger-ui/src/components/SearchTracePage/url.test.js
+++ b/packages/jaeger-ui/src/components/SearchTracePage/url.test.js
@@ -94,7 +94,7 @@ describe('SearchTracePage/url', () => {
             [trace1]: span2,
           },
         })
-      ).toBe(`/search?span=${span0}%2C${span1}%40${trace0}&span=${span2}%40${trace1}&traceID=${trace2}`);
+      ).toBe(`/search?span=${span0}%20${span1}%40${trace0}&span=${span2}%40${trace1}&traceID=${trace2}`);
     });
   });
 
@@ -116,7 +116,7 @@ describe('SearchTracePage/url', () => {
 
     it('converts multiple spans to traceID and spanLinks', () => {
       expect(
-        getUrlState(`span=${span0}%2C${span1}%40${trace0}&span=${span2}%40${trace1}&traceID=${trace2}`)
+        getUrlState(`span=${span0}%20${span1}%40${trace0}&span=${span2}%40${trace1}&traceID=${trace2}`)
       ).toEqual({
         traceID: expect.arrayContaining([trace0, trace1, trace2]),
         spanLinks: {
--- a/packages/jaeger-ui/src/components/SearchTracePage/url.test.js
+++ b/packages/jaeger-ui/src/components/SearchTracePage/url.test.js
@@ -126,6 +126,17 @@ describe('SearchTracePage/url', () => {
       });
     });
 
+    it('converts span param without spanIDs to just traceID', () => {
+      expect(getUrlState(`span=${span0}%20${span1}%40${trace0}&span=%40${trace1}&traceID=${trace2}`)).toEqual(
+        {
+          traceID: expect.arrayContaining([trace0, trace1, trace2]),
+          spanLinks: {
+            [trace0]: `${span0} ${span1}`,
+          },
+        }
+      );
+    });
+
     it('handles duplicate traceIDs', () => {
       expect(
         getUrlState(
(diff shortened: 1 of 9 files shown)


packages/jaeger-ui/src/components/SearchTracePage/url.test.js
```diff
@@ -0,0 +1,187 @@
...
+
+    it('converts spanLink and other traceID to traceID and span', () => {
+      expect(
+        getUrl({
+          traceID: [trace0, trace2],
+          spanLinks: {
+            [trace0]: span0,
+          },
+        })
+      ).toBe(`/search?span=${span0}%40${trace0}&traceID=${trace2}`);
```
guo0693: Just FYI, in our use case, you won't find a mix of span@trace1&traceId=${trace2}, it's OK for UI can support that.
We should add a test case for:
```
${span1_1}%20${span1_2}%40${trace1}
&${span2_1}%40${trace2}
&%40${trace3}
```

packages/jaeger-ui/src/components/TracePage/url/ReferenceLink.test.js
```diff
@@ -45,7 +45,7 @@ describe(ReferenceLink, () => {
 
     it('render for external trace', () => {
       const component = shallow(<ReferenceLink reference={externalRef} focusSpan={focusMock} />);
-      const link = component.find('a[href="/trace/trace2/uiFind?=span2"]');
+      const link = component.find('a[href="/trace/trace2?uiFind=span2"]');
```
guo0693: +1

packages/jaeger-ui/src/components/TracePage/url/index.tsx
```diff
@@ -12,19 +12,25 @@
...
+
 import prefixUrl from '../../../utils/prefix-url';
 
 import { TNil } from '../../../types';
 
 export const ROUTE_PATH = prefixUrl('/trace/:id');
 
-export function getUrl(id: string) {
-  return prefixUrl(`/trace/${id}`);
+export function getUrl(id: string, uiFind?: string): string {
```
guo0693: Not very familiar with tsx syntax, is "uiFind?" expected with "?"

everett980: the `?` just means it's optional. not all trace urls are deep links.

packages/jaeger-ui/src/components/SearchTracePage/url.test.js
everett980: Ah, I didn't realize spanIDs were optional.
I added a check & a test for that.

packages/jaeger-ui/src/components/SearchTracePage/url.test.js
```diff
@@ -0,0 +1,198 @@
...
+
+    it('preserves traceID without spanLinks', () => {
+      expect(
+        getUrl({
+          traceID: trace0,
+        })
+      ).toBe(`/search?traceID=${trace0}`);
+    });
+
+    it('converts spanLink and traceID to traceID and span', () => {
```
yurishkuro: `to traceID and span` or just `to span`?

PR #539, opened 2020-03-05
Fix span links for leading 0s trace ID

everett980:
- When `?span=${id}@${traceID}` fetches a trace such that:
```
trace.traceID = `0+${traceID}`
```
the deep linking would not be passed on.

- Check span links for `trace.traceID` with leading 0s stripped out if `trace.traceID` is not present in `spanLinks`

### DIFF
--- a/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.js
+++ b/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.js
@@ -208,7 +208,7 @@ export class UnconnectedSearchResults extends React.PureComponent<SearchResultsP
                   linkTo={getLocation(
                     trace.traceID,
                     { fromSearch: searchUrl },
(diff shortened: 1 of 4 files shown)


packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.js
```diff
@@ -208,7 +208,7 @@ export class UnconnectedSearchResults extends React.PureComponent<SearchResultsP
                   linkTo={getLocation(
                     trace.traceID,
                     { fromSearch: searchUrl },
-                    spanLinks && spanLinks[trace.traceID]
+                    spanLinks && (spanLinks[trace.traceID] || spanLinks[trace.traceID.replace(/^0*/, '')])
```
yurishkuro: I suggest adding a unit test for this use case.

PR #541, opened 2020-03-09
Correct trace name resolution

swapster:
Related to [#461](https://github.com/jaegertracing/jaeger-ui/pull/461).

There is a problem with resolution of a trace name for traces without spans with empty refs (when a "root span" is a span with remote parent).

This PR fixes this problem as suggested in a [discussion](https://github.com/jaegertracing/jaeger-ui/pull/461#discussion_r333851003). To detect a "root span" we:

1. Find all spans that either have no parent reference, or a parent reference that is not found among all trace's spans. 
2. Sort found spans by start time.
3. Take a first ("the earliest") span and make a trace name from this span's data.

Additionally, there are some unit tests for a different cases.

### DIFF
--- a/packages/jaeger-ui/src/model/find-trace-name.test.js
+++ b/packages/jaeger-ui/src/model/find-trace-name.test.js
@@ -57,7 +57,7 @@ describe('getTraceName', () => {
       ],
     },
   ];
-  const spansWithMultipleRoots = [
+  const spansWithMultipleRootsDifferentByStartTime = [
     {
       spanID: firstSpanId,
       startTime: t + 200,
@@ -69,16 +69,52 @@ describe('getTraceName', () => {
       ],
     },
     {
-      spanID: secondSpanId, // root span
+      spanID: secondSpanId, // may be a root span
       startTime: t + 100,
+      references: [
+        {
+          spanID: missingSpanId,
+          refType: 'CHILD_OF',
+        },
+      ],
     },
     {
-      spanID: thirdSpanId, // root span
+      spanID: thirdSpanId, // root span (as the earliest)
       startTime: t,
       operationName,
       process: {
         serviceName,
       },
+      references: [
+        {
+          spanID: missingSpanId,
+          refType: 'CHILD_OF',
+        },
+      ],
+    },
+  ];
+  const spansWithMultipleRootsWithOneWithoutRefs = [
+    {
+      spanID: firstSpanId,
+      startTime: t + 200,
+      references: [
+        {
+          spanID: thirdSpanId,
+          refType: 'CHILD_OF',
+        },
+      ],
+    },
+    {
+      spanID: secondSpanId, // root span (as a span without any refs)
+      startTime: t + 100,
+      operationName,
+      process: {
+        serviceName,
+      },
+    },
+    {
+      spanID: thirdSpanId, // may be a root span
+      startTime: t,
       references: [
         {
           spanID: missingSpanId,
@@ -161,7 +197,11 @@ describe('getTraceName', () => {
   });
 
   it('returns an id of root span with the earliest startTime', () => {
-    expect(getTraceName(spansWithMultipleRoots)).toEqual(fullTraceName);
+    expect(getTraceName(spansWithMultipleRootsDifferentByStartTime)).toEqual(fullTraceName);
+  });
+
+  it('returns an id of root span without any refs', () => {
+    expect(getTraceName(spansWithMultipleRootsWithOneWithoutRefs)).toEqual(fullTraceName);
   });
 
   it('returns an id of root span with remote ref', () => {
--- a/packages/jaeger-ui/src/model/find-trace-name.test.js
+++ b/packages/jaeger-ui/src/model/find-trace-name.test.js
@@ -18,7 +18,7 @@ describe('getTraceName', () => {
   const firstSpanId = 'firstSpanId';
   const secondSpanId = 'secondSpanId';
   const thirdSpanId = 'thirdSpanId';
-  const remoteSpanId = 'remoteSpanId';
+  const missingSpanId = 'missingSpanId';
 
   const serviceName = 'serviceName';
   const operationName = 'operationName';
@@ -79,7 +79,7 @@ describe('getTraceName', () => {
       },
       references: [
         {
-          spanID: remoteSpanId,
+          spanID: missingSpanId,
           refType: 'CHILD_OF',
         },
       ],
@@ -115,7 +115,7 @@ describe('getTraceName', () => {
       },
       references: [
         {
-          spanID: remoteSpanId,
+          spanID: missingSpanId,
           refType: 'CHILD_OF',
         },
       ],
--- a/packages/jaeger-ui/src/model/find-trace-name.test.js
+++ b/packages/jaeger-ui/src/model/find-trace-name.test.js
@@ -1,4 +1,4 @@
-// Copyright (c) 2018 Uber Technologies, Inc.
+// Copyright (c) 2020 Uber Technologies, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/packages/jaeger-ui/src/model/find-trace-name.test.js
+++ b/packages/jaeger-ui/src/model/find-trace-name.test.js
@@ -21,7 +21,6 @@ describe('getTraceName', () => {
   const missingSpanId = 'missingSpanId';
 
   const currentTraceId = 'currentTraceId';
-  const anotherTraceId = 'anotherTraceId';
 
   const serviceName = 'serviceName';
   const operationName = 'operationName';
@@ -89,7 +88,7 @@ describe('getTraceName', () => {
       references: [
         {
           spanID: missingSpanId,
-          traceID: anotherTraceId,
+          traceID: currentTraceId,
         },
       ],
     },
@@ -104,7 +103,7 @@ describe('getTraceName', () => {
       references: [
         {
           spanID: missingSpanId,
-          traceID: anotherTraceId,
+          traceID: currentTraceId,
         },
       ],
     },
@@ -139,7 +138,7 @@ describe('getTraceName', () => {
       references: [
         {
           spanID: missingSpanId,
-          traceID: anotherTraceId,
+          traceID: currentTraceId,
         },
       ],
     },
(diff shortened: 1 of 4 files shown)


packages/jaeger-ui/src/model/find-trace-name.test.js
```diff
@@ -0,0 +1,172 @@
...
+  const thirdSpanId = 'thirdSpanId';
+  const remoteSpanId = 'remoteSpanId';
+
+  const serviceName = 'serviceName';
+  const operationName = 'operationName';
+
+  const spansWithNoRoots = [
+    {
+      spanID: firstSpanId,
+      startTime: 1583758690000,
```
yurishkuro: I recommend defining a constant `t` and using deltas like `t+10` to vary the timestamps.

packages/jaeger-ui/src/model/find-trace-name.test.js
```diff
@@ -0,0 +1,172 @@
...
+    {
+      spanID: thirdSpanId, // root span
+      startTime: 1583758670000,
+      operationName,
+      process: {
+        serviceName,
+      },
+      references: [
+        {
+          spanID: remoteSpanId,
```
yurishkuro: nit: this looks more like `missingSpanId` rather than `remote`.

packages/jaeger-ui/tsconfig.lint.json
```diff
@@ -30,7 +30,6 @@
     "src/api/jaeger.js",
     "src/components/SearchTracePage/SearchResults/ResultItem.markers.js",
     "src/model/order-by.js",
-    "src/model/trace-viewer.js",
```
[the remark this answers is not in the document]
swapster: I will be linted since this config file [inherits another config](https://github.com/jaegertracing/jaeger-ui/blob/e968cdb07c61dedaaf3e1dfda888d8ef596b0491/packages/jaeger-ui/tsconfig.lint.json#L23) with [include](https://github.com/jaegertracing/jaeger-ui/blob/e968cdb07c61dedaaf3e1dfda888d8ef596b0491/packages/jaeger-ui/tsconfig.json#L23) option.
Please correct me if I'm wrong.

packages/jaeger-ui/src/model/find-trace-name.test.js
```diff
@@ -0,0 +1,172 @@
...
+          refType: 'CHILD_OF',
+        },
+      ],
+    },
+    {
+      spanID: thirdSpanId,
+      startTime: 1583758670000,
+      references: [
+        {
+          spanID: firstSpanId,
```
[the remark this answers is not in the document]
swapster: Yes, it's exactly a loop.
I just dont' know how else I can fake a case when there is no root span persists in a trace. No references - it's a root. Ref to a missing span - it's a root too. Maybe now this test case has no sense and it's better to remove it at all.

packages/jaeger-ui/src/model/find-trace-name.test.js
```diff
@@ -0,0 +1,172 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
```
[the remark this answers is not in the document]
swapster: [Fixed](https://github.com/jaegertracing/jaeger-ui/pull/541/commits/5c633c2699233e572330c5cc1ca77fcbf2e3113e)

packages/jaeger-ui/src/model/find-trace-name.test.js
swapster: Yes, its' better, [fixed](https://github.com/jaegertracing/jaeger-ui/pull/541/commits/08d8dbb7e98f7cf739d1d291d8e507f42e2d5b71).

packages/jaeger-ui/src/model/find-trace-name.test.js
swapster: Thanks, [fixed](https://github.com/jaegertracing/jaeger-ui/pull/541/commits/579300519e0a8d2f73ab2f09bcf1187439b38da1).

packages/jaeger-ui/src/model/trace-viewer.tsx
```diff
@@ -12,8 +12,22 @@
...
+  const allSpanIDs = spans.map(sp => sp.spanID);
+  const rootSpan = spans
+    .filter(sp => {
+      if (!sp.references || !sp.references.length) {
+        return true;
+      }
+      const parentIDs = sp.references.filter(r => r.refType === 'CHILD_OF').map(r => r.spanID);
+
+      // no parent from trace
+      return !parentIDs.some(pID => allSpanIDs.includes(pID));
```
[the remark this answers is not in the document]
swapster: Just didn't think about it. But yeap, dictionary makes sense here.
[Changed](https://github.com/jaegertracing/jaeger-ui/pull/541/commits/52e3794141f2978fbacd9955404f8f587db822f7) an array to a dict.

packages/jaeger-ui/src/model/trace-viewer.tsx
```diff
@@ -12,8 +12,22 @@
...
+    })
+    .sort((sp1, sp2) => sp1.startTime - sp2.startTime)[0];
```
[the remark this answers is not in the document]
swapster: I missed it.

[Added](https://github.com/jaegertracing/jaeger-ui/pull/541/commits/c914bcfcb199c0c4433e03081777b31d88b92717) a parents check to a `sort()` method (also, left untouched comparing by a `startTime` since, I think, it's better than not to sort multiple "root" spans at all if there are several of them).

Also, [changed](https://github.com/jaegertracing/jaeger-ui/pull/541/commits/a6f67ba80ada6ef9f65dfe9201bd2952e4790a48) a couple of tests.

packages/jaeger-ui/src/model/find-trace-name.test.js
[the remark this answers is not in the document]
yurishkuro: It's the only one that returns empty string, let's keep it.

packages/jaeger-ui/src/model/trace-viewer.tsx
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
yurishkuro:
```suggestion
// Copyright (c) 2020 The Jaeger Authors
```

packages/jaeger-ui/src/model/find-trace-name.test.js
```diff
@@ -0,0 +1,225 @@
+// Copyright (c) 2020 Uber Technologies, Inc.
```
yurishkuro:
```suggestion
// Copyright (c) 2020 The Jaeger Authors
```

packages/jaeger-ui/src/model/find-trace-name.test.js
```diff
@@ -0,0 +1,225 @@
...
+      process: {},
+      references: [
+        {
+          spanID: thirdSpanId,
+          refType: 'CHILD_OF',
+        },
+      ],
+    },
+    {
+      spanID: thirdSpanId, // root span
```
yurishkuro:
```suggestion
      spanID: thirdSpanId, // effective root span, since its parent is missing
```

packages/jaeger-ui/src/model/find-trace-name.test.js
```diff
@@ -0,0 +1,225 @@
...
+  const secondSpanId = 'secondSpanId';
+  const thirdSpanId = 'thirdSpanId';
+  const missingSpanId = 'missingSpanId';
+
+  const serviceName = 'serviceName';
+  const operationName = 'operationName';
+
+  const t = 1583758670000;
+
+  const spansWithNoRoots = [
```
yurishkuro:
```suggestion
  // Note: this trace has a loop S1 <- S2 <- S3 <- S1, which is the only way
  // to make the algorithm return an empty string as trace name.
  const spansWithNoRoots = [
```

packages/jaeger-ui/src/model/trace-viewer.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+// eslint-disable-next-line import/prefer-default-export
+export function getTraceName(spans: Span[]) {
+  const allTraceSpans: spansDict = spans.reduce((dict, span) => ({ ...dict, [span.spanID]: span }), {});
+  const rootSpan = spans
+    .filter(sp => {
+      if (!sp.references || !sp.references.length) {
+        return true;
+      }
+      const parentIDs = sp.references.filter(r => r.refType === 'CHILD_OF').map(r => r.spanID);
```
yurishkuro: What if you dropped filters for CHILD_OF? It's possible to have an async trace using FOLLOWS_FROM. I don't think the algorithm is actually sensitive to the type of reference. If anything,  I would filter on `r.spanContext.traceID` being the same as the current span, because a reference to an external trace is _definitely_ not a parent.

swapster: Hm. Did I understand correctly, `span.references` **can't contain** refs to a children? I mean, if some span listed as a reference, it can be either a parent or a some missing ref, but not a child?
If so, I suggest to change this logic[ by this way](https://github.com/jaegertracing/jaeger-ui/pull/541/commits/9aa5b4bd17cd30b0ce655483aafc8e0c98e63038). There is no need to calculate a dict of all trace spans anymore, since `some()` can check traceID [by itself](https://github.com/jaegertracing/jaeger-ui/pull/541/commits/9aa5b4bd17cd30b0ce655483aafc8e0c98e63038#diff-2968d836188925a4a82346df3642203bR25). If there are no situations when `spans` (which is an argument to this function) contains more data then all `.references`, this approach looks more correct.

P. S. "I didn't find `spanContext` in a [types](https://github.com/jaegertracing/jaeger-ui/blob/master/packages/jaeger-ui/src/types/trace.tsx), think it should be `r.traceID` or `r.span.traceID` :)"

swapster: Sorry, I was wrong.

There may be situations when there are references with same `traceID` **but** with incorrect (missing) `spanID`. It this case such span won't define as a root but it can be a root.

Updated a [logic](https://github.com/jaegertracing/jaeger-ui/pull/541/commits/8d3eb81a6d3b1d9f0dda577dde8d671326f821e5) and a [tests](https://github.com/jaegertracing/jaeger-ui/pull/541/commits/3b2cbae16293e78ef7ee8c04b3bfc80cc48e91ce). Now it should work correct.

PR #544, opened 2020-03-13
Usage of a helper for trace name when preparing trace data

swapster:
In continue of #541.

Now there is a `getTraceName()` method that can deal with a "root spans" with missing references. But it doesn't use when traces data transforms for a "Search" and "Compare" pages.
Seems like it should be using here.

### DIFF
--- a/packages/jaeger-ui/src/model/transform-trace-data.test.js
+++ b/packages/jaeger-ui/src/model/transform-trace-data.test.js
@@ -76,23 +76,7 @@ describe('transformTraceData()', () => {
       ],
       startTime,
       duration,
-      tags: [
-        {
-          key: 'http.url',
-          type: 'string',
-          value: 'https://example.com/getSome',
-        },
-        {
-          key: 'http.method',
-          type: 'string',
-          value: 'get',
-        },
-        {
-          key: 'http.status_code',
-          type: 'float64',
-          value: 200,
-        },
-      ],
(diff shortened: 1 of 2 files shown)


bogdanov1609:
Any updates here?

swapster:
> Can there be a unit test?

Added some [tests](https://github.com/jaegertracing/jaeger-ui/pull/544/commits/50bb1911345fcdaf0758004f1de23d83592b1f75).

packages/jaeger-ui/src/model/transform-trace-data.test.js
```diff
@@ -53,3 +53,191 @@ describe('deduplicateTags()', () => {
...
+      {
+        key: 'http.method',
+        type: 'string',
+        value: 'get',
+      },
+      {
+        key: 'http.status_code',
+        type: 'float64',
+        value: 200,
+      },
```
yurishkuro, line 191: Nit: are these tags relevant to the functionality being tested? They don't look like it. Could you please trim the fixtures to minimally required data?

swapster: Sounds reasonable, thanks :)
[Removed](https://github.com/jaegertracing/jaeger-ui/pull/544/commits/246db5249943e2e0bccfff64bdff967c78dc3e0a) unnecessary tags.

PR #549, opened 2020-03-24
DDG: Add path agnostic decorations action & reducer

everett980:
- Lays the groundwork to add "path agnostic" decorations to DDG nodes. For instance: current alert count, or time since last deployment.

- Adds action and reducer to fetch data for a given service or operation. Calculates maximum for each decoration to later be used for a color gradient on nodes.


yurishkuro:
Are there any visuals coming with this PR?

everett980:
> Are there any visuals coming with this PR?

Not in this one. This is just the functionality to fetch and store decoration values (and the max value for each decoration).
Next I'm working on the side bar to select a decoration, as well as display additional information for a selected node.
Then I'm working on node coloring based on the selected decoration.

Issue #546, opened 2020-03-18
DDG should support client root span


cykl:
## Requirement - what kind of business use case are you trying to solve?

I want to display the DDG of a trace started by a batch job. 

## Problem - what in Jaeger blocks you from solving the requirement?

Root spans of traces started by batch job are often client spans, ie. span.kind = client, but current DDG implementation requires a server root span. If this condition is not fulfilled DDG is not shown and an error message is displayed: _There are no dependencies. No traces were found that contain client and any other service where span.kind is ‘server’._

I believe client root span ares valid use cases and should be supported.

## Minimal test case

Following code produces a trace matching my issue:

```java
import io.jaegertracing.internal.JaegerTracer;
import io.jaegertracing.internal.reporters.RemoteReporter;
import io.jaegertracing.internal.samplers.ConstSampler;
import io.jaegertracing.thrift.internal.senders.UdpSender;
import io.opentracing.Tracer;
import io.opentracing.tag.Tags;

public class ClientRootSpan {

    public static void main( String[ ] args ) throws InterruptedException {
        try (var clientTracer = createTracer( "client" )) {
            var clientSpan = clientTracer.buildSpan( "GET" )
                    .withTag( Tags.SPAN_KIND, "client" )
                    .withTag( Tags.HTTP_METHOD, "GET" )
                    .start( );

            Thread.sleep( 100 );

            try (var serverTracer = createTracer( "server" )) {
                var serverSpan = serverTracer.buildSpan( "GET" )
                        .asChildOf( clientSpan )
                        .withTag( Tags.SPAN_KIND, "server" )
                        .withTag( Tags.HTTP_METHOD, "GET" )
                        .start( );

                Thread.sleep( 50 );
                serverSpan.finish( );
            }

            clientSpan.finish( );
        }
    }

    static Tracer createTracer( String serviceName ) {
        return new JaegerTracer.Builder( serviceName )
                .withSampler( new ConstSampler( true ) )
                .withReporter( new RemoteReporter.Builder( ).withSender( new UdpSender( ) ).build( ) )
                .build( );
    }
}
```

Trace is properly displayed in timeline and graph view:

![image](https://user-images.githubusercontent.com/1764725/76965472-a1a67000-6924-11ea-8a2f-a20b043bad89.png)

![image](https://user-images.githubusercontent.com/1764725/76965491-ac610500-6924-11ea-80d2-3d5bfa0f9f40.png)

However, DDG generation fails:

[...]

PR #557, opened 2020-04-04
DDG: Remove kind.server filter and validate the case of service calling itself

### LINKED ISSUE #554: DDG for search results is still wrong
The HotROD graph is missing DB nodes. The redis and mysql services do show in the search results, as well as in the regular DAG graph, but not in DDG:

image

rubenvp8510:
- Fixes #554 

- I removed the span.kind == server filter, instead validate when the service calls itself multiple times and remove all of those recursive calls and only preserve one of them.

Attached an image.

Screenshot from 2020-04-04 09-35-55

### DIFF
--- a/packages/jaeger-ui/src/model/ddg/transformTracesToPaths.test.js
+++ b/packages/jaeger-ui/src/model/ddg/transformTracesToPaths.test.js
@@ -145,11 +145,6 @@ describe('transform traces to ddg paths', () => {
     );
   });
 
-  it('errors if span has ancestor id not in trace data', () => {
-    const traces = makeTraces(makeTrace([rootSpan, followsFocalSpan], missTraceID));
-    expect(() => transformTracesToPaths(traces, focalSvc)).toThrowError(/Ancestor spanID.*not found/);
-  });
-
   it('skips trace without data', () => {
     const traces = {
       ...makeTraces(shortTrace),
--- a/packages/jaeger-ui/src/model/ddg/transformTracesToPaths.test.js
+++ b/packages/jaeger-ui/src/model/ddg/transformTracesToPaths.test.js
@@ -22,6 +22,14 @@ describe('transform traces to ddg paths', () => {
     })),
     attributes: [{ key: 'exemplar_trace_id', value: trace.data.traceID }],
   });
+
+  const addExemplarTraceIDs = (path, traces) => {
+    traces.forEach(trace => {
+      path.attributes.push({ key: 'exemplar_trace_id', value: trace.data.traceID });
+    });
+    return path;
+  };
+
   const makeSpan = (spanName, parent, kind, operationName, processID) => ({
     hasChildren: true,
     operationName: operationName || `${spanName} operation`,
@@ -186,12 +194,11 @@ describe('transform traces to ddg paths', () => {
       focalSvc
     );
 
-    const path = makeExpectedPath([rootSpan, focalSpan, spanServiceAServer], serverClientTrace);
-    path.attributes.push({ key: 'exemplar_trace_id', value: kindlessTrace.data.traceID });
-
     expect(new Set(result)).toEqual(
       new Set([
-        path,
+        addExemplarTraceIDs(makeExpectedPath([rootSpan, focalSpan, spanServiceAServer], serverClientTrace), [
+          kindlessTrace,
+        ]),
         makeExpectedPath([rootSpan, focalSpan, spanServiceAServer, otherSpanServiceAServer], twoServersTrace),
         makeExpectedPath([rootSpan, focalSpan, spanServiceBServer], clientServerTrace),
       ])
@@ -204,8 +211,9 @@ describe('transform traces to ddg paths', () => {
     const trace1 = makeTrace([rootSpan, focalSpan, otherSpan], 'trace1');
     const trace2 = makeTrace([rootSpan, focalSpan, otherSpan], 'trace2');
     const { dependencies: result } = transformTracesToPaths(makeTraces(trace1, trace2), focalSvc);
-    const path = makeExpectedPath([rootSpan, focalSpan, otherSpan], trace1);
-    path.attributes.push({ key: 'exemplar_trace_id', value: trace2.data.traceID });
-    expect(new Set(result)).toEqual(new Set([path]));
+
+    expect(new Set(result)).toEqual(
+      new Set([addExemplarTraceIDs(makeExpectedPath([rootSpan, focalSpan, otherSpan], trace1), [trace2])])
+    );
   });
 });
--- a/packages/jaeger-ui/src/model/ddg/transformTracesToPaths.test.js
+++ b/packages/jaeger-ui/src/model/ddg/transformTracesToPaths.test.js
@@ -205,6 +205,20 @@ describe('transform traces to ddg paths', () => {
     );
   });
 
+  it('support client span root', () => {
+    const spanServiceAClient = makeSpan('SpanA2', undefined, 'client', 'opA', 'serviceA');
+    const spanServiceAServer = makeSpan('SpanA1', spanServiceAClient, 'server', 'opA', 'serviceA');
+    const clientServerTrace = makeTrace([spanServiceAClient, spanServiceAServer], 'clientServerTraceID');
+    spanServiceAServer.hasChildren = false;
+    const { dependencies: result } = transformTracesToPaths(
+      makeTraces(clientServerTrace),
+      clientServerTrace.data.processes[spanServiceAClient.processID].serviceName
+    );
+    expect(new Set(result)).toEqual(
+      new Set([makeExpectedPath([spanServiceAClient, spanServiceAServer], clientServerTrace)])
+    );
+  });
+
   it('dedupled paths', () => {
     const otherSpan = makeSpan('other-span', focalSpan);
     otherSpan.hasChildren = false;
--- a/packages/jaeger-ui/src/model/ddg/transformTracesToPaths.test.js
+++ b/packages/jaeger-ui/src/model/ddg/transformTracesToPaths.test.js
@@ -22,10 +22,10 @@ describe('transform traces to ddg paths', () => {
     })),
     attributes: [{ key: 'exemplar_trace_id', value: trace.data.traceID }],
   });
-  const makeSpan = (spanName, parent, kind) => ({
+  const makeSpan = (spanName, parent, kind, operationName, processID) => ({
     hasChildren: true,
-    operationName: `${spanName} operation`,
-    processID: `${spanName} processID`,
+    operationName: operationName || `${spanName} operation`,
+    processID: processID || `${spanName} processID`,
     references: parent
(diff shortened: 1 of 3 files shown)


rubenvp8510:
Still need to fix some tests.

packages/jaeger-ui/src/model/ddg/transformTracesToPaths.tsx
```diff
@@ -35,22 +38,33 @@ function transformTracesToPaths(
...
-            .filter(span => span.tags.find(({ key, value }) => key === 'span.kind' && value === 'server'))
-            .map(({ processID, operationName: operation }) => ({
-              service: data.processes[processID].serviceName,
-              operation,
-            }));
-
+            if (reducedSpans.length > 0) {
+              const headSpan = reducedSpans[reducedSpans.length - 1];
+              // Transition inside the same service ServiceA -> ServiceA
+              if (headSpan.processID === span.processID) {
```
everett980, line 51: @rubenvp8510 sorry I couldn't get to this sooner

If I'm reading this correctly, it captures every change in `processID`. Then, if there isn't a change in `processID` and the current span is of kind `server`:
1) a parent `client` span that was a different process than _its_ parent span gets replaced with the `server` span
2) a parent `server` span is kept and the current span is added.

I'm wondering if we need to do the replacement in case 1, or if it would be better to simply add any span of kind `server` or any span with a different `processID` than the previous reduced span.
i.e.:
```
if (headSpan.processID !== span.processID || isKindServer(span) {
  reducedSpans.push(span);
}
```
The only case this is different is if `process a` calls `process b` which emits a `client` span as its first span, which is the parent of a `server` span in the same process. IMO this case is strange, possibly caused by bad instrumentation, and I'd prefer to surface the scenario to potentially fix it rather than bury it.

@yurishkuro @vprithvi I'd like to hear your opinions on the two approaches to handling this case.

everett980: @rubenvp8510 Update: I have discussed the two approaches with Yuri and Prithvi and we agree to not replace the ingress client spans.

Here's an example, which, for brevity, I will represent a single span as `processID,operationName,[s]erver/[c]lient`
e.g.:
```
{
  processID: 0,
  operationName: 'a',
  span.kind: 'client',
}
```
would be `0,a,c`

original:
`0,a,c 1,a,s 1,b,c 1,c,s, 2,a,c, 2,b,s 2,c,c`
would be reduced to
`0,a,c 1,a,s 1,c,s 2,a,c 2,b,s` by my proposed method
versus
`0,a,c 1,a,s 1,c,s 2,b,s` in the PR

The differences is that the client ingress span (`2,a,c`) is not dropped.

rubenvp8510: Hi @everett980 I think it is OK not loses the ingress client span. it just wasn't clear to me what would be the correct way of derivative the paths on those cases, but now it makes more sense. I'll do the pertinent changes.

rubenvp8510:
This also fixes https://github.com/jaegertracing/jaeger-ui/issues/546 I added a test for that case.

packages/jaeger-ui/src/model/ddg/transformTracesToPaths.tsx
rubenvp8510: I did the changes, this is ready for another review.

Thank you!

packages/jaeger-ui/src/model/ddg/transformTracesToPaths.test.js
```diff
@@ -22,10 +22,18 @@ describe('transform traces to ddg paths', () => {
     })),
     attributes: [{ key: 'exemplar_trace_id', value: trace.data.traceID }],
   });
-  const makeSpan = (spanName, parent, kind) => ({
+
+  const addExemplarTraceIDs = (path, traces) => {
+    traces.forEach(trace => {
+      path.attributes.push({ key: 'exemplar_trace_id', value: trace.data.traceID });
```
yurishkuro, line 28: @everett980 do we support span IDs for DDG exemplars?

packages/jaeger-ui/src/model/ddg/transformTracesToPaths.test.js
```diff
@@ -22,10 +22,18 @@ describe('transform traces to ddg paths', () => {
     })),
     attributes: [{ key: 'exemplar_trace_id', value: trace.data.traceID }],
   });
-  const makeSpan = (spanName, parent, kind) => ({
+
+  const addExemplarTraceIDs = (path, traces) => {
```
yurishkuro, line 26: Why do we use `const`+lambda instead of `function`? I find function a lot more readable & explicit than a lambda.

packages/jaeger-ui/src/model/ddg/transformTracesToPaths.tsx
```diff
@@ -35,22 +38,27 @@ function transformTracesToPaths(
...
+              isKindServer(span)
+            ) {
+              reducedSpans.push(span);
+            }
+            return reducedSpans;
+          }, []);
+          const path: TDdgPayloadEntry[] = spans.map(({ processID, operationName: operation }) => ({
+            service: data.processes[processID].serviceName,
+            operation,
+          }));
```
yurishkuro, line 61: I find this algorithm quite hard to follow, not to mention that it seems to do a lot of redundant work by constantly recalculating the paths from each leaf upwards over and over. Have we tried implementing it as a classic BFS on a tree?

rubenvp8510: We can do that, but one think I would like to avoid is using recursion, it will lead to bad performance (see https://github.com/jaegertracing/jaeger-ui/pull/326).

Can we do that in a separate PR?

yurishkuro: ok

packages/jaeger-ui/src/model/ddg/transformTracesToPaths.test.js
everett980: @yurishkuro not yet. I asked @vprithvi about it but it wasn't a priority at the time.
Once my inflight work for the trace quality v2 page is landed, it'll be easier to add to the DDG as that page has many deep links and deep link utils.
However url length restrictions are already hit without deep linking, and deep linking would double the length so even then it won't be a super trivial change.
Let's create a ticket for it and put a pin it for now.

packages/jaeger-ui/src/model/ddg/transformTracesToPaths.test.js
rubenvp8510: I just followed the other parts of this code, I think there is no special reason for using lambda.

packages/jaeger-ui/src/model/ddg/transformTracesToPaths.tsx
rubenvp8510: I created another version of this algorithm using a depth breadth search tree algorithm, I'll update this PR soon.

rubenvp8510: Updated the algorithm, I haven't reused the tree , I can do it in another PR if it's required.

packages/jaeger-ui/src/model/ddg/transformTracesToPaths.tsx
```diff
@@ -30,53 +32,55 @@ function transformTracesToPaths(
...
-                service === focalService && (!focalOperation || operation === focalOperation)
-            )
+      data.spans.forEach(span => spanMap.set(span.spanID, span));
+      const tree = getTraceSpanIdsAsTree(data);
+      tree.paths((pathIds: string[]) => {
+        const paths = pathIds.reduce((reducedSpans: Span[], id: string): Span[] => {
+          if (id === '__root__') {
+            return reducedSpans;
+          }
+          const span = spanMap.get(id);
```
yurishkuro, line 42: Q: why does `tree.paths()` generate paths as `ids[]` instead of `spans[]`? The tree descent does operate on spans, and the paths generator does not know what the paths will be used for, so returning IDs instead of spans seems unnecessary, only requiring you to re-map IDs back to spans.

rubenvp8510: This is a good point, and actually the reason I didn't do in that way is that I'm using the same tree used in transform-trace-data, which in each node only have the `spanID`.

We can of course change that to store the whole span on the node, but that will require other changes.

yurishkuro: fair enough

packages/jaeger-ui/src/model/ddg/transformTracesToPaths.tsx
```diff
@@ -30,53 +32,55 @@ function transformTracesToPaths(
...
-          if (
-            path.some(
-              ({ service, operation }) =>
-                service === focalService && (!focalOperation || operation === focalOperation)
-            )
+      data.spans.forEach(span => spanMap.set(span.spanID, span));
+      const tree = getTraceSpanIdsAsTree(data);
+      tree.paths((pathIds: string[]) => {
+        const paths = pathIds.reduce((reducedSpans: Span[], id: string): Span[] => {
+          if (id === '__root__') {
```
yurishkuro, line 39: please use `TREE_ROOT_ID` constant from `packages/jaeger-ui/src/selectors/trace.js`

packages/jaeger-ui/src/model/ddg/transformTracesToPaths.tsx
```diff
@@ -30,53 +32,55 @@ function transformTracesToPaths(
...
+            return reducedSpans;
+          }
+          const span = spanMap.get(id);
+          if (!span) throw new Error(`Ancestor spanID ${id} not found in trace ${traceID}`);
+          if (reducedSpans.length === 0) {
+            reducedSpans.push(span);
+          } else if (
+            reducedSpans[reducedSpans.length - 1].processID !== span.processID ||
+            isKindServer(span)
```
yurishkuro, line 48: the whole algorithm expressed in these two lines, awesome!

 cc @vprithvi

packages/jaeger-ui/src/model/ddg/transformTracesToPaths.test.js
```diff
@@ -52,7 +60,7 @@ describe('transform traces to ddg paths', () => {
         (result, span) => ({
           ...result,
           [span.processID]: {
-            serviceName: `${span.spanID.split(' ')[0]} service`,
+            serviceName: span.processID,
```
everett980, line 63: This doesn't fully test that the `serviceName`s are populated. change this to
```
serviceName: ${span.processID}-name,
```

rubenvp8510: Done. 

Thanks for the review!

PR #574, opened 2020-05-12
Improve get trace name performance

### LINKED ISSUE #562: Support loading traces with large number of spans.
We are trying to port our tracing infrastructure to Jaeger, but some of our workflows end up with traces that have upwards 40K spans. Trying to load these spans in the Jaeger UI doesn't work.

The fact that once, you open a trace all of the spans are loaded and are expanded. In the case of traces with a very large number of spans, Jaeger UI fails to load the traces.

Instead of loading all of the spans within a trace at once, load them only when they are clicked on, i.e. don't load or show any of the children, when one clicks on the span, then only fetch and show the children of the clicked span.
This can be done selectively for large spans only as well since that would lead to no behavior change in case of smaller traces, whereas great improvement in the case of massive ones.
(We have a similar solution in case of our current in-house tracing solution.)

cc: @annanay25

everett980:
- Closes #562 
- Reduces execution time of `getTraceName` for a trace with 11.5k spans from 20s to 250ms

- Replace `reduce` of `spans` into `allTraceSpans` with a set of `allIDs`
- Replace multiple loops with a single `for` loop
- Memoization remains beneficial as it improves the render time of the gantt chart and other trace views by 3x (from 750ms to 250ms for test trace).


yurishkuro:
👏 👏 👏

PR #586, opened 2020-05-25
Avoid the horizontal scrollbar (KeyValueTable)

smanolloff:
scroll-before

The "copy" icon should not cause a horizontal scroll to appear on the expanded KeyValue table.
There is also no need to treat the values as linguistic composites of non-breakable words - they can just as likely be just strings of symbols, for which break-word would actually degrade, rather than improve visual perception.
After the change:
scroll-after


smanolloff:
I want to point out that I cannot explain why this fix works, it was a trial-and-error.
This fix also is meant to work with #584 so the layout is even more "tight".

yurishkuro:
I think it's a reasonable change, don't see much value in breaking on words only, it could always cause scrolling issues.

PR #587, opened 2020-05-25
Don't rely on json-markup for non-json strings

### LINKED ISSUE #571: Remove quotes around values
Currently, all values are surrounded by double quotes. This is redundant and, more importantly, confusing, when the message itself contains such quotes:

Screenshot 2020-05-11 at 15 33 10

smanolloff:

Resolves #571 

The presence of those quotes around strings is not only unnecessary (the value type is inferred from the text [color](https://github.com/jaegertracing/jaeger-ui/blob/master/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/index.css#L30-L41)), but also confusing (when the value itself contains quotes)

Added a separate function in case the value is a non-json string.

Before:
Screenshot 2020-05-25 at 22 00 15

After:
Screenshot 2020-05-25 at 21 56 38

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.test.js
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.test.js
@@ -46,9 +46,10 @@ describe('<KeyValuesTable>', () => {
   let wrapper;
 
   const data = [
-    { key: 'span.kind', value: 'client' },
-    { key: 'omg', value: '{not-a-json' },
-    { key: 'numeric', value: 123456789 },
+    { key: 'span.kind', value: 'client', expected: 'client' },
+    { key: 'omg', value: 'mos-def', expected: 'mos-def' },
+    { key: 'numericString', value: '12345678901234567890', expected: '12345678901234567890' },
+    { key: 'numeric', value: 123456789, expected: '123456789' },
     { key: 'jsonkey', value: JSON.stringify({ hello: 'world' }) },
   ];
 
@@ -138,7 +139,7 @@ describe('<KeyValuesTable>', () => {
     expect(el.length).toBe(data.length);
     el.forEach((valueDiv, i) => {
       if (data[i].key !== 'jsonkey') {
-        expect(valueDiv.html()).toMatch(data[i].value.toString());
+        expect(valueDiv.html()).toMatch(data[i].expected);
       }
     });
   });
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.test.js
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.test.js
@@ -138,7 +138,7 @@ describe('<KeyValuesTable>', () => {
     expect(el.length).toBe(data.length);
     el.forEach((valueDiv, i) => {
       if (data[i].key !== 'jsonkey') {
-        expect(valueDiv.html()).toMatch(`"${data[i].value}"`);
+        expect(valueDiv.text()).toBe(data[i].value);
       }
     });
   });
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.test.js
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.test.js
@@ -47,8 +47,8 @@ describe('<KeyValuesTable>', () => {
 
   const data = [
     { key: 'span.kind', value: 'client' },
-    { key: 'omg', value: 'mos-def' },
-    { key: 'numericString', value: '12345678901234567890' },
+    { key: 'omg', value: '{not-a-json' },
+    { key: 'numeric', value: 123456789 },
     { key: 'jsonkey', value: JSON.stringify({ hello: 'world' }) },
   ];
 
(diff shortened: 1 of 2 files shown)


yurishkuro:
Thanks for the PR. Please make sure the CI is green and all commits are signed.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.tsx
```diff
@@ -25,16 +25,43 @@ import './KeyValuesTable.css';
...
+    <div dangerouslySetInnerHTML={markup} />
+  );
+}
+
+function formatValue(value: any) {
+  let content;
+
   // if the value is a string representing actual json object or array, then use json-markup
-  if (typeof value === 'string' && jsonObjectOrArrayStartRegex.test(value)) {
+  if (typeof value === 'string') {
```
yurishkuro, line 55: Why does this logic need to change if the goal is to only remove the quotes? Attempting to parse _any_ value as JSON led to bugs previously, where, for example, a string that looks like a number would get converted to number and change representation / precision.

smanolloff: but it has not changed - see a few lines below.
Basically, the boolean `&&` has been split into two separate conditionals, due to the need to explicitly handle non-json strings.

yurishkuro: ok, thanks. Could you please add unit tests that cover the new conditions? There are only two lines that are not covered: https://codecov.io/gh/jaegertracing/jaeger-ui/compare/3335512110c60ace5b089e1b61d187a6d6ec4000...6c76e930f59843f41817effbf3d57c3f3a999367/diff

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.test.js
```diff
@@ -47,8 +47,8 @@ describe('<KeyValuesTable>', () => {
 
   const data = [
     { key: 'span.kind', value: 'client' },
-    { key: 'omg', value: 'mos-def' },
-    { key: 'numericString', value: '12345678901234567890' },
```
yurishkuro, line 51: please keep this, it's a real use case that cased problems in the past.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.test.js
```diff
@@ -138,7 +138,7 @@ describe('<KeyValuesTable>', () => {
     expect(el.length).toBe(data.length);
     el.forEach((valueDiv, i) => {
       if (data[i].key !== 'jsonkey') {
-        expect(valueDiv.html()).toMatch(`"${data[i].value}"`);
+        expect(valueDiv.html()).toMatch(data[i].value.toString());
```
yurishkuro, line 141: Could we change the data table to contain the exact expectation instead of relying on toString()?

smanolloff: I did try to make this test more meaningful by _comparing against the actual text_ instead of _matching against part of the DOM_, but it did not work:
```
// ----------- this works
valueDiv.html()
// =>  "<div class="ub-inline-block"><div class="json-markup"><span class="json-markup-string">some text</span></div></div>"

valueDiv.text()
// => "some text"

// ----------- this not
valueDiv.html()
// =>  "<div class="ub-inline-block"><div><div class="json-markup"><span class="json-markup-number">some text</span></div></div></div>"

valueDiv.text()
// => ""

```
I know it is somehow related to the shallow representation of the DOM inside this test, but I am really not familiar with react and its test frameworks, so I am at a dead end.
If you can help me get `"some text"` in the second example, I would implement the test change. Apart from that, testing with a regular expression against the DOM is just as bad as the call to `toString()`, if you ask me, so the latter itself doesn't really degrade the overall quality of the test.

yurishkuro: sorry, maybe I didn't explain clearly. The test is data driven from the `data` table. I suggest adding the expected values to that table, e.g. `{ key: 'numeric', value: 123456789, expected: "123456789" }` and comparing with those, instead of `value.toString()` (which is not as obvious in terms of what it would produce).

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.test.js
```diff
@@ -46,9 +46,10 @@ describe('<KeyValuesTable>', () => {
...
 
   const data = [
-    { key: 'span.kind', value: 'client' },
-    { key: 'omg', value: 'mos-def' },
-    { key: 'numericString', value: '12345678901234567890' },
+    { key: 'span.kind', value: 'client', expected: 'client' },
+    { key: 'omg', value: 'mos-def', expected: 'mos-def' },
+    { key: 'numericString', value: '12345678901234567890', expected: '12345678901234567890' },
+    { key: 'numeric', value: 123456789, expected: '123456789' },
     { key: 'jsonkey', value: JSON.stringify({ hello: 'world' }) },
```
yurishkuro, line 53: sorry, how does this one pass if it doesn't have `expected` field?

smanolloff: https://github.com/jaegertracing/jaeger-ui/blob/52e64dd2531354e78f1059c9cce2c89574c663cf/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.test.js#L141

PR #599, opened 2020-07-17
span bar row size fix

### LINKED ISSUE #589: Individual spans in trace view grow in size on hover in Chromium
When hovering over individual spans in the trace view, spans seem to grow in horizontal height when the gray bar on the side is hovered or when a trace is clicked. I could only reproduce this behavior in Google Chrome, not in Firefox. I have attached videos of the issue below:

Does not work in Google Chrome: https://streamable.com/6zgtif
Works in Firefox: https://streamable.com/7nwlal

## System info

OS: Ubuntu 20.04
Google Chrome version: 83.0.4103.61
Firefox version: 76.0.1
Jaeger version: 1.18.0

vankop:
resolves #589 

bug 👇 (black inner bar is custom critical path info, it does not affect a bug)
ezgif com-optimize (1)

Steps to reproduce:
- use chrome 84
- hover on SpanTreeOffset element, move cursor

What happens:
Not sure why but pseudo element `:before` increases parent's height on `:hover`, so [here](https://github.com/ozonru/jaeger-ui/blob/master/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/ListView/index.tsx#L357) after rendering jaeger thinks that element height changed and fall into [recomputing y positions](https://github.com/jaegertracing/jaeger-ui/blob/master/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/ListView/index.tsx#L372), this happens on each span hover so span height is growing infinitely 

I fixed size of `indentGuide`, this actually improves perf a little bit (no [layout](https://www.html5rocks.com/en/tutorials/internals/howbrowserswork/#Layout) happens)

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanTreeOffset.css
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanTreeOffset.css
@@ -28,7 +28,7 @@ limitations under the License.
   /* The size of the indentGuide is based off of the iconWrapper */
   padding-right: calc(0.5rem + 12px);
   height: 100%;
-  border-left: 1px solid transparent;
+  border-left: 3px solid transparent;
   display: inline-flex;
 }
 
@@ -39,15 +39,11 @@ limitations under the License.
 }
 
 .SpanTreeOffset--indentGuide.is-active {
-  /* The size of the indentGuide is based off of the iconWrapper */
-  padding-right: calc(0.5rem + 11px);
-  border-left: 0px;
+  border-left: 3px solid darkgrey;
 }
 
 .SpanTreeOffset--indentGuide.is-active:before {
-  content: '';
-  padding-left: 3px;
-  background-color: darkgrey;
+  background-color: transparent;
 }
 
 .SpanTreeOffset--iconWrapper {


everett980:
@vankop Can you please attach a gif of hovering over the indent guidelines with your changes?

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanTreeOffset.css
```diff
@@ -39,15 +39,11 @@ limitations under the License.
...
-  padding-right: calc(0.5rem + 11px);
-  border-left: 0px;
+  border-left: 3px solid darkgrey;
 }
 
 .SpanTreeOffset--indentGuide.is-active:before {
-  content: '';
-  padding-left: 3px;
-  background-color: darkgrey;
+  background-color: transparent;
```
everett980, line 46: I believe a psuedoelement without `content` is not generated at all.

Please remove this `:before` entirely if it is not necessary.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanTreeOffset.css
```diff
@@ -39,15 +39,11 @@ limitations under the License.
 }
 
 .SpanTreeOffset--indentGuide.is-active {
-  /* The size of the indentGuide is based off of the iconWrapper */
-  padding-right: calc(0.5rem + 11px);
-  border-left: 0px;
+  border-left: 3px solid darkgrey;
```
everett980, line 42: I think you can get away with just `border-color: darkgrey` to avoid reproducing the other two characteristics (`3px` and `solid`)

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanTreeOffset.css
vankop: padding is a part of element size https://github.com/jaegertracing/jaeger-ui/blob/09d763f6b1d262118c15244c49c16d326e6b518d/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanTreeOffset.css#L37 this is needed to show tiny lines (lines when `:not(:hover)`)

PR #598, opened 2020-07-17
Update lodash from 4.17.15 to 4.17.19

objectiser:
Bump lodash dependency to 4.17.19, which should have been done in https://github.com/jaegertracing/jaeger-ui/pull/597

Signed-off-by: Gary Brown <gary@brownuk.com>

PR #603, opened 2020-07-21
Fixed missing 'types/node' dependency

### LINKED ISSUE #601: Update to lodash introduced dependency that is causing "yarn list" to fail

Able to run `yarn list`

Since #598 , the `yarn list` command fails with:

```
error Couldn't find package "@types/node@*" required by "@types/object-hash@^1.3.0" on the "npm" registry.
```
Unclear why this dependency entry was introduced, as not directly related to lodash?

objectiser:
Fixes #601 

Not sure how this didn't get generated in `yarn.lock` as part of #598?

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1539,6 +1539,11 @@
   dependencies:
     moment "*"
 
+"@types/node@*":
+  version "14.0.24"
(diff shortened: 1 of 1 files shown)


yurishkuro:
how did you update the lock file this time?

objectiser:
@yurishkuro Using the same commands (`nvm use`, `yarn install`) - only difference may have been the `.yarnrc` was pointing to an internal npm registry (mirror).

Issue #365, opened 2017-05-05
Show running version in UI


hwinkel:
Is there a way to show the actual version (or git hash before release)  (UI, Server) in the UI ?

jan25:
Is this still a valid issue? If no one is working on this, i could pick this up

tiffon:
@jan25 This is still a valid issue, and it would be awesome if you could pick it up!

For now, the version of the UI is available to JS, but it is not displayed anywhere. Maybe as a footer on the sear h page would work? Open to other approaches, too!

The version of jaeger-query would need to be injected, by jaeger-query, before it is available to JS. Similar to the base path. So that's a bit of work that needs to be done in jaeger core.

jan25:
@tiffon Footer would be a good idea. What do you think about putting it under 'About Jaeger' dropdown?
Or i was also thinking like a tiny badge near 'Jaeger UI' title in nav bar, it would show UI version and when you hover over badge the tiny tooltip may show the jaeger-query version.

I found a way to inject jaeger-query version from StaticFileHandler on server side. We are currently using regex replace to inject stuff into index.html from there. Would you know if there are better ways to inject data from Go into html?

alanisaac:
I recently ran into this issue, struggling to figure out what version of the UI I had running without digging into my infrastructure / deployment.  I took a quick stab at implementing the suggestion to put the UI version in the "About Jaeger" dropdown, and opened #606.

I'm more than happy to play around with other ideas, and / or continue on with trying to enhance it with the commit hash or jaeger-query version.

yurishkuro:
Resolved by #606.

PR #606, opened 2020-07-28
Added jaeger ui version to about menu

alanisaac:
Signed-off-by: Alan Pinkert <alan.pinkert@gmail.com>

- Part of #365 

Adds the version of Jaeger UI to the about menu as suggested in this comment https://github.com/jaegertracing/jaeger-ui/issues/365#issuecomment-491203207.  There are many suggestions in the issue thread about where to place the version, this is simply a stab at one of them.

image


packages/jaeger-ui/src/constants/default-config.tsx
```diff
@@ -16,6 +16,8 @@ import deepFreeze from 'deep-freeze';
 
 import { FALLBACK_DAG_MAX_NUM_SERVICES } from './index';
 
+const { version } = require('../../package.json');
```
yurishkuro, line 19: It might be misleading/confusing to display the UI project version rather than the backend release version (which implies a certain UI version).

jan25: +1. We may need to supply release version to the CONFIG here https://github.com/jaegertracing/jaeger-ui/blob/2a552379c92606c530304cbec33ff4f89d832cad/packages/jaeger-ui/public/index.html#L18-L27

This is replaced in server code with a JSON object.

yurishkuro: alternative solution that does not touch the config: https://github.com/jaegertracing/jaeger/pull/2547

yurishkuro: Would you like to give it another try? If you sync to jaeger-ui/master, the index.html now gets the following top level function:
```javascript
// Jaeger version data is embedded by the query-service via search/replace.
function getJaegerVersion() {
  const DEFAULT_VERSION = {"gitCommit":"", "gitVersion":"", "buildDate":""};
  const JAEGER_VERSION = {"gitCommit":"0f88343172c0bb16ad5c1805d18e20d59176ea7e","gitVersion":"v1.20.0","buildDate":"2020-10-09T17:42:48Z"};
  return JAEGER_VERSION;
}
```

Considering that there are several data elements here, I would recommend adding them at the bottom of the About menu, e.g.
  * Jaeger v1.20.0
  * Commit 2a55237
  * Build 2020-10-09T17:42:48Z
  * Jaeger UI v1.9.0

alanisaac: Made the updates as requested, though this does make an assumption about the correct short hash for git being 7 characters.  While this is the default, my understanding is that's not always the case if there are unambiguous revisions.  

I think the more technically correct way to get the short hash might be to use `git rev-parse --short <SHA>` at the source and inject it instead, though this would involve more changes on the Jaeger query side.

image

On a side note, in testing with Jaeger, I found that the [default recommended way to run](https://github.com/jaegertracing/jaeger/blob/master/CONTRIBUTING.md#running-local-build-with-the-ui) of `make run-all-in-one` does not include the `$(BUILD_INFO)` necessary to inject the version information, so the about menu appears blank when running locally.  Is this desired or not?

yurishkuro: Looks awesome. One other idea I had is to make versions actually clickable, to that vN.N.N links would take you to the respective release notes page and the commit hash to that commit (the latter is less important than the versions). But this can be done in a separate PR, I would like to merge this first.

Regarding `make run-all-in-one`, sounds like something we could improve in the make target, although when running from a local copy getting the versions is a bit fuzzy.

yurishkuro (approved):
Thanks a lot!

Issue #374, opened 2019-04-28
Remove UNSAFE lifecyle methods from React components


tiffon:
### Requirement - what kind of business use case are you trying to solve?

Remain compatible with React upgrades.

### Problem - what in Jaeger blocks you from solving the requirement?

Some code needs to be changed. See the `UNSAFE_*` methods in [The Component Lifecycle](https://reactjs.org/docs/react-component.html#the-component-lifecycle).

### Proposal - what do you suggest to solve the problem or improve the existing situation?

Do not use the unsafe methods.

tklever:
By my check, branch master no longer has any component methods of componentWillMount, componentWillUpdate, or componentWillReceiveProps or with an UNSAFE_ prefix.

I think this issue is safe to close.

yurishkuro:
Thanks @tklever , that's great!

PR #610, opened 2020-08-06
update legacy react lifecycle methods to indicate unsafe status

tklever:
Following the react migration path laid out here https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html

- Start of work for Issue #374 

- Adding UNSAFE to legacy lifecycle methods for further review and update

### DIFF
--- a/packages/jaeger-ui/src/components/App/Page.tsx
+++ b/packages/jaeger-ui/src/components/App/Page.tsx
@@ -42,7 +42,8 @@ export class PageImpl extends React.Component<TProps> {
     trackPageView(pathname, search);
   }
 
-  componentWillReceiveProps(nextProps: TProps) {
+  // eslint-disable-next-line camelcase
+  UNSAFE_componentWillReceiveProps(nextProps: TProps) {
     const { pathname, search } = this.props;
     const { pathname: nextPathname, search: nextSearch } = nextProps;
     if (pathname !== nextPathname || search !== nextSearch) {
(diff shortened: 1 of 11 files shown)


yurishkuro:
Excuse my lack of React knowledge, but what would this accomplish? Wouldn't this result in the methods not being called when needed?

tklever:
> Excuse my lack of React knowledge, but what would this accomplish? Wouldn't this result in the methods not being called when needed?

Actually, somewhat the opposite. In the future, the UNSAFE_* versions of these methods are the only ones that will be called.

According to to the "Gradual Migration Plan" at https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html, in react 16.3, these method names are simply aliases (although, IMO the UNSAFE_* is a nice visual indicator that there is some debt hanging around if nothing else).

In later versions of 16, the legacy methods (like componentDidMount), will throw a deprecated warning in development mode, whereas the UNSAFE_* ones will not. UNSAFE_* in this case acknowledging, "yes we're doing it the old way, and it's with intent"

In React 17+, the legacy methods (like componentDidMount) will not be called at all. Only the UNSAFE_* version will be executed in the component lifecycle.

If you're looking for a short list of benefits as I see them, this change allows the project to upgrade React through 16.X and into 17.X without any major code changes, while still giving a pretty clear indicator that there is potential refactoring to be done. This would allow component refactoring to happen at a schedule decoupled from any needed React upgrades.

yurishkuro:
👍

yurishkuro:
@rubenvp8510 would you be able to review?

rubenvp8510:
This LGTM as a starting point to address the migration of those methods.

yurishkuro:
Thanks!

PR #612, opened 2020-08-11
refactor UNSAFE_componentWillReceiveProps to use componentDidUpdate

tklever:
- This work is part of what needs to be done to close Issue #374 and a child of PR: #610

- Migrating these UNSAFE_componentWillRecieveProps calls to componentDidUpdate in line with recommendations provided at https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops, stating "If you need to perform a side effect (for example, data fetching or an animation) in response to a change in props, use componentDidUpdate"


rubenvp8510:
This looks good.

PR #619, opened 2020-08-18
Archive notifier tests

tklever:
- Adding test coverage to an existing component

- Adding unit tests for component ArchiveNotifier
- A few minor cleanup items within the component itself

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/ArchiveNotifier/index.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/ArchiveNotifier/index.tsx
@@ -26,15 +26,8 @@ enum ENotifiedState {
   Outcome = 'ENotifiedState.Outcome',
 }
 
-// const NOTIFIED_PROGRESS = 'NOTIFIED_PROGRESS';
-// const NOTIFIED_OUTCOME = 'NOTIFIED_OUTCOME';
-
-// type NotifiedState = 'NOTIFIED_PROGRESS' | 'NOTIFIED_OUTCOME' | null;
-
 type Props = {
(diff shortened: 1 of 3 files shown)


packages/jaeger-ui/src/components/TracePage/ArchiveNotifier/index.tsx
```diff
@@ -73,7 +66,7 @@ function updateNotification(oldState: ENotifiedState | null, nextState: ENotifie
   const { acknowledge, archivedState } = props;
   if (nextState === ENotifiedState.Outcome) {
     if (archivedState && archivedState.error) {
-      const error = typeof archivedState.error === 'string' ? archivedState.error : archivedState.error;
+      const { error } = archivedState;
```
yurishkuro, line 69: what changed that makes it unnecessary to check for string type?

tklever: No change, if you look at the true and false side of the ternary, the true and false condition are the same.

`? archivedState.error : archivedState.error`

`archivedState.error` could be anything or nothing and this if statement will always return `archievedState.error`.

If I had to guess, this code might predate the partial adoption of typescript, typescript's transpiler will break if you put an inappropriate value in the error property. There might have been a check here that got refactored, but at current, it's an unnecessary code branch.

tklever: Just checked, this line was added in #194, it's always been this way. Probably just something that wasn't easy to spot without tests, hard to speculate.

PR #622, opened 2020-08-23
Adding tests around existing ArchiveNotifier duck

tklever:
- Adding tests around existing ArchiveNotifier/duck code

- Adding tests around ArchiveNotifier/duck
- Updated a redux middleware to return next() (keeps redux promise chains intact)


yurishkuro:
Thanks!

Issue #628, opened 2020-08-26
Span bar broken on click


Stono:
Hey, 
I've just excitedly deployed 1.19 (which has ui version 1.10) due to it having a fix for the span hover issue (https://github.com/jaegertracing/jaeger-ui/pull/599).

Whilst that has been fixed, there's another bug now that when we click spans sometimes they load empty:

<img width="1674" alt="Screenshot 2020-08-26 at 17 35 35" src="https://user-images.githubusercontent.com/1486729/91331491-ba179200-e7c2-11ea-8fe6-2b2758537c64.png">

On top of that sometimes multiple spans expand when you click one of them:

<img width="1680" alt="Screenshot 2020-08-26 at 17 37 31" src="https://user-images.githubusercontent.com/1486729/91331598-d9162400-e7c2-11ea-8543-4c043fae2cff.png">

So unfortunately all in all its even worse than 1.18! 

I'm on chrome `84.0.4147.135` however it's the same on safari too.

yurishkuro:
cc @rubenvp8510

robinator8:
Reproduced on Ubuntu 18.04.4 LTS:
* chrome version 84.0.4147.135
* Firefox version 79.0

Azim-Palmer:
Reproduced on Windows

- chrome version 84.0.4147.135

objectiser:
@everett980 @tiffon Any chance you could take a look? @rubenvp8510 is on PTO until Monday.

yurishkuro:
@tklever I suspect it's the React refactoring that caused this issue. I reverted the UI prior to your first change and it seems to work.

tklever:
Oh no! Let me know what you find, maybe I can get some tests around it to ensure it doesn't regress further

yurishkuro:
Sadly we don't have anyone available at the moment to troubleshoot this (people are on vacations). The reported issue is easily reproduced - running all-in-one (which does self-tracing), and then expanding any span shows the blank space. I suspect one of those callback methods isn't being called at the right time, since the renaming wasn't semantically fully equivalent.

tklever:
It's #612 that causes it. I'll keep digging to get more specific / find a fix.

tklever:
Specifically it's the changes in #612 to  `packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx`, reverting just that file gets rid of the emptiness.

Looks like @rubenvp8510 already has a fix.

[...]

PR #629, opened 2020-08-27
Fixes TraceTimelineViewer span details render issue

rubenvp8510:
- This fixes Issue #628 , a  regression when  refactored `VirtualizedTraceView` to remove deprecated method `componentWillReceiveProps` and use the `componentDidUpdate` instead. 

- The component was using `componentWillReceiveProps` to derive data from properties and store in the instance attributes to avoid recompute it in each render call, With the migration, that process was changed to `componentDidUpdate`, but this is not correct, `RowState` computation on that method doesn't trigger a new render process, so we ended up seeing something not in sync with the `RowState`.  This change uses memoization for `rowState` to avoid recompute it each time.

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
@@ -17,6 +17,7 @@ import cx from 'classnames';
 import { connect } from 'react-redux';
 import { bindActionCreators, Dispatch } from 'redux';
 import { withRouter, RouteComponentProps } from 'react-router-dom';
+import _isEqual from 'lodash/isEqual';
 
 // import { History as RouterHistory, Location } from 'history';
 
@@ -148,8 +149,8 @@ function getCssClasses(currentViewRange: [number, number]) {
 }
 
 const memoizedGenerateRowStates = memoizeOne(generateRowStatesFromTrace);
-const memoizedViewBoundsFunc = memoizeOne(createViewedBoundsFunc);
-const memoizedGetCssClasses = memoizeOne(getCssClasses);
+const memoizedViewBoundsFunc = memoizeOne(createViewedBoundsFunc, _isEqual);
+const memoizedGetCssClasses = memoizeOne(getCssClasses, _isEqual);
 
 // export from tests
 export class VirtualizedTraceViewImpl extends React.Component<VirtualizedTraceViewProps> {
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
@@ -20,6 +20,7 @@ import { withRouter, RouteComponentProps } from 'react-router-dom';
 
 // import { History as RouterHistory, Location } from 'history';
 
+import memoizeOne from 'memoize-one';
 import { actions } from './duck';
 import ListView from './ListView';
 import SpanBarRow from './SpanBarRow';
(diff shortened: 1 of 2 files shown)


objectiser:
Tested with hotrod example, seems to fix the problem.

yurishkuro:
@tklever please review

tklever:
I pulled this code locally and I can no longer replicate the "empty box" problem. This looks good (to my limited expertise).

I cooked up a couple tests last night to lock in @rubenvp8510's work in this PR, specifically to ensure that the memoization he added is preserved and can't be accidentally removed. When this PR (or some version of it) lands, I'll submit some unit tests so we (or I in this case :-( ) can't reintroduce this regression in the future.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
```diff
@@ -223,6 +202,28 @@ export class VirtualizedTraceViewImpl extends React.Component<VirtualizedTraceVi
...
+  getViewedBounds(): ViewedBoundsFunctionType {
+    const { currentViewRangeTime, trace } = this.props;
+    const [zoomStart, zoomEnd] = currentViewRangeTime;
+
+    return memoizedViewBoundsFunc({
+      min: trace.startTime,
+      max: trace.endTime,
+      viewStart: zoomStart,
+      viewEnd: zoomEnd,
+    });
```
josejulio, line 224: This is probably not going to work as expected, the `memoize` does a shallow comparison on the arguments, and IIUC, this is always calling it with a new object (because is always being defined as a new object here).


You probably need to define a [custom equality function ](https://github.com/alexreardon/memoize-one#custom-equality-function) or memoize a function that accept the `currentViewRangeTime` and `trace` params (if these are more stable)

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
```diff
@@ -223,6 +202,28 @@ export class VirtualizedTraceViewImpl extends React.Component<VirtualizedTraceVi
...
   }
 
+  getRowStates(): RowState[] {
+    const { childrenHiddenIDs, detailStates, trace } = this.props;
+    return memoizedGenerateRowStates(trace, childrenHiddenIDs, detailStates);
+  }
+
+  getClippingCssClasses(): string {
+    const { currentViewRangeTime } = this.props;
+    return memoizedGetCssClasses(currentViewRangeTime);
```
josejulio, line 212: Is `currentViewRangeTime` stable? Or the object is created in every render?

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
```diff
@@ -223,6 +202,28 @@ export class VirtualizedTraceViewImpl extends React.Component<VirtualizedTraceVi
     }
   }
 
+  getRowStates(): RowState[] {
+    const { childrenHiddenIDs, detailStates, trace } = this.props;
+    return memoizedGenerateRowStates(trace, childrenHiddenIDs, detailStates);
```
josejulio, line 207: Are these props stable?

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
rubenvp8510: You're right, good catch, Thanks

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
rubenvp8510: Those props are from the redux state, so I consider it stable.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
rubenvp8510: Same as above.

josejulio (approved):

rubenvp8510:
@yurishkuro Could you approve this change please? Thanks!

yurishkuro (approved):

PR #647, opened 2020-10-06
Improve duration formatting

### LINKED ISSUE #319: Use time units in the Web UI which are easy to read by humans

Always use time units in the Web UI which are easy to read by humans.

We use Jaeger to trace some complex, long-running offline jobs in which a seconds granularity in the Web UI is too fine-grained for human consumption.

Currently, durations seem to be shown in either milliseconds (“ms”) or seconds (“s”). For traces of long-running requests/jobs a duration like “26904.23s” is not very readable.

Instead of showing a duration like “26904.23s”, a duration like “7.47 h” or even “7 h 28 m 24.23 s” would be better for human consumption.

jamesfer:
Resolves #319

Improves the duration formatting in the UI so that larger durations are easier to read. I've listed some examples below comparing the old and new formatting:

Old | New | Comment
----- | ----- | ----
0ms | 0μs | Smallest displayed unit used to be ms, now it is μs
1.5s | 1.5s | Durations less than 1 minute are displayed in decimals
1.5m | 1m 30s | Units are split instead of using decimal points
183840s | 2d 3h 4m | No more then 3 units are displayed as going all the way to ms on a day long duration seemed overkill
173040 | 2d 4m | Units that are 0 are hidden

Each of these behaviours are open for discussion, I just tried to pick something that looked sensible.

In addition, this PR also removes the `inputUnit` parameter of the `formatDuration` function. I did this because the parameter would significantly complicate the implementation of the function and it isn't used anywhere I could find in the code base. The duration is always assumed to be given in microseconds which is consistent with the behaviour of other functions in the file. If a duration of a different unit is ever used, then it is probably best to create a function that will convert it to microseconds before passing it to the `formatDuration` function.

### DIFF
--- a/packages/jaeger-ui/src/utils/date.test.js
+++ b/packages/jaeger-ui/src/utils/date.test.js
@@ -1,0 +1,51 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { formatDuration, ONE_MILLISECOND, ONE_SECOND, ONE_MINUTE, ONE_DAY } from './date.tsx';
+
+describe('formatDuration', () => {
+  it('keeps microseconds the same', () => {
+    expect(formatDuration(1)).toBe('1μs');
+  });
+
+  it('formats values in different durations', () => {
+    const input = 13 * ONE_SECOND + 256 * ONE_MILLISECOND + 777;
+    expect(formatDuration(input)).toBe('13s 256ms 777μs');
+  });
+
+  it('rounds the number of microseconds', () => {
+    const input = 256 * ONE_MILLISECOND + 0.7;
+    expect(formatDuration(input)).toBe('256ms 1μs');
+  });
+
+  it('displays a maximum of 3 units and rounds the last one', () => {
+    const input = 10 * ONE_MINUTE + 13 * ONE_SECOND + 256 * ONE_MILLISECOND + 777;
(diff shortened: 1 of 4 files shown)


yurishkuro:
Thanks for the PR. 

I am not sold on multi-unit form being the best representation. E.g. it's much clearer to show 1.5s than 1s 500 ms, but that's the easy one (and also sub-second values are normal decimals that require no tricks, unlike h:m:s that use different multiples). I think two nearby units could be ok, like 4d 2h, but not 4d 2m (irrelevant precision). More than two units seems like overkill always.

jamesfer:
They sound like good points. I'll change the number of units displayed to 2 so `2d 3h 52m` would turn into `2d 4h` (due to rounding) and `4d 0h 3m 20s` would be displayed as just `4d`.

I agree with you on using decimal points for seconds, milliseconds and microseconds as they are decimal based, but I think splitting the units is clearer on minutes, hours and days due to the odd numbers of units. Would the following displays be better?
* 1.5μs
* 1.5ms
* 1.5s
* 1m 30s
* 1h 30m
* 1d 12h

yurishkuro:
all these examples look good to me.

jamesfer:
@yurishkuro Just updated the PR to match those examples

packages/jaeger-ui/src/utils/date.test.js
```diff
@@ -0,0 +1,71 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
yurishkuro, line 1:
```suggestion
// Copyright (c) 2020 The Jaeger Authors
```

jamesfer: Done 👍

yurishkuro:
trying to fix code coverage run #649

packages/jaeger-ui/src/utils/date.test.js
```diff
@@ -0,0 +1,71 @@
...
+
+  // it('formats values in different durations', () => {
+  //   const input = 13 * ONE_SECOND + 256 * ONE_MILLISECOND + 777;
+  //   expect(formatDuration(input)).toBe('13s 257ms');
+  // });
+  //
+  // it('rounds the number of microseconds', () => {
+  //   const input = 256 * ONE_MILLISECOND + 0.7;
+  //   expect(formatDuration(input)).toBe('256ms 1μs');
+  // });
```
yurishkuro, line 30: these look like they should be deleted?

yurishkuro (approved):
Thanks!

PR #650, opened 2020-10-09
Add placeholder for Jaeger version

yurishkuro:
Part of #365, and depends on https://github.com/jaegertracing/jaeger/pull/2547.


rubenvp8510:
This looks good, I'm assuming that we still need to do some work on the query-service side.

rubenvp8510:
Nevermind, I saw the PR on the query service side.

PR #655, opened 2020-10-13
Explain "self time" in graph view

### LINKED ISSUE #653: Add explanation about "Self Time" in "Trace Graph" view
I don't understand what `Self Time` means.

It seems jaeger not documented about `Self Time`.

Add explanation in `Help` tip.
trace_graph

yurishkuro:
Fixes #653


yurishkuro:
Travis was successful , but didn't update the PR, force merging

https://travis-ci.org/github/jaegertracing/jaeger-ui/builds/735219248

PR #656, opened 2020-10-29
Prevent DAG crashes because of empty service name string

### LINKED ISSUE #634: Error: Can not create element with invalid string ID ``
**Describe the bug**
Attempting to access the DAG view on the Jaeger UI results in a fully blank white page and the error:
```
Error: Can not create element with invalid string ID ``
...
```
We have a few Kubernetes environments with Jaeger (managed with Jaeger Operator, production strategy, external Elasticsearch for storage) collecting traces from multiple Spring Boot microservices (shipping Zipkin spans using Spring Cloud Sleuth) and Kong API Gateway / Ingress Controller (also shipping Zipkin spans, using kong-plugin-zipkin.) In each environment we see this error when accessing the DAG view on the Jaeger UI, all other views are fine.

In the Force Directed Graph I can see a node that has no name or connected edges, but still causes other nodes to be highlighted when it is selected. (Shown below.) When tested in a fresh environment this node was not present at first, at which time the DAG was able to load successfully. This graph node has eventually appeared in every environment. Given this behaviour and the error message, it appears that the two are linked.

**To Reproduce**
1. Deploy Jaeger with Operator with production strategy and external Elasticsearch storage
2. Deploy microservices and ingress controller sending traces to Jaeger collector
3. Successfully view DAG
4. Wait some time for graph node with no name to appear in Force Directed Graph
5. Attempt to view DAG

**Expected behavior**
Directed acyclic graph to load successfully.

**Screenshots**

The suspect node with no name
Screen Shot 2020-07-30 at 4 26 24 PM (1)

Error message in browser console
Screen Shot 2020-07-30 at 6 33 58 PM

Jaeger:
 - OS: EKS-Optimized Amazon Linux 2 (Kubernetes worker node)
 - Jaeger version: 1.18.0
 - Deployment: Kubernetes Operator, production strategy, external Elasticsearch
UI:
 - OS: macOS Mojave 10.14.6
 - Browser: Google Chrome 84.0.4147.89

**What troubleshooting steps did you try?**
Try to follow https://www.jaegertracing.io/docs/latest/troubleshooting/ and describe how far you were able to progress and/or which steps did not work.

**Additional context**
There is already jaegertracing/jaeger-ui#596 open about this error, which I have commented on, but it seems to me to possibly not entirely be a UI issue.
Assistance on both the error and the source of this unnamed node are appreciated.

rubenvp8510:

- Fixes #634 

- With this change the DAG won't show a node with an empty service, this prevents DAG view crashes because Cytoscape doesn't allow an empty string as an ID.

### DIFF
--- a/packages/jaeger-ui/src/components/DependencyGraph/DAG.js
+++ b/packages/jaeger-ui/src/components/DependencyGraph/DAG.js
@@ -35,6 +35,11 @@ export default class DAG extends React.Component {
     serviceCalls: [],
   };
 
+  constructor(props) {
+    super(props);
+    this.cytoscapeRef = React.createRef();
+  }
+
   componentDidMount() {
     const { serviceCalls } = this.props;
     const nodeMap = {};
@@ -55,8 +60,9 @@ export default class DAG extends React.Component {
         });
       }
     });
+
     cytoscape({
-      container: document.getElementById('cy'),
+      container: this.cytoscapeRef.current,
       boxSelectionEnabled: false,
       autounselectify: true,
       layout: {
@@ -104,6 +110,7 @@ export default class DAG extends React.Component {
           left: 0,
           top: 0,
         }}
+        ref={this.cytoscapeRef}
       />
     );
   }
--- a/packages/jaeger-ui/src/components/DependencyGraph/DAG.js
+++ b/packages/jaeger-ui/src/components/DependencyGraph/DAG.js
@@ -41,17 +41,19 @@ export default class DAG extends React.Component {
     const nodes = [];
     const edges = [];
     serviceCalls.forEach(d => {
-      if (!nodeMap[d.parent]) {
-        nodes.push({ data: { id: d.parent } });
-        nodeMap[d.parent] = true;
+      if (d.parent.trim().length !== 0 && d.child.trim().length !== 0) {
+        if (!nodeMap[d.parent]) {
+          nodes.push({ data: { id: d.parent } });
+          nodeMap[d.parent] = true;
+        }
+        if (!nodeMap[d.child]) {
+          nodes.push({ data: { id: d.child } });
+          nodeMap[d.child] = true;
+        }
+        edges.push({
+          data: { source: d.parent, target: d.child, label: `${d.callCount}` },
+        });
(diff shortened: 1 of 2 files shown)


packages/jaeger-ui/src/components/DependencyGraph/DAG.js
```diff
@@ -41,17 +41,19 @@ export default class DAG extends React.Component {
     const nodes = [];
     const edges = [];
     serviceCalls.forEach(d => {
-      if (!nodeMap[d.parent]) {
-        nodes.push({ data: { id: d.parent } });
-        nodeMap[d.parent] = true;
+      if (d.parent.trim().length !== 0 && d.child.trim().length !== 0) {
```
yurishkuro, line 44: What are the semantics of `d.parent` and `d.child`? Could we not ensure elsewhere that they are not empty?

Could you add a unit test to demonstrate that data leads to an issue you're trying to fix?

rubenvp8510: The way I was able to reproduce the issue is instrumenting a service with an space, (or empty string)

The context of this is `d.parent` and `d.child` are service names, I don't think there are other plkaces where we ensure that the service name is not empty, indeed I can see the empty service name in the search bar.

Is this considered an instrumentation error? Should we indicate it in some way?


I'll add a couple of tests.

yurishkuro: imo empty or whitespace service name should be considered an instrumentation error, and we could build a sanitizer to fix it during ingestion.

However, w.r.t. this code that looks things up in a map, an empty or whitespace string is still a string, so why is the code failing on that?

rubenvp8510: I agree that we should build a sanitizer that fix this during ingestion, but also this view should not fail even if there are empty strings.

The error is not on this part, but on cytoscape, which doesn't allow empty strings nor white spaces as node IDs.

yurishkuro (approved):

yurishkuro:
Thanks!

Issue #657, opened 2020-11-05
Empty screen when click on a trace loaded from JSON


rubenvp8510:
## Requirement - what kind of business use case are you trying to solve?
Loading a JSON file and see the trace details of some selected trace.

## Problem - what in Jaeger blocks you from solving the requirement?

There is an issue when traces are loaded from JSON files, the result page shows all load traces, but when click on a trace appears a blank page. (and error in browser console).

Steps to reproduce:
1) On search page, select JSON File
2) Load json file traces
3) Click on a trace

Attached screenshot with the error.
![Screenshot from 2020-11-05 11-55-16](https://user-images.githubusercontent.com/3799716/98278155-d0c62c00-1f5d-11eb-96e4-b08fdce06e0f.png)

GoingMyWay:
See this answer: https://stackoverflow.com/a/43922390/5046896

PR #658, opened 2020-11-05
Pass a function that doesn't return anything to FileUpload component

rubenvp8510:

- Fixes https://github.com/jaegertracing/jaeger-ui/issues/657

- In this PR: https://github.com/jaegertracing/jaeger-ui/pull/622/files a Redux middleware was updated to return the value of `next(..)` in order to preserve the Promises chain, which I think is correct.

- The problem is that ` Upload.Dragger ` and our component `FileUploader` expect a function that returns nothing `(...)=>void`  and the `Upload` component which is used by `Upload.Dragger` expect a function that returns nothing or a custom request, not a Promise, returning a Promise breaks the underling `Upload` component.  `Upload.Dragger`  (and our component `FileUploader`) has type definition for their parameters and Typescript is suppose to do some type checks here , but the search page was not migrated to TS.

Wondering if may be we should do the migration and enforce the types more to prevent future errors like this one. May be in a subsequent PR.

### DIFF
--- a/packages/jaeger-ui/src/components/SearchTracePage/index.js
+++ b/packages/jaeger-ui/src/components/SearchTracePage/index.js
@@ -104,7 +104,11 @@ export class SearchTracePageImpl extends Component {
                   {!loadingServices && services ? <SearchForm services={services} /> : <LoadingIndicator />}
                 </TabPane>
                 <TabPane tab="JSON File" key="fileLoader">
-                  <FileLoader loadJsonTraces={loadJsonTraces} />
+                  <FileLoader
+                    loadJsonTraces={(fileList: FileList) => {
+                      loadJsonTraces(fileList);
+                    }}
+                  />
                 </TabPane>
               </Tabs>
             </div>


yurishkuro:
Thanks for the fix!

> Wondering if may be we should do the migration and enforce the types more to prevent future errors like this one. May be in a subsequent PR.

Finishing the migration would be great, but difficult to staff. Could there be a unit test to catch this issue?

yurishkuro (approved):

yurishkuro:
there is a linter error in CI
```
[prettier-lint] $ prettier --list-different '{.,scripts}/*.{js,json,md,ts,tsx}' 'packages/*/{src,demo/src}/**/!(layout.worker.bundled|react-vis).{css,js,json,md,ts,tsx}' 'packages/*/*.{css,js,json,md,ts,tsx}'
[eslint       ] $ eslint --cache 'scripts/*.{js,ts,tsx}' 'packages/*/src/**/*.{js,ts,tsx}' 'packages/*/*.{js,ts,tsx}'
[prettier-lint] packages/jaeger-ui/src/components/SearchTracePage/index.js
[prettier-lint] error Command failed with exit code 1.
[prettier-lint] info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
ERROR: "prettier-lint" exited with 1.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
The command "yarn lint" exited with 1.
```

rubenvp8510:
> Thanks for the fix!
> 
> > Wondering if may be we should do the migration and enforce the types more to prevent future errors like this one. May be in a subsequent PR.
> 
> Finishing the migration would be great, but difficult to staff. Could there be a unit test to catch this issue?

Agree, it is difficult at this point, I can do a unit test for this case, my concern is if there is any other places where this could be a potential problem. Well, for now I think we are fine with a unit test here.

yurishkuro:
> Well, for now I think we are fine with a unit test here.

yeah, but it didn't catch the issue with return value.

PR #664, opened 2020-11-29
Switch CI to GitHub Actions

yurishkuro:


yurishkuro:
Don't understand why the workflow run is not showing under Checks. Successful run here: https://github.com/yurishkuro/jaeger-ui/runs/1468905502?check_suite_focus=true.

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,27 @@
...
+  unit-tests:
+    strategy:
+      fail-fast: true
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2-beta
+        with:
+          node-version: '10'
+      - name: install yarn
```
Ashmita152, line 15: Do we need to separately install yarn. I used to think action/setup-node will do that too.

yurishkuro: it's how travis build was setup, to ensure the latest version of yarn

PR #672, opened 2020-12-25
Display references unless it's a single CHILD_OF

### LINKED ISSUE #358: Include references in Span Details

When tracing through Kafka, follows-from reference type is often used. 

Jaeger UI currently does not distinguish child-of and follows-from references in any way.

At minimum, add a section to Span Details listing the references and their types. It will also help dealing with multi-parent spans or references to other traces.

yurishkuro:
Resolves #358


yurishkuro:
Normal span with CHILD_OF
image

yurishkuro:
Same span with reference changed to FOLLOWS_FROM
image

yurishkuro:
same span with two references
image

PR #676, opened 2021-01-12
Rearrange nav controls

### LINKED ISSUE #563: Search in the top menu is easy to confuse with a button
image

The Search comes right after the textbox for trace ID and looks like a button, whereas it's actually the currently selected menu item, so clicking on it does nothing. Instead, to search by trace ID one needs to hit Enter, which is not obvious. Some users are confused by that.

A better placement for this text box would be on the right, as shown in this duplicate ticket #670

image

yoave23:
- Resolves #563

- move the search by id input to the right and differentiate the jaeger-ui link display from the navigation tabs

image


packages/jaeger-ui/src/components/App/TopNav.tsx
```diff
@@ -29,6 +29,8 @@ import { ConfigMenuItem, ConfigMenuGroup } from '../../types/config';
 import { getConfigValue } from '../../utils/config/get-config';
 import prefixUrl from '../../utils/prefix-url';
 
+import JaegerHomeLink from '../../img/jaeger-home-link.svg';
```
yurishkuro, line 32: what is this for & why didn't we need it before?

packages/jaeger-ui/src/components/App/TraceIDSearchInput.tsx
```diff
@@ -37,7 +37,12 @@ class TraceIDSearchInput extends React.PureComponent<Props> {
   render() {
     return (
       <Form layout="horizontal" onSubmit={this.goToTrace} className="TraceIDSearchInput--form">
-        <Input autosize={null} name="idInput" placeholder="Lookup by Trace ID..." />
+        <Input
+          autosize={null}
+          name="idInput"
+          placeholder="Lookup by Trace ID..."
+          prefix={<Icon type="search" />}
```
yurishkuro, line 44: nice touch!

packages/jaeger-ui/src/components/App/TopNav.tsx
yoave23: The upper left link looked like another tab, but the behavior is not the same as in tabs - there is no selected state nor content. We changed it to an image to make it clearer that it’s the ‘home’ link and that it differs from the controls next to it.
image

yoave23: your choice :)
this is how it looks with the logo (personally i don't like it that much):
image
on the other hand, we can go with a simple text label and a different font, just to distinguish it from the other controls

yurishkuro: Let's do a text label / different font.

yoave23:
kept the same font but changed the size and weight
image

yurishkuro (approved):
thanks!

PR #677, opened 2021-01-13
Change UI configuration to a Javascript file

th3M1ke:
- resolves #123 

- add ability to use JS file as a configuration file
- backend implementation PR: https://github.com/jaegertracing/jaeger/pull/2707

### DIFF
--- a/packages/jaeger-ui/public/index.html
+++ b/packages/jaeger-ui/public/index.html
@@ -18,7 +18,12 @@
       // Jaeger UI config data is embedded by the query-service via search-replace.
       // This is later merged with defaults into the redux `state.config` via
       // src/utils/config/get-config.js.
+      // JAEGER_CONFIG_JS
+      // the line above may be replaced by user-provided JS file that should define a UIConfig function.
       function getJaegerUiConfig() {
+        if(typeof window.UIConfig === 'function') {
+          return UIConfig();
+        }
         const DEFAULT_CONFIG = null;
         const JAEGER_CONFIG = DEFAULT_CONFIG;
         return JAEGER_CONFIG;


th3M1ke:
@yurishkuro I remember, that we discuss memoization of the `getConfig()` function [here,](https://github.com/jaegertracing/jaeger/pull/2707#discussion_r549876846) but I found one test, which checks for depreciation each time we invoke `getConfig()`. Is this relevant? What is the use case for this?

yurishkuro:
> but I found one test, which checks for depreciation each time we invoke getConfig(). Is this relevant? What is the use case for this?

Don't know what the use case would be, given that at present the config is static as far as the UI is concerned. Maybe it was envisioning loading (and reloading) the config from the server dynamically?  Where is the test?

yurishkuro (approved):

th3M1ke:
Test 'processes deprecations every time `getConfig` is invoked' located 'packages/jaeger-ui/src/utils/config/get-config.test.js:113'

yurishkuro:
So from what I understand, there were already some historical config format changes such that certain properties have been renamed or moved to different location, and that's what `process-deprecation.ts` module is doing. In terms of memoization, the deprecation logic should run at the same time as the UIConfig() function is being called, i.e. the combined result should be memoized. I think the calls to getConfig() are the ones that should be memoized.

packages/jaeger-ui/public/index.html
```diff
@@ -18,7 +18,12 @@
       // Jaeger UI config data is embedded by the query-service via search-replace.
       // This is later merged with defaults into the redux `state.config` via
       // src/utils/config/get-config.js.
+      // JAEGER_CONFIG_JS
+      // the line above may be replaced by user-provided JS file that should define a UIConfig function.
       function getJaegerUiConfig() {
+        if(typeof window.UIConfig === 'function') {
+          return UIConfig();
```
bobrik, line 25: Should the result be memorized? I see a dozen log lines with the following config:

```js
function UIConfig() {
    console.log("loading config");
}
```

yurishkuro: yes - https://github.com/jaegertracing/jaeger-ui/pull/677#issuecomment-760386379

th3M1ke:
@yurishkuro I suggest move the discussion about memoization here https://github.com/jaegertracing/jaeger-ui/pull/679

Issue #675, opened 2021-01-10
Support for showing tags for collapsed spans


bobrik:
## Requirement - what kind of business use case are you trying to solve?

I'm trying to reduce the number of distinct operations and make it easier to aggregate and analyze traces with many spans of the same type, while retaining the ability to see what happens in the trace without expanding too many spans.

## Problem - what in Jaeger blocks you from solving the requirement?

I have a trace with many similar operations like "load module by name" or "run command X with args", here's an example:

<img width="1440" alt="image" src="https://user-images.githubusercontent.com/89186/104133873-25f63380-533b-11eb-9a66-c50958a3c99b.png">

This leads to many distinct operations, since every module gets its own. Trace stats are suffering from this, not allowing me to do group by (operation, module name), because both things are lumped into one string of `operationName`:

<img width="1440" alt="image" src="https://user-images.githubusercontent.com/89186/104134031-0c092080-533c-11eb-8aea-b7bbfc27dc30.png">

It also makes it hard to search for "any module load above Xms". It's still doable if I attach the same tag to all module loading operations, but it doesn't feel right.

If I were to not include the module name into operation name, the trace would've looked like this:

![image](https://user-images.githubusercontent.com/89186/104133928-753c6400-533b-11eb-8f41-30c9d961f85e.png)

This is much harder to understand, since I have 2000+ spans for module loading and I'm trying to figure out whether some of them are getting loaded for no reason.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

I suggest to designate some tag to be showed in the timeline:

![image](https://user-images.githubusercontent.com/89186/104134085-686c4000-533c-11eb-940d-ecefa3dd804b.png)

I think it solves all the outlined problems at expense of requiring a new tag to be attached, which feels like an okay compromise. For spans that do not opt into having a new tag it's a no-op.

## Any open questions to address

I'm happy to consider other options.

yurishkuro:
I agree this is a valuable feature. Another common situation where it can be useful is on outbound HTTP calls, which are often represented simply with span name `GET` because instrumentation does not have an easy way of extracting low-cardinality name from the URL. But if we had a feature like this, those spans could be annotated with the full URL just in the display.

The thing we need to design is how the UI should know which tag to pull into display for different types of spans (NB: somewhat relevant discussion in OpenTelemetry: https://github.com/open-telemetry/opentelemetry-specification/pull/810). Some options are:

[...]

PR #678, opened 2021-01-15
WIP: Support for customizing the display of operation names

bobrik:
See https://github.com/jaegertracing/jaeger-ui/issues/675 for backstory. Marking it as WIP as it's far from ready.

With the following config file:

```js
function UIConfig() {
    return {
        "traceProcessor": async function (trace) {
            for (const span of trace.spans) {
                const tags = {};

                for (const kv of span.tags) {
                    tags[kv.key] = kv.value;
                }

                if ("http.method" in tags) {
                    span.operationLabel = `${tags["http.method"]} → ${tags["http.status_code"]}`;
                }
            }
        },
    };
}
```

I can get the following view:

image

This doesn't change the stats view, as expected:

image

This solves the immediate issue, but I'd like to be able to do `async` fetches in my `traceProcessor` to allow augmenting traces with data that is not immediately available. One use-case is having a customer specific SLO and adding a `warning` to spans if they breach the SLO. Given that the call chain leading to `traceProcessor` is not `async`, this currently doesn't work. I'm not a front-end expert to figure this out by myself and it would be nice to get some hints.

Let me know if this is the right approach.

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBarRow.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBarRow.tsx
@@ -104,6 +104,7 @@ export default class SpanBarRow extends React.PureComponent<SpanBarRowProps> {
       duration,
       hasChildren: isParent,
       operationName,
+      operationLabel,
       process: { serviceName },
     } = span;
     const label = formatDuration(duration);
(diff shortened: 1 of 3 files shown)


colelawrence:
I don't understand why this is described as "far from ready" the code is simple enough. I think it might be a good idea to clone the span into a separate data structure and use something akin to `try { transformerFn(spanCopy) } catch (err) {}` in case the internal structure of the span changes (we don't want to corrupt the UI with screwed up data if there is a mapping error).

I would also consider naming this API as `unstableTraceTransformer(trace)` until there's a data model that can be stabilized. I think it would also suffice as well to skip documentation to start getting feedback.

yurishkuro:
I think OP listed some of the reasons why it's "not ready", one is a lack of async behavior.

I also don't like such a brute force "transform trace" backdoor, it's dangerous. The specific behavior here is defining a label for operation name, which is a substantially more narrow use case than what transform-trace allows. Instead of having such UDF that can mutate the trace, and considering that the UI still needs to have knowledge about the label field, I would rather have a smaller scoped hook like `computeOperationNameLabel(span, trace) -> string` (and maybe it can be made async, this is a bit beyond my React knowledge).

The thing I do like about a code-based hook is that it unblocks this use case for motivated users, instead of waiting for us to decide to implement more configuration-driven approaches that were discussed in the ticket.

PR #679, opened 2021-01-15
Add memoizaion for getConfig

th3M1ke:
- invoke JS config each time when `getConfig()` run
- resolve discussion from https://github.com/jaegertracing/jaeger-ui/pull/677 
- Add memoization for `getConfig()`

### DIFF
--- a/packages/jaeger-ui/src/utils/config/get-config.test.js
+++ b/packages/jaeger-ui/src/utils/config/get-config.test.js
@@ -59,6 +59,7 @@ describe('getConfig()', () => {
     let getJaegerUiConfig;
 
     beforeEach(() => {
+      getConfig.apply({}, []);
       embedded = {};
       getJaegerUiConfig = jest.fn(() => embedded);
       window.getJaegerUiConfig = getJaegerUiConfig;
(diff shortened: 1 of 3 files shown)


packages/jaeger-ui/src/utils/config/get-config.test.js
```diff
@@ -59,6 +59,7 @@ describe('getConfig()', () => {
     let getJaegerUiConfig;
 
     beforeEach(() => {
+      getConfig.apply({}, []);
```
yurishkuro, line 62: what does this do?

yurishkuro (approved):
lgtm

yurishkuro:
I wonder if #678 requires the opposite, i.e. always re-evaluating the config. @bobrik PTAL

bobrik:
I don't think #678 requires re-evaluation. As long as I can have a function stored in a config key, it should be fine.

bobrik (approved):

packages/jaeger-ui/src/utils/config/get-config.test.js
th3M1ke: This is reset memoization cache

https://github.com/alexreardon/memoize-one/issues/43

yurishkuro: thx. I wish they would've fixed that ticket by adding an explicit method like reset(), much better for readability.

PR #681, opened 2021-01-22
Generic web analytics tracking implementation

### LINKED ISSUE #652: generic analytics tracking (in addition to GA)

We would like to be able to track events in a generic way (i.e with a non GA solution) when running jaeger-UI inside an iframe 

Jaeger-UI tracking is tightly coupled to GA

We were thinking about adding a [post message](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) to the tracking flow so it would be possible to register for events invoked by the UI.
the new flow will be configurable via the UI-config file

th3M1ke:
- resolves #652 

- GA module was transferred to a separate class and will be invoked once UI configuration property `UIConfig.tracking.gaID` will be added
- if will be provided `customWebAnalytics` implementation via UI configuration property `UIConfig.tracking.customWebAnalytics` it will be used as web analytic in jaeger UI
- if any of UI configuration will not be provided - empty mock(`() => {}`) will be used
- only one web analytic system can work. Priority: Custom Web Analytic -> Google Analytics -> Mock
- `UIConfig.tracking.customWebAnalytics` format(also described in `IWebAnalytics` in types):
```
const CustomWebAnalytics: IWebAnalytics = {
    init: () => {},
    trackPageView: () => {},
    trackError: () => {},
    trackEvent: () => {},
    context: null,
    isEnabled: () => false,
};
```

`UIconfig.js`:
```
function UIConfig() {
  return {
     ...other properties
     tracking: {
       customWebAnalytics: function () {
         return {
           init: () => { /* initialization actions */ },
           trackPageView: (pathname, search) => {},
           trackError: (description) => {},
           trackEvent: (category, action, labelOrValue, value) => {},
           context: null,
           isEnabled: () => false,
         }
       }
     }
  }
}
```

### DIFF
--- a/packages/jaeger-ui/src/constants/default-config.tsx
+++ b/packages/jaeger-ui/src/constants/default-config.tsx
@@ -85,6 +85,7 @@ export default deepFreeze(
       tracking: {
         gaID: null,
         trackErrors: true,
+        customWebAnalytics: null,
       },
     },
     // fields that should be individually merged vs wholesale replaced
--- a/packages/jaeger-ui/src/types/tracking.tsx
+++ b/packages/jaeger-ui/src/types/tracking.tsx
@@ -14,6 +14,11 @@
 
 import { RavenStatic } from 'raven-js';
 import { TNil } from '.';
+import { Config } from './config';
+
+export interface IWebAnalyticsFunc {
+  (config: Config, versionShort: string, versionLong: string): IWebAnalytics;
+}
 
 export default interface IWebAnalytics {
   init: () => void;
--- a/packages/jaeger-ui/src/types/tracking.tsx
+++ b/packages/jaeger-ui/src/types/tracking.tsx
@@ -1,0 +1,30 @@
+// Copyright (c) 2021 Uber Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { RavenStatic } from 'raven-js';
+import { TNil } from '.';
+
+export default interface IWebAnalytics {
+  init: () => void;
+  context: boolean | RavenStatic | null;
+  isEnabled: () => boolean;
+  trackPageView: (pathname: string, search: string | TNil) => void;
+  trackError: (description: string) => void;
+  trackEvent: (
+    category: string,
+    action: string,
+    labelOrValue?: string | number | TNil,
+    value?: number | TNil
+  ) => void;
+}
--- a/packages/jaeger-ui/src/types/tracking.tsx
+++ b/packages/jaeger-ui/src/types/tracking.tsx
@@ -1,4 +1,4 @@
-// Copyright (c) 2021 Uber Technologies, Inc.
+// Copyright (c) 2021 The Jaeger Authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/packages/jaeger-ui/src/utils/tracking/index.tsx
+++ b/packages/jaeger-ui/src/utils/tracking/index.tsx
@@ -15,6 +15,7 @@
 import { TNil } from '../../types';
 import { IWebAnalyticsFunc } from '../../types/tracking';
 import GA from './ga';
+import NoopWebAnalytics from './noopWebAnalytics';
 import getConfig from '../config/get-config';
 
 const TrackingImplementation = () => {
@@ -40,15 +41,6 @@ const TrackingImplementation = () => {
     versionShort = 'unknown';
     versionLong = 'unknown';
   }
-
-  const NoopWebAnalytics: IWebAnalyticsFunc = () => ({
-    init: () => {},
-    trackPageView: () => {},
-    trackError: () => {},
-    trackEvent: () => {},
-    context: null,
-    isEnabled: () => false,
-  });
 
   let webAnalyticsFunc = NoopWebAnalytics;
 
--- a/packages/jaeger-ui/src/utils/tracking/index.tsx
+++ b/packages/jaeger-ui/src/utils/tracking/index.tsx
@@ -12,71 +12,47 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import _get from 'lodash/get';
-import queryString from 'query-string';
-import ReactGA from 'react-ga';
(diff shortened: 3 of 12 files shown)


packages/jaeger-ui/src/utils/tracking/index.tsx
```diff
@@ -12,71 +12,47 @@
...
-  action: 499,
-  category: 149,
-  label: 499,
-};
+const TrackingImplementation = () => {
+  const config = getConfig();
 
-// Util so "0" and "false" become false
-const isTruish = (value?: string | string[]) => Boolean(value) && value !== '0' && value !== 'false';
+  const GenericWebAnalytics: IWebAnalytics = {
```
yurishkuro, line 22: maybe `NoopWebAnalytics`?

packages/jaeger-ui/src/utils/tracking/index.tsx
```diff
@@ -12,71 +12,47 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import _get from 'lodash/get';
-import queryString from 'query-string';
-import ReactGA from 'react-ga';
-import Raven, { RavenOptions, RavenTransportOptions } from 'raven-js';
```
yurishkuro, line 18: I didn't realize we were using raven!

packages/jaeger-ui/src/utils/tracking/index.tsx
```diff
@@ -12,71 +12,47 @@
...
 
-// In test mode if development and envvar REACT_APP_GA_DEBUG is true-ish
-const isDebugMode =
-  (isDev && isTruish(process.env.REACT_APP_GA_DEBUG)) ||
-  isTruish(queryString.parse(_get(window, 'location.search'))['ga-debug']);
+  if (config.tracking && config.tracking.customWebAnalytics) {
+    webAnalytics = config.tracking.customWebAnalytics(config) as IWebAnalytics;
+  } else if (config.tracking && config.tracking.gaID) {
+    webAnalytics = new GA(config);
+  }
```
yurishkuro, line 37: Another option instead of forced priority order is to allow enabling both GA and custom WA via a wrapper implementation that would dispatch events to both. But I don't feel strongly about it, the priority order seems fine.

packages/jaeger-ui/src/types/tracking.tsx
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) 2021 Uber Technologies, Inc.
```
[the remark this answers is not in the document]
th3M1ke: No, I believe I just copied it form some file...

packages/jaeger-ui/src/constants/default-config.tsx
```diff
@@ -85,6 +85,7 @@ export default deepFreeze(
       tracking: {
         gaID: null,
         trackErrors: true,
+        customWebAnalytics: null,
```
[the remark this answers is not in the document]
th3M1ke: If I understand the question correct, my answer is yes it will look like

```
customWebAnalytics: function (config?) {
         return {
           init: () => { /* initialization actions */ },
           trackPageView: (pathname, search) => {},
           trackError: (description) => {},
           trackEvent: (category, action, labelOrValue, value) => {},
           context: null,
           isEnabled: () => false,
         }
       }
```

packages/jaeger-ui/src/utils/tracking/index.tsx
th3M1ke: 🤷‍♂️

th3M1ke:
> I'm confused - all this functionality now seems to only exist in the GA implementation of IWebAnalytics, but wasn't it a general functionality that would be applicable to any other implementation? I would've expected only the reporting part to be swapped out, not all the data preparation / massaging.

Do you suggest making a transport layer? if so, I do not clearly understand how should it look like. Currently, I lot of logic goes with GA and data preparation to be sent. 

let's make alignment about the implementation itself:

1. We have the next function that currently using:
  - trackPageView
  - trackError
  - trackEvent

`trackPageView` can be a part of trackEvent?
`trackError` - can be used for web analytics and, also we can pass the custom logger

Does this fit for you? Or can we reduce them to `trackError` and `trackEvent`?


2. Once we add only the transport layer users can not modify incoming data and map it according to their needs
3. Is invoking custom WA good from your perspective?
```function UIConfig() {
  return {
     ...other properties
     tracking: {
       customWebAnalytics: function () {
         return {
           init: () => { /* initialization actions */ },
           trackPageView: (pathname, search) => {},
           trackError: (description) => {},
           trackEvent: (category, action, labelOrValue, value) => {},
           context: null,
           isEnabled: () => false,
         }
       }
     }
  }
}
```


> Another option instead of forced priority order is to allow enabling both GA and custom WA via a wrapper implementation that would dispatch events to both. But I don't feel strongly about it, the priority order seems fine.

Sure we can init them both in parallel. Wrappers could be implemented inside `trackPageView`, `trackError`, `trackEvent` in `packages/jaeger-ui/src/utils/tracking/index.tsx`


> it would be good to ramp up some test coverage, 0.19% drop is pretty big.

Will add tests as we align about implementation if it is ok :)

packages/jaeger-ui/src/constants/default-config.tsx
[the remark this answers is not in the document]
yurishkuro: I meant whether we have a type definition for Config somewhere. Maybe not.

yurishkuro:
> Do you suggest making a transport layer?

Well, depends on how you look at it. Take the logic in the init method, for example - it determines app version, checks some cookies, configures raven, etc. A good 80-90 lines of code, of which maybe 5 are related to GA. So shouldn't that same functionality be available if you're reporting events to another provider?

th3M1ke:
@yurishkuro,  I'm not sure about transport. From my perspective transport layers remove flexibility, and add complexity. We need to expose adapters for different analytic systems or provide such ability to users.

About errors. Not all systems provide specific error tracking function. So you have to track them as events. But for errors user can provide a custom logger via configuration.

As for raven, we can provide it as a configuration and the user can decide to use it or not.

yurishkuro:
ok. We can refactor that later, right now it's not worse than it was before, so I'd be ok to merge.

th3M1ke:
🤝  Let me increase test coverage shortly!

th3M1ke:
@yurishkuro can you take a look at codecov analysis? https://codecov.io/gh/jaegertracing/jaeger-ui/pull/681/diff

There are gaps under google analytics implementation. Those lines are related to development. Should I also cover them as well?

yurishkuro:
the rule of thumb is to leave the code with the cumulative coverage no worse than before

yurishkuro:
seems the checks are stuck

yurishkuro:
https://codecov.io/github/jaegertracing/jaeger-ui/commit/aa603193a1ed5daf343ce4d44ceda2d721ded953

Is it possible to run GA tests in both debug and non-debug mode? This will improve the coverage

yurishkuro:
+0.13%  https://codecov.io/github/jaegertracing/jaeger-ui/commit/491353a892a932b76b22c67c9a5a8e2e48e3c97b

PR #688, opened 2021-02-16
Improved SearchForm test suite

achesin:
Signed-off-by: Amanda Chesin <Amanda.Chesin1@aexp.com>

- Adding test coverage around existing functionality.

- Added new tests for SearchForm, including coverage for mapDispatchToProps, the operation dropdown, etc.


packages/jaeger-ui/src/components/SearchTracePage/SearchForm.test.js
```diff
@@ -414,6 +432,24 @@ describe('<SearchForm>', () => {
...
+    wrapper = shallow(<SearchForm {...defaultProps} selectedService="svcC" />);
+    const field = wrapper.find(`Field[name="operation"]`);
+    expect(field).toMatchSnapshot();
+  });
+
+  it('enables operation when known service selected', () => {
+    wrapper = shallow(<SearchForm {...defaultProps} selectedService="svcB" />);
+    const field = wrapper.find(`Field[name="operation"]`);
+    expect(field).toMatchSnapshot();
+  });
```
yurishkuro, line 452: If I understand this correctly, these new tests are merely checking a single property on the `field`. Wouldn't it be simpler to test that single property directly, like `expect(field.foo).equalTo('bar')`, than to compare with a whole snapshot? The intent would be a lost more obvious and maintenance easier.

achesin: Hi, I reworked those tests to test the single property instead of using the snapshots. Please let me know what you think, thanks!

yurishkuro: Thanks!

What about this specific test? Is it aiming to check only the boolean condition, or also the list of operations that are included in the dropdown?

achesin: I had wanted to write a test to check the boolean condition as well as the list of operations included. I removed the above test since I didn't think it was what you wanted, but I could reword the name of the test and include it if you think it's important to check all of those values as a snapshot. Or I could write another test to just check the list of operations that are included once a service is selected- whatever would work best for your style.

achesin: Oops, meant to remove that snapshot test in the last commit, I'm new to open source contributions so my bad!

yurishkuro (approved):
Thanks!

PR #691, opened 2021-02-18
Add dependabot/codeql/mergify

jpkrohling:
This PR adds the capability of automatic merge of npm package updates, as long as they pass the regular tets.

Signed-off-by: Juraci Paixão Kröhling <juraci@kroehling.de>

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -1,0 +1,8 @@
+pull_request_rules:
+  - name: automatic merge for dependabot pull requests
+    conditions:
+      - author~=^dependabot(|-preview)\[bot\]$
+    actions:
+      merge:
+        method: squash
+        commit_message: title+body
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "daily"
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,32 @@
+name: "CodeQL"
+
+on:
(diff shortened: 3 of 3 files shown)


.github/mergify.yml
```diff
@@ -0,0 +1,12 @@
+pull_request_rules:
+  - name: automatic merge for dependabot pull requests
+    conditions:
+      - author~=^dependabot(|-preview)\[bot\]$
+      - "status-success=unit-tests"
+      - "status-success=CodeQL Analyze (javascript)"
+      - "status-success=DCO"
+      - "status-success=WIP"
```
yurishkuro, line 8: Listing all merge conditions like this is kind of unfortunate. If we add another required check in the future, we'd need to remember to add it here too.

Does it not support "merge on green" where "green" is defined by the required checks on the repo, whatever they are?

yurishkuro (approved):

.github/mergify.yml
jpkrohling: It might now, I think we could determine the conditions on the repo settings and mergify just won't be able to merge it. We could give it a try afterward.

jpkrohling: >Note that Mergify will always respect the branch protection settings. When the conditions match and the merge action runs, Mergify waits for the branch protection to be validated before merging the pull request.

It might just work, I'll remove this from this PR.

.github/mergify.yml
```diff
@@ -0,0 +1,8 @@
+pull_request_rules:
+  - name: automatic merge for dependabot pull requests
+    conditions:
+      - author~=^dependabot(|-preview)\[bot\]$
```
yurishkuro, line 4: I assume this will require manual approval before auto-merging.

We need to be more careful with this repo than with compiled languages, as we do not have full code coverage, and UI features sometimes have subtle functional differences.

jpkrohling: Yes. I removed this requirement, but just reinstated based on your comment.

image

PR #699, opened 2021-02-19
Preparing release v1.13.0

albertteoh:
Signed-off-by: albertteoh <albert.teoh@logz.io>

- Prepares jaeger-ui for release of v1.13.0

- Updates changelog and package.json appropriately.


yurishkuro (approved):

jpkrohling:
I would like to get #702 in before this release.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBarRow.test.js
```diff
@@ -42,7 +42,7 @@ describe('<SpanBarRow>', () => {
       serviceName: 'rpc-service-name',
     },
     showErrorIcon: false,
-    getViewedBounds: () => ({ start: 0, end: 1 }),
+    getViewedBounds: () => ({ start: 10, end: 1 }),
```
yurishkuro, line 45: what does it mean to start from 10 through to 1?

albertteoh: The motivation for this change is to add test coverage for this code block which was blocking this PR from passing builds: https://github.com/jaegertracing/jaeger-ui/blob/df17e8d80e6f630ed3ed744a80dbf27c0418718c/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBarRow.tsx#L118-L119

I don't know how this code coverage issue appeared given my only changes were to the changelog, but in any case, thought I'd try to fix it as it seems to be the best option for me to unblock the release (please advise if I'm wrong).

The above code block uses the start and end view bounds to determine where the hint label should appear when mousing over the span timeline. So if there's more whitespace to the left of the span timeline than the right, show the label to the left, otherwise, show it to the right.

Screen Shot 2021-02-23 at 10 56 09 am
Screen Shot 2021-02-23 at 10 55 28 am

Realistically, these bounds seem to range between 0 and 1 and so 10 is a little extreme to be honest, but it does the job to get the line coverage.

yurishkuro: > seem to range between 0 and 1 and so 10 is a little extreme to be honest

And out of range, most importantly. I am not familiar with the test, but having 10 here is just going to confuse someone else at some time in the future.

I would remove this change. We can ignore the code cov diff for the purpose of this PR, since the diff is coming from a flaky test.

albertteoh: Agree and reverted. The "Squash and merge" button is greyed out for me due to the codecov failure; how do we ignore it and proceed with merging this PR?

yurishkuro: booked a ticket for flaky coverage #704

yurishkuro (approved):

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBarRow.test.js
yurishkuro: I merged it - probably requires admin rights on the repo.

albertteoh:
Thanks @yurishkuro!

PR #703, opened 2021-02-22
Allow a tag to be displayed next to operation name (revamp #487)

bboreham:
This PR takes the change in #487 from @lookfwd, rebases it against latest, and applies the suggested changes from @everett980.

> ## Which problem is this PR solving?
> * There is a need to display additional information for each _Span_ in a prominent position (next to operation name). For example, with the following `config.json`:
> 
> ```
> {
>   "opLabelTag": "http.status_code"
> }
> ```
> 
> we get:
> 8E0BC66E-B4F4-4598-A945-FDF80FE2052C
>
> ## Short description of the changes
>
> Adds a configuration option `opLabelTag` which, when set, shows the value of the specified `opLabelTag` _Tag_ for a _Span_, if set, in parenthesis next to the operation name in the trace view.

I signed the extra commits as myself because I don't want to impersonate someone without permission, but the code is from @everett980. Some small changes were needed to let tests pass.

### DIFF
--- a/packages/jaeger-ui/src/model/link-patterns.tsx
+++ b/packages/jaeger-ui/src/model/link-patterns.tsx
@@ -20,6 +20,7 @@ import { encodedStringSupplant, getParamNames } from '../utils/stringSupplant';
 import { getParent } from './span';
 import { TNil } from '../types';
 import { Span, Link, KeyValuePair, Trace } from '../types/trace';
+import pluckTruthy from '../utils/ts/pluckTruthy';
 
 type ProcessedTemplate = {
   parameters: string[];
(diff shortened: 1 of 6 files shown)


yurishkuro:
@th3M1ke would you be able to review this?

th3M1ke:
@yurishkuro sure! will review later today

packages/jaeger-ui/src/model/link-patterns.tsx
```diff
@@ -157,12 +158,34 @@ export function computeTraceLink(linkPatterns: ProcessedLinkPattern[], trace: Tr
...
 }
 
+function getParameterValues(
+  parameters: string[],
+  items: KeyValuePair[],
+  findInAncestorsFn?: (_: string) => KeyValuePair | undefined,
+  onMissingFn?: (_: string) => void
+) {
+  const parameterValues: Record<string, any> = {};
+  const allParameters = parameters.every(parameter => {
```
th3M1ke, line 168: Why we need `allParameters`?

th3M1ke, line 168:
```suggestion
  parameters.forEach(parameter => {
```

To be able to render different paraments(from different scopes. HTTP calls statuses and Redis's driverID, for example)

bboreham: This code is from @everett980 at https://github.com/jaegertracing/jaeger-ui/pull/487/files#r356227730.

What happens to return values if I make that change?
Are you proposing that the "all elements match otherwise nothing" behaviour be changed?
That was quite deliberate in the original PR; @lookfwd [added a test for it](https://github.com/jaegertracing/jaeger-ui/pull/487/files#diff-5366cdb1f6c0803f57f6dc374b480ba6abc8897a82ab0ab7bc517532bd2d5ec9R95)

jpkrohling:
What's the state of this PR?

bboreham:
I was awaiting further comment.
I asked a couple of questions for clarification at https://github.com/jaegertracing/jaeger-ui/pull/703#discussion_r582821264

jpkrohling:
@rubenvp8510 could you please review this one?

drewcorlin1:
This is quite an old what, any chance of this getting revived? I'd be happy to help out

PR #714, opened 2021-03-07
Proposal: Render user tooltips near operation

FauxFaux:

Feature proposal: Allow the client library to label each span, and show that label prominently, but not as prominently as the operation name. This allows some higher cardinality data into the UI, in a way that isn't three clicks away for every span, and doesn't mess with the core data model, or the low-cardinality indexing on operation name.

`span.tooltip` is an intentionally awful straw-man proposal for a tag name, suggestions welcome.

In the single trace view, look for a tag named `span.tooltip`, and show it next to the operation (called "endpoint name" in some places), both in the sidebar, and in the hover on the bar graph (next to the time). Note, it is not part of the operation name, which is still e.g. `pg.query:SELECT`.

image


bobrik:
This seems like a less general version of #675.

PR #716, opened 2021-03-08
Tabular View of Spans WIP 

vvvprabhakar:
Added a tabular view of spans 
Where spans can sort based on the columns in the table  
Filters can be applied on spans 

The reason for the creation of this view is to find out the most time taken spans in a trace which have more than ~20K spans which we are facing currently 

External Packages Used : 
[react-table ](https://www.npmjs.com/package/react-table)
[materail-ui/core*](https://www.npmjs.com/package/@material-ui/core) 

Screenshot 2021-03-26 at 13 45 05

Screenshot 2021-03-26 at 13 49 04

Screenshot 2021-03-26 at 13 45 22


packages/jaeger-ui/src/components/TracePage/TracePageHeader/AltViewOptions.tsx
```diff
@@ -46,6 +47,10 @@ const MENU_ITEMS = [
     viewType: ETraceViewType.TraceStatistics,
     label: 'Trace Statistics',
   },
+  {
+    viewType: ETraceViewType.TraceSpansView,
+    label: 'Trace Spans View',
```
yurishkuro, line 52:
```suggestion
    label: 'Trace Spans Table',
```

package.json
```diff
@@ -74,5 +74,15 @@
     "hooks": {
       "pre-commit": "npm-run-all -ln --parallel lint test"
     }
+  },
+  "dependencies": {
+    "@material-ui/core": "^4.11.3",
+    "@material-ui/icons": "^4.11.2",
```
yurishkuro, line 80: Not too familiar with this, does it mean you're using a different widgets library that wasn't used previously?

yurishkuro (commented):
Could you add screenshots or even better screencast?

package.json
vvvprabhakar: Yeah we needed some icons while loading , for tables sort icons and all 
Can i share you some screenshot how it looks

joe-farro:
Historically we've used react-icons which has a ton of icons to including
material design icons.

https://react-icons.github.io/react-icons/icons?name=md

vvvprabhakar:
@joe-farro @yurishkuro 

Can i use  `@material-ui/core` for TablePagination , TableSortLabel? 
Do u have any other alternative

vvvprabhakar:
And What do u say about styling, do u want any changes in styling

packages/jaeger-ui/src/components/SearchTracePage/SearchForm.test.js
```diff
@@ -473,7 +437,7 @@ describe('mapStateToProps()', () => {
   let state;
 
   beforeEach(() => {
-    state = { router: { location: { search: '' } } };
+    state = { router: { location: { serach: '' } } };
```
rubenvp8510, line 440: is this change intentional?

rubenvp8510:
Not totally convinced of introducing another widget library, I'll try to figure out what other alternatives we have here.

packages/jaeger-ui/src/components/SearchTracePage/SearchForm.test.js
vvvprabhakar: no it was not intentional

rubenvp8510:
What is the real motivation behind the use of `material-ui` ? Is a dependency of `react-table` ? If we can use only `react-table` I think I'm fine with that, or if there is another lightweight option that allows sorting, better!

vvvprabhakar:
I will come up with new PR without any dependencies and just the `react-table`

Issue #721, opened 2021-03-31
Feature request: Treat different `service.instance.id` the same as different `service.name` when viewing a trace


nikclayton-dfinity:
## Requirement - what kind of business use case are you trying to solve?

The current trace view does a great job when a trace spans multiple distinct services. The spans are clearly delineated by colour so you can which span is associated with which service.

However, it's possible to have a service architecture where different members of the service are communicating amongst themselves as well as with other services.

Per https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/resource/semantic_conventions#service each member of the service should have a different `service.instance.id`.

So if I have A and B, both members of service "foo", and A is gossiping data to B, the trace showing that gossip will have spans with

```
service.name: foo
service.instance.id: A
...
```

and spans with

```
service.name: foo
service.instance.id: B
...
```

Currently, when viewing that trace, Jaeger renders each span in the same colour, making it difficult to see which instance is associated with each span.

If I change my spans to emit:

```
service.name: foo.A
```

and

```
service.name: foo.B
```

then I get the UI that I want -- spans from different processes implementing the same service are clearly delineated -- but the data in the `service.name` field is now wrong.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

When rendering traces, Jaeger should colour spans according to not just the `service.name` field, but the `service.instance.id` field as well.

And ideally, the `service.namespace` field would be included too, since the combination of `service.name`, `service.namespace`, and `service.instance.id` are supposed to be globally unique.

nikclayton-dfinity:
I'm working on a PR for this.

nikclayton-dfinity:
https://github.com/jaegertracing/jaeger-ui/pull/722

Issue #582, opened 2020-05-22
Better naming for operations and grouping by operation type in trace summary


romilpunetha:
## Requirement - what kind of business use case are you trying to solve?

The operation names for mysql and mongo currently are the entire queries itself. Would be good to have the query as an attribute, with the operation name as `mysql.query`,`mongo.query`, `redis.query`, etc. This would simplify grouping not only by service name but also by operation types, which would also be good to have on the trace summary.

## Problem - what in Jaeger blocks you from solving the requirement?

Jaeger groups spans only by service name. It would be good to have if there's another section in the summary that groups by operation type. This would help better visualize the trace. eg. Lets say a particular trace spans across 5 services. The trace summarizes the names of these services, along with info like `15 http calls`, `10 mysql.query calls`, `3 mongo.query calls`... and so on.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

There's already `db.type` getting captured as an attribute. Possibly that could be used for better naming a db operation. Then, the UI could show a grouped version of these operations. This would allow a developer to better identify what the trace exactly contains without having to deep dive in it. Knowing that the trace spans across might not be that useful, but knowing that it does contain a lot of DB calls makes a difference.

romil-punetha:
@annanay25

jkowall:
I know this issue is quite old, and I was trying to replicate the issue here with the operation names. I'm not seeing this on my Jaeger instances across a few different applications. I'm not sure if this is related to the instrumentation we are using Otel, versus what you might have been using. @romilpunetha can you please post a screenshot showing the issue you are describing? 

Thanks

Issue #121, opened 2017-11-09
[traceview] Allow grouping of spans by functional areas


yurishkuro:
Let's say a trace starts in a business service and then reaches into a storage that branches into many different spans (e.g. a call to Cassandra). Sometimes users don't want to see all the minute details of the storage interactions, they just want to see a single span for the call to storage. They manually collapse the top span representing a\ call to the storage, but if there are many of storage calls it's a lot of clicking in the trace before one can get to a desired picture.

How can we improve this?

bayesianmind:
We (The Catalyst team) would love something like this. Ideally this would even support multiple levels of collapse:
- Interservice spans like calls to databases
- Service API invocation with any proxy layers like Muttley or any of the upcoming RPC proxy layers for GRPC / Goldstar compliance
- Functional areas like you mention

One idea for implementing this would be that spans can emit metadata tags with whatever hierarchy makes sense to them:
context[0] = "myservice-mysql"
context[1] = "myservice"
context[2] = "myservice+proxies"
context[3] = "myFunctionalArea"

Then there could be a 'Collapse one level' button in the UI which collapses one level at a time.

objectiser:
Just wondering if this could start at the service level - so initially when the user views a trace they see the activity on a per service basis, with all internal spans hidden - and then expanding a service may then show the functional areas grouped - and then expanding on those can delve into more detail?

For traces involving a large number of services, this would atleast provide a high level overview of the end to end service interactions.

It may also be good eventually if some services could be configured as not displayed by default - so if users want to focus primarily on the business services only they can exclude databases, etc. But this may require per user UI config/preferences.

divdavem:
I have opened #217 that could be an idea to help solving this issue.

PR #722, opened 2021-04-02
WIP: Distinguish traces by service namespace and instance.id

nikclayton-dfinity:
When displaying / distinguishing traces by name or colour consider the
service.namespace and service.instance.id tags if present.

This helps distinguishes traces that span multiple process that make up
the same logical service where different instances have different IDs.

https://github.com/jaegertracing/jaeger-ui/issues/721

This is not yet ready to merge, I've opened the PR because I think there are a few ways to proceed and I wanted to get consensus on whatever approach is acceptable to the Jaeger maintainers. I'll add comments to this PR with additional information.

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TracePageHeader/SpanGraph/CanvasSpanGraph.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TracePageHeader/SpanGraph/CanvasSpanGraph.tsx
@@ -21,7 +21,8 @@ import { TNil } from '../../../../types';
 import './CanvasSpanGraph.css';
 
 type CanvasSpanGraphProps = {
-  items: { valueWidth: number; valueOffset: number; serviceName: string }[];
+  // TODO: Unclear why this is not the `SpanItem` type
+  items: { valueWidth: number; valueOffset: number; serviceName: string, cacheKey: string, rgbColor: [number, number, number] }[];
   valueWidth: number;
 };
 
--- a/packages/jaeger-ui/src/components/TracePage/TracePageHeader/SpanGraph/render-into-canvas.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TracePageHeader/SpanGraph/render-into-canvas.tsx
@@ -25,7 +25,7 @@ export const MAX_ITEM_HEIGHT = 6;
 
 export default function renderIntoCanvas(
   canvas: HTMLCanvasElement,
-  items: { valueWidth: number; valueOffset: number; serviceName: string }[],
+  items: { valueWidth: number; valueOffset: number; serviceName: string, cacheKey: string, rgbColor: [number, number, number] }[],
   totalValueWidth: number,
   getFillColor: (serviceName: string) => [number, number, number]
 ) {
@@ -50,9 +50,11 @@ export default function renderIntoCanvas(
     if (width < MIN_ITEM_WIDTH) {
       width = MIN_ITEM_WIDTH;
     }
-    let fillStyle = fillCache.get(serviceName);
+    let fillStyle = fillCache.get(items[i].cacheKey);
+    const color = items[i].rgbColor;
+    console.log(color);
     if (!fillStyle) {
-      fillStyle = `rgba(${getFillColor(serviceName)
+      fillStyle = `rgba(${color
         .concat(ITEM_ALPHA)
         .join()})`;
       fillCache.set(serviceName, fillStyle);
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
@@ -150,6 +150,29 @@ function getCssClasses(currentViewRange: [number, number]) {
   });
 }
 
+export function getHexColorForSpan(span: Span) {
+  return colorGenerator.getColorByKey(getColorKey(span));
+}
+
+export function getRGBColorForSpan(span: Span) {
+  return colorGenerator.getRgbColorByKey(getColorKey(span));
+}
+
+/**
+ * 
+ * @param span 
+ * @returns The colour key for the span, based on the span's `service.name`,
+ *   `service.namespace`, and `service.instance.id` values.
+ */
+export function getColorKey(span: Span) {
+  const { serviceName } = span.process;
+  const service_namespace_kv = span.process.tags.find(kv => kv.key === "service.namespace");
+  const service_namespace = service_namespace_kv ? service_namespace_kv.value : "";
+  const service_instance_id_kv = span.process.tags.find(kv => kv.key === "service.instance.id");
+  const service_instance_id = service_instance_id_kv ? service_instance_id_kv.value : "";
+  return `${serviceName}:${service_namespace}:${service_instance_id}`
+}
+
 const memoizedGenerateRowStates = memoizeOne(generateRowStatesFromTrace);
 const memoizedViewBoundsFunc = memoizeOne(createViewedBoundsFunc, _isEqual);
 const memoizedGetCssClasses = memoizeOne(getCssClasses, _isEqual);
@@ -341,7 +364,7 @@ export class VirtualizedTraceViewImpl extends React.Component<VirtualizedTraceVi
     if (!trace) {
       return null;
     }
-    const color = colorGenerator.getColorByKey(serviceName);
+    const color = getHexColorForSpan(span);
     const isCollapsed = childrenHiddenIDs.has(spanID);
     const isDetailExpanded = detailStates.has(spanID);
     const isMatchingFilter = findMatchesIDs ? findMatchesIDs.has(spanID) : false;
@@ -353,8 +376,9 @@ export class VirtualizedTraceViewImpl extends React.Component<VirtualizedTraceVi
       const rpcSpan = findServerChildSpan(trace.spans.slice(spanIndex));
       if (rpcSpan) {
         const rpcViewBounds = this.getViewedBounds()(rpcSpan.startTime, rpcSpan.startTime + rpcSpan.duration);
+ 
         rpc = {
-          color: colorGenerator.getColorByKey(rpcSpan.process.serviceName),
+          color: getHexColorForSpan(rpcSpan),
           operationName: rpcSpan.operationName,
           serviceName: rpcSpan.process.serviceName,
           viewEnd: rpcViewBounds.end,
@@ -416,7 +440,7 @@ export class VirtualizedTraceViewImpl extends React.Component<VirtualizedTraceVi
     if (!trace || !detailState) {
       return null;
     }
-    const color = colorGenerator.getColorByKey(serviceName);
+    const color = getHexColorForSpan(span);
     return (
       <div className="VirtualizedTraceView--row" key={key} style={{ ...style, zIndex: 1 }} {...attrs}>
         <SpanDetailRow
--- a/packages/jaeger-ui/src/model/transform-trace-data.tsx
+++ b/packages/jaeger-ui/src/model/transform-trace-data.tsx
@@ -66,6 +66,16 @@ export function orderTags(spanTags: KeyValuePair[], topPrefixes?: string[]) {
   return orderedTags;
 }
 
+// TODO: This duplicates getColorKey and should be refactored
+export function getServiceFromSpan(span: Span) {
+  const { serviceName } = span.process;
+  const service_namespace_kv = span.process.tags.find(kv => kv.key === "service.namespace");
+  const service_namespace = service_namespace_kv ? service_namespace_kv.value : "";
+  const service_instance_id_kv = span.process.tags.find(kv => kv.key === "service.instance.id");
+  const service_instance_id = service_instance_id_kv ? service_instance_id_kv.value : "";
+  return `${serviceName}:${service_namespace}:${service_instance_id}`
+}
+
 /**
  * NOTE: Mutates `data` - Transform the HTTP response data into the form the app
  * generally requires.
@@ -130,8 +140,10 @@ export default function transformTraceData(data: TraceData & { spans: SpanData[]
     if (!span) {
       return;
     }
-    const { serviceName } = span.process;
-    svcCounts[serviceName] = (svcCounts[serviceName] || 0) + 1;
+    // Service instance is uniquely identified by name + namespace + instance id
+    const service = getServiceFromSpan(span);
+
+    svcCounts[service] = (svcCounts[service] || 0) + 1;
     span.relativeStartTime = span.startTime - traceStartTime;
     span.depth = depth - 1;
     span.hasChildren = node.children.length > 0;
--- a/packages/jaeger-ui/src/utils/color-generator.tsx
+++ b/packages/jaeger-ui/src/utils/color-generator.tsx
@@ -74,6 +74,8 @@ export class ColorGenerator {
    * If the key has been used already, it will
    * use the same color.
    */
+  // HexColor might make sense as its own type. Or a Color type with rgb,rgba,
+  // hex() methods
   getColorByKey(key: string) {
     const i = this._getColorIndex(key);
     return this.colorsHex[i];
@@ -84,6 +86,7 @@ export class ColorGenerator {
    * it with a color if the key is not recognized.
    * @return {number[]} An array of three ints [0, 255] representing a color.
    */
+  // TODO: RgbColor might make sense as its own type
   getRgbColorByKey(key: string): [number, number, number] {
     const i = this._getColorIndex(key);
     return this.colorsRgb[i];
--- a/packages/jaeger-ui/src/components/TracePage/TraceGraph/OpNode.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceGraph/OpNode.tsx
@@ -34,6 +34,7 @@ type Props = {
   percentSelfTime: number;
   operation: string;
   service: string;
+  cacheKey: string,
   mode: string;
 };
 
@@ -66,7 +67,7 @@ export function round2(percent: number) {
 
 export default class OpNode extends React.PureComponent<Props> {
   render() {
-    const { count, errors, time, percent, selfTime, percentSelfTime, operation, service, mode } = this.props;
+    const { count, errors, time, percent, selfTime, percentSelfTime, operation, service, cacheKey, mode } = this.props;
 
     // Spans over 20 % time are full red - we have probably to reconsider better approach
     let backgroundColor;
@@ -77,7 +78,7 @@ export default class OpNode extends React.PureComponent<Props> {
       backgroundColor = [255, 0, 0, percentSelfTime / 100].join();
     } else {
       backgroundColor = colorGenerator
-        .getRgbColorByKey(service)
+        .getRgbColorByKey(cacheKey)
         .concat(0.8)
         .join();
     }
--- a/packages/jaeger-ui/src/components/TracePage/TracePageHeader/SpanGraph/index.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TracePageHeader/SpanGraph/index.tsx
@@ -20,6 +20,7 @@ import TickLabels from './TickLabels';
 import ViewingLayer from './ViewingLayer';
 import { TUpdateViewRangeTimeFunction, IViewRange, ViewRangeTimeUpdate } from '../../types';
 import { Span, Trace } from '../../../../types/trace';
+import { getColorKey, getRGBColorForSpan } from '../../TraceTimelineViewer/VirtualizedTraceView';
 
 const DEFAULT_HEIGHT = 60;
 const TIMELINE_TICK_INTERVAL = 4;
@@ -36,6 +37,8 @@ type SpanItem = {
   valueOffset: number;
   valueWidth: number;
   serviceName: string;
+  cacheKey: string,
+  rgbColor: [number, number, number];
 };
 
 function getItem(span: Span): SpanItem {
@@ -43,6 +46,8 @@ function getItem(span: Span): SpanItem {
     valueOffset: span.relativeStartTime,
     valueWidth: span.duration,
     serviceName: span.process.serviceName,
+    cacheKey: getColorKey(span),
+    rgbColor: getRGBColorForSpan(span),
   };
 }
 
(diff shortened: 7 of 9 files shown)


nikclayton-dfinity:
First, here's what Jaeger does with my service without this PR. In this example the trace spans two different process that are part of the same service, and they are gossiping information between them.

In this set of screenshots (without this PR) Jaeger only looks at the service name, and doesn't distinguish between the different instances involved.

Screenshot 2021-04-02 at 11 24 30
Screenshot 2021-04-02 at 11 24 50
Screenshot 2021-04-02 at 11 25 02

nikclayton-dfinity:
This is the same thing, but with this PR.

Screenshot 2021-04-02 at 11 25 15
Screenshot 2021-04-02 at 11 25 20
Screenshot 2021-04-02 at 11 25 28

nikclayton-dfinity:
The service is `replica`, and each instance of that service is reporting a different `service.instance.id` tag value (they don't have a `service.namespace`).

These fields are from the OpenTelemetry resource semantic conventions at https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/resource/semantic_conventions#service.

nikclayton-dfinity:
The code's the simplest thing to prove the concept, but isn't production ready -- some code is duplicated, some information is passed lower down the call tree than I'd like, and it can probably be made more general.

I think the best solution is probably to:

1. Introduce a new class, `Service` (maybe in `types/traces.tsx`) that can be initialised from a `Span` (`Service` is a very generic name, perhaps `SpanService` or something else would be better?)
2. Add methods to this type to return the `name`, `namespace`, and `instance.id` tags
3. Add a `display` method to this type that returns `${name}:${namespace}:${instance_id}`
4. Add `color_hex` and `color_rgb` methods to this type that returns the correct color to use
5. Pass an instance of this new object everywhere the `serviceName` string is currently expected.

Another way to achieve a similar thing would be to do the above, but instead of a `Service` class, convert the existing `Process` type in to a class with those methods -- it already holds the `serviceName` field so this would be a natural extension of that. And then pass that down.

Or do that, but on the existing `Span` type (convert it to a class, add methods, etc).

Or something else? I'm not very familiar with the existing design decisions in the code, so I'm happy to take whatever approach the maintainers think will best fit.

Optional other cleanup work, either in this PR, or a followup:

I had to spend a bit of time dealing with the different representations of color -- sometimes it's a hex string, sometimes it's an RGB triple. A Color type with `as_hex` and `as_rgb` methods might be useful.

The type `CanvasSpanGraphProps` as an `items` field which duplicates the type definition of `SpanItem`. It wasn't clear why it doesn't use `SpanItem` directly.

yurishkuro (approved):
I'm interested in the user flow changes. You seem to be making color decisions by default, based on a specific tag that may not be applicable to every user. I think this behavior needs to be customizable, not default. Customization can be done via site configuration or vis user preferences, which is more appropriate here?

yurishkuro:
cc @albertteoh

nikclayton-dfinity:
> You seem to be making color decisions by default, based on a specific tag that may not be applicable to every user.

Sort of, but in exactly the same way that Jaeger currently does it.

Conceptually, Jaeger takes all the spans from a trace, and puts each span in to a group. The group gets a colour that Jaeger determines.

Before this PR, Jaeger puts spans in to groups based on the span's service name.

At it's heart, all this PR does is change the logic so that instead of `${serviceName}` being the string that's used to determine the group, `${serviceName}:${serviceNamespace}:${serviceInstanceId}` is the string that's used to determine the group.

So if spans don't have a namespace and instance ID then the behaviour is unchanged (you can see that in the first set of screenshots). Everything goes in to one group.

If the spans do have a namespace and/or instance ID then you get more groups.

But the mechanism that Jaeger uses to assign a colour to a group is *not* changed by this PR.

yurishkuro:
> Sort of, but in exactly the same way that Jaeger currently does it.

I understand that, but the main difference is that the `service.name` is a fundamental property in Jaeger data model, you can't even use Jaeger SDKs without providing a service name, so treating is specially in the UI it justified. The properties you're considering are not standard, not produces by many instrumentations. On the other hand, the actual coloring mechanism is generic, not dependent on the particular span attributes, so hardcoding the exact list of its attributes for color coding doesn't sit well with me. At minimum it should be site-configurable, & ideally it should be use-controlled (e.g. what if I do have those attributes but I don't want to see different colors for the same service?)

albertteoh (commented):
> what if I do have those attributes but I don't want to see different colors for the same service?

I think this is a good point in that choosing the attributes used to determine span colours is technically changing existing behaviour.

Moreover, a configurable span colouring scheme could even open up interesting use cases that allow users to determine their own colouring scheme based on other span attributes (e.g. by deployment region instead of instance), that may even be completely independent of service names.

packages/jaeger-ui/src/model/transform-trace-data.tsx
```diff
@@ -130,8 +140,10 @@ export default function transformTraceData(data: TraceData & { spans: SpanData[]
     if (!span) {
       return;
     }
-    const { serviceName } = span.process;
-    svcCounts[serviceName] = (svcCounts[serviceName] || 0) + 1;
+    // Service instance is uniquely identified by name + namespace + instance id
+    const service = getServiceFromSpan(span);
```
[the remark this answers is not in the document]
nikclayton-dfinity: Yes, I agree.

nikclayton-dfinity:
> The properties you're considering are not standard, nor produced by many instrumentations. 

That's true -- they're an experimental recommendation from a sibling CNCF project.

I'm coming from the perspective that I do think it's reasonable to expect (and as a user trying to use these projects in production) that different CNCF projects should, where practical, out-of-the-box, integrate cohesively with one another.

So it would be reasonable for Jaeger to -- with no additional user configuration required -- recognise and respond to OpenTelemetry semantic conventions on spans.

Whereas if there were other conventions recommended by non-CNCF projects, it would be reasonable for the user to have to explicitly configure Jaeger to support those conventions.

Again, with the caveat that these OpenTelemetry semantic conventions are currently marked as experimental.

> On the other hand, the actual coloring mechanism is generic, not dependent on the particular span attributes, so hardcoding the exact list of its attributes for color coding doesn't sit well with me.

I don't think this is true. As you note earlier, Jaeger already hardcodes the exact list of attributes used for color coding (specifically, it only uses one, the service name). This change (in whatever its final form looks like) just extends that list.

> At minimum it should be site-configurable, & ideally it should be use-controlled (e.g. what if I do have those attributes but I don't want to see different colors for the same service?)

and @albertteoh wrote:

> Moreover, a configurable span colouring scheme could even open up interesting use cases that allow users to determine their own colouring scheme based on other span attributes (e.g. by deployment region instead of instance), that may even be completely independent of service names.

There are a few open issues that touch on this:

- https://github.com/jaegertracing/jaeger-ui/issues/582
- https://github.com/jaegertracing/jaeger-ui/issues/121

which have been open for some time (almost 3.5 years in one case) with no code that implements either of them being written.

It's also not something I've got the time to tackle at the moment.

OTOH, my PR, when cleaned up, would:

- Fix this particular use case
- Put the code for determining the grouping in a single place, making subsequent work for #582 or #121 more tractable (the code is probably not the difficult bit there, deciding how the user will configure it is)

tiffon:
It's a really interesting idea to have more expressive heuristics around span coloring. I've always thought using service name or anything with similar cardinality is a losing battle.

Great point that this hasn't changed since pre v1.0 (IIRC). 

OTOH, we now have the ability to specify the UI configuration via JaveScript, which makes it a ton easier to empower interesting configurations.

I agree with Yuri's points, and making it configurable and unchanged unless configuration indicates otherwise seems like a good idea, keeping the current behavior and opting in for the enhancement, for now at least, seems practical.

It's be nice to have it user controlled, as in per user, but that's _a lot more work_ than say making it an element of the ui configuration, which can now be a JS file, and is therefore low hanging fruit for this type of thing. 🥝 

A simple field on JS configurations like the following might work:

```tsx
getSpanViewModifiers(...): { color?: number | string, colorId?: string } | void
```

The thinking for the fields is: `color` is either a number which is the actual color or a valid color according to CSS styling spec, eg `"#c00"` etc.

The color ID would just be any value and would have a color associated with the ID in a consistent manner, based on internal determinations of what to color a given ID, ie akin to a `colorHash(...)` so to speak.

Alternatively, and more conservatively, the config could specify a pattern that would determine the color ID for the span. E.g.

```
#{serviceName}|#{span.operation}
```

Or

```
#{span.tags.component}
```

```
#{process.tags.serviceInstanceId}
```

The function approach has the benefit of being able to do things like color based on error or potentially even duration. It has more risk re expanding api contract, though, and we would probably want to pass a simple view onto a span not the span itself if we were to go this route.

Another thing, with these approaches we could create presets, so to speak, so the JS config could have minimal work required to fit your use case of the OTEL semantics, given the version, eg

```tsx
{
  // etc...
  viewModifiers: { span: ['preset:otel-v0.1.2', 'process'] }
}
```

Anyway, a few alternative ideas — what do you think?

tiffon:
Now that I've re-read this, I can see the PR is also proposing changes to various aggregations, like the service tags in the search results. That makes sense in that how can you have one aggregated data point that has more than one visual representation. OTOH, it makes the stakes a lot higher. 

I still recommend leveraging JS config for custom stuff and also proving presets for common scenarios like OTEL, istio,etc. But, if this is to change aggregations in search results it should change all of them, consistently, and therefore the trace DAG, diffs, and DDG, too. 

That's a large scope. Is there a simplification that would get it to 80% of your use case or maybe more without being so sweeping in scope?

nikclayton-dfinity:
@tiffon 

> OTOH, we now have the ability to specify the UI configuration via JaveScript,

Did you actually mean JavaScript there, or just JSON?

> That's a large scope. Is there a simplification that would get it to 80% of your use case or maybe more without being so sweeping in scope?

Probably not. In many ways the colour change is a bit of a red herring (no pun intended). The actual change is how Jaeger groups spans within a trace. The colouring changes naturally fall out of that.

So, here are the two options I think:

## 1: Allow the user to specify the span grouping through the config

```
{
  // etc...
  spanGroupKey: { fields: [...] }
}
```

?

The current behaviour would be:

```
{
  // etc...
  spanGroupKey: { fields: ['service.name'] }
}
```

The otel behaviour would be:

```
{
  // etc...
  spanGroupKey: { fields: ['service.name', 'process.service.namespace', 'process.service.instance.id'] }
}
```

Pro:
- Most general approach
- Resolves two other open issues

Con:
- More code, unrelated to the specific problem I'm trying to solve
- Doesn't do the right thing out of the box.

## 2: Don't make this user configurable yet, but do make grouping decisions based on other properties of the span.

For example, an otel span **must** have an `otel.library.name` field. So if that field exists then Jaeger could automatically group by the name/namespace/id triple.

Pro:
- Simpler to implement
- Solves my specific problem
- Does the right thing for OpenTelemetry spans with no additional user configuration

Con:
- The two other open issues stay open

nikclayton-dfinity:
Requested re-review of the most recent comments on the options for proceeding.

yurishkuro:
I would vote for a version of option 1. There are two main changes required by your use case:
  1. the ability to provide a lambda for determining the color of a node instead of hardcoding it to `service.name`. This sounds like the main chunk of work. I am not sure if it's possible to have a single API for such lambda, i.e. what would be the argument to it, the Span object, or something else?
  2. the ability to configure the lambda in JSON or JS configuration.

I don't agree with your assessment that Option 1 "Doesn't do the right thing out of the box" - it does, because the "right thing" (while subjective) is to be backwards compatible with the existing UI behavior by default, not to change it. People need to opt-in into the new behavior.

Regarding configuration, your proposal `spanGroupKey: { fields: ['service.name'] }` LGTM, however, as @tiffon mentioned, I would start simpler and provide just a few pre-defined configurations like `service-name` and `otel-service-instance` instead of building a fully dynamic config that must parse field names from the array.

nikclayton-dfinity:
PTAL.

No documentation or test updates yet, I'll tackle those when there's broad approval for this approach.

To try it out, edit `default-config.tsx` and add a `spanGroupKey` property, either:

```
spanGroupKey: { preset: 'otel-ef4612d' },
```

or

```
spanGroupKey: { tags: ['serviceName', 'service.namespace', 'service.instance.id'] },
```

(those two are functionally the same thing) and the UI will group and colour spans according to those fields, in that order. Obviously, you'll need some traces with spans that set those fields. If you don't then adjust the value for `tags` as necessary.

BTW, using `yarn start` is there a straightforward way to pass in a path to a config file to load? I couldn't see any after a cursory check, and couldn't find anything mentioned in the documentation.

Some things I'm still not overly happy about, and would appreciate suggestions for:

- The property name `spanGroupKey`. It's functional, but maybe there's a better name.

- I add a `group` property to the `Span` type, which is a very generic name. Maybe there's something less generic that would be a better fit?

nikclayton-dfinity:
Ping @yurishkuro @tiffon @albertteoh

packages/jaeger-ui/src/components/TracePage/TracePageHeader/SpanGraph/CanvasSpanGraph.tsx
```diff
@@ -21,7 +21,14 @@ import { TNil } from '../../../../types';
...
 
 type CanvasSpanGraphProps = {
-  items: { valueWidth: number; valueOffset: number; serviceName: string }[];
+  // TODO: Unclear why this is not the `SpanItem` type
+  items: {
+    valueWidth: number;
+    valueOffset: number;
+    serviceName: string;
+    spanGroup: string;
+    rgbColor: [number, number, number];
```
yurishkuro, line 30: I am not sure I follow this:
  * why do we need both spanGroup and color?
  * shouldn't color be always calculated rather than captured in the item?

nikclayton-dfinity: `spanGroup` is the key in `fillCache` that `renderIntoCanvas` uses, `rgbColor` is the color to use.

This could go either way, I think. The previous code didn't pass in the colour, instead, it passed in a function to get the colour (`getColor`, removed here from line 53).

To me, that felt a bit weird -- the caller of this code knows both the function to call, and the value of the parameter it will be called with. So it made sense to move that responsibility in to `getItem` in `SpanGraph/index.tsx`.

I actually think the best way to do this is to make the color a property on the `Span` type, populate it when a span is created, and have `getItem` fetch the color the same way it fetches the other `Span` properties.

I can move it back if you prefer though.

When I was first navigating this code I found the indirection through the `getItem` function confusing. For example, I don't understand the design tradeoff between making it a function vs. an `asSpanItem()` method on the `Span` type.

And (as noted in the comment in the code here) why `items` explicitly lists the fields it accepts, instead of just being `items: SpanItem[]`.

Or (and this would also be a larger refactoring) why `renderIntoCanvas` takes an array of `SpanItem` and not an array of `Span`. The `SpanItem` type is a strict subset of the `Span` type, so this seems like an unnecessary additional level of indirection (unless maybe it makes tests easier...? I haven't checked there yet).

yurishkuro: I don't know the answers to all "why"s, but some general design considerations I could offer are:
  * Color is not a property of the span, it is a function of the span attributes and a given display mode. Span contains the data, that same data could be displayed differently in different views or based on different use selections.
  * Different components / parts of the code declare "span" only as wide as they need it to be. The type declaration defines the contract that the code expects the data to satisfy. This decouples different components from depending on each other or on the shared data model.

packages/jaeger-ui/src/components/TracePage/TracePageHeader/SpanGraph/render-into-canvas.tsx
```diff
@@ -50,11 +50,10 @@ export default function renderIntoCanvas(
     if (width < MIN_ITEM_WIDTH) {
       width = MIN_ITEM_WIDTH;
     }
-    let fillStyle = fillCache.get(serviceName);
+    let fillStyle = fillCache.get(items[i].spanGroup);
+    const color = items[i].rgbColor;
```
yurishkuro, line 54: move inside if() to reduce scope

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
```diff
@@ -150,6 +150,14 @@ function getCssClasses(currentViewRange: [number, number]) {
   });
 }
 
+export function getHexColorForSpan(span: Span) {
```
yurishkuro, line 153: a) what's the difference between hex color and rgb color? Can we use just one of these?
b) can we find a better place for this/these function(s)? components should not depend on each other, but you're importing from this module into other components

packages/jaeger-ui/src/model/span.tsx
```diff
@@ -24,3 +25,89 @@ export function getParent(span: Span) {
...
+/** Shape of data required to determine a span's group. */
+type SpanGroupData = {
+  process: {
+    serviceName: string;
+    tags: Array<KeyValuePair>;
+  };
+  tags: Array<KeyValuePair>;
+};
+
+/**
```
yurishkuro, line 38: love the docs!

packages/jaeger-ui/src/model/span.tsx
```diff
@@ -24,3 +25,89 @@ export function getParent(span: Span) {
...
+ *
+ * - `tags` (string[]) - set the group according to the values of the tags
+ *   listed in this array. Process tags are searched first, followed by span
+ *   tags. The special value `serviceName` corresponds to the service's name.
+ *
+ * If `preset` is specified then `tags` is ignored.
+ *
+ * # Available presets
+ *
+ * - {@link https://github.com/open-telemetry/opentelemetry-specification/blob/ef4612dd0b1f4266e6eca7177290ee2a5f087199/specification/resource/semantic_conventions/README.md|otel-ef4612d},
```
yurishkuro, line 56: a) for better readability of the doc, I would move the link to details and explicitly list the actual constant first, e.g.
```
 * - "otel-xxx" - Equivalent to a combination of tags: ['serviceName', 'service.namespace', 'service.instance.id]`.
 *    @see [OTEL semantic conventions](...link...)
b) `otel-ef4612d` is a confusing present name, and not descriptive of the preset itself, I suggest `otel-service-instance`
c) it would be better to define the default as a preset too, e.g. `service-name`

File: packages/jaeger-ui/src/model/span.tsx [hunk 7]
```diff
@@ -24,3 +25,89 @@ export function getParent(span: Span) {
...
+ *   ef4612d. Equivalent to setting
+ *   `tags: ['serviceName', 'service.namespace', 'service.instance.id]`.
+ *
+ * # Default
+ *
+ * If no configuration value is provided the default is to use `serviceName`.
+ *
+ * @param {Span} span The span to determine the group for
+ */
+export function getSpanGroupFromSpan(span: SpanGroupData) {
```
yurishkuro, line 67: Will memoize work for this function? I am concerned with performance impact of constant config parsing and recalculation. Ideally we should "pre-compile" the configuration object into a lambda that can be evaluated against a span without any additional parsing.

packages/jaeger-ui/src/model/trace-dag/DenseTrace.tsx
```diff
@@ -22,6 +22,7 @@ function convSpans(spans: Span[]) {
   const ids: string[] = [];
   spans.forEach(span => {
     const { spanID: id, operationName: operation, process, references, tags: spanTags } = span;
+    const spanGroup = span.group;
```
yurishkuro, line 25: used only once, unnecessary const, makes it harder to grok the actual use of group

packages/jaeger-ui/src/model/trace-dag/DenseTrace.tsx
```diff
@@ -46,6 +47,7 @@ function convSpans(spans: Span[]) {
       operation,
       parentID,
       service,
+      spanGroup,
```
yurishkuro, line 50:
```suggestion
      span.group,
```

packages/jaeger-ui/src/types/trace.tsx
```diff
@@ -67,6 +67,7 @@ export type Span = SpanData & {
   references: NonNullable<SpanData['references']>;
   warnings: NonNullable<SpanData['warnings']>;
   subsidiarilyReferencedBy: Array<SpanReference>;
+  group: string;
```
yurishkuro, line 70: I would prefer to call this `colorGroup` to make the meaning explicit

packages/jaeger-ui/src/utils/color-generator.tsx
```diff
@@ -74,6 +74,8 @@ export class ColorGenerator {
    * If the key has been used already, it will
    * use the same color.
    */
+  // HexColor might make sense as its own type. Or a Color type with rgb,rgba,
+  // hex() methods
   getColorByKey(key: string) {
```
yurishkuro, line 79:
```suggestion
  getColorByKey(key: string): string {
```

packages/jaeger-ui/src/model/span.tsx
```diff
@@ -24,3 +25,89 @@ export function getParent(span: Span) {
...
+  let groupKeyFields = [];
```
yurishkuro, line 68:
```suggestion
  let groupKeyFields: string[] = [];
```

packages/jaeger-ui/src/model/span.tsx
```diff
@@ -24,3 +25,89 @@ export function getParent(span: Span) {
...
+  const preset = getConfigValue('spanGroupKey.preset');
+  if (preset !== undefined) {
+    switch (preset) {
+      case 'otel-ef4612d':
+        groupKeyFields = ['serviceName', 'service.namespace', 'service.instance.id'];
+        break;
+      default:
+        // Give the user a chance to discover the problem
+        // eslint-disable-next-line no-console
+        console.warn(`Unknown value '${preset}' for 'spanGroupKey.preset' config option`);
```
yurishkuro, line 79: This would be producing a log for every span, instead of just once. However, if we introduce memoization of the config -> lambda parsing then it would only log once

yurishkuro (commented):
@albertteoh do you have someone who can review this?

albertteoh:
> @albertteoh do you have someone who can review this?

I've reached out to my team to see if someone can help take a look.

albertteoh:
> I've reached out to my team to see if someone can help take a look.

@Danafrid, @th3M1ke or @yoave23 please take a look when you get a chance.

PR #723, opened 2021-04-02
Update jest to produce a junit report for tests

kevinearls:
Signed-off-by: Kevin Earls <kearls@redhat.com>

## Which problem is this PR solving?
- It would be useful to be able to produce a JUnit type report in CI when tests are run.

## Short description of the changes
- Add a jest-junit reporter to packages/jaeger-ui/package.json to product a junit.xml report whenever tests are run.

PR #726, opened 2021-04-06
[chore]: gitignore the generated unit test report

meenal06:
Signed-off-by: Meenal Trivedi <meenaltrivedi6102@gmail.com>

- Since the PR #723, there is a junit report generated each time with pre-commit hooks, this PR adds them into the gitignore.


yurishkuro (approved):
good call

PR #730, opened 2021-04-08
[feat]: add upload widget into the search form tab 

### LINKED ISSUE #683: Improve Search box UI and UX
Improving the usability of the search in Jaeger, by unifying between the two tabs - saving the user an extra click and updating the UI.

Currently, there are several issues in the Search box UI:
- The search button is not prominent enough
- In order to upload a JSON file, the user has to navigate to a different tab inside the search
- The search fields can be grouped better (see min duration, max duration)
- Ui-wise the search can look lighter and not take as much vertical space, which is helpful when there are many results on the screen

See attached proposal for the redesign - unifying between the search and upload functionalities, whilst keeping the drag and drop functionality - all in the same view. Also, a cleaner UI and a more prominent search button.

**Before:**
Screen Shot 2021-01-24 at 17 05 06

**Suggested:**

https://user-images.githubusercontent.com/75564311/105634579-4e1b8180-5e67-11eb-8284-392ebaea6420.mov

Would appreciate your thoughts on the proposal.

meenal06:
- fixes #683
- reference https://github.com/jaegertracing/jaeger-ui/issues/683#issuecomment-814222170

- Combined upload json and search form page into a single tab

## Screenshot
Screenshot 2021-04-09 at 1 18 49 AM

### DIFF
--- a/packages/jaeger-ui/src/components/SearchTracePage/FileLoader.css
+++ b/packages/jaeger-ui/src/components/SearchTracePage/FileLoader.css
@@ -14,37 +14,12 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-.SearchTracePage--row {
-  display: flex;
-  flex-grow: 1;
+.FileLoader--upload {
+  background-color: #11939a;
+  color: white;
+  float: right;
 }
 
-.SearchTracePage--column {
-  display: flex;
-  flex-direction: column;
-  flex-grow: 1;
-  padding: 1rem 0.5rem;
-}
-
-.SearchTracePage--column:first-child {
-  padding-left: 1rem;
-}
-
-.SearchTracePage--column:last-child {
-  padding-right: 1rem;
-}
-
-.SearchTracePage--fileload {
-  height: 8rem;
-}
-
-.SearchTracePage--find {
-  background-color: #f5f5f5;
-  border: 1px solid #e6e6e6;
-  padding: 1rem;
-}
-
-.SearchTracePage--logo {
-  display: block;
-  margin: 13rem auto 0 auto;
+.FileLoader--helper-text {
+  padding-top: 1rem;
 }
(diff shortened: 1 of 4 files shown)


yurishkuro:
I am not sure this is an improvement. Uploading a file to the UI is not the primary use case, and we're sacrificing a bunch of primary top-level real estate to it. 

I think the main issue raised in #683 has been fixed, but changing upload from a separate tab to something else is not a priority imo.

yurishkuro:
I am curious to see what others think. The second tab never bothered me, and #683 was just one person's opinion (plus I think it was more on the side of compacting the search form, which was done, the upload change was mixed in but really is a separate ask).

jpkrohling:
>Uploading a file to the UI is not the primary use case, and we're sacrificing a bunch of primary top-level real estate to it.

Agree with this.

rubenvp8510:
> I am not sure this is an improvement. Uploading a file to the UI is not the primary use case, and we're sacrificing a bunch of primary top-level real estate to it.

I agree, this makes sense,  +1
> 
> I think the main issue raised in #683 has been fixed, but changing upload from a separate tab to something else is not a priority imo.

In  #683  search form compaction is an improvement, but I think we leave the upload in a separate tab, is something that I don't think is a big deal.

meenal06:
So should I close this PR? Also, can this be added as an outreachy contribution or not :confused: ?

yurishkuro:
Yes, let's close the PR.

@jpkrohling might be able to answer re outreachy qualification (I'm +1 to count it in).

jpkrohling:
Sorry for missing this: of course it counts! The goal is not to get the most PRs merged but contribute to the project. Sometimes, it's getting a PR merged, other times it's bringing clarity in some way, like what happened here.

meenal06:
Thanks @jpkrohling !

PR #734, opened 2021-04-13
[fix] Add null check for span.logs in search/filter-spans

achesin:
- Hello! This is a small PR for a bug I found in on the Trace Page. In the Trace Page Search Bar when you type in something that does not correspond with a tag, the page crashes and goes all white. That is because there was nothing checking to make sure span.logs was populated.

- I added a check to make sure span.logs is not equal to null and a corresponding test.


yurishkuro (approved):
thanks

PR #759, opened 2021-05-26
Add optional tag actions

### LINKED ISSUE #729: improve the UX and discoverability of the link patterns
provide help for creating new link patterns for tags, services, and logs

Currently, there's no indication for users on how to create link patterns in the app

We'd like to extend the UI in a way that will allow directing users to a reference page (either the jaeger-ui config manual or any other page that will be defined). We'd also like to be able to use the link as a javascript function invoker if we'd like to make a call to action like open a custom modal using the dynamic UI config that was recently added (https://github.com/jaegertracing/jaeger-ui/pull/677)

https://user-images.githubusercontent.com/3125539/114016220-49197c00-9873-11eb-9bd4-78f04550d41c.mov

## Any open questions to address

nope

yoave23:
- Resolves #729 
-  Add the ability to configure custom actions for tags, services and processes by adding a function to the UI-config
- Add a configurable link to the [link patterns config section](https://www.jaegertracing.io/docs/1.22/frontend-ui/#link-patterns)

https://user-images.githubusercontent.com/3125539/119652907-5bee0d00-be2f-11eb-9010-09b5d9b037c7.mov

config used in the example:
```
"tagsActions": [{
          title: 'this link was configured in the ui-config',
          action: function(tag, span) {
            const { key, value } = tag;
            const url = `/search?service=${span.process.serviceName}&tags={"${key}":"${value}"}`;

            window.open(url);
          }
        },{
          title: 'display an alert',
          icon: 'setting',
          action: function(tag, span) {
            alert(JSON.stringify(span))
          }
        }]
```


yurishkuro:
Please make sure all commits are signed per contributing guide.

I like the functionality, except for the legend:
image
I think the legend is misplaced since it implies that the end user of the UI can actually go and configure the links, which is not the case in a typical enterprise setup.

I think this might also need a future extension to allow setting some filters on the individual tag actions, to be able to control which tags they should apply to. In fact, could we not introduce a new top-level section to the config but instead integrate this into the existing LinkPatterns configuration, which already supports selection of tags? We could extend it to make the selection optional or "*"

yoave23:
@yurishkuro sorry for getting back to this PR so late
> I like the functionality, except for the legend:
I think the legend is misplaced since it implies that the end user of the UI can actually go and configure the links, which is not the case in a typical enterprise setup.

this is a link to the [link-patterns docs](https://www.jaegertracing.io/docs/latest/frontend-ui/#link-patterns) which can be overridden in the config

> I think this might also need a future extension to allow setting some filters on the individual tag actions, to be able to control which tags they should apply to. In fact, could we avoid introducing a new top-level section to the config but instead integrate this into the existing LinkPatterns configuration, which already supports selection of tags? We could extend it to make the selection optional or "*"

we can avoid another top-level section in the config and use a "select all" option such as "*". 
please note that we only display those links when a user hovers the tag as opposed to other links which are always displayed.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/index.tsx
```diff
@@ -150,6 +158,12 @@ export default function SpanDetail(props: SpanDetailProps) {
               focusSpan={focusSpan}
             />
           )}
+        <div>
+          <a href={linkPatternsUrl} target="_blank" rel="noopener noreferrer">
+            <Icon type="info-circle-o" className="SpanDetail--docsIcon" />
+            Learn how to configure links
```
jpkrohling, line 164: Is this being shown within every span?

jpkrohling:
@esnible, what do you think about this from a UX perspective?

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/index.tsx
yoave23: it is, any other idea would be appreciated :)
we can also remove that part from this PR and do it in a separate one

jpkrohling: I think it would be too much to have this as part of every single span in a trace. We might have hundreds/thousands of them...

yoave23: ok, we'll try to find a better location for it or just remove it from this PR

yurishkuro: As I already wrote earlier, this information shouldn't be in the UI because the end user of the UI cannot change the configuration, only the operator can.

yurishkuro:
@yoave23 are you still interested in finishing this PR?

PR #764, opened 2021-06-04
Preparing release v1.14.0

jpkrohling:
Signed-off-by: Juraci Paixão Kröhling <juraci@kroehling.de>

PR #765, opened 2021-06-04
Fix jaeger-ui package version

jpkrohling:
Follow up from #764, which missed this version bump.

﻿Signed-off-by: Juraci Paixão Kröhling <juraci@kroehling.de>


yurishkuro:
was it oversight or missing instruction?

jpkrohling:
Oversight. Perhaps the order of the items in the instructions could be better, but I should have re-read after creating the original PR.

PR #768, opened 2021-06-17
fix deep dependency graph not show in uiEmbed=v0 mode

### LINKED ISSUE #719: In uiEmbed=v0, deep dependency graph, the result is blank
When I use uiEmbed=v0, traces has a result, but deep dependency graph has no result. If I remove uiEmbed=v0, deep dependency graph will display the result
image
image
image

leroy-chen:
fix deep dependency graph not show in uiEmbed=v0 mode

-  Resolves #719 

- also add Page--content class to Content in uiEmbed=v0 mode


packages/jaeger-ui/src/components/App/Page.tsx
```diff
@@ -52,7 +52,7 @@ export class PageImpl extends React.Component<TProps> {
 
   render() {
     const { embedded } = this.props;
-    const contentCls = cx({ 'Page--content': !embedded });
+    const contentCls = cx({ 'Page--content': true });
```
yurishkuro, line 55: I do not have enough context to understand the impact of this change. It looks like there was an explicit decision NOT to show this in the embed mode, there must have been a reason for that.

jpkrohling: cc @rubenvp8510

leroy-chen: when i directly set the classname to Page-content, it'll show an nav height blank on top of window, and when scroll the page,the layout displayed messy ,so i add an new classname copyed from 'Page--content' ,just delete the "top: var(--nav-height);" property,it seem's ok.

yurishkuro:
@rubenvp8510 could you review?

packages/jaeger-ui/src/components/App/Page.css
```diff
@@ -31,3 +31,12 @@ limitations under the License.
   right: 0;
   top: var(--nav-height);
 }
+
+.Page--content--embedded {
+  display: flex;
```
rubenvp8510, line 36: The rules of` .Page--content--embedded` and .`Page--content`  are the same, except for   `top: var(--nav-height);`

We preserve the class  `.Page--content` , removing the last rule and create a new one , something like this 

```
.Page--content--no-embedded {
     top: var(--nav-height);
}
```

Then we can use those two classes in the component like this:

`cx({ 'Page--content': true, 'Page--content--no-embedded': !embedded });`

Whad do you think?

rubenvp8510 (commented):
The CI is failing, run `yarn prettier` to fix it :)

packages/jaeger-ui/src/components/App/Page.css
leroy-chen: fixed, @rubenvp8510 PTAL

PR #782, opened 2021-07-07
[fix]: deep dependency header overlaps page header (#731)

### LINKED ISSUE #731: [bug] Navbar overlapping and css overflow in deep dependency section
## Current Behaviour

The deep dependency graph section overlaps and cutout the navigation bar when scrolled

## Expected Behaviour

Remove the css overflow and intersection of nav-bar

## Screenshot
Screenshot 2021-04-09 at 2 03 33 AM

stardotcode:
- Resolves #731

- Bump up the `z-index` of page header element so that the deep dependency header doesn't overlap when scrolling down.

### DIFF
--- a/packages/jaeger-ui/src/components/App/Page.css
+++ b/packages/jaeger-ui/src/components/App/Page.css
@@ -19,7 +19,7 @@ limitations under the License.
   padding: 0;
   position: fixed;
   width: 100%;
-  z-index: 3;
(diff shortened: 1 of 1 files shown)


rubenvp8510 (approved):

albertteoh (approved):
Thanks @stardotcode! I tried it out and looks like this has fixed the problem.

PR #817, opened 2021-09-23
Enable regexes and functions in link patterns

yoave23:
Unblocked by JSON config support from #123 

Allow matching link patterns with regular expressions or functions

example with regex, this will apply the pattern all tags:
```
{
  type: 'tags',
  text: 'hello world',
  key: /./,
  url: 'https://github.com/jaegertracing',
}
```
example for function config:
```
{
  type: 'tags',
  text: 'hello world',
  key: (key: string) => key.startsWith('comp'),
  url: 'https://github.com/jaegertracing',
}
```


yurishkuro (approved):
Please add documentation somewhere.

FWIW, I see that the README has a `UI Configuration` section that simply points to the website, I wonder if it would've been better to do the opposite, i.e. keep docs in this repo (e.g. plexus plugin has an extensive README, as well as tracking API) and point the docs here.

yoave23:
> Please add documentation somewhere.

can it wait till I submit a PR for #729? should be ready by the end of the week

PR #867, opened 2022-01-14
Widen date input to avoid clipping, fixes #864

### LINKED ISSUE #864: Date input is clipped in custom time selection
**Describe the bug**

Date input is not fully visible (Macbook Pro 13", maximized window):

image

**To Reproduce**

1. Set date format to ISO8601:

image

2. Open Jaeger UI
3. Pick custom time range 

**Expected behavior**

Date can be entered without issues.

 - Jaeger version: 1.29, jaeger-ui build from the tip

**Additional context**

I think we can use `input` with `type="datetime-local"` rather than separate inputs for date and time. This also gives a nice picker, at least in Chrome:

image

I'm happy to open a PR if you agree.

bobrik:

Resolves #864.

Before:

Lookback

After:

Lookback

### DIFF
--- a/packages/jaeger-ui/src/components/SearchTracePage/SearchForm.js
+++ b/packages/jaeger-ui/src/components/SearchTracePage/SearchForm.js
@@ -378,7 +378,7 @@ export class SearchFormImpl extends React.PureComponent {
             }
           >
             <Row gutter={16}>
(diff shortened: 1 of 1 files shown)


yurishkuro (approved):
Thank you!

Issue #871, opened 2022-01-25
Monitor Tab: Numbers in the Legend


nofar9792:
**Describe the bug**
Cannot see the whole number in the legend when the number is greather than 1000

**Expected behavior**
See the while number

**Screenshots**
![Screen Shot 2022-01-25 at 16 01 10](https://user-images.githubusercontent.com/11076023/151014788-0d3b465b-1e15-4da1-ba1c-bad5fbc8a085.png)

and i would like to fix it

esnible:
I don't recall seeing this view in Jaeger UI.  How do I navigate to it?

yurishkuro:
@esnible I assume it's the Monitoring tab introduced here #2954. I don't think the documentation part of this was finished.

PR #873, opened 2022-01-26
Fix Bug: Cannot see the whole numbers in the legend

nofar9792:
Signed-off-by: nofar9792 <nofar.cohen@logz.io>

Resolves https://github.com/jaegertracing/jaeger-ui/issues/871 

Added some margin-left when the number is high

### DIFF
--- a/packages/jaeger-ui/src/components/Monitor/ServicesView/serviceGraph.tsx
+++ b/packages/jaeger-ui/src/components/Monitor/ServicesView/serviceGraph.tsx
@@ -150,52 +150,68 @@ export class ServiceGraphImpl extends React.PureComponent<TProps> {
     const noDataComponent = this.generatePlaceholder('No Data');
     const apiErrorComponent = this.generatePlaceholder('Couldn’t fetch data');
 
-    const Plot = (
-      <XYPlot
-        margin={{ bottom: 25 }}
-        onMouseLeave={() => this.setState({ crosshairValues: [] })}
-        width={width}
-        height={this.height - 74}
-        yDomain={yDomain}
-      >
-        {showHorizontalLines ? <HorizontalGridLines /> : null}
-        <XAxis tickFormat={tickFormat} tickTotal={Math.floor(width / 60)} />
-        <YAxis />
-        {this.renderLines()}
-        <Crosshair values={this.state.crosshairValues}>
-          <div style={{ width: 140 }}>
-            {this.state.crosshairValues[0] &&
-              `${new Date(this.state.crosshairValues[0].x).toLocaleDateString()} ${new Date(
-                this.state.crosshairValues[0].x
-              ).toLocaleTimeString()}`}
-            {this.state.crosshairValues.reverse().map((d: TCrossHairValues) =>
-              showLegend ? (
-                <div key={d.label}>
-                  P{d.label * 100}: {d.y}
-                </div>
-              ) : (
-                <div key={d.label}>{d.y}</div>
-              )
-            )}
-          </div>
-        </Crosshair>
-        {showLegend ? (
-          <DiscreteColorLegend
-            className="legend-label"
-            orientation="horizontal"
-            items={this.getData()
-              .map((d: ServiceMetricsObject, idx: number) => ({
-                color: this.colors[idx],
-                title: `${d.quantile * 100}th`,
-              }))
-              .reverse()}
-          />
-        ) : null}
-      </XYPlot>
-    );
+    const Plot = () => {
+      let maxValue = 0;
+
+      if (metricsData && Array.isArray(metricsData)) {
+        const allMaxMetrics = metricsData.map(x => x.max);
+        maxValue = Math.max(...allMaxMetrics);
+      } else if (metricsData) {
+        maxValue = metricsData.max;
+      }
+
+      const pixelPerNumber = 6;
+      const minNumberNumbersToDisplay = 3;
+      const additionalPixelWidth =
+        (Math.round(maxValue).toString(10).length - minNumberNumbersToDisplay) * pixelPerNumber;
+
+      return (
+        <XYPlot
+          margin={{ bottom: 25, left: 40 + (additionalPixelWidth < 0 ? 0 : additionalPixelWidth) }}
+          onMouseLeave={() => this.setState({ crosshairValues: [] })}
+          width={width}
+          height={this.height - 74}
+          yDomain={yDomain}
+        >
+          {showHorizontalLines ? <HorizontalGridLines /> : null}
+          <XAxis tickFormat={tickFormat} tickTotal={Math.floor(width / 60)} />
+          <YAxis />
+          {this.renderLines()}
+          <Crosshair values={this.state.crosshairValues}>
+            <div style={{ width: 140 }}>
+              {this.state.crosshairValues[0] &&
+                `${new Date(this.state.crosshairValues[0].x).toLocaleDateString()} ${new Date(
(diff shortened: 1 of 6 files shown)


packages/jaeger-ui/src/components/Monitor/ServicesView/serviceGraph.tsx
```diff
@@ -150,52 +150,68 @@ export class ServiceGraphImpl extends React.PureComponent<TProps> {
...
+        const allMaxMetrics = metricsData.map(x => x.max);
+        maxValue = Math.max(...allMaxMetrics);
+      } else if (metricsData) {
+        maxValue = metricsData.max;
+      }
+
+      const pixelPerNumber = 6;
+      const minNumberNumbersToDisplay = 3;
+      const additionalPixelWidth =
+        (Math.round(maxValue).toString(10).length - minNumberNumbersToDisplay) * pixelPerNumber;
```
yurishkuro, line 166: Hm, I wonder, instead of trying to display numbers of very different lengths, wouldn't it be better to change from always displaying milliseconds to using human-readable values like 1.5s, 125ms, which we already do for trace latency in the search results?

nofar9792: @afishler  WDYT?

albertteoh:
@nofar9792 can you please share a screenshot of what the graph and y-axis labels look like after these changes?

nofar9792:
> @nofar9792 can you please share a screenshot of what the graph and y-axis labels look like after these changes?

Thanks Albert, I added 🙏

packages/jaeger-ui/src/components/Monitor/ServicesView/serviceGraph.tsx
nofar9792: @yurishkuro I converted the numbers to another human readable time unit.
Please review my PR again 🙏

nofar9792:
@AlonMiz

nofar9792:
@yurishkuro ?

yurishkuro:
Are the before/after screenshots still accurate? I was expecting to see axis ticks like `1s, 2s`, not to have `Sec` added to the axis title.

nofar9792:
> Are the before/after screenshots still accurate? I was expecting to see axis ticks like `1s, 2s`, not to have `Sec` added to the axis title.

 We think it's redundant to write it in the ticks
@afishler

nofar9792:
> Are the before/after screenshots still accurate? I was expecting to see axis ticks like `1s, 2s`, not to have `Sec` added to the axis title.

 We think it's redundant to write it in the ticks
@afishler

albertteoh:
> We think it's redundant to write it in the ticks

What about when latencies are 1ms or 1hr?

afishler:
> > We think it's redundant to write it in the ticks
> 
> What about when latencies are 1ms or 1hr?

Well, they cannot be different units on the same diagram. Once the units are set they will be the same for the whole diagram.

That's why putting the units in the title should be enough. So if the function sets the units to ms it will be all ms, if it is seconds it will be all seconds
Screenshot at Feb 23 17-57-40

albertteoh:
> That's why putting the units in the title should be enough. So if the function sets the units to ms it will be all ms, if it is seconds it will be all seconds

Ah, I see, so the title's units are dynamic based on the latency data. If that's the case, it makes sense to me.

nofar9792:
> > That's why putting the units in the title should be enough. So if the function sets the units to ms it will be all ms, if it is seconds it will be all seconds
> 
> Ah, I see, so the title's units are dynamic based on the latency data. If that's the case, it makes sense to me.

@yurishkuro @albertteoh so WDYT?

packages/jaeger-ui/src/components/Monitor/ServicesView/index.tsx
```diff
@@ -249,12 +270,13 @@ export class MonitorATMServicesViewImpl extends React.PureComponent<TProps, Stat
                 metrics.serviceError.service_latencies_95
               }
               loading={metrics.loading}
-              name="Latency, ms"
+              name={`Latency, ${convertTimeUnitToShortTerm(displayTimeUnit)}`}
```
albertteoh, line 273: I prefer putting the units in brackets like "Latency (ms)" or "Latency (min)", instead of a comma. 

What do you/others think?

packages/jaeger-ui/src/utils/date.tsx
```diff
@@ -40,6 +40,14 @@ const UNIT_STEPS: { unit: string; microseconds: number; ofPrevious: number }[] =
   { unit: 'μs', microseconds: 1, ofPrevious: 1000 },
 ];
 
+const timeUnitToShortTermMapper = {
+  milliseconds: 'ms',
+  seconds: 'Sec',
+  minutes: 'Min',
+  hours: 'Hrs',
+  days: 'Days',
```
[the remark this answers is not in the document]
albertteoh: It just occurred to me that `min` could unfortunately be confused with "minimum", so it's worth considering the plural form: `ms`, `secs`, `mins`, `hrs`, `days`.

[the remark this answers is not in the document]
yurishkuro: https://physics.nist.gov/cuu/Units/outside.html

packages/jaeger-ui/src/utils/date.tsx
```diff
@@ -40,6 +40,14 @@ const UNIT_STEPS: { unit: string; microseconds: number; ofPrevious: number }[] =
   { unit: 'μs', microseconds: 1, ofPrevious: 1000 },
 ];
 
+const timeUnitToShortTermMapper = {
+  milliseconds: 'ms',
```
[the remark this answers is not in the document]
nofar9792: moment.js doesn't support it and in my cases (the only usages), we don't need microseconds

packages/jaeger-ui/src/utils/date.tsx
```diff
@@ -145,25 +153,46 @@ export function formatRelativeDate(value: any, fullMonthName: boolean = false) {
...
+      return durationInTimeUnit >= 1;
+    })!;
+};
+
+export function convertTimeUnitToShortTerm(timeUnit: string) {
+  const shortTimeUnit = (timeUnitToShortTermMapper as any)[timeUnit];
+
+  if (shortTimeUnit) return shortTimeUnit;
+
+  return '';
```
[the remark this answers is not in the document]
nofar9792: if i don't have```shortTimeUnit```, I don't want to return ```undefined```

packages/jaeger-ui/src/components/Monitor/ServicesView/index.tsx
nofar9792: done

albertteoh (approved):
LGTM overall, thanks! It looks like there are some code coverage issues that need addressing.

albertteoh:
Just noticed, the description's "After" screenshot shows "Latency (s)" and the scale on  the left goes up to a max 390. 

I would have expected the units to be minutes already?

nofar9792:
> Just noticed, the description's "After" screenshot shows "Latency (s)" and the scale on the left goes up to a max 390.
> 
> I would have expected the units to be minutes already?

This data is different

nofar9792:
@albertteoh @yurishkuro What prevents us from merging it

yurishkuro:
@albertteoh can you finish the review for this?

Issue #889, opened 2022-02-23
Monitor Tab - Request rate number more readable


nofar9792:
## Requirement - what kind of business use case are you trying to solve?
Truncate request rate number to two decimal places:
![Screen Shot 2022-01-25 at 16 13 00](https://user-images.githubusercontent.com/11076023/155326369-ef35ebdc-54c6-411b-b52b-73bb2f86eda6.png)

Request rate that is lower than 0.1 can be written as "< 0.1" 
![Screen Shot 2022-01-25 at 16 07 35](https://user-images.githubusercontent.com/11076023/155326463-1ca735c1-4b82-4339-a77a-010ae40be475.png)

## Proposal - what do you suggest to solve the problem or improve the existing situation?
format the request rate value

I would like to do it

PR #890, opened 2022-02-23
Monitor Tab - Request rate number more readable

nofar9792:
 Resolves https://github.com/jaegertracing/jaeger-ui/issues/889

- Formating the request rate value

before:
Screen Shot 2022-01-25 at 16 07 35

after:
image

before:
Screen Shot 2022-01-25 at 16 13 00

after:
Screen Shot 2022-01-25 at 16 10 24

### DIFF
--- a/packages/jaeger-ui/src/components/Monitor/ServicesView/operationDetailsTable/index.tsx
+++ b/packages/jaeger-ui/src/components/Monitor/ServicesView/operationDetailsTable/index.tsx
@@ -34,6 +34,15 @@ type TState = {
   hoveredRowKey: number;
 };
 
+function formatValue(value: number) {
+  if (value < 0.1) {
+    return `< 0.1 req/s`;
+  }
+
+  // Truncate number to two decimal places
+  return `${value.toString().match(/^-?\d+(?:\.\d{0,2})?/)![0]} req/s`;
+}
+
 export class OperationTableDetails extends React.PureComponent<TProps, TState> {
   state = {
     hoveredRowKey: -1,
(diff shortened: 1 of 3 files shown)


packages/jaeger-ui/src/components/Monitor/ServicesView/operationDetailsTable/index.tsx
```diff
@@ -34,6 +34,15 @@ type TState = {
   hoveredRowKey: number;
 };
 
+function formatValue(value: number) {
+  if (value < 0.1) {
+    return `< 0.1 req/s`;
+  }
+
+  // Truncate number to two decimal places
+  return `${value.toString().match(/^-?\d+(?:\.\d{0,2})?/)![0]} req/s`;
```
yurishkuro, line 43: Nit: this is not so much a question of decimal places, but of precision. E.g. 2.666667 is just as useless (too much precision) as 26666.67, but this code will allow the latter, whereas I would prefer to represent it as 26.6k - never more than 3 digits of precision

afishler: Good point and totally accepted. However, will be better just to go this path right now, and just improve on another iteration. Your comment definitely enhances the solution, but the current step will already provide an improvement which is immediate.

nofar9792: [yurishkuro](https://github.com/yurishkuro) ?

yurishkuro (approved):

PR #891, opened 2022-02-23
Use One node per resource as default option

### LINKED ISSUE #600: [Feature] change the default layout selection for deep dependency graph view (view=ddg)

can we change the default choice of `layout selection` to be `one node per resource` in this section:
Screen Shot 2020-07-16 at 3 32 48 PM
Current default is `show paths to focal node`

PawBud:
- Resolves #600 

- the default option for graph density has been replaced with **One node per resource**

https://user-images.githubusercontent.com/48093317/155391399-710429a6-7760-468a-ae6a-c9864ec4cce2.mp4

### DIFF
--- a/packages/jaeger-ui/src/components/DeepDependencies/Header/LayoutSettings/index.test.js
+++ b/packages/jaeger-ui/src/components/DeepDependencies/Header/LayoutSettings/index.test.js
@@ -23,7 +23,7 @@ import { EDdgDensity } from '../../../../model/ddg/types';
 
 describe('LayoutSettings', () => {
   const props = {
-    density: EDdgDensity.PreventPathEntanglement,
+    density: EDdgDensity.MostConcise,
     setDensity: jest.fn(),
     showOperations: true,
     toggleShowOperations: jest.fn(),
--- a/packages/jaeger-ui/src/components/DeepDependencies/url.test.js
+++ b/packages/jaeger-ui/src/components/DeepDependencies/url.test.js
@@ -140,11 +140,11 @@ describe('DeepDependencyGraph/url', () => {
       });
     });
 
-    it("defaults `density` to 'ppe'", () => {
+    it("defaults `density` to 'mc'", () => {
       const { density: unused, ...rest } = expectedParams;
       const { density: alsoUnused, ...rv } = acceptableParams;
       parseSpy.mockReturnValue(rv);
-      expect(getUrlState(getSearch())).toEqual({ ...rest, density: 'ppe' });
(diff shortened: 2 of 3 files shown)


yurishkuro:
It was stated in the ticket that we cannot just change the default, instead we can add a config option to control what default to use.

PawBud:
> It was stated in the ticket that we cannot just change the default, instead we can add a config option to control what default to use.

Oops, I misread it, Thanks a lot for pointing it out. I'll issue a commit adding that feature.

PawBud:
> It was stated in the ticket that we cannot just change the default, instead we can add a config option to control what default to use.

Screen Shot 2022-02-24 at 5 39 14 PM
Felt like a dropdown would be much more minimalistic, thoughts @yurishkuro?

yurishkuro:
If you're suggesting to remove descriptions from the drop down than I would be against it, it would make usability worse.

PawBud:
> If you're suggesting to remove descriptions from the drop down than I would be against it, it would make usability worse.

I was actually asking about the UI to keep a graph density option as default?

yurishkuro:
Sorry, I don't follow. We don't want to change the default, only to allow it to be configurable.

PawBud:
> Sorry, I don't follow. We don't want to change the default, only to allow it to be configurable.

Yes which basically means that the user should choose what the  default option should be? @yurishkuro

yurishkuro:
not the end user of the UI, but the operator of Jaeger installation: https://www.jaegertracing.io/docs/1.31/frontend-ui/#configuration

PawBud:
> not the end user of the UI, but the operator of Jaeger installation: https://www.jaegertracing.io/docs/1.31/frontend-ui/#configuration

Got it, Thanks a lot for clarifying!

PawBud:
@yurishkuro I need to add a "setDefaultOption" for the UI config https://github.com/jaegertracing/jaeger-ui/blob/main/packages/jaeger-ui/src/types/config.tsx#L41-L66 ?

yurishkuro:
yes
```
  deepDependencies?: {
    menuEnabled?: boolean;
    defaultDisplayMode: ...
  };
```

PawBud:
> yes
> 
> ```
>   deepDependencies?: {
>     menuEnabled?: boolean;
>     defaultDisplayMode: ...
>   };
> ```
I am having trouble writing the tests, can you help me? @yurishkuro

PawBud:
> 

Let me have a proper look, will get back to u soon

PawBud:
Closing this, as the issue is resolved !!

Issue #892, opened 2022-02-24
Monitor Tab - show 95 Latency in a more readable time-unit


nofar9792:
## Problem - what in Jaeger blocks you from solving the requirement?
In other places on this page, we show the latencies in a readable time unit.
We should do it too in P95 Latency
![Screen Shot 2022-01-25 at 16 09 10](https://user-images.githubusercontent.com/11076023/155535064-a25b7cc6-7b4d-46bb-bff0-8f0ce127068d.png)
>

## Proposal - what do you suggest to solve the problem or improve the existing situation?
convert the time

I would like to do it

PR #893, opened 2022-02-24
Monitor Tab - show 95 Latency in a more readable time-unit

nofar9792:
Resolves https://github.com/jaegertracing/jaeger-ui/issues/892

before:
Screen Shot 2022-01-25 at 16 09 10

after:
Screen Shot 2022-02-24 at 0 31 09

### DIFF
--- a/packages/jaeger-ui/src/components/Monitor/ServicesView/operationDetailsTable/__snapshots__/index.test.js.snap
+++ b/packages/jaeger-ui/src/components/Monitor/ServicesView/operationDetailsTable/__snapshots__/index.test.js.snap
@@ -799,7 +799,7 @@ exports[`<OperationTableDetails> render latency in seconds in the table 1`] = `
                         <div
                           className="table-graph-avg"
                         >
-                          8 sec
+                          8Sec
                         </div>
                       </div>
                     </td>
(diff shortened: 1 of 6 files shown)


yurishkuro:
Don't have a problem with this PR itself, but I am not a fan of this `timeConversion` function, it reduces precision more than necessary. E.g. 5900ms would be displayed as 5s - this is very misleading.

Could we fix this function to:
  1. use rounding, not floor function
  2. keep at least 2 significant digits
  3. use lower-case units

nofar9792:
> Don't have a problem with this PR itself, but I am not a fan of this `timeConversion` function, it reduces precision more than necessary. E.g. 5900ms would be displayed as 5s - this is very misleading.
> 
> Could we fix this function to:
> 
> 1. use rounding, not floor function
> 2. keep at least 2 significant digits
> 3. use lower-case units

lower-case units? change ```timeConversion``` or use in my case ```toLowerCase```>

nofar9792:
> change timeConversion. It's weird that it mixes cases between us/ms and Sec/Hrs



> change timeConversion. It's weird that it mixes cases between us/ms and Sec/Hrs

OK, it is going to affect this place too:
Screen Shot 2022-02-24 at 22 56 07

nofar9792:
> change timeConversion. It's weird that it mixes cases between us/ms and Sec/Hrs

done

nofar9792:
@[yurishkuro](https://github.com/yurishkuro) ping

yurishkuro (approved):

yurishkuro (approved):

Issue #897, opened 2022-02-27
Monitor Tab: Cannot choose another timeframe


nofar9792:
**Describe the bug**
Cannot choose another timeframe

**To Reproduce**
1. Choose timeframe in monitor tab
2.Choose another timeframe in monitor tab

**Expected behavior**
See the chosen timeframe

**Screenshots**

https://user-images.githubusercontent.com/11076023/155885867-93aaed88-830d-46d8-a03e-fc8daea2f8df.mov

I'll fix it

PR #898, opened 2022-02-27
Monitor Tab: Cannot choose another timeframe

nofar9792:
 Resolve (https://github.com/jaegertracing/jaeger-ui/issues/897)

before:

https://user-images.githubusercontent.com/11076023/155885940-fd35d6e5-26a0-489b-8c28-04e3ecf53659.mov

after:

https://user-images.githubusercontent.com/11076023/155885945-e397059e-d938-4ce7-914a-b97b6dc86eba.mov

The issue was the calculation of the timeframe. I did in the constructor and not on every new state

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -13,7 +13,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2.4.0
-      - uses: actions/setup-node@v2.5.1
+      - uses: actions/setup-node@v3
         with:
           node-version: "10"
       - name: install yarn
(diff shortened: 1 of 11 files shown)


albertteoh (dismissed):
LGTM, thanks for fixing! Curious, is there a test we could write to prevent this from happening again?

nofar9792:
@albertteoh i added the test + small change in the implementation

albertteoh (approved):

nofar9792:
@yurishkuro @albertteoh I don't understand why the test coverage decreased.
My changes are not related to ```packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx```

albertteoh:
> I don't understand why the test coverage decreased.

I don't understand the root cause either, but I have observed flakiness in codecov in the past with jaeger-ui PRs. The workaround was to close and re-open the PR to see if it resolves the issue, which I'll try now for you. 

I'd be keen to learn if anyone else has insights on why the codecov check is failing.

albertteoh:
FWIW, a link to what codecov _thinks_ is where the coverage regression exists, which doesn't seem to be related to this PR: https://app.codecov.io/gh/jaegertracing/jaeger-ui/compare/898/changes.

Any ideas @yurishkuro?

Issue #908, opened 2022-03-10
Monitor Tab - Error rate graph in operation table is always a straight line


nofar9792:
**Describe the bug**
The graph is always a straight line, we can't see the values

**Expected behavior**
The values should affect the graph

**Screenshots**
The bug:
<img width="178" alt="Screen Shot 2022-03-10 at 23 49 42" src="https://user-images.githubusercontent.com/11076023/157762020-fa8ebf95-4685-472d-ac4b-a37c89898977.png">

How it's supposed to look:
<img width="194" alt="Screen Shot 2022-03-10 at 23 49 28" src="https://user-images.githubusercontent.com/11076023/157762033-96c85683-619d-4d0d-9271-712afaefa144.png">

nofar9792:
@yurishkuro ?

PR #909, opened 2022-03-10
Monitor Tab - Error rate graph in operation table is always a straight line

nofar9792:
Signed-off-by: nofar9792 <nofar.cohen@logz.io>

Resolves https://github.com/jaegertracing/jaeger-ui/issues/908

The values of this graph are between 0 to 1, but the yDomain is hardcode 1-100

before:
Screen Shot 2022-03-10 at 23 49 42

after:
Screen Shot 2022-03-10 at 23 49 28


albertteoh (approved):

albertteoh:
Thanks!

PR #927, opened 2022-04-13
Add error icon to errored services in trace list view

### LINKED ISSUE #925: Highlight services that has error in trace list view

Highlight the service tag that has an error in the trace list view.

image

When in the trace list view, it currently shows how many errors exist in the trace, but it does not tell the user which service has an error.

image
Add an error icon to the service if error tag exists.

If this is acceptable, I have the code written already :)

Jaskerv:
Resolves #925 
- Add an error icon to errored services in `ResultItem.tsx`.
image
- Refactor existing constants to state.

### DIFF
--- a/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ResultItem.tsx
+++ b/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ResultItem.tsx
@@ -19,7 +19,8 @@ import { Link } from 'react-router-dom';
 import { sortBy } from 'lodash';
 import moment from 'moment';
 
-import { trackConversions, EAltViewActions } from './index.track';
+import IoAlert from 'react-icons/lib/io/alert';
+
 import * as markers from './ResultItem.markers';
 import ResultItemTitle from './ResultItemTitle';
 import colorGenerator from '../../../utils/color-generator';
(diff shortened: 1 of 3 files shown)


packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ResultItem.tsx
```diff
@@ -19,6 +19,8 @@ import { Link } from 'react-router-dom';
 import { sortBy } from 'lodash';
 import moment from 'moment';
 
+import IoAlert from 'react-icons/lib/io/alert';
```
yurishkuro, line 22: I am not seeing this used in the other parts of the UI, how do we render (!) in the trace view?

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ResultItem.tsx
```diff
@@ -38,10 +40,43 @@ type Props = {
...
+
+    const numErredSpans = spans.filter(sp => {
+      const hasError = sp.tags.some(isErrorTag);
+      if (hasError) {
+        erroredServices.add(sp.process.serviceName);
+      }
+      return hasError;
+    }).length;
+
+    this.state = {
```
yurishkuro, line 71: just curious, how/where is this `.state` field associated with the `State` type?

Jaskerv:
```
export default class ResultItem extends React.PureComponent<Props, State> {
```
It is associated here :)

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ResultItem.tsx
Jaskerv: I'm using the same one as in line 152 of `SpanBarRow.tsx` to keep things consistent.
```
{showErrorIcon && <IoAlert className="SpanBarRow--errorIcon" />}
```

yurishkuro:
Could you please add or modify a unit test to add coverage the new conditional code?

PR #932, opened 2022-04-16
Fix: service operations fetch by URL query params

### LINKED ISSUE #920: Trace Search: setting operation query param not populating menu item
**Describe the bug**
Setting the operation query parameter to the trace search URL does not populate operation menu item.

**To Reproduce**
1. `cd docker-compose/monitor`
2. `docker-compose up`
3. Open browser to http://localhost:16686
4. Open the Monitor tab
5. Select a service
6. Mouse over an operation then click on "View traces"
7. A new browser tab should open with the trace search page. Note the URL contains all the trace search parameters
8. Note the service is correctly selected in the menu item, but operation has no selection and only contains the "all" operation.
9. Selecting the service again triggers a population of the respective service's operations.

**Expected behavior**
All menu items correct prefilled and trace search results based on those inputs after clicking on "View traces".

**Screenshots**

https://user-images.githubusercontent.com/26584478/161751063-db331063-02c2-4c8a-b358-66a35fbf1361.mp4

 - OS: MacOS
 - Jaeger version: latest
 - Deployment: local laptop

FloydJohn:
Resolves #920

If query parameters are provided to the search page, the system now fetches the correct set of operations linked to the specified service.

This fixes the current behavior on fetching the operations linked to the last searched service, ignoring all specified URL parameters.

### DIFF
--- a/packages/jaeger-ui/src/components/SearchTracePage/index.js
+++ b/packages/jaeger-ui/src/components/SearchTracePage/index.js
@@ -62,7 +62,7 @@ export class SearchTracePageImpl extends Component {
       fetchMultipleTraces(needForDiffs);
     }
     fetchServices();
-    const { service } = store.get('lastSearch') || {};
+    const { service } = urlQueryParams || store.get('lastSearch') || {};
     if (service && service !== '-') {
       fetchServiceOperations(service);
     }
--- a/packages/jaeger-ui/src/components/SearchTracePage/index.js
+++ b/packages/jaeger-ui/src/components/SearchTracePage/index.js
@@ -62,7 +62,10 @@ export class SearchTracePageImpl extends Component {
       fetchMultipleTraces(needForDiffs);
     }
     fetchServices();
-    const { service } = urlQueryParams || store.get('lastSearch') || {};
+    let { service } = store.get('lastSearch') || {};
+    if (urlQueryParams && urlQueryParams.service) {
+      service = urlQueryParams.service;
+    }
     if (service && service !== '-') {
       fetchServiceOperations(service);
     }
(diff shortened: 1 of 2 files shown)


packages/jaeger-ui/src/components/SearchTracePage/index.js
```diff
@@ -62,7 +62,7 @@ export class SearchTracePageImpl extends Component {
       fetchMultipleTraces(needForDiffs);
     }
     fetchServices();
-    const { service } = store.get('lastSearch') || {};
+    const { service } = urlQueryParams || store.get('lastSearch') || {};
```
yurishkuro, line 65: I'm not a JS expert, but this looks odd to me - queryParams is an object, right? So wouldn't it evaluate to true and short circuit the rest even if the objects doesn't have a 'service' property?

I would like to see a unit tests.

FloydJohn: I think that queryParams can be undefined (if no query is passed) or an object otherwise. After this comment I played a bit with the UI and it seems that if a param is passed (i.e. `end`), then the `service` param is mandatory.

To be absolutely sure of this behavior though I updated the code to explicitly overwrite the `service` variable only if the `service` param is provided, in a way that supports both `undefined` and `{}` values of `urlQueryParams`.
I also added and updated unit test to verify that:
1. If `urlQueryParams={}`, the system fetches the operations for the stored service
2. If `urlQueryParams` includes `service`, the system works as expected

As I'm quite new to this project and to open source contributions in general, let me know if there's anything I should improve.

Thanks!

yurishkuro (approved):
thanks!

PR #943, opened 2022-05-05
Support new user analytics events in the Monitor Tab

### LINKED ISSUE #942: Monitor Tab - Support new user analytics events in Jaeger
We want to be able to track a few addition SPM tracing events 

currently Jaeger doesn't have a necessary events for SPM tracing

We suggest to implement this list of events and assign them to the appropriate UI controls.

The list of required events:

- Clicked the services dropdown and selected an option
- Clicked the timeframe dropdown and selected an option
- Clicked on the link under the services list to go to trace search based on service
- Clicked the button in the operations list to go to trace search based on the operation
- Searched the operations by using the search box
- Clicked the operations list header to sort

VladislavBryukhanov:
Resolves #942 

Added new tracking events to SPM tracing functionality

### DIFF
--- a/packages/jaeger-ui/src/components/Monitor/ServicesView/index.track.tsx
+++ b/packages/jaeger-ui/src/components/Monitor/ServicesView/index.track.tsx
@@ -1,0 +1,34 @@
+// Copyright (c) 2017 The Jaeger Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import _debounce from 'lodash/debounce';
+import { trackEvent } from '../../../utils/tracking';
+
+const CATEGORY_SPM = 'jaeger/ux/trace/spm';
+
+const ACTION_SELECT_SERVICE = 'select-service';
+const ACTION_SELECT_TIMEFRAME = 'select-timeframe';
+const ACTION_VIEW_ALL_TRACES = 'view-all-traces';
+const ACTION_SEARCH_OPERATION = 'search-operation';
+
+export const trackViewAllTraces = () => trackEvent(CATEGORY_SPM, ACTION_VIEW_ALL_TRACES);
+export const trackSelectService = (service: string) =>
+  trackEvent(CATEGORY_SPM, ACTION_SELECT_SERVICE, service);
+export const trackSelectTimeframe = (timeframe: string) =>
+  trackEvent(CATEGORY_SPM, ACTION_SELECT_TIMEFRAME, timeframe);
+
+export const trackSearchOperationDebounced = _debounce(
+  (searchQuery: string) => trackEvent(CATEGORY_SPM, ACTION_SEARCH_OPERATION, searchQuery),
+  1500
+);
--- a/packages/jaeger-ui/src/components/Monitor/ServicesView/index.track.tsx
+++ b/packages/jaeger-ui/src/components/Monitor/ServicesView/index.track.tsx
@@ -15,12 +15,12 @@
 import _debounce from 'lodash/debounce';
 import { trackEvent } from '../../../utils/tracking';
 
-const CATEGORY_SPM = 'jaeger/ux/trace/spm';
+export const CATEGORY_SPM = 'jaeger/ux/trace/spm';
 
-const ACTION_SELECT_SERVICE = 'select-service';
-const ACTION_SELECT_TIMEFRAME = 'select-timeframe';
-const ACTION_VIEW_ALL_TRACES = 'view-all-traces';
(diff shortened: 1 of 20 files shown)


packages/jaeger-ui/src/components/Monitor/ServicesView/index.track.tsx
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) 2017 The Jaeger Authors.
```
yurishkuro, line 1:
```suggestion
// Copyright (c) 2022 The Jaeger Authors.
```

packages/jaeger-ui/src/components/Monitor/ServicesView/index.track.tsx
```diff
@@ -0,0 +1,34 @@
...
+export const ACTION_VIEW_ALL_TRACES = 'view-all-traces';
+export const ACTION_SEARCH_OPERATION = 'search-operation';
+
+export const trackViewAllTraces = () => trackEvent(CATEGORY_SPM, ACTION_VIEW_ALL_TRACES);
+export const trackSelectService = (service: string) =>
+  trackEvent(CATEGORY_SPM, ACTION_SELECT_SERVICE, service);
+export const trackSelectTimeframe = (timeframe: string) =>
+  trackEvent(CATEGORY_SPM, ACTION_SELECT_TIMEFRAME, timeframe);
+
+export const trackSearchOperationDebounced = _debounce(
```
yurishkuro, line 31: organizationally, it seems this should be just `trackSearchOperation`, and the debouncing can be done by the caller.

packages/jaeger-ui/src/components/Monitor/ServicesView/index.track.tsx
VladislavBryukhanov: fixed

packages/jaeger-ui/src/components/Monitor/ServicesView/index.track.tsx
VladislavBryukhanov: refactored

packages/jaeger-ui/src/components/Monitor/ServicesView/index.track.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+import _debounce from 'lodash/debounce';
+import { trackEvent } from '../../../utils/tracking';
+
+export const CATEGORY_SPM = 'jaeger/ux/trace/spm';
+
+export const ACTION_SELECT_SERVICE = 'select-service';
+export const ACTION_SELECT_TIMEFRAME = 'select-timeframe';
+export const ACTION_VIEW_ALL_TRACES = 'view-all-traces';
+export const ACTION_SEARCH_OPERATION = 'search-operation';
```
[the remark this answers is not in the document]
VladislavBryukhanov: I refactored the tracking methods usage and use category instead, is it clearer?

packages/jaeger-ui/src/components/Monitor/ServicesView/index.tsx
```diff
@@ -68,6 +75,8 @@ type TDispatchProps = {
   fetchAllServiceMetrics: (serviceName: string, query: MetricsAPIQueryParams) => void;
 };
 
+const trackSearchOperationDebounced = _debounce(searchQuery => trackSearchOperation(searchQuery), 1000);
```
yurishkuro, line 78: Just curious, why is it necessary to introduce a delay here? Aren't events reported async in the background anyway?

yurishkuro:
lgtm overall. Please make sure all commits are signed (often easier to squash them and resubmit as one).

packages/jaeger-ui/src/components/Monitor/ServicesView/index.tsx
VladislavBryukhanov: it is important to prevent dummy information reporting if user search for some specific name. 
For  example if the target name is "my-metrics" I attempt to track and send report only once with completed searching query, but without debounce there are 10 separate report requests with partial searching query will be send. The cause of this behaviour is that the onChange handler fires every time a new character is typed into the input, that's why these 10 reports are sent, without debounce:
"m"
"my"
"my-"
...
"my-metric"
"my-metrics"

yurishkuro: Are you saying a pending debounce gets canceled if another one is issued? That was not my understanding of the function.

VladislavBryukhanov: regarding 1-st point - yes, you are right
regarding 2-nd point - the search field filters data only locally on front-end and doesn't sent any requests to the backend

pavolloffay:
Is there any screenshot or mockup for this feature?

yurishkuro:
@pavolloffay it's an international observability, aka user tracking

nofar9792:
@yurishkuro can we merge it or do we need to fix something before?

yurishkuro:
we need DCO check (signed commits) to pass before I can merge.

VladislavBryukhanov:
DCO issue resolved @yurishkuro

packages/jaeger-ui/src/components/Monitor/ServicesView/index.tsx
yurishkuro: So on point 1 - if debounce merely delays, won't you still end up with the same incremental logs?

yurishkuro (approved):

Issue #945, opened 2022-05-09
Monitor Tab - pages crashed when searching for traces


nofar9792:
**Describe the bug**
Monitor tab crashes when search tab is in loading state

**To Reproduce**
1.  click on "find traces"
2. clink on monitor tab

**Expected behavior**
show monitor tab on loading state

**Screenshots**

https://user-images.githubusercontent.com/11076023/167506977-3762a772-2403-4f64-a862-2214ffb76ea1.mov

PR #946, opened 2022-05-09
Monitor tab - Fix monitor tab crashing

nofar9792:
Resolves  https://github.com/jaegertracing/jaeger-ui/issues/945

if the xDomain is empty we get an error from react-vis lib


nofar9792:
@yurishkuro @albertteoh

albertteoh (approved):
Would be good if someone with react experience can look at this as well.

nofar9792:
> Would be good if someone with react experience can look at this as well.

@AlonMiz @galangel @VladislavBryukhanov

galangel (commented):
lookin good

PR #951, opened 2022-05-13
Use red on scatterplot for traces if any spans have an error=true tag

esnible:
Signed-off-by: Ed Snible <snible@us.ibm.com>

Show error traces in red on the scatterplot.

To test this it can be handy to have a source of good-looking traces with errors.  I used synthentic data generated by `docker run -it --env JAEGER_COLLECTOR_URL=http://docker.for.mac.localhost:14268 omnition/synthetic-load-generator`.

Note that this is my first Jaeger-UI PR.  When I tried to commit I got a warnings.  Resolving them needed some help: I didn't know that I needed to run `yarn prettier` to clean my code, nor `yarn lint` to lint it.  Is this in the docs?  I couldn't find it.

jaeger-with-red-errors

### DIFF
--- a/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.js
+++ b/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.js
@@ -39,6 +39,7 @@ import reduxFormFieldAdapter from '../../../utils/redux-form-field-adapter';
 
 import type { FetchedTrace } from '../../../types';
 import type { SearchQuery } from '../../../types/search';
+import type { KeyValuePair } from '../../../types/trace';
 
 import './index.css';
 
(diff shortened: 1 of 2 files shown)


packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.js
```diff
@@ -167,6 +170,7 @@ export class UnconnectedSearchResults extends React.PureComponent<SearchResultsP
                   traceID: t.traceID,
                   size: t.spans.length,
                   name: t.traceName,
+                  color: t.spans.some(sp => sp.tags.some(isErrorTag)) ? 'red' : '#12939A',
```
rubenvp8510, line 173: Can we use constants for the colors?

esnible: Can you advise me on how we do this elsewhere?  For example https://github.com/jaegertracing/jaeger-ui/blob/main/packages/jaeger-ui/src/components/DependencyGraph/DAG.js#L80 uses them inline, so I did as well.

The shade of blue we use, #12939A, seems to come from https://github.com/uber/react-vis but I couldn't actually figure out if it was coming from .css or .js, and which file.

PR #965, opened 2022-06-24
Store preferred span name column width in localStorage

bobrik:
Different developers tend to look at different kinds of traces, meaning that they often prefer some set width of span name column in trace view.

Without this patch, the width is set at 25% and it resets on refresh. With this patch the width is remembered in localStorage and preserved on refresh, making it less necessary to adjust.

Posting it without a backing issue since it's a tiny change.

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/duck.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/duck.tsx
@@ -52,7 +52,7 @@ export function newInitialState(): TTraceTimeline {
     detailStates: new Map(),
     hoverIndentGuideIds: new Set(),
     shouldScrollToFirstUiFindMatch: false,
-    spanNameColumnWidth: 0.25,
+    spanNameColumnWidth: parseFloat(localStorage.getItem('spanNameColumnWidth') || '0.25'),
     traceID: null,
   };
 }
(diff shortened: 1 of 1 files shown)


yurishkuro (approved):
thanks

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/duck.tsx
```diff
@@ -52,7 +52,7 @@ export function newInitialState(): TTraceTimeline {
     detailStates: new Map(),
     hoverIndentGuideIds: new Set(),
     shouldScrollToFirstUiFindMatch: false,
-    spanNameColumnWidth: 0.25,
+    spanNameColumnWidth: parseFloat(localStorage.getItem('spanNameColumnWidth') || '0.25'),
```
yurishkuro, line 55: prettier linter seems to be failing on this

bobrik: I think it's failing below on a missing semicolon. I amended the commit.

bobrik: It is now unhappy about `scripts/get-changelog.js`, which I haven't touched, and I'm not sure what to do with that.

yurishkuro: I pushed a fix to it https://github.com/jaegertracing/jaeger-ui/commit/ec17f7896d669b21c3b34c9c9fd6946becd9687b, rebasing

bobrik:
Thanks!

PR #972, opened 2022-07-20
Bump moment to 2.29.4

zhuje:
Signed-off-by: Jenny Zhu <jezhu@redhat.com>

- Bumping moment to 2.29.4

-

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,8 +36,7 @@
     "**/js-yaml": "3.13.1",
     "**/lodash": "4.17.21",
     "**/node-gyp/tar": "2.2.2",
-    "**/prettier": "1.18.2",
-    "**/moment": "2.29.4"
+    "**/prettier": "1.18.2"
   },
   "workspaces": {
     "packages": [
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -36,7 +36,8 @@
     "**/js-yaml": "3.13.1",
     "**/lodash": "4.17.21",
     "**/node-gyp/tar": "2.2.2",
-    "**/prettier": "1.18.2"
+    "**/prettier": "1.18.2",
+    "**/moment": "2.29.4"
```
yurishkuro, line 40: Why do we need this here? I think lock file is sufficient.

yurishkuro: dependabot never inserts dependencies into package.json when it bumps the versions

yurishkuro: https://github.com/yarnpkg/yarn/issues/4986#issuecomment-395036563

* delete dependency from the lock file
* run `yarn` - it will re-add it with a newer version

yurishkuro (approved):

Issue #390, opened 2019-06-10
Visualizing follows_from references with horizontal stacking


richard-fine:
## Requirement - what kind of business use case are you trying to solve?

It is sometimes necessary to trace processes which involve a long sequence of sequential spans (perhaps using the follows_from reference). For example, imagine a trace of a system which is given a DAG of thousands of short-lived work items, and spawns N threads, each of which begins executing work items in dependency order. In terms of tracing what actually happened, we would like to record a span for each individual work item, and we would like to be able to visualize the parallelism across the threads, so we can identify sync points in the DAG.

## Problem - what in Jaeger blocks you from solving the requirement?

Jaeger-UI does not currently support visualizing long sequences of spaces (e.g using the follows_from ref) in an efficient manner. Every row in the chart always only displays a single span, meaning that a long sequence of short spans shows as a large amount of wasted space:

![image](https://user-images.githubusercontent.com/588194/59233574-f8865880-8be0-11e9-9cba-8d5503f68e3b.png)

This is pretty low information density. It makes it harder to see that the spans are sequential, and makes it harder to spot when threads are work-starved due to a sync point.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

Ideally, find some way to horizontally stack spans in a disjoint follows_from sequence. For example, here is a trace of a similar process in the Chrome tracing UI:

![image](https://user-images.githubusercontent.com/588194/59233821-f2dd4280-8be1-11e9-83cb-356d3d89ea59.png)

This is much more compact, and makes it very easy to see how successfully the process is parallelized. Detail about an individual span can be seen by clicking it, in a similar way to Jaeger.

## Any open questions to address

* What would happen to the duration markers, currently displayed to the right of a span, when they would overlap a horizontally stacked neighboring span? (Maybe it's OK to just drop durations in that context - you can see it in the detail view when you click on the span anyway).
* What happens when there is more than one operation name in the sequence? Right now the operation name is shown once for the whole row.
* What happens when the trace contains spans that are `child_of` the spans in the sequence?

yurishkuro:
Ironically, one of the oldest, still unresolved, issues in OpenTracing, https://github.com/opentracing/opentracing-go/issues/3, is about how to model sibling spans. Your use case is a lot better / clearer scenario, assuming your individual work units make other downstream calls (otherwise it's all in process and distributed tracing may not be the best tool).

[...]

PR #976, opened 2022-08-05
Add trace visualization as a FlameGraph

### LINKED ISSUE #525: Support trace view as a flame graph
I am implementing jaeger, it really looks nice.
But I got a problem.
I want to collect spans on one line in tracing view, but I cannot find way.

[As-IS]
image

[To-Be]
image

Maybe I can't find the doc... could you help ?

pavelpashkovsky:
We added the ability to visualize a trace as a flamegraph and thought it might be worth contributing upstream! Would love feedback on if it resolves this issue: https://github.com/jaegertracing/jaeger-ui/issues/390

Resolves #525

https://user-images.githubusercontent.com/7372044/183039257-26fd8cee-1d57-411a-a31c-032a2d9a0948.mp4

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceFlamegraph/index.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceFlamegraph/index.tsx
@@ -1,4 +1,4 @@
-// Copyright (c) 2017 Uber Technologies, Inc.
+// Copyright (c) 2022 The Jaeger Authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/packages/jaeger-ui/src/components/TracePage/TraceFlamegraph/index.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceFlamegraph/index.tsx
@@ -18,7 +18,7 @@ import '@pyroscope/flamegraph/dist/index.css';
 import './index.css';
 
 const TraceFlamegraph = ({ trace }: any) => {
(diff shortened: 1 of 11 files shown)


yurishkuro (commented):
Thanks for contributing! 🥳 

A couple of questions on the view:
  1. The table on the left seems sorted by Self column - is it possible to indicate that visually?  I can't tell if the sorting is by Self or Total column, can only infer it from the data
  2. There is a Total/Self cell in the table that shows "< 0.01 sec" - is it informative? As I understand, "total" here refers to an implied span outside of the actual spans (judging by the flame graph itself, the top purple row), so under what condition could Total/Self be non-zero?

Lastly, please make sure all commits in the PR are signed (it's a CNCF licensing requirement).

packages/jaeger-ui/src/components/TracePage/TraceFlamegraph/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+  const convertedProfile = convertJaegerTraceToProfile(trace.data);
+
+  return (
+    <div className="Flamegraph-wrapper">
+      <FlamegraphRenderer colorMode="light" profile={convertedProfile} />
+    </div>
+  );
+};
+
+export default TraceFlamegraph;
```
nofar9792, line 30: maybe add some basic tests like "does not explode" and snapshot?

yurishkuro (dismissed):

yurishkuro:
Thank you!

bobrik:
There are some kinks with this: #1012.

PR #977, opened 2022-08-11
Bump @pyroscope/flamegraph to 0.18.4

pavelpashkovsky:
Signed-off-by: pavelpashkovsky <pavelpashkovsky@gmail.com>

- Bumping `@pyroscope/flamegraph` to `0.18.4`

- As a follow up to some of the comments from [[#976]()](https://github.com/jaegertracing/jaeger-ui/pull/976):
☑︎ Added indication of which column is sorted visually
☑︎ When "self" is 0 then table will show 0.00

### DIFF
--- a/packages/jaeger-ui/package.json
+++ b/packages/jaeger-ui/package.json
@@ -42,7 +42,7 @@
   },
   "dependencies": {
     "@jaegertracing/plexus": "0.2.0",
-    "@pyroscope/flamegraph": "0.18.4-1379-557f0b4.0",
+    "@pyroscope/flamegraph": "0.18.4",
     "@types/classnames": "^2.2.7",
     "@types/deep-freeze": "^0.1.1",
     "@types/history": "^4.7.2",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1517,10 +1517,10 @@
     universal-user-agent "^2.0.0"
     url-template "^2.0.8"
 
-"@pyroscope/flamegraph@0.18.4-1379-557f0b4.0":
-  version "0.18.4-1379-557f0b4.0"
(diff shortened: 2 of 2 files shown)


packages/jaeger-ui/package.json
```diff
@@ -42,7 +42,7 @@
   },
   "dependencies": {
     "@jaegertracing/plexus": "0.2.0",
-    "@pyroscope/flamegraph": "0.18.4-1379-557f0b4.0",
+    "@pyroscope/flamegraph": "0.18.4",
```
yurishkuro, line 45: while I have you, I wanted to ask - in the last merge I am seeing that the numbers in the flame graph & table are in "units", which is pretty confusing. It sounds like a technical reason (library may not know the units of numbers), but couldn't we pass some parameter to it to make it more obvious?  Also, I'm just curious under what conditions the numbers in the flamegraph are not time units - if they are always time than it seems like the graph could always display them with proper time units like s, ms, etc.

yurishkuro (approved):

packages/jaeger-ui/package.json
```diff
@@ -42,7 +42,7 @@
   },
   "dependencies": {
     "@jaegertracing/plexus": "0.2.0",
-    "@pyroscope/flamegraph": "0.18.4-1379-557f0b4.0",
```
yurishkuro, line 45: Where is the source repository for this module?

packages/jaeger-ui/package.json
Rperry2174: The flamegraph component can be used for visualizing other arbitrary data that does have [other units](https://github.com/pyroscope-io/pyroscope/blob/main/packages/pyroscope-flamegraph/src/format/format.ts#L19-L35).

Makes sense though that for traces it will always be time so made a Pr to fix this:
https://github.com/pyroscope-io/pyroscope/pull/1387

It's available in [v0.18.5](https://www.npmjs.com/package/@pyroscope/flamegraph)

PR #981, opened 2022-08-18
Reproducible asset-manifest.json

TripleDogDare:
https://github.com/jaegertracing/jaeger/pull/3880#pullrequestreview-1076673563

Reconfigures the webpack-manifest-plugin to use a stable sorted serialization format.

### DIFF
--- a/packages/jaeger-ui/config-overrides.js
+++ b/packages/jaeger-ui/config-overrides.js
@@ -16,6 +16,8 @@
 const fs = require('fs');
 const { addBabelPlugin, addLessLoader } = require('customize-cra');
 const lessToJs = require('less-vars-to-js');
+const ManifestPlugin = require('webpack-manifest-plugin');
+const stringify = require('json-stable-stringify');
 
 function useEslintRc(config) {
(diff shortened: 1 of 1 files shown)


packages/jaeger-ui/config-overrides.js
```diff
@@ -16,6 +16,8 @@
 const fs = require('fs');
 const { addBabelPlugin, addLessLoader } = require('customize-cra');
 const lessToJs = require('less-vars-to-js');
+const ManifestPlugin = require('webpack-manifest-plugin');
```
yurishkuro, line 19: I am not familiar with the webpack ecosystem, but isn't it possible to configure this plugin via config file, rather than with explicit code like that?

yurishkuro: https://www.npmjs.com/package/webpack-manifest-plugin

TripleDogDare: Everything I know about it I learned today. So this is probably the best I've got at the moment.

yurishkuro (approved):

PR #983, opened 2022-08-22
Resolves #982

### LINKED ISSUE #982: [Bug]: /trace/traceID loss of precision
### What happened?

i guess the package 'isomorphic-fetch' may loss of precision(int64) while proxy to load data from jaeger
jaeger-ui: v1.14.0
jaeger: v1.24.0

### Steps to reproduce

1. start jaeger
2. change packages/jaeger-ui/src/api/jaeger.js:42 && start jaeger-ui 
image
4.access /trace/4c5d7a066c367635704f06207dd56b45 page
image
image
5.aceess api directly
image

6.we can find out the value of "liveroom_sid" changed 
780177723283783680 -> 780177723283783700

### Expected behavior

jaeger : response int64  value "780177723283783680"
jaeger-ui: receive int64 value "780177723283783680" (actual it is )

### Relevant log output

_No response_

### Screenshot

_No response_

### Additional context

_No response_

### Jaeger backend version

v1.24.0

### SDK

_No response_

### Pipeline

_No response_

### Stogage backend

_No response_

### Operating system

_No response_

### Deployment model

_No response_

### Deployment configs

_No response_

SealinGp:
avoid value precision loss
Signed-off-by: seazhang <sealingp@163.com>

- 
Resolves https://github.com/jaegertracing/jaeger-ui/issues/982
- avoid display value precision loss in case of value overflow Number.MAX_VALUE

### DIFF
--- a/packages/jaeger-ui/src/model/transform-trace-data.test.js
+++ b/packages/jaeger-ui/src/model/transform-trace-data.test.js
@@ -76,7 +76,13 @@ describe('transformTraceData()', () => {
       ],
       startTime,
       duration,
-      tags: [],
+      tags: [
+        {
+          key: "bigintkey",
+          type: "int64",
+          value: 780177723283783680
+        },
+      ],
       processID: 'p1',
     },
     {
--- a/packages/jaeger-ui/src/model/transform-trace-data.tsx
+++ b/packages/jaeger-ui/src/model/transform-trace-data.tsx
@@ -88,7 +88,18 @@ export default function transformTraceData(data: TraceData & { spans: SpanData[]
   const max = data.spans.length;
   for (let i = 0; i < max; i++) {
     const span: Span = data.spans[i] as Span;
-    const { startTime, duration, processID } = span;
+    const { startTime, duration, processID, tags } = span;
+
+    // avoid display value precision loss in case of value overflow Number.MAX_VALUE
+    if(tags) {
+      for(let j = 0; j < tags.length; j++) {
+        const tag = tags[j]
+        if(Number.isInteger(tag.value) && !Number.isSafeInteger(tag.value)) {
+          tag.value = BigInt(tag.value).toString();
+        }
+      }   
+    }
+
     //
     let spanID = span.spanID;
     // check for start / end time for the trace


packages/jaeger-ui/src/model/transform-trace-data.tsx
```diff
@@ -88,7 +88,18 @@ export default function transformTraceData(data: TraceData & { spans: SpanData[]
...
   for (let i = 0; i < max; i++) {
     const span: Span = data.spans[i] as Span;
-    const { startTime, duration, processID } = span;
+    const { startTime, duration, processID, tags } = span;
+
+    // avoid display value precision loss in case of value overflow Number.MAX_VALUE
+    if(tags) {
+      for(let j = 0; j < tags.length; j++) {
+        const tag = tags[j]
+        if(Number.isInteger(tag.value) && !Number.isSafeInteger(tag.value)) {
```
yurishkuro, line 97: What is the type of tag.value at this point? If it's a string, no need to convert, if it's a number then the precision may already be lost. I would've thought the fix needs to be on the server side such that big numeric numbers are sent as strings in JSON.

packages/jaeger-ui/src/model/transform-trace-data.test.js
```diff
@@ -76,7 +76,13 @@ describe('transformTraceData()', () => {
       ],
       startTime,
       duration,
-      tags: [],
+      tags: [
+        {
+          key: "bigintkey",
+          type: "int64",
+          value: 780177723283783680
```
yurishkuro, line 83: a) this isn't a test by itself since it's not checking anything
b) defining data in code is not equivalent to parsing the data from a JSON

packages/jaeger-ui/src/model/transform-trace-data.tsx
SealinGp: >  "if it's number then the precision may already be lost" 

 but i got the right console output after i converted it to bigint, it confused me.
image

api return 
image

i'll consider output string when it is big numeric numbers. it's a way to solve it after all.

SealinGp: - should server side return below data instead
```json
{
    "key": "xx_id",
    "type": "string",
    "value": "780177723283783680"
}
```

- or should jaeger json model return value as json string ?
https://github.com/jaegertracing/jaeger/blob/main/model/json/model.go#L100
```go
type KeyValue struct {
	Key   string      `json:"key"`
	Type  ValueType   `json:"type,omitempty"`
	Value interface{} `json:"value"` //json output string ?
}
```

all ways seems to be strange in some logic way 😂

yurishkuro:
I believe this needs to be addressed on the server side - https://github.com/jaegertracing/jaeger/issues/3958

PR #1001, opened 2022-10-20
Upgrade many dependencies

mmorel-35:
- This is upgrading all possible dependencies without breaking build and tests

NB : I have not done any integration tests, only those actually provided

Please squash the commits


.eslintrc.js
```diff
@@ -44,7 +44,7 @@ module.exports = {
   ],
   rules: {
     /* general */
-    'arrow-parens': [1, 'as-needed'],
+    'arrow-parens': [1, 'always'],
```
yurishkuro, line 47: it looks like this change is responsible for the most styling noise in the PR, what was the motivation?

yurishkuro:
Thanks for taking on this issue!

After looking at the diff, here's my suggestion on how to move this forward:

* I suggest removing all pure styling changes. The PR is currently nearly impossible to review in GH Web UI, it crumbles. But more importantly, the styling changes obscure the real code changes that may have been necessary due to dependency upgrades. This is just a bad practice to combine such changes.
* I would like to understand the scope of the change. The title says "all deps", but I see that typescript version is still 3.x (dependabot has been trying to upgrade to 4.x for some time, there are some things that are breaking). And the other major one is react, which still seems to be at `"react": "^16.3.2"`.

.eslintrc.js
mmorel-35: The responsible is prettier v2. I rollbacked it’s upgrade and git back to the initial configuration.

yurishkuro:
@mmorel-35 could you elaborate on your method of selecting the versions to upgrade to?  

For example, 
```diff
-    "eslint": "5.14.1",
+    "eslint": "5.16.0",
```
but the latest version is `5.40.1`

yurishkuro (approved):

mmorel-35:
> I suggest removing all pure styling changes. The PR is currently nearly impossible to review in GH Web UI, it crumbles. But more importantly, the styling changes obscure the real code changes that may have been necessary due to dependency upgrades. This is just a bad practice to combine such changes.

Ok, I just did it

> I would like to understand the scope of the change. The title says "all deps", but I see that typescript version is still 3.x (dependabot has been trying to upgrade to 4.x for some time, there are some things that are breaking). And the other major one is react, which still seems to be at "react": "^16.3.2"

The title and description are not enoughly clear so let me try to explain. I tried to upgrade every possible dependencies except for the huge ones like webpack , react and typescript. 
This is a preparation work so it becomes easier to deal with those big ones.

From what I have seen some dependencies like react-vis and react-vis-force  are unmaintained and depend on old version of  react so there is going to be a need to integrate it inside of jaeger of move to another project

I could also see that jaegertracing/plexus is a worker-loader which is deprecated in v5, see https://v4.webpack.js.org/loaders/worker-loader/
Jaeger ui  is also relying on customize-cra which is depending on babel-plugin which is deprecated. 

So there are several questions to address before being able to upgrade to latest versions of thoses

mmorel-35:
> @mmorel-35 could you elaborate on your method of selecting the versions to upgrade to?
> 
> For example,
> 
> ```diff
> -    "eslint": "5.14.1",
> +    "eslint": "5.16.0",
> ```
> 
> but the latest version is `5.40.1`

I just used dependabot on my own repository and that is the pr I received

yurishkuro:
Thanks!

PR #1003, opened 2022-10-21
Provide token for codecov-action

mmorel-35:
Signed-off-by: Matthieu MOREL <matthieu.morel35@gmail.com>

- Lack of codecov token

- This provides a way to use the codecov action with the organisation token [as recommended by codecov](https://github.com/codecov/codecov-action#usage)

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -24,5 +24,6 @@ jobs:
       - name: Upload coverage to codecov.io
         uses: codecov/codecov-action@v3
         with:
+          token: ${{ secrets.CODECOV_TOKEN }}
           fail_ci_if_error: true
           verbose: true


.github/workflows/unit-tests.yml
```diff
@@ -24,5 +24,6 @@ jobs:
       - name: Upload coverage to codecov.io
         uses: codecov/codecov-action@v3
         with:
+          token: ${{ secrets.CODECOV_TOKEN }}
```
yurishkuro, line 27: they say for public repos no token is needed. Also, if we did this ^, then PRs won't be able to upload coverage because secrets are only passed for runs on the primary repo, not on forks.

mmorel-35: Even if jaegertracing and I don’t have a secret Token in our settings but with the token field defined the step is passing while without the other dependabot pr are failing , why is that happening then ?

yurishkuro: The problem is that those PRs are failing sporadically, it's not consistent on the presence of the token. It seems to be related to rate limiting on Codecov side (see https://github.com/codecov/codecov-action/issues/837)

PR #1013, opened 2022-11-05
Upgrade many dependencies part 2

mmorel-35:

It upgrades several more dependencies, including webpack 5 and react 16.14
It also increases the number of PR allowed to dependabot on npm dependencies


packages/jaeger-ui/src/components/DeepDependencies/Header/HopsSelector/__snapshots__/Selector.test.js.snap
```diff
@@ -115,7 +115,7 @@ exports[`Selector renders buttons with expected text and classNames 1`] = `
     >
       4
     </button>
-    <Component
+    <Memo(ChevronDown)
```
yurishkuro, line 118: is there a link to docs that explains this change?

mmorel-35: I couldn't find one, all that I found was https://github.com/enzymejs/enzyme/blob/master/docs/guides/migration-from-2-to-3.md and https://github.com/enzymejs/enzyme/blob/master/CHANGELOG.md but it seems there is not much about those Memo()

yurishkuro:
thanks

PR #1053, opened 2022-11-12
Upgrade to antd/v3.9.0 to avoid loading fonts from alicdn

### LINKED ISSUE #187: Remove the dependency on `alicdn.com`

yurishkuro:
Resolves #187

Despite the concerns in the original ticket, it does not seem to affect the bundle size, because SVG icons are inlined, so the full font file is not added to the bundle.
```
$ du -h cmd/query/app/ui/actual/
3.8M	cmd/query/app/ui/actual/
```


yurishkuro:
@albertteoh the most changes are in packages/jaeger-ui/src/components/Monitor/ServicesView/operationDetailsTable/__snapshots__/index.test.js.snap. How much are the icons used in there? I'm surprised antd inlines the SVGs instead of defining them in the style. If this happens many times per row, it might result in a lot of bulk. But if it's only in the headers of the table then should be ok.

albertteoh:
> How much are the icons used in there?

Yes, those icons only exist in the table's header.

Based on the diff, the changes relate to just the table header (the "info" icon and the up/down caret icons for sorting the rows). My understanding of why there are a lot of changes is that there's ~10 snapshots (from what I see in the diff), with each snapshot having 2*5 up/down caret updates + 1 "info" icon; then each icon change taking up ~20 lines of changes.

So a back of the envelope calc is about 10 * 11 * 20 ~= 2200 which is around the vicinity of the line count in the diff on that file (2471).

PR #1117, opened 2022-12-14
Prevent wrapping of the timeline tick label text

SOF3:
When the timeline is more than one minute, the ticks would line wrap:
image

Setting `width` to `max-content` solves this problem:
image


SOF3:
What about using `white-space: nowrap`? Should have the equivalent effect.

yurishkuro:
That sounds better

yurishkuro:
Sorry for delay, thanks for contributing.

Issue #998, opened 2022-10-09
Upgrade React version - LFX Mentorship


yurishkuro:
For the application process please refer to [this issue](https://github.com/jaegertracing/jaeger/issues/4457).

- Description: Jaeger UI is built on React. While we are seemingly already on v18.x of React, the upgrade was not done across the board and some other dependencies are still lagging behind, e.g. `"@types/react": "16.8.7"`. It's also blocking upgrades of other dependencies. This project is likely to involve a substantial amount of code contribution, as certain upgrade require fixing the code to use the new APIs, and sometimes we may run into dependencies that are EOL and need to be replaced altogether.
- Expected Outcome: Ideal outcome is to have _all_ dependencies upgraded to the latest versions (with the help of @dependabot) and fix all deprecation warnings during the build. But incremental progress towards that goal is also acceptable.
- Recommended Skills: Javascript, Typescript, NPM, Yarn, Vite.js
- Mentor(s):
  - Yash Sharma (yashrsharma44@meta.com)
  - Yuri Shkuro (@yurishkuro)
- Selected mentee:
  - @anshgoyalevil

ShaswatPrabhat:
Is this dependent on other upgrades like the ones we have for graph libs ? @yurishkuro

yurishkuro:
@ShaswatPrabhat I don't know

yurishkuro:
I think #1173 is only part of the story. A lot of dependabot attempts to upgrade other deps are failing, and among them several are failing due to an indirect upgrade of `@types/react` (e.g. from "16.8.7" to "16.14.34"), which causes a lot of type compatibility errors. I assume the versioning of `@types/react` should match that of React itself, i.e. we need to target `@types/react@18.x`

mszabo-wikia:
Looks like there's a fun issue... the `jaeger-ui` subpackage is relying on a transitive dependency on `@types/react` for its typechecking. This is currently locked to version 16, and yarn by default hoists it to the root. The current TypeScript has no qualms with this, but TypeScript >= 3.7 starts to complain that there are newer React typings in Plexus' dependencies which aren't compatible with typings in some of the older dependencies for `jaeger-ui` such as `antd` or `react-router-dom`. The typings were updated for React 18 on the `antd` v4 release line, but that one won't work with TypeScript 3.5.3.

I'm thinking we could roll back the react typing versions for now and try to see how far ahead we can get with a TS upgrade.

yurishkuro:
[...]

PR #1173, opened 2023-02-04
Upgrade to React 18

Sergio-Mira:
Upgrade to React 18.

- Resolves https://github.com/jaegertracing/jaeger-ui/issues/998

- Upgrade to React 18


packages/jaeger-ui/package.json
```diff
@@ -85,10 +85,10 @@
     "prop-types": "^15.5.10",
     "query-string": "^6.3.0",
     "raven-js": "^3.22.1",
-    "react": "^16.14.0",
-    "react-circular-progressbar": "^2.0.3",
-    "react-dimensions": "^1.3.0",
-    "react-dom": "^16.14.0",
+    "react": "^18.2.0",
```
yurishkuro, line 88: Wow, that looks suspiciously easy, were there really no breaking changes?

Sergio-Mira: The breaking changes of lifecycle etc have been postponed to upcoming versions of React 18 and are still warnings afaik, and that is why its an easier upgrade. As far as I understand and can test the project still can use dependencies which are not yet fully React 18 compatible (like react-vis in this case).

Would love to hear how to move this PR forward and what we can do to test it, what edge cases of libs we think might break.

(I've also submitted changes to migrate from old lifecycle events in react-vis here https://github.com/uber/react-vis/pull/1473)

packages/jaeger-ui/src/setupTests.js
```diff
@@ -20,7 +20,7 @@ rafPolyfill();
 
 /* eslint-disable import/no-extraneous-dependencies */
 const Enzyme = require('enzyme');
-const EnzymeAdapter = require('enzyme-adapter-react-16');
+const EnzymeAdapter = require('@wojtekmaj/enzyme-adapter-react-17');
```
yurishkuro, line 23: I can see this is another future PITA to be solved -- https://testing-library.com/docs/react-testing-library/migrate-from-enzyme/

yurishkuro (approved):
did a brief test of different views, don't see any immediate issues, so let's give it a try

metasal1:
Any plans to merge?

yurishkuro:
@metasal1 merge what? This PR was merged in Feb

Issue #578, opened 2020-05-16
Make the trace ID and span ID accessible to different types of link patterns


omallo:
## Requirement - what kind of business use case are you trying to solve?

I would like to be able to link from the Jaeger UI to some log storage backend in order to search for logs of an individual span.

Note: This can already be done on a trace level but not on a span level.

## Problem - what in Jaeger blocks you from solving the requirement?

The Jaeger UI provides different types of [link patterns](https://www.jaegertracing.io/docs/1.18/frontend-ui/#link-patterns) but the span ID seems not to be exposed and the trace ID is only exposed for links of type `traces`.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

In order to accommodate my specific use case, and maybe also others, I would suggest the following extensions to the Jaeger UI:
 - Make the already existing `traceID` field also accessible to link patterns of the types `process`, `tags`, `logs`.
 - Introduce a new `spanID` field and make it accessible to link patterns of the types `process`, `tags`, `logs`.

## Any open questions to address

Does the proposed solution make sense for the members of the project?

yurishkuro:
makes total sense, I'm surprised it's not the case already.

bforbis:
I would also like to add that having a `spanStartTime` and `spanEndTime` would be useful to include for patterns of type `process`, `tags`, and `logs`.

The use case I have is that I have an ElasticSearch documentID as a tag in my span, but it would be very useful to also know the timestamps as it will let me create a deep link directly to the ES daily index that contains that ES document without having to search across a potentially large date range (equal to the jaeger trace retention period?).

yurishkuro:
PRs are welcome.

ashutosh887:
@yurishkuro @omallo I would like to work on this
please assign

yurishkuro:
Feel free to work on it. We do not assign issues - https://github.com/jaegertracing/jaeger/blob/main/CONTRIBUTING_GUIDELINES.md#assigning-issues

7908837174:
Hi @yurishkuro, I'd love to contribute to issue #578! I'll work on making traceID and spanID accessible to different link patterns in Jaeger UI. Please let me know if I can start!

PR #1178, opened 2023-02-04
Make traceID, startTime, endTime, duration and traceName available fo…

MUI-Pop:
This change makes trace related keys like traceID, traceName, startTime, endTime and duration be available for Link Patterns.

It partially solves part of https://github.com/jaegertracing/jaeger-ui/issues/578

- Pass 'trace' object as additional parameter to LinksPatterns GetLink, there by trace variable would be accessible when computing link patterns for process, tags and logs type.


packages/jaeger-ui/src/model/link-patterns.test.js
```diff
@@ -354,18 +384,42 @@ describe('computeLinks()', () => {
       url: 'http://example.com/?myKey=#{myOtherKey}&myKey=#{myKey}',
       text: 'second link (#{myOtherKey})',
     },
+    {
+      type: 'tags',
+      key: 'myThirdKey',
+      url: 'http://example.com/?myKey1=#{myKey}&myKey=#{myThirdKey}&traceID=#{traceID}&startTime=#{startTime}',
```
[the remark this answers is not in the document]
yurishkuro: Reading more into the existing code I am not a big fan of how it automatically ascends into ancestor spans when looking for tags. https://github.com/jaegertracing/jaeger-ui/issues/1179

packages/jaeger-ui/src/model/link-patterns.tsx
```diff
@@ -110,6 +110,22 @@ export function getParameterInAncestor(name: string, span: Span) {
     }
     currentSpan = getParent(currentSpan);
   }
+
+  return undefined;
+}
+
+export function getParameterInTrace(name: string, trace: Trace | undefined) {
```
yurishkuro, line 117: looks like an overlap with `computeTraceLink`, can we find a way to do this only once?

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
```diff
@@ -316,7 +316,10 @@ export class VirtualizedTraceViewImpl extends React.Component<VirtualizedTraceVi
     return DEFAULT_HEIGHTS.detail;
   };
 
-  linksGetter = (span: Span, items: KeyValuePair[], itemIndex: number) => getLinks(span, items, itemIndex);
+  linksGetter = (span: Span, items: KeyValuePair[], itemIndex: number) => {
```
yurishkuro, line 319: can this be reflected in `VirtualizedTraceView.test.js` ?

yurishkuro (commented):
Overall looks good, but I have some suggestions, some to address in this PR, and a larger scoped issue https://github.com/jaegertracing/jaeger-ui/issues/1179

Q: why does the description say that "It partially solves part of #578"? Could you please list which aspects it does not solve?

MUI-Pop:
I said it partial because the current change makes trace related data accessible to logs, process and tags link pattern types but not span related data.

packages/jaeger-ui/src/model/link-patterns.test.js
```diff
@@ -354,18 +384,42 @@ describe('computeLinks()', () => {
...
+      text: 'third link (#{myThirdKey}) for traceID - #{traceID}',
+    }
   ].map(processLinkPattern);
 
   const spans = [
     { depth: 0, process: {}, tags: [{ key: 'myKey', value: 'valueOfMyKey' }] },
     { depth: 1, process: {}, logs: [{ fields: [{ key: 'myOtherKey', value: 'valueOfMy+Other+Key' }] }] },
+    { depth: 1, process: {}, tags: [{ key: 'myThirdKey', value: 'valueOfThirdMyKey' }] },
```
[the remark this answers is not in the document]
MUI-Pop: Fixed - added a new to existing span.

packages/jaeger-ui/src/model/link-patterns.test.js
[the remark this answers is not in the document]
MUI-Pop: Fixed - now trace level are accessible only with `trace.<data>` qualification in template variables.

packages/jaeger-ui/src/model/link-patterns.tsx
MUI-Pop: Addressed the feedback, let me know if it looks OK.

packages/jaeger-ui/src/model/link-patterns.tsx
```diff
@@ -119,20 +140,17 @@ function callTemplate(template: ProcessedTemplate, data: any) {
...
-  );
 
   linkPatterns
     .filter(pattern => pattern.type('traces'))
     .forEach(pattern => {
       const parameterValues: Record<string, any> = {};
       const allParameters = pattern.parameters.every(parameter => {
-        const key = parameter as keyof Trace;
-        if (validKeys.includes(key)) {
+        const val = getParameterInTrace(parameter, trace);
```
yurishkuro, line 149: it looks like `validKeys` will now be re-calculated for every pattern and every parameter. Perhaps this needs a helper class:
```
const traceParams = TraceParams(trace);
// in a loop
  const val = traceParams.getParameter(parameter);
```

packages/jaeger-ui/src/model/link-patterns.tsx
```diff
@@ -110,6 +110,27 @@ export function getParameterInAncestor(name: string, span: Span) {
...
+}
+
+const getValidTraceKeys = memoize(10)((trace: Trace) => {
+  const validKeys = (Object.keys(trace) as (keyof Trace)[]).filter(
+    key => typeof trace[key] === 'string' || typeof trace[key] === 'number'
+  );
+  return validKeys;
+});
+
+export function getParameterInTrace(name: string, trace: Trace | undefined) {
```
yurishkuro, line 124: it looks like the other similar functions return `{key, value}`, not just one value. Also, you later check `if(val)`, but it's possible that the value is 0 or false. Returning a dictionary vs. undefined would fix that as well.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.test.js
```diff
@@ -430,4 +431,22 @@ describe('<VirtualizedTraceViewImpl>', () => {
...
+    });
+
+    afterAll(() => {
+      getLinksSpy.mockRestore();
+    });
+
+    it('calls getLinks with expected params', () => {
+      const span = trace.spans[1];
+      instance.linksGetter(span, span.tags, 0);
+      expect(getLinksSpy).toHaveBeenCalledWith(span, span.tags, 0, trace);
```
yurishkuro, line 449: it would be better to validate the actual value returned, which, as I understand, is `[{url, text}]`

packages/jaeger-ui/src/model/link-patterns.tsx
MUI-Pop: There is a helper function `getValidTraceKeys()` which solves memoize the returned value for a trace object, let me know if that is not ok.

yurishkuro: ah, indeed, I forgot the memoize - probably ok then.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.test.js
MUI-Pop: I didn't find a clean way to initialize `linkPatterns` for tests, so I exported the `const processedLinks` in `link-patterns.tsx`.

packages/jaeger-ui/src/model/link-patterns.tsx
```diff
@@ -169,16 +188,28 @@ export function computeLinks(
...
-          // do not look in ancestors for process tags because the same object may appear in different places in the hierarchy
-          // and the cache in getLinks uses that object as a key
-          entry = getParameterInAncestor(parameter, span);
+        let entry;
+
+        if (parameter.startsWith('trace.')) {
+          const traceKV = getParameterInTrace(parameter.split('trace.')[1], trace);
+          if (traceKV) {
+            entry = traceKV;
+          }
```
yurishkuro, line 197:
```suggestion
          entry = getParameterInTrace(parameter.split('trace.')[1], trace);
```

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.test.js
```diff
@@ -430,4 +431,40 @@ describe('<VirtualizedTraceViewImpl>', () => {
...
+    beforeEach(() => {
+      getLinks.processedLinks.splice(0, getLinks.processedLinks.length);
+    });
+
+    afterAll(() => {
+      getLinks.processedLinks.splice(0, getLinks.processedLinks.length);
+      getLinks.processedLinks.push(...origLinkPatterns);
+    });
+
+    it('calls getLinks with expected params', () => {
```
yurishkuro, line 450: stale title

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.test.js
```diff
@@ -22,6 +22,7 @@ import { DEFAULT_HEIGHTS, VirtualizedTraceViewImpl } from './VirtualizedTraceVie
 import traceGenerator from '../../../demo/trace-generators';
 import transformTraceData from '../../../model/transform-trace-data';
 import updateUiFindSpy from '../../../utils/update-ui-find';
+import * as getLinks from '../../../model/link-patterns';
```
yurishkuro, line 25: nit
```suggestion
import * as linkPatterns from '../../../model/link-patterns';
```

yurishkuro (approved):
lgtm, just a couple minor comments.

> I didn't find a clean way to initialize linkPatterns for tests, so I exported the const processedLinks in link-patterns.tsx.

yeah, the API feels unnecessarily complicated, it could've just accepted processed patterns as optional arg, since the methods to process patterns are exported anyway.

yurishkuro (approved):

yurishkuro:
Thanks!  🎉   

Could you also make a PR in the documentation repo to describe this change in https://www.jaegertracing.io/docs/1.42/frontend-ui/#link-patterns ?

PR #1210, opened 2023-02-23
Bump Plexus React typings to React 18

mszabo-wikia:
Contributes towards #998

Bump Plexus React typings to React 18 and also update the package's `peerDependencies` accordingly. I am not entirely sure why Dependabot was trying to update these typings separately because bumping them together appears to yield a successful build.


yurishkuro:
🎉

Issue #1199, opened 2023-02-14
[Feature]: 


yurishkuro:
### Requirement

https://github.com/jaegertracing/jaeger-ui/pulls?q=is%3Apr+is%3Aopen+label%3Adependencies

### Problem

https://github.com/jaegertracing/jaeger-ui/pulls?q=is%3Apr+is%3Aopen+label%3Adependencies

SaarthakMaini:
I think this would be good for reference:

https://docs.github.com/en/code-security/dependabot/working-with-dependabot/troubleshooting-dependabot-errors

For the following PR:

https://github.com/jaegertracing/jaeger-ui/pull/1200

I think the problem is in the 

- [packages/plexus/demo/src/SimpleGraph.tsx](https://github.com/jaegertracing/jaeger-ui/blob/f4127669cabe16d661427acad2e4035a32947338/packages/plexus/demo/src/SimpleGraph.tsx)
- [packages/plexus/demo/src/UxEdges.tsx](https://github.com/jaegertracing/jaeger-ui/blob/f4127669cabe16d661427acad2e4035a32947338/packages/plexus/demo/src/UxEdges.tsx)
- [packages/plexus/demo/src/index.tsx](https://github.com/jaegertracing/jaeger-ui/blob/f4127669cabe16d661427acad2e4035a32947338/packages/plexus/demo/src/index.tsx)

ReactDOM.render is no longer supported in React 18. Use createRoot instead. Until you switch to the new API, your app will behave as if it’s running React 17.

Please refer to this [blog](https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#updates-to-client-rendering-apis:~:text=ReactDOM.render%20is%20no%20longer%20supported%20in%20React%2018.%20Use%20createRoot%20instead.%20Until%20you%20switch%20to%20the%20new%20API%2C%20your%20app%20will%20behave%20as%20if%20it%E2%80%99s%20running%20React%2017.) for more details

If possible, please explain how we can help in cleaning this up and keeping the Jaeger UI up to date?

Thank You!

yurishkuro:
@SaarthakMaini thanks for offering to help. I suspect your analysis above may apply to some upgrades, but others might require other changes.

[...]

Issue #818, opened 2021-09-23
Typescript version is outdated


yoave23:
## Requirement - what kind of business use case are you trying to solve?

We'd like to upgrade the typescript version from 3.5.3 to the latest one (4.4)

## Problem - what in Jaeger blocks you from solving the requirement?

the current version is missing key features

Related: #998

yurishkuro:
If you have cycles & expertise, go for it.

yoave23:
@yurishkuro sure, we'll take it

PR #1212, opened 2023-02-25
Migrate build tooling to Vite

mszabo-wikia:
- Contributes towards #818, #1199

`jaeger-ui` currently uses build tooling based on `react-scripts` (née CRA) 2.1.3, which dates back to early 2019. Since `react-scripts` is an opinionated wrapper around other tooling such as `jest` and `webpack`, it in turn keeps those locked on outdated versions.

There are a few options I considered here:

* **Do nothing** - ignore the outdated `react-scripts` and try to upgrade other things without touching it. This is what I tried first, when I tested out migrating some tests to `@testing-library` from `enzyme`. Unfortunately, I found that the outdated `jest` / `jsdom` combo precluded effective use of that library as some events and APIs were simply not available.
* **Try to update `react-scripts`**. Unfortunately, there is a problem here - although `react-scripts` is intended to be a zero-configuration wrapper around CRA tooling, the project attempts to modify its settings via two different packages (`react-scripts-rewired` and `customize-cra`) that power a custom buildscript which attempts to manipulate the webpack configs normally abstracted away by `react-scripts`. Given that all this hassle is done only to perform some mundane tasks like integrating with AntD's Less styles, or making the build step respect the project's ESLint config, I am not sure if staying with `react-scripts` and trying to customize it with a bag of tricks is a sustainable approach.
* **Eject the `react-scripts` and work with that**. `react-scripts` has an escape hatch that dumps the build tooling it uses internally in the project, should you decide that you want to roll your own build. The problem I saw with this is that this simply leaves us with a bunch of messy webpack config files that we then need to maintain going forward.
* **Migrate to a different build tool**. There are some promising tools out there that try to strike a reasonable balance between ease of configuration and use and extensibility, such as Vite and Parcel. I ended up settling on this approach after exhausting the above options, using Vite as the bundler of choice. I also gave Parcel a try but unfortunately I ran into issues trying to get it to transpile the AntD Less stylesheet, so I ended up not exploring it further for now.

So, per the above points, this PR reconfigures the build step to use Vite rather than `react-scripts`.

There are a few caveats I ran into:
* Vite by default doesn't interpret JSX tags in .js files. This proved surprisingly difficult to override, so I renamed the few affected files to `.tsx` (if migrating didn't require sweeping changes) or `.jsx` otherwise.
* Consuming the plexus build resulted in some errors due to ambiguous exports, so I reconfigured it to output ESM rather than UMD.
* The abandoned `react-dimensions` package, used to automatically size/resize the scatterplot diagram on the trace search page, threw an error at runtime due to gratuitous use of `this`. I replaced it with a simple hooks-based implementation.
* The `rc-menu` component used internally by AntD would crash the entire page due to it merrily using `require()` at runtime and assuming it would be present in the environment. I ended up just providing a basic `require()` stub, since it was only using it to load a `MutationObserver` polyfill. Fortunately, the problematic code was removed in newer versions of the component.
* The customized `react-scripts` setup used a Babel plugin to automatically import styles whenever an AntD React component was imported. This also seems doable with Vite; for simplicity and to reduce the number of build deps, I haven't configured it though. If the bundle size impact is considered prohibitive, this can be revisited.

Overall, I am cautiously optimistic about Vite - the developer experience seems significantly better as the HMR server is notably snappier, at least on my machine. Most of the above issues stemmed from the fact that many dependencies are outdated (and therefore more likely to contain code that doesn't play that well with present tooling). This change also allowed me to migrate the test suite from Jest v23 to Jest v28 with minimal changes. Jest v29 unfortunately doesn't yet work as it doesn't play well with the old TypeScript version currently used by the project.

It may be worth noting that the `react-scripts` build step would also run ESLint at build time. This is doable with Vite as well, I'm just not sure if it's worth it, since the linter is already executed separately in continuous integration. It therefore may not be necessary to have it slow down the build.


mszabo-wikia:
Any feedback is welcome. If there is something I'm not, it's an expert in modern frontend.

packages/jaeger-ui/src/components/Monitor/ServicesView/index.tsx
```diff
@@ -455,7 +457,7 @@ export default connect(
   mapStateToProps,
   mapDispatchToProps
 )(
-  reduxForm({
```
mszabo-wikia, line 458: These type errors popped up after removing `react-scripts`. What's weird about them is that they seem valid—the prop types of `MonitorATMServicesViewImpl` don't actually match what's coming in from the HOC, so I am not sure why these were not failing before.

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.tsx
```diff
@@ -16,15 +16,14 @@
```
mszabo-wikia, line 16: This file was apparently using Flow types, so I ended up converting it to TS instead of just renaming to `.jsx`.

package.json
```diff
@@ -37,12 +37,7 @@
       "packages/*"
     ],
     "nohoist": [
-      "**/customize-cra",
-      "**/customize-cra/**",
-      "**/react-scripts",
-      "**/react-scripts/**",
-      "**/react-app-rewired",
-      "**/react-app-rewired/**"
+      "**/parse5"
```
mszabo-wikia, line 40: This one is here because it contains type definitions with syntax that TypeScript 3.5 cannot comprehend. It could conceivably be removed once TS is upgraded.

yurishkuro:
@mszabo-wikia I will need to look deeper into Vite, but so far seems like a reasonable choice.

However, do you think it's possible to partition the changes in this PR such that the changes to the source code (typing, renaming) could be committed separately, before making changes to the build?

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ScatterPlot.jsx
```diff
@@ -0,0 +1,120 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
yurishkuro, line 1: It would be good to undo this file's change and redo it such that Git would recognize the move, otherwise it's impossible to see what changed.

packages/jaeger-ui/src/utils/tracking/ga.tsx
```diff
@@ -28,11 +28,11 @@ const isTruish = (value?: string | string[]) => {
 };
 
 const GA: IWebAnalyticsFunc = (config: Config, versionShort: string, versionLong: string) => {
-  const isProd = process.env.NODE_ENV === 'production';
-  const isDev = process.env.NODE_ENV === 'development';
-  const isTest = process.env.NODE_ENV === 'test';
+  const isProd = __APP_ENVIRONMENT__ === 'production';
+  const isDev = __APP_ENVIRONMENT__ === 'development';
+  const isTest = __APP_ENVIRONMENT__ === 'test';
```
yurishkuro, line 33: nit: I would prefer to future-proof this a bit and isolate all env var access to a util module, so that we could change it in one place instead of many files. The "isolate" part can be a standalone PR merged first.

mszabo-wikia:
@yurishkuro Thanks for the quick review. I split up the changeset so that it's hopefully easier to review :)

yurishkuro:
> I split up the changeset so that it's hopefully easier to review :)

Do you mean you split the in the commits only? I was referring to making independent PRs, because some of the refactoring of types etc. seems like it would be applicable to the current build system as well, so I would prefer to merge them independently, and only they have a PR that changes the build system. This way it will be easier to revert changes should the need arise.

mszabo-wikia:
Ah sorry, I misunderstood. Absolutely, I'll try and split out those standalone parts into separate PRs.

mszabo-wikia:
Going to close this since most changes have been split out. I hope to have a PR up for the final build system switch momentarily.

PR #1219, opened 2023-02-27
Make Plexus output ESM

mszabo-wikia:
Split from https://github.com/jaegertracing/jaeger-ui/pull/1212

Plexus currently outputs a Universal Module Definition, which seems to cause issues during the Vite build as it results in "ambiguous exports". Have Plexus output ESM instead to avoid the issue.

### DIFF
--- a/packages/plexus/webpack-factory.js
+++ b/packages/plexus/webpack-factory.js
@@ -195,13 +195,14 @@ function makeCommonProdConfig() {
 function makeWorkerConfig() {
   const layoutDir = join(__dirname, 'src/LayoutManager');
   const config = {
+    experiments: {
+      outputModule: true,
+    },
     output: {
       path: layoutDir,
       publicPath: '/',
       filename: '[name].bundled.js',
-      library: 'layout.worker.bundled',
-      libraryTarget: 'umd',
-      umdNamedDefine: true,
+      libraryTarget: 'module',
     },
     entry: {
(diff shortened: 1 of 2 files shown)


yurishkuro:
what is the implication of this? My brief read-up indicates that UMD has broader compatibility. What are we losing by switching to ESM?

packages/plexus/webpack-factory.js
```diff
@@ -195,13 +195,14 @@ function makeCommonProdConfig() {
...
   const layoutDir = join(__dirname, 'src/LayoutManager');
   const config = {
+    experiments: {
+      outputModule: true,
+    },
     output: {
       path: layoutDir,
       publicPath: '/',
       filename: '[name].bundled.js',
-      library: 'layout.worker.bundled',
```
yurishkuro, line 202: I remember there was some weirdness with this layout.worker - did you try running the deep dependency graphs after this change?

mszabo-wikia: Yeah, I tested rendering a DDG from a proxied-in Jaeger deployment that already had one, and it seems to load and render okay.

mszabo-wikia:
@yurishkuro As far as I understand, the current build system will just consume this dependency via ESM and bundle/transform it as needed; same for the Vite production build. The main error this should fix is in the Vite dev run, where it seems that the ESM interop code that wraps the output in the UMD case isn't handled well and results in the above error.

yurishkuro (approved):

PR #1222, opened 2023-02-27
Convert SearchResults and index components to TS

mszabo-wikia:

As with https://github.com/jaegertracing/jaeger-ui/pull/1218, convert SearchResults and the root index component to use TS and the .tsx file extension rather than plain .js. TS was chosen here since SearchResults was already using Flow types, and index is trivially convertible.

Adding `@types/react-form` as a new dev dependency to provide typings for `redux-form` in `SearchResults` in turn revealed type errors in the usage of `redux-form` in the `ServicesView` component, so update that component accordingly. Likewise, the conversion from Flow to TS revealed some typing errors in child components.

Also update the list of JS files to be included in the TS linting process - since we now have two new TS files to be built and linted, the JS files that these import in turn need to be explicitly listed in tsconfig.lint.json, otherwise `tsc` complains that they are not included in the build.


packages/jaeger-ui/tsconfig.lint.json
```diff
@@ -28,11 +28,29 @@
   "files": [
```
yurishkuro, line 28: this feels kinda insane that we need to list the files like this. Isn't there a way to use a glob like `**/*.js` ?

yurishkuro (approved):

packages/jaeger-ui/tsconfig.lint.json
mszabo-wikia: That works as well. I just wanted to be explicit here (ideally I'd imagine this list will eventually go away anyways as things are converted to TS).

yurishkuro: my concern is more about missing files here and, as a result, accumulating more lint issues over time.

mszabo-wikia: AIUI `tsc` won't actually lint these files, since they are JS and we don't have `checkJs` specified in `compilerOptions`—it will however complain if they are absent and a TS file tries to import them.

yurishkuro:
please resolve conflict

yurishkuro (approved):

PR #1223, opened 2023-02-27
Remove use of `react-dimensions` package in ScatterPlot component

mszabo-wikia:

This package is abandoned and doesn't play well with Vite, because one of its dependencies uses `this` to access the global context in a way that apparently fails in a module context. Replace this dependency with a simple hooks-based implementation.

### DIFF
--- a/packages/jaeger-ui/package.json
+++ b/packages/jaeger-ui/package.json
@@ -88,7 +88,6 @@
     "raven-js": "^3.22.1",
     "react": "^18.2.0",
     "react-circular-progressbar": "^2.1.0",
-    "react-dimensions": "^1.3.1",
     "react-dom": "^18.2.0",
     "react-ga": "^3.3.1",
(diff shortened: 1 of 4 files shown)


yurishkuro (commented):
could you please describe how this was tested (screenshots?)

mszabo-wikia:
So what I did here was checking the dimensions of the scatterplot in a normal situation (fullscreen window) and verifying it's properly set to an absolute value:
Screenshot 2023-03-01 at 01 34 57

Then I reduced the window size and verified that the scatterplot width is reduced as the window shrinks:
Screenshot 2023-03-01 at 01 35 11

Then I restored the original window size (fullscreen) and verified that the scatterplot expands again to fill the available width.

I compared this with the behavior on an existing Jaeger 1.42 instance—the dimensions seem to match.

yurishkuro (approved):
Thanks!

PR #1224, opened 2023-02-28
Abstract access to injected globals

mszabo-wikia:

Create a wrapper module for accessing globals injected by the build system. This makes it easier to rename or adjust those later.

### DIFF
--- a/packages/jaeger-ui/src/utils/constants.tsx
+++ b/packages/jaeger-ui/src/utils/constants.tsx
@@ -1,0 +1,35 @@
+// Copyright (c) 2023 The Jaeger Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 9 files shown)


packages/jaeger-ui/src/utils/constants.tsx
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) 2023 Uber Technologies, Inc.
```
yurishkuro, line 1:
```suggestion
// Copyright (c) 2023 The Jaeger Authors.
```

packages/jaeger-ui/src/utils/version/get-version.test.js
```diff
@@ -34,9 +34,9 @@ describe('getVersion()', () => {
     console.warn = oldWarn;
   });
 
-  describe('`window.getVersion` is not a function', () => {
+  describe('`window.getJaegerVersion` is not a function', () => {
```
yurishkuro, line 37: what is the origin of this change? I only see the test change, not the main code.

mszabo-wikia: It looks like this test was testing for the wrong function name - it should be `window.getJaegerVersion` as per https://github.com/jaegertracing/jaeger-ui/blob/dda3b97edccac392324ddfc4d90487377c1ca47e/packages/jaeger-ui/src/utils/version/get-version.tsx.

yurishkuro (approved):
👍

Issue #1211, opened 2023-02-23
[Feature]: Additional configuration to prevent the creation of a new tab and use of file upload controls


Binrix:

Binrix:
Thanks a lot! I'm currently working on the implementation of the feature.

Binrix:
@yurishkuro To push my changes do I need special rights from you to do so? When I try to push my changes git says that permission have been denied.

yurishkuro:
You need to fork the repo, push changes to a branch in the fork, and create a pull request. When you push to a fork, git will print a URL to create a PR.

yurishkuro:
@Binrix see https://github.com/jaegertracing/jaeger/blob/main/CONTRIBUTING_GUIDELINES.md#making-a-change

udita-0707:
@yurishkuro, could you please confirm whether this issue has already been resolved? I'd be interested in working on it if it's not already resolved.

yurishkuro:
I believe this has been resolved by the attached PRs.

PR #1229, opened 2023-03-02
Feature/config option for new page and file upload controls

Binrix:
This pull request resolves the issue #1211 

Add two new properties to the configuration. One property is being used to prevent the creation of a new page. The other one is used for blocking the file upload control and preventing the user to access the file system.

Sorry for the 2 contributors! Both are me, I somehow messed up the credentials.


packages/jaeger-ui/src/constants/default-config.tsx
```diff
@@ -27,49 +27,49 @@ export default deepFreeze(
...
-            {
-              label: `Build ${getVersion().buildDate}`,
-            },
-            {
-              label: `Jaeger UI v${version}`,
-            },
-          ],
-        },
-      ],
+       menu: [
```
yurishkuro, line 30: why this formatting change?

yurishkuro (commented):
Please fix conflicts.

I will approve the DCO check because all commits do include proper signature. You may want to ensure your git config matches your GitHub full name & email.

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ResultItemTitle.test.js
```diff
@@ -86,6 +86,11 @@ describe('ResultItemTitle', () => {
       expect(wrapper.find(Link).length).toBe(0);
     });
 
+    it('<Link> does not target _blank because forbidNewPage is false', () => {
```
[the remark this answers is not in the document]
Binrix: You are totally right, thanks for pointing this out. I will fix this.

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ResultItemTitle.tsx
```diff
@@ -77,7 +79,9 @@ export default class ResultItemTitle extends React.PureComponent<Props> {
       wrapperProps.to = linkTo;
       WrapperComponent = Link;
       if (targetBlank) {
-        wrapperProps.target = '_blank';
+        if(!forbidNewPage) {
```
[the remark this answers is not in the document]
Binrix: I will change this one, thanks for pointing this out.

packages/jaeger-ui/src/constants/default-config.tsx
Binrix: Somehow my IDE probably messed up. I will restore the old formatting!

packages/jaeger-ui/src/components/TracePage/TracePageHeader/AltViewOptions.test.js
```diff
@@ -68,6 +68,16 @@ describe('AltViewOptions', () => {
     expect(wrapper).toMatchSnapshot();
   });
 
+  it('json links are not displayed because forbidnewPagePage is true', () => {
```
yurishkuro, line 71: This seems like unrelated coupling, JSON can still be shown, just open in the same window (and require Back button to get back into the UI)

packages/jaeger-ui/src/components/TracePage/TraceSpanView/index.tsx
```diff
@@ -22,6 +22,7 @@ import { TNil } from '../../../types';
 import { Trace, Span } from '../../../types/trace';
 import { timeConversion } from '../../../utils/date';
 import prefixUrl from '../../../utils/prefix-url';
+import { getConfigValue } from '../../../utils/config/get-config';
```
yurishkuro, line 25: is the import needed? Doesn't look like it's used

yurishkuro:
@Binrix do you plan to follow-up on this?

Binrix:
@yurishkuro Yes, I will now continue to work on this. I'm done with my final exam and I'm back from the holidays.

packages/jaeger-ui/src/components/SearchTracePage/index.jsx
```diff
@@ -246,9 +253,10 @@ const stateServicesXformer = memoizeOne(stateServices => {
 
 // export to test
 export function mapStateToProps(state) {
-  const { embedded, router, services: stServices, traceDiff } = state;
+  const { embedded, router, services: stServices, traceDiff, config } = state;
   const query = getUrlState(router.location.search);
   const isHomepage = !Object.keys(query).length;
+  const { forbidNewPage, disableFileUploadControl } = config;
```
yurishkuro, line 259: what guarantees that `config` won't be undefined here?

packages/jaeger-ui/src/components/TracePage/TraceJSON/index.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import { useEffect, useState } from "react";
+
+const TraceJSON = ({ trace, raw }: any) => {
```
yurishkuro, line 4: Is this related to this PR? Looks like new functionality, can we split it?

Binrix: This is part of the comment about the dropdown entries "Trace Json" and "Trace Json (unadjusted)" where you suggested to add a button to go back when the user clicks on one of the entries.  I just didn't know how to solve the issue with the json window because it was automatically opened and outside of react. I was not able to add a button, so I added a new page where the json data is displayed. If necessary, it can be splitted, yes.

packages/jaeger-ui/src/components/SearchTracePage/index.jsx
Binrix: As far as I was able to see, the default configuration is always being loaded (as long as the file is available under the path "constants/default-config"). So it should be defined. Would it be better when I take care of the case that config is being undefined? Like with an if and if it is undefined providing default values?

packages/jaeger-ui/src/components/TracePage/TraceJSON/index.tsx
yurishkuro: I would like to keep the PR focused on one thing - in this case preventing opening new windows.

I did not mean that we need to implement a Back button - it already exists in the browsers. Just showing the JSON in the current window is enough.

Binrix: Unfortunately, there is no back button. The Jaeger UI is embedded inside a global area surrounded with a notification area that provides access to other apps. I can do it in a way that the JSON will be opened in the current window. So we would remove this change from this PR and then I create a new PR with this solution?

Binrix: Here is an image of one of the instruments:

98_Roche_Cobas_6800_system_01

yurishkuro: Do you need JSON view in your setup? The alternative would be to have an option to disable that functionality. If it's still required, let's do it in a separate PR. I don't have strong objections to rendering JSON ourselves, the main concern is that when traces are very large, rendering JSON manually can choke up the browser, whereas using browser-native view for JSON may still work ok. Browsers also tend to have extensions for viewing JSON with extra features like syntax coloring, expand/collapse, search, etc.- not the things we want to start adding to Jaeger as a custom JSON view.

Binrix: Thanks for the response. Then I will add a new option in the configuration to deactivate the json view because it is not needed (as far as I know).

yurishkuro (dismissed):

yurishkuro:
Could you squash the commits and make sure that the new commit is signed?

yurishkuro (changes requested):
When I did a sanity check testing, searching for traces and then clicking on one of them opened the trace in a new tab. That's not how the UI works currently, something in this PR is changing it, please address.

yurishkuro:
@Binrix I pulled one piece of this into a different PR #1464 to make the changes more manageable and independent. Could you do the same with the JSON Viewer change?

yurishkuro:
Please rebase / resolve conflicts

yurishkuro:
I ran prettier, but the CI is still not happy

yurishkuro (commented):
Hi,

Sorry it took so long to get back to this. This diff make me very uncomfortable because it introduces so much spaghetti code that keeps passing and checking `forbidNewPage` everywhere. We already have an established pattern of passing `targetBlank: Boolean` property, e.g. in ResultItemTitle. Can we not use that as a control mechanism (adding it to the components where necessary) and limiting the impact of `forbidNewPage`  config value only to some top-level pages?

I suppose one argument against that is that it does not preclude someone making a change in the future where they will set `targetBlank=true` explicitly. So then another alternative would be to go the opposite way and push the rendering of `target=` attribute all the way down to a util function which will check the config flag in exactly one place in the code. I am not completely sure what is involved in that, e.g. does it mean we have to push down config to all components? Even so, pushing config seems more preferable than adding `forbidNewPage` to random component properties here and there, for reasons that will immediately become quite obscure to people who will read the code. But with the util function we will have a single well-defined place, with git blame explaining why was done this way.

Binrix:
Did the PR grow so much because of the things the bot did?

Binrix:
I think I will create a new branch and a new PR because it is somehow messed up with the changes of the bots and mine (as you can see 40 files changes, when I only touchet less then 20).  Furthermore the name of the branch does not match anymore

Binrix:
Opened as new PR #1521

PR #1244, opened 2023-03-08
Bump TypeScript to 3.8.3

mszabo-wikia:
- Contributes towards #818

This is a conservative bump to minimize the amount of changes needed in tooling and dependencies. Version 3.8.3 was chosen because it is the version that implements support for type-only exports and imports (i.e. `import type { TFoo } from '@types/foo';`, which is the major source of incompatibility with upgraded type definitions in newer dependency versions.

Notable changes:
* As noted in #998, the `@types/react` version used to type-check `jaeger-ui` is a transitive dependency on version `16.8.7`, rather than the expected `18`. Unfortunately, both newer `16.x` type definitions as well as `18.x` type definitions are incompatible with various dependencies such as `antd`. As a workaround, downgrade the typing versions for now and add an explicit dependency on them to the project. Only index.tsx was using a React 18-specific API (createRoot), so convert it back to JS until the typings can be updated again.
* TypeScript now enforces that `composite` projects must also generate declaration files, since that's what the project references system uses. Make it so. This required a small adjustment to the ErrorMessage component, as the types of its Message and Details sub-components could not be reflected in this declaration file; this was trivially fixable by converting them to named exports instead (which is also consistent with other areas of the codebase).
* Make Plexus a project reference in the root tsconfig, per the longstanding todo and per the changes above.


yurishkuro:
lgtm, any concerns from your side?  Any additional testing required? Bundle size changes?

mszabo-wikia:
The bundle size should be unaffected (seems like it's indeed 1648 kB both on `main` and this branch). I think overall this should be a fairly safe change as the TS upgrade didn't necessitate widespread code changes.

yurishkuro:
👍

PR #1247, opened 2023-03-09
Bump antd to 3.26.20

mszabo-wikia:
- Contributes towards #1199 

With the TS update it now seems possible to update `antd`. There are two eventual benefits:
* antd v4 contains React 18-compatible typings,
* antd v4 supports [treeshakeable SVG icons](https://github.com/ant-design/ant-design/blob/4.24.8/docs/react/migration-v4.en-US.md#icon-upgrade), reducing the bundle size.

The [migration guide](https://github.com/ant-design/ant-design/blob/4.24.8/docs/react/migration-v4.en-US.md) recommends updating to the latest v3 release before going to v4.

Apart from ubiquitous snapshot changes, a few adjustments were necessary in components and tests:
* The test case for the `<Page />` component that aimed to verify that the header was not rendered in embedded mode was only working by accident since 53937eb5acded15f41292f2f1dc0b30eb35fb907—the `Header` component didn't have a display name, so the test wouldn't find it even though the test setup was no longer correct. Fix the test setup to setup the `embedded` prop accordingly.
* antd's `Menu` component is now wrapped in a `Context.Consumer`, so tests using enzyme's shallow rendering need to `dive` into it to be able to find the child components they are looking for.
* The `Upload` component has more strictly typed props now, so the signature of the `loadJsonFile` prop had to be adjusted accordingly. The handling code was only accessing the `file` property of `FileList`, and the real type supplied by the event handler also has a `file` property, so no code changes were required apart from the typing update.
* The `Table` component has one more internal wrapper (each passing down the `className` prop to the next), so the test in `TraceSpanView` had to be updated accordingly.
* The `onCell` event handler for `Table`s is now strictly typed, so `DetailTable` needed an update to return an empty object `{}` if it intends to set no props. This is consistent with the upstream [examples](https://github.com/react-component/table/blob/83cce50ddb0a55d63bb0f435d3f4ff7e3301ebad/docs/examples/colspan-rowspan.tsx#L25).
* `KeyboardShortcutsHelp` was setting a spurious `align` prop on `Modal`, which was already not part of its typings in 3.9.0 but was still present as a prop type. This caused a type error now that the prop type was removed.
* In manual testing, I observed that the name selectors (e.g. "Group By") on the trace statistics view were not functioning—they would immediately close when clicking to open them. I thought this was a regression caused by the update, but experienced the same behavior on `main`. This was fixed by deferring event handler registration to ensure the event handlers don't end up catching the same event that triggered the original state change.

## Screenshots
Screenshot 2023-03-09 at 23 05 09
Screenshot 2023-03-09 at 23 05 28
Screenshot 2023-03-09 at 23 05 45
Screenshot 2023-03-09 at 23 06 07


mszabo-wikia:
Good idea, filed https://github.com/jaegertracing/jaeger-ui/pull/1251 for that.

PR #1256, opened 2023-03-12
Update TypeScript and ESLint to latest versions

### LINKED ISSUE #818: Typescript version is outdated

mszabo-wikia:
- Resolves #818

Update TypeScript, ESLint and `@typescript-eslint` to the latest versions.

Notable changes:
* Update the list of core ESLint rules that need to be disabled because `@typescript-eslint` provides TypeScript-specific equivalents for them, and change existing suppressions referencing the now-disabled core rules. Group these together in the ESLint config for clarity.
* Fix references to unimported types in `serviceGraph.tsx` and `types.tsx`.
* It seems that the type of `defaultConfig` and the user-defined config wasn't correctly inferred as `Config` before the upgrade. Fix various type errors resulting from this:
  * `'__mergeFields'`, i.e. the list of config options where user-specified values should be merged with the default object values in `defaultConfig`, was defined via `defineProperty` and so seemingly inferred as an unknown type. This now causes a type error when iterating over its keys to try and merge the relevant property values, as TypeScript can't guarantee that the property values are objects. Make it a separate named export instead, which also allows for strictly typing the allowable property names.
  * Make the `url` field of the `ConfigMenuItem` type optional as this was now recognized as a type error. This matches documentation and actual usage.
  * Rename the `dagMaxServicesLen` property definition to `dagMaxNumServices` as that is what the actual code uses.
  * Fix the `linkPatterns` config option typedef to be an array, as per actual usage.
 * Update `handleError` in Plexus to correctly take either an `ErrorEvent` or a `MessageEvent` to match updated typings for the Web Worker API.
 * Update `catch` clauses in `readJsonFile` because TypeScript 4.4+ [types the error in `catch` clauses as `unknown` by default](https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#using-unknown-in-catch-variables). We don't expect anything outside of an `Error` to be thrown here, so we can restore the previous behavior.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -48,7 +48,6 @@ module.exports = {
       },
       plugins: ['@typescript-eslint'],
       rules: {
-        'no-unused-vars': 0,
         '@typescript-eslint/naming-convention': [
           'error',
           {
@@ -57,7 +56,16 @@ module.exports = {
             prefix: ['I'],
           },
         ],
+
+        // Disable ESLint core rules for which @typescript-eslint provides TypeScript-specific equivalents.
+        'no-unused-vars': 0,
         '@typescript-eslint/no-unused-vars': 1,
+        'no-use-before-define': 0,
+        '@typescript-eslint/no-use-before-define': 1,
+        'no-redeclare': 0,
+        '@typescript-eslint/no-redeclare': 1,
+        'no-shadow': 0,
+        '@typescript-eslint/no-shadow': 1,
 
         // Disable prop type checks for TSX components, as prop type validation is expected
         // to be handled by TypeScript there. Stray prop types in components converted from Flow
@@ -79,6 +87,7 @@ module.exports = {
     /* general */
     'arrow-body-style': 0,
     'arrow-parens': [1, 'as-needed'],
+    'class-methods-use-this': 0,
     'comma-dangle': 0,
     'lines-between-class-members': ['error', 'always', { exceptAfterSingleLine: true }],
     'no-continue': 0,
--- a/packages/jaeger-ui/src/components/DeepDependencies/Graph/DdgNodeContent/index.tsx
+++ b/packages/jaeger-ui/src/components/DeepDependencies/Graph/DdgNodeContent/index.tsx
@@ -111,6 +111,7 @@ export function getNodeRenderer({
   return function renderNode(vertex: TDdgVertex, _: unknown, lv: TLayoutVertex<any> | null) {
     const { isFocalNode, key, operation, service } = vertex;
     return (
+      // eslint-disable-next-line @typescript-eslint/no-use-before-define
       <DdgNodeContent
         focalNodeUrl={isFocalNode ? null : getUrl({ density, operation, service, ...extraUrlArgs }, baseUrl)}
         focusPathsThroughVertex={focusPathsThroughVertex}
--- a/packages/jaeger-ui/src/types/config.tsx
+++ b/packages/jaeger-ui/src/types/config.tsx
@@ -18,13 +18,13 @@ import { TNil } from '.';
 
 export type ConfigMenuItem = {
   label: string;
-  url: string;
+  url?: string;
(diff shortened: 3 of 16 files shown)


.eslintrc.js
```diff
@@ -57,7 +56,16 @@ module.exports = {
             prefix: ['I'],
           },
         ],
+
+        // Disable ESLint core rules for which @typescript-eslint provides TypeScript-specific equivalents.
```
yurishkuro, line 60: Just curious, what's the harm in keeping them? if TS flags them first, then fixing the issue would also fix ESLint issue, would it not? (basically, I prefer fewer deviations from defaults if possible)

packages/jaeger-ui/src/components/DeepDependencies/Graph/DdgNodeContent/index.tsx
```diff
@@ -111,6 +111,7 @@ export function getNodeRenderer({
   return function renderNode(vertex: TDdgVertex, _: unknown, lv: TLayoutVertex<any> | null) {
     const { isFocalNode, key, operation, service } = vertex;
     return (
+      // eslint-disable-next-line @typescript-eslint/no-use-before-define
```
yurishkuro, line 114: it's good to have a comment explaining the reason (I assume it's a false positive)

packages/jaeger-ui/src/types/config.tsx
```diff
@@ -84,15 +84,15 @@ export type Config = {
...
     // before the DAG dependency view is disabled. Too many services
     // cause the DAG view to be non-responsive.
-    dagMaxServicesLen?: number;
+    dagMaxNumServices?: number;
   };
 
   // menu controls the dropdown menu in the top-right corner of the UI.
   // When populated, this element completely overrides the default menu.
-  menu: (ConfigMenuGroup | ConfigMenuItem)[];
+  menu: readonly (ConfigMenuGroup | ConfigMenuItem)[];
```
yurishkuro, line 95: curious why these specific entries need `readonly` and not all others.

.eslintrc.js
mszabo-wikia: The problem is that these can cause false positives with the new version of `@typescript-eslint`. E.g. removing the suppression for core's `no-shadow` rules results in a plethora of bogus reports about TS type definitions. The recommendation of silencing the corresponding core rules comes from the [plugin itself](https://typescript-eslint.io/rules/no-shadow/).

mszabo-wikia: Another way to fix it could be to make the root eslintrc also extend the `recommended` config from `@typescript-eslint`, because that config already does this toggling for us. This is what plexus does already. I'm thinking this could be done in a followup PR as it might cause some new errors.

yurishkuro (approved):

yurishkuro:
@mszabo-wikia can we also close #998 ?

packages/jaeger-ui/src/types/config.tsx
mszabo-wikia: It seems it's specifically for `array` type entries. TS is apparently now better at detecting that the object passed to `deepFreeze` is recursively `readonly`.

.eslintrc.js
yurishkuro: +1 -- #1258

yurishkuro:
🥳 🥳 🥳 🥳 🥳 🥳 🥳 🥳 🥳 🥳 🥳 🥳 🥳 🥳 🥳 🥳 🥳 🥳

yurishkuro:
Another awesome PR! It looks like it's going to unblock a lot of pending dependabot upgrades, several of them already turned green after failing CI for months. I'll be merging them today (unfortunately, every time I merge one I need to wait for the CI to run on all others, takes a lot of time).

mszabo-wikia:
Awesome, that's great to see :)

Issue #1260, opened 2023-03-12
[Deps]: Upgrade redux-actions to 3.x


yurishkuro:
### Requirement

Keep dependencies up to date

### Problem

dependabot PR #1195 is failing with errors like `Cannot find module 'redux-actions'`.

Could be related to another PR #1169 that upgrades redux itself.

ashutosh887:
@yurishkuro Is this resolved? I'm willing to work on this

yurishkuro:
I don't think it's resolved, it's probably a part of more complicated upgrade process that 
@mszabo-wikia was working on in #1261 and before that.

ashutosh887:
@mszabo-wikia @yurishkuro 
Should I work on this?

ashutosh887:
Please let me know what all need to be updated

M00nNight:
@yurishkuro Hello sir I just new to this so try to contribute in this repository I just figure out the redux-action version 
so i just manually update redux-action version.
 which in the file path "https://github.com/jaegertracing/jaeger-ui/blob/main/packages/jaeger-ui/package.json"
 line no.64   [@types/redux-actions": "2.2.1"] to this  [@types/redux-actions": "^3.0.0"]

if this solution not right please give feedback 🙏

yurishkuro:
@M00nNight and do the tests pass after you make that change?

prathamesh-mutkure:
@yurishkuro, I'd like to work on this, can you assign me this?

yurishkuro:
feel free to work on it, but we do not assign issues. https://github.com/jaegertracing/jaeger/blob/main/CONTRIBUTING_GUIDELINES.md#assigning-issues

prathamesh-mutkure:
Hey @yurishkuro, there's actually some problem with `redux-actions@3.0.0` + `jest`

Jest is unable to find the redux-actions package, and this issue seems to be unresolved for now

Should we wait for a solution or we can upgrade to `v2.6.5` for now

Parship999:
@yurishkuro I think upgradation of redux-actions to version 3.0.3 can be done by:
- first update the package and the typescript types
- Update import statements in all files using redux-actions

https://www.npmjs.com/package/redux-actions#installation
I'd like to work on this issue

Parship999:
For anyone working on this issue, these are some helpful references:
- https://www.npmjs.com/package/redux-actions
- https://github.com/redux-utilities/redux-actions/releases

PR #1261, opened 2023-03-12
[WIP] Upgrade antd to 4.24.8

mszabo-wikia:
- Contributes towards #1199 

Initial upgrade work for the antd v4 migration.

Notable changes:
* AntD v4 uses individual icon imports rather than a generic `Icon` component with a string `type`. This way, it doesn't have to include the full SVG sprite, only the used icons, reducing the gzipped bundle size by ~110 kB. The migration is done mostly automatically by `@ant-design/codemod-v4` except for some dynamic icon usages that were updated by hand in 4b79a6877242a8f1514ad2e730eeef1c851ed99c.
* AntD v4 requires a newer version of React typings, which also required an explicit Yarn resolution to avoid pulling in other versions of the typings.
* React refs for AntD `Input` components need a typing update.
* The `NumberInput` component now guarantees to return `number | null`, so some custom type-checking code related to that can be removed.
* Typings changed around the `Table` component around row selection and sorting logic, requiring appropriate updates. In particular, sorting by multiple sortable columns now reports all columns that are active in sorting, so event handling code must also accept an array.
* The trace search form needs a workaround to function with `redux-form`, as AntD `Form`s no longer expose an `onSubmit` event that this library could hook into. For now, this is done by attaching the submit callback directly to the submit button instead as a click handler (which, somehow, still catches Enter keypresses on the search form).

## Known issues
* There are some visual regressions that will need styling fixes.
* Tests need updating.


mszabo-wikia:
Looking at `redux-form`, this library does not actually seem very useful. Redux itself [recommends not to put form state i the Redux store](https://redux.js.org/faq/organizing-state#should-i-put-form-state-or-other-ui-state-in-my-store), which the author of this library also [seems to have come to agree with](https://www.npmjs.com/package/redux-form). Out of all forms in the project using this library, only the trace search form sets `destroyOnUnmount: false` to actually persist the form data even after the form itself is unmounted, and even there the form state can be reconstructed from URL parameters, which the mapping logic already does. Nothing else seems to be operating on the raw form data other than the respective form components, either.

It could be potentially worth it (as a separate project) to just remove this.

yurishkuro:
@mszabo-wikia does redux-form need to be tackled in this PR or incidentally related? I opened another issue #1260 for Redux upgrade, it probably makes sense to tackle all Redux-related changed there (assuming they can be decoupled)

packages/jaeger-ui/src/components/SearchTracePage/SearchForm.jsx
```diff
@@ -276,9 +276,11 @@ export class SearchFormImpl extends React.PureComponent {
     const opsForSvc = (selectedServicePayload && selectedServicePayload.operations) || [];
     const noSelectedService = selectedService === '-' || !selectedService;
     const tz = selectedLookback === 'custom' ? new Date().toTimeString().replace(/^.*?GMT/, 'UTC') : null;
+
     return (
-      <Form layout="vertical" onSubmit={handleSubmit}>
```
yurishkuro, line 280: When the handler was previously on the form itself, what did it mean? Could one submit the form by hitting Enter from any field?

packages/jaeger-ui/src/components/TracePage/TraceSpanView/index.tsx
```diff
@@ -174,7 +174,7 @@ export default class TraceSpanView extends Component<Props, State> {
     return (
       <div>
         <h3 className="title--TraceSpanView"> Trace Tabular View</h3>
-        <Row type="flex" style={{ marginTop: '8px' }}>
```
yurishkuro, line 177: what did `type="flex"` do? Seems important.

yurishkuro (commented):
I would prefer to go through all visual changes to make sure they are still sensible for Jaeger UI. Although, just the icon changes seem fine to accept on code review, if that's the extent.

.gitignore
```diff
@@ -13,11 +13,14 @@ coverage
 .nyc_output
 junit.xml
 
+# IDE-related files and directories
```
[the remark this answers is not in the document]
mszabo-wikia: Oh, I didn't know about the global option, that makes sense. I'll probably rebase away this commit since I only added it in the first place to be able to run the codemod, which wouldn't start with uncommitted files present.

packages/jaeger-ui/src/components/SearchTracePage/SearchForm.jsx
mszabo-wikia: Pretty much. Surprisingly though, that still works for me locally even with the handler moved to the submit button.

packages/jaeger-ui/src/components/TracePage/TraceSpanView/index.tsx
mszabo-wikia: The prop was removed and it seems it's no longer needed. In v3, it was used to [control whether to set `display: flex`](https://github.com/ant-design/ant-design/blob/3.26.0/components/grid/style/index.less#L13) on the row but in v4 this seems to be [unconditional](https://github.com/ant-design/ant-design/blob/4.24.8/components/grid/style/index.less).

mszabo-wikia:
I think redux-form need not be changed in this PR—the submit button event handler seems like a reasonable workaround for now, and the other forms don't seem to be using `handleSubmit` in this problematic way.

For the visual regressions, the main issue I noticed so far is a header size bug due to some Menu component updates; this should be fixable with some CSS. The icons should be the same as their design shouldn't have changed as far as I see, only the way they are referenced.

mszabo-wikia:
Closing this as there is great work ongoing in https://github.com/jaegertracing/jaeger-ui/pull/1636 and followups.

PR #1262, opened 2023-03-13
Add totalMemory in config for trace graph

### LINKED ISSUE #1249: [Feature]: Adding Config for TotalMemory in LayoutManager for TraceGraph
### Requirement

As a Jaeger operator,
I want to have `TotalMemory` option in config so that we have the flexibility to tune it for trace graph with large trace (15K+ spans).

### Problem

Today, we face the issue that the trace graph doesn't show up for large traces (with ~15K+ spans).

Screenshot 2023-03-10 at 12 14 25 AM

Error message:
```
Coordinator.js:115 Viz worker onerror 
Cannot enlarge memory arrays. Either (1) compile with  -s TOTAL_MEMORY=X  with X higher than the current value 16777216, (2) compile with  -s ALLOW_MEMORY_GROWTH=1  which allows increasing the size at runtime but prevents some optimizations, (3) set Module.TOTAL_MEMORY to a higher value before the program runs, or (4) if you want malloc to return NULL (0) instead of this abort, compile with  -s ABORTING_MALLOC=0  
```

I think the reason is the allocated memory for viz is not enough.
So when I provided a larger number for memory like below in code [here](https://github.com/jaegertracing/jaeger-ui/blob/main/packages/jaeger-ui/src/components/TracePage/TraceGraph/TraceGraph.tsx#L132), then the trace graph can show up successfully. 
```
// default 16MB
// 32MB: 33554432
this.layoutManager = new LayoutManager({ totalMemory: 33554432, useDotEdges: true, splines: 'polyline' });
```

### Proposal

I can fix by provide a large `totalMemory` when initializing `LayoutManager` for trace graph, but I don't think hardcode the value in the code (TraceGraph.tsx) is a good way, and default value is good enough for most use cases.

So I would like to have `totalMemory` exposed in the [config](https://github.com/jaegertracing/jaeger-ui/blob/main/packages/jaeger-ui/src/types/config.tsx), so that anyone has the flexibility to tune the `totalMemory` themselves if needed.

Like:
```
export type Config = {
  ....
  traceGraph?: {
     layoutManager?: {
        totalMemory?: number; 
     };
  };
  ...
};
```

If this proposal is ok with you, I can implement it and submit a PR.

### Open questions

_No response_

ChenX1993:
Trace graph has a problem to renders large traces (with ~15K+ spans), and gives the error "Cannot enlarge memory arrays". Need to give users flexibility to control the totalMemory of LayoutManager used for TraceGraph.

- Resolves #1249

Trace graph has a problem to renders large traces (with ~15K+ spans), and gives the error "Cannot enlarge memory arrays". This can be fixed by providing a larger number of memory when initializing the `LayoutManager` in trace graph, like below:
```
// default 16MB
this.layoutManager = new LayoutManager({ totalMemory: 33554432, useDotEdges: true, splines: 'polyline' });
```
But instead of hardcoding the memory value in code, we want to give users the flexibility to control the totalMemory of LayoutManager for TraceGraph.

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/index.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/index.tsx
@@ -58,6 +58,7 @@ import updateUiFind from '../../utils/update-ui-find';
 import TraceStatistics from './TraceStatistics/index';
 import TraceSpanView from './TraceSpanView/index';
 import TraceFlamegraph from './TraceFlamegraph/index';
+import { TraceGraphConfig } from '../../types/config';
 
 import './index.css';
 
@@ -82,6 +83,7 @@ type TReduxProps = {
   searchUrl: null | string;
   trace: FetchedTrace | TNil;
   uiFind: string | TNil;
+  traceGraphConfig?: TraceGraphConfig;
 };
 
 type TProps = TDispatchProps & TOwnProps & TReduxProps;
@@ -326,6 +328,7 @@ export class TracePageImpl extends React.PureComponent<TProps, TState> {
       id,
       uiFind,
       trace,
+      traceGraphConfig,
       location: { state: locationState },
     } = this.props;
     const { slimView, viewType, headerHeight, viewRange } = this.state;
@@ -401,6 +404,7 @@ export class TracePageImpl extends React.PureComponent<TProps, TState> {
           ev={this.traceDagEV}
           uiFind={uiFind}
           uiFindVertexKeys={graphFindMatches}
+          traceGraphConfig={traceGraphConfig}
         />
       );
     } else if (ETraceViewType.TraceStatistics === viewType && headerHeight) {
@@ -435,6 +439,7 @@ export function mapStateToProps(state: ReduxState, ownProps: TOwnProps): TReduxP
   const archiveEnabled = Boolean(config.archiveEnabled);
   const { state: locationState } = router.location;
   const searchUrl = (locationState && locationState.fromSearch) || null;
+  const { traceGraph: traceGraphConfig } = config;
 
   return {
     ...extractUiFindFromState(state),
@@ -444,6 +449,7 @@ export function mapStateToProps(state: ReduxState, ownProps: TOwnProps): TReduxP
     id,
     searchUrl,
     trace,
+    traceGraphConfig,
   };
 }
 
--- a/packages/jaeger-ui/src/constants/default-config.tsx
+++ b/packages/jaeger-ui/src/constants/default-config.tsx
@@ -101,6 +101,11 @@ const defaultConfig: Config = {
     },
     docsLink: 'https://www.jaegertracing.io/docs/latest/spm/',
   },
+
+  traceGraph: {
+    layoutManagerMemory: undefined,
+  },
+
   deepDependencies: {
     menuEnabled: false,
   },
--- a/packages/jaeger-ui/src/types/config.tsx
+++ b/packages/jaeger-ui/src/types/config.tsx
@@ -73,6 +73,15 @@ export type MonitorConfig = {
   docsLink?: string;
 };
 
+export type TraceGraphConfig = {
+  // layoutManagerMemory controls the total memeory available for the GraphViz
+  // Emscripten module instance. The value should be a power of two.
+  // The default of 16MB should be sufficient for most cases — only consider
+  // using a larger number if you run into the error "Cannot enlarge memory arrays".
+  // See https://github.com/jaegertracing/jaeger-ui/issues/1249 for background
+  layoutManagerMemory?: number;
+};
+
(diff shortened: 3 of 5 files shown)


packages/jaeger-ui/src/types/config.tsx
```diff
@@ -164,4 +164,14 @@ export type Config = {
     menuEnabled?: boolean;
     menuLabel?: string;
   };
+
+  // traceGraph controls the trace graph under trace page
+  traceGraph?: {
```
yurishkuro, line 169: please move this above L157, I don't think this is "experimental"

packages/jaeger-ui/src/constants/default-config.tsx
```diff
@@ -108,6 +108,9 @@ const defaultConfig: Config = {
     menuEnabled: false,
     menuLabel: 'Trace Quality',
   },
+  traceGraph: {
```
yurishkuro, line 111: please move above L104

packages/jaeger-ui/src/components/TracePage/index.tsx
```diff
@@ -435,6 +438,7 @@ export function mapStateToProps(state: ReduxState, ownProps: TOwnProps): TReduxP
   const archiveEnabled = Boolean(config.archiveEnabled);
   const { state: locationState } = router.location;
   const searchUrl = (locationState && locationState.fromSearch) || null;
+  const graphLayoutManagerMemory = config.traceGraph?.layoutManagerMemory;
```
yurishkuro, line 441: can we add a test for this?

packages/jaeger-ui/src/components/TracePage/index.tsx
```diff
@@ -82,6 +82,7 @@ type TReduxProps = {
   searchUrl: null | string;
   trace: FetchedTrace | TNil;
   uiFind: string | TNil;
+  graphLayoutManagerMemory: number | undefined;
```
yurishkuro, line 85: I'm not an expect in React, so a Q: why does this setting need to be in the properties (which causes more code changes) instead of only passed from config to `<TraceGraph>`? Or even looked up from config by `TraceGraph` directly?

packages/jaeger-ui/src/types/config.tsx
ChenX1993: Moved.

I originally put it below L157 because I thought it was "undocumented" in https://www.jaegertracing.io/docs/1.42/frontend-ui/

packages/jaeger-ui/src/components/TracePage/index.tsx
ChenX1993: Added a test.
It is not possible to test any UI change with this option, so I just followed the existing tests to test the propagation from  redux state to component properties.

packages/jaeger-ui/src/components/TracePage/index.tsx
ChenX1993: I haven't worked on React for a while, so I might be wrong.


1>
I observed in this repo:
* The config is passed into React components via Redux.
* Only top/page level components (like TracePage in this example) subscribe to Redux, and most sub-components don't.

So I am trying to follow the same style here - have `TracePage` get config from redux and pass to `TraceGraph` via property.

2>
Since we let component `TracePage` subscribe to Redux, then we need to make sure what we need from Redux is exposed as component property by `mapStateToProps` function defined in the component.

We can either expose the entire config object, or sub fields/objects,  but either requires code change - adding into component property definition.

The reason I exposed `graphLayoutManagerMemory` (now I changed to `traceGraphConfig`) instead of entire config object is due component re-render.
If any property value of the component changes, it causes component to re-render. The entire config object contains other fields/sub objects that current component doesn't need at all, which might cause unnecessary re-render.

yurishkuro (approved):
thanks

ChenX1993:
The unit-tests failed at `Upload coverage to codecov.io` step. 
Error message:
```
['error'] There was an error running the uploader: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 404 - {'detail': ErrorDetail(string='Unable to locate build via Github Actions API. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}
[2023-03-13T23:48:23.044Z] ['verbose'] The error stack is: Error: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 404 - {'detail': ErrorDetail(string='Unable to locate build via Github Actions API. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}
    at main (/snapshot/repo/dist/src/index.js)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

Not very sure about the reason.

yurishkuro:
sometimes codecov hiccups, I reran the job and tests were fine.

PR #1274, opened 2023-03-15
Batch Download of traces

### LINKED ISSUE #663: Batch Download of traces

Our traces contain data that is covered by GDPR (european user privacy regulations) and, thus developers do not have direct access to all traces. If an error occurs, relevant traces has to be filtered out and the relevant ones are forwarded to the developer.

Another use case:
It would be nice to just attach files with relevant traces to a Jira Ticket. HTTP Links are nice but they are are only available for a certain time, which conflicts with some of our documentation requirements, and everybody has to have access to the jaeger instance.

The process of "exporting" the traces is cumbersome. For each trace, you have to go to the json view and save it.

It would be nice to have:

* [ ] A download trace button in the trace view
* [ ] A download traces button in the search view to download all found traces -- #1274
  * [ ] An option to download traces as a zip file with one JSON file per trace in the archive ([explained below][1])
* [ ] A download selected traces button in the search view ([proposal below][1])

[1]: https://github.com/jaegertracing/jaeger-ui/issues/663#issuecomment-1470940708

Katarzyna-B:
Batch Download of traces
- Resolves #663

- Add a button with download functionality


yurishkuro:
Thanks for the PR. I have a couple of comments, but I posted them in the ticket to solicit opinions from people who voted for this feature: https://github.com/jaegertracing/jaeger-ui/issues/663#issuecomment-1470940708

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/DownloadResults.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export default function DownloadResults(props: Props) {
+    const {onDownloadResultsClicked} = props;
+    const downloadBtn = (
+        <Button className="ub-ml2" htmlType="button" onClick={onDownloadResultsClicked}>
+            Download Results
+        </Button>
+    );
+    return (
+        <>
```
yurishkuro, line 31: what does `<>` represent? Is it needed?

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.tsx
```diff
@@ -110,6 +112,17 @@ export class UnconnectedSearchResults extends React.PureComponent<SearchResultsP
...
 
+  onDownloadResultsClicked = () => {
+    const file = new Blob([JSON.stringify(this.props.tracesToDownload)], {type : "application/json"});
+    const element = document.createElement("a");
+    element.href = URL.createObjectURL(file);
+    element.download = "traces-" + Date.now() + ".json";
+    document.body.appendChild(element);
+    element.click();
+    URL.revokeObjectURL(element.href)
+    element.remove()
```
yurishkuro, line 123: Reading this as not a web developer, is this the preferred way of executing downloads from a page? Do you have some pointer to documentation describing it?

packages/jaeger-ui/src/reducers/trace.js
```diff
@@ -114,7 +115,7 @@ function searchDone(state, { meta, payload }) {
   }
   const traces = { ...state.traces, ...resultTraces };
   const search = { ...state.search, results, state: fetchedState.DONE };
-  return { ...state, search, traces };
+  return { ...state, search, traces, tracesToDownload: payloadData};
```
yurishkuro, line 118: nit: I don't think this reducer knows _why_ it's returning the full payload, so naming the output `tracesToDownload` is not accurate. Can we name it something like `rawTraces`?

yurishkuro (commented):
please make sure all commits are signed (see CONTRIBUTING.md)

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/DownloadResults.tsx
Katarzyna-B: **<> </>**  
This is called the Empty tag in React. This is a shorter syntax for React Fragments.
Documentation: https://react.dev/reference/react/Fragment
It is needed. If you remove this tag you can get exception like 

> [AltViewOptions.tsx](react-dom.development.js:14887 Uncaught Error: Objects are not valid as a React child (found: object with keys {downloadBtn})).....

It is used already in this project. See for example AltViewOptions.tsx

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.tsx
Katarzyna-B: I am not primarily a web developer. But when I was looking for the best way of executing downloads from a page I found bunch of similar solution how you can do it.  Here some references:

- https://www.techiediaries.com/javascript-blobs-object-urls/ (Date: 02 Oct 2022)
- https://theroadtoenterprise.com/blog/how-to-download-csv-and-json-files-in-react (27 Aug, 2021)

element.click() --> it's simulate a click 
URL.revokeObjectURL(element.href) --> "it is recommended to release object URLs to optimize performance and reduce memory consumption."  In the first article it is good explained.

packages/jaeger-ui/src/reducers/trace.js
Katarzyna-B: Good point. I will do it.

yurishkuro (approved):

yurishkuro:
please fix lint errors

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/DownloadResults.tsx
Katarzyna-B: I was wrong. I got the exception which I described above, but the lint complained about this part of code:
> .../SearchResults/DownloadResults.tsx
[eslint       ]   29:10  error  Fragments should contain more than one child - otherwise, there’s no need for a Fragment at all  react/jsx-no-useless-fragment

I changed the code in a different wait and it works locally without the described error.

PR #68, opened 2017-08-28
Virtualized scrolling for trace detail view

tiffon:
# TODO

- [x] Tests
- [x] Additional documentation
- [x] Demo for team for sign-off

# Description

Started with [react-virtualized](https://github.com/bvaughn/react-virtualized), but issues around flashes of un-rendered regions while scrolling and dynamically sized / periodically changing content lead to revival of old project that essentially does the same thing: `src/components/TracePage/TraceTimelineViewer/ListView/*`. 

`ListView` is more optimized to our needs (and less versatile). These optimizations include:

- Taking the approach of rendering less often and larger amounts of items per render
- Optimizing for scrolling in both directions instead of only the current scroll direction
- Having a concept of minimum and regular ["overscan"](https://github.com/bvaughn/react-virtualized/blob/master/docs/overscanUsage.md), which means if the currently rendered list of items meets a minimum buffer, don't render more, but if it doesn't, then render enough meet the minimum buffer and some extra, as well

# Performance

## Flashes of un-rendered content

- Smaller traces (~500 spans or less)
  - Not really an issue in prod or PR 68
- Larger traces (~5k spans)
  - PR 68 is on par with prod, possible slightly better
  - Still fairly easy to see flashes of un-rendered content
  - Issue was quite aggravated with react-virtualized (hence, a custom solution is implemented)
- Massive traces (~15k spans or more)
  - Flashes of un-rendered content is a significant problem for the release currently in prod
  - PR 68 performance doesn't vary much between 5k and 15k+ traces

 ## Performance on a 5,000 span trace

 _Sample size of 1._

- /trace/8d224b61611c807f
- Duration: 5.83s
- Services: 11
- Depth: 11
- Spans: 5,013
- Performance recording enabled (degrades performance)

**Initial loading (in milliseconds)**

Task | Prod - PR 53 (ms) | PR 68 (ms) | Ratio | Reduction
-- | -- | -- | -- | --
**Total** | **6,964** | **1,899** | **0.27** | **73%**
Loading | 45 | 32 ||
**Scripting** | **3,051** | **1,198** | **0.39** | **61%**
**Rendering** | **3,126** | **430** | **0.14** | **86%**
Painting | 116 | 34 ||
Other | 626 | 205 ||

Note: Ticket #61 (Use canvas instead of SVG for trace mini-map) will likely further improve initial loading.

**Expanding details**

- Prod : 3753 ms
- PR 68: 418 ms (0.11)

**Text search ("gopro")**

- Prod : 5496 ms
[...]

PR #1284, opened 2023-03-18
Remove unused Redux selectors

mszabo-wikia:
- Contributes towards making coverage reports more consistent

Some of the inconsistency of coverage results is due to randomly generated test data in trace.test.js. While investigating, however, I noticed that most of the Redux selectors in this directory were only referenced by each other and by tests - it seems they were used at some point (they're in use in the initial commit) but were progressively removed by refactorings. Since selectors only serve to encapsulate and reuse logic to lookup data from the Redux store[1], they are not useful when not imported. As such, take this opportunity to remove the dead code.

---
[1] https://redux.js.org/usage/deriving-data-selectors


mszabo-wikia:
The other file where coverage changes randomly between local test runs is https://github.com/jaegertracing/jaeger-ui/blob/main/packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx. Not yet sure why.

yurishkuro:
> it seems they were used at some point (they're in use in the initial commit) but were progressively removed by refactorings

Any idea what would compel people to change away from using selectors? Do you think it's just general unawareness that they exist?

yurishkuro (approved):

mszabo-wikia:
Looks like https://github.com/jaegertracing/jaeger-ui/pull/68 was the PR that removed most usages, it shifted part of the state management into duck.js (duck.tsx). There isn't much context specifically why though.

PR #1293, opened 2023-03-22
Make coverage results more stable

mszabo-wikia:
- Makes coverage checks more stable

Coverage results are currently flaky because
src/components/TracePage/index.test.js uses randomly generated trace data to integration test the page, which sometimes covers code paths in tableValues.tsx not covered by its own unit tests. Attempt to make things more stable by extending unit test coverage for tableValues.tsx.

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
@@ -209,7 +209,7 @@ function calculateContent(trace: Trace, span: Span, allSpans: Span[], resultValu
 
           tempSelf = onlyOverlay(overlayWithout, allChildrenWithout, tempSelf, span);
           const diff = span.relativeStartTime + span.duration - earliestLongerAsParent.relativeStartTime;
-          tempSelf -= diff;
+          tempSelf = Math.max(0, tempSelf - diff);
         }
       } else if (longerAsParent) {
         // span is longer as Parent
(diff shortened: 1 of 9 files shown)


mszabo-wikia:
Looks like there's 1 line left still

yurishkuro:
Wanna fix it before merging?

Also, how are you finding which lines' coverage changes? Codecov.io keeps breaking for me, refuses to show diffs.

mszabo-wikia:
Yeah, I just need to come up with a good trace that can be used to test it—it's been proving elusive so far.

For determining coverage,
```
$ yarn test --coverage --collectCoverageFrom src/components/TracePage/TraceStatistics/tableValues.tsx
```
is useful to observe how coverage results for this single file change over multiple executions of the entire suite, and
```
$ yarn test --coverage --collectCoverageFrom src/components/TracePage/TraceStatistics/tableValues.tsx src/components/TracePage/index.test.js
```
does the same but only running a single test file.

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
```diff
@@ -209,7 +209,7 @@ function calculateContent(trace: Trace, span: Span, allSpans: Span[], resultValu
 
           tempSelf = onlyOverlay(overlayWithout, allChildrenWithout, tempSelf, span);
           const diff = span.relativeStartTime + span.duration - earliestLongerAsParent.relativeStartTime;
-          tempSelf -= diff;
+          tempSelf = Math.max(0, tempSelf - diff);
```
mszabo-wikia, line 212: Apparently a bug - the final test case returns a negative value without this guard. The example is somewhat contrived as it hinges on two child spans firing off exactly at trace start time—which is unlikely, but that seems to be the only way to reach line 34.

yurishkuro (approved):
🎉 🙇

PR #1295, opened 2023-03-23
Render HTTP header tags as comma-separated strings

### LINKED ISSUE #1294: [Feature]: Better rendering of OTEL array attributes (e.g. HTTP headers)
### Requirement

As a developer, I want to use Jaeger all-in-one locally while I work, so I can both verify my code is correctly instrumented and also use traces to help with debugging. When I work on HTTP services, I want to include HTTP headers as trace tags to quickly verify correct content negotiation, caching, keep-alive, etc.

### Problem

The OpenTelemetry [semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-request-and-response-headers) for HTTP request and response headers says:
> The attribute value MUST consist of either multiple header values as an array of strings or a single-item array containing a possibly comma-concatenated string, depending on the way the HTTP library provides access to headers.

Jaeger displays array values as pretty-printed JSON arrays, which isn't well suited to HTTP headers, which rarely have more than one value:

image

### Proposal

As a special case for HTTP request and response tags, Jaeger should display the array value as a comma-separated list of un-decorated strings (i.e. no array brackets, no quotes), the same as other HTTP tags like `http.user_agent` are displayed.

### Open questions

_No response_

wjrogers:

Resolves #1294

Change `formatValue` to check whether the (de-serialized) value is an array and, if so, whether the key should be rendered as a comma-separated list of strings. When both conditions are true, render the value like so:

```html
<div class="json-markup">
  <span class="json-markup-string">value1</span>, <span class="json-markup-string">value2</span>
</div>
```
image

Currently, tag keys starting with `http.request.header.` and `http.response.header.` render as string lists. Please let me know if there are any other attributes I should include.

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.test.js
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.test.js
@@ -53,7 +53,11 @@ describe('<KeyValuesTable>', () => {
     { key: 'numeric', value: 123456789, expected: '123456789' },
     { key: 'jsonkey', value: JSON.stringify(jsonkeyValue), expected: JSON.stringify(jsonkeyValue, null, 4) },
     { key: 'http.request.header.accept', value: ['application/json'], expected: 'application/json' },
-    { key: 'http.response.header.set_cookie', value: JSON.stringify(['name=mos-def', 'code=12345']), expected: 'name=mos-def, code=12345' },
+    {
+      key: 'http.response.header.set_cookie',
+      value: JSON.stringify(['name=mos-def', 'code=12345']),
+      expected: 'name=mos-def, code=12345',
+    },
   ];
 
   beforeEach(() => {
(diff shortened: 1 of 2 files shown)


packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.test.js
```diff
@@ -126,9 +128,27 @@ describe('<KeyValuesTable>', () => {
...
     expect(el.length).toBe(data.length);
     el.forEach((valueDiv, i) => {
-      if (data[i].key !== 'jsonkey') {
+      if (data[i].expected) {
         expect(valueDiv.html()).toMatch(data[i].expected);
       }
     });
   });
+
+  it('renders HTTP headers as multiple string spans', () => {
```
yurishkuro, line 137: This test feels very complicated & hard to maintain. Could we not use the `.expected` field above to define the exact output we want for the two new test cases?

wjrogers: Sure. I was figuring out how the tests work and maybe got carried away. Will change it.

yurishkuro (approved):
👍

yurishkuro:
please run the linter, `yarn prettier`

PR #1391, opened 2023-05-03
Support in-page-search in trace view on key=value pairs

bobrik:
Previously one could only filter on `key` or `value`. This wasn't useful when one needed to find spans with `result=false` in a trace view.

With this change one can use `key=value` as a search term, for example:

* `http.method=GET`

Before (no matches):

image

After (a successful match):

image

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageSearchBar.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageSearchBar.tsx
@@ -67,12 +67,12 @@ export function TracePageSearchBarFn(props: TracePageSearchBarProps & { forwarde
         </p>
         <p>
           When matching key-value pairs, the substring search is applied separately against the key, the
-          value, and the concatenated <code>"key=value"</code> string. The latter allows searching for exact
-          matches like <code>http.status_code=200</code>.
+          value, and the concatenated <code>&quot;key=value&quot;</code> string. The latter allows searching
+          for exact matches like <code>http.status_code=200</code>.
         </p>
         <p>
           To preclude certain key-value pairs from participating in the matching, prefix the key with the
-          minus <code>'-'</code> sign, e.g., <code>-http.status_code</code>.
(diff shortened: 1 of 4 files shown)


yurishkuro:
a bit dirty (i.e. would also match on `value="a=b"`), but seems useful regardless. +1

yurishkuro:
Overall the in-page search functionality is not very discoverable. When it was just a direct string match it was less of an issue, but if supporting a=b format it would be good to have some help tip somewhere.

Q: in the backend search form we support multiple space-separated tag queries, would be nice to allow the same syntax for in-page search. I believe we even have a parser for the syntax in the UI.

bobrik:
I added some tests for the `key=value` syntax.

Here's the tooltip I added in the second commit:

image

Using the same syntax as the tag filter on the search page would mean that we lose `-field` syntax, so it might not be a desired change. Trace view search also looks at more things than just tags, which might not be compatible with what we have in the regular search. I'd also say that it's outside of my expertise, since I'm not a frontend person.

bobrik:
@yurishkuro, any comments on the latest update?

packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageSearchBar.tsx
```diff
@@ -56,49 +56,79 @@ export function TracePageSearchBarFn(props: TracePageSearchBarProps & { forwarde
...
+        <p>
+          When matching key-value pairs, the substring search is applied separately against
+          the key, the value, and the <code>"key=value"</code> string.
+          The latter allows searching for exact matches like <code>http.status_code=200</code>.
+        </p>
+        <p>
+          To preclude certain key-value pairs from participating in the matching, prefix the key
+          with the minus <code>'-'</code> sign, e.g., <code>-http.status_code</code>.
+        </p>
+      </div>
```
yurishkuro, line 77: I rewrote the tooltip to make the explanation a bit more clear and to remove the styling applied to some key words which was difficult to read on the dark background.

yurishkuro:
Sorry it took a while to review, a busy month.

PR #1500, opened 2023-06-14
Trace statistics table is now using antd Table component

### LINKED ISSUE #1181: [Bug]: trace statistics and trace span table are using different look & feel
### What happened?

When switching between views of a single trace I expect the UI to use similar visual design. Instead, these two views seem to use different table widgets. Obvious differences:
  * span table rows are much taller, for no reason
  * sorting indicators are different
  * statistics table can be sorted by any column, not so for spans table

### Steps to reproduce

open any trace (e.g. from HotROD) and flip between Trace Statistics and Spans Table views

### Expected behavior

look & feel should be similar

### Relevant log output

_No response_

### Screenshot

_No response_

### Additional context

_No response_

### Jaeger backend version

_No response_

### SDK

_No response_

### Pipeline

_No response_

### Stogage backend

_No response_

### Operating system

_No response_

### Deployment model

_No response_

### Deployment configs

_No response_

GLVSKiriti:
Trace Statistics table is not using antd Table component like Trace span Table so now code is modified such that TraceStatistics table now uses antd Table component

Resolves #1181 

@yurishkuro Once review this PR and let me now any changes should I do


packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -392,79 +262,331 @@ export default class TraceStatistics extends Component<Props, State> {
...
+              this.togglePopup(name)
+    }
 
-  render() {
+    const columns: ColumnProps<ITableSpan>[] = [
+      {
+        title: 'Name',
+        dataIndex: 'name',
+        sorter: (a,b) => {
+          if(a.type === 'undefined'){
```
yurishkuro, line 277: * why is this checking the `.type` field but main sorting is done on `.name`?
* the sorter looks nearly identical across all columns, can it not me simplified with a helper function that takes a lambda to extract the value to sort?

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -392,79 +262,331 @@ export default class TraceStatistics extends Component<Props, State> {
...
+            return -1
+          }
+          if(b.type === 'undefined'){
+            return 1
+          }
+          return a.name.localeCompare(b.name)
+        },
+        render: (name: string,row: ITableSpan) => {
```
yurishkuro, line 285: is it necessary to have so much boiler-plate code for each column? can it be DRYed?

yurishkuro:
Did you model your code after another code in Jaeger? If so, please point to the existing code.

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
GLVSKiriti: --> .type field checking is done to fix the row whose type is undefined at bottom irrespective of sorting i.e, In the previous trace statistics view "without tag: component" row is fixed at bottom irrespective of sorting see the (below picture in subgroup of frontend row)
--> Yep I simplified the sorter code and onCell code by using helper functions and committed the changes
DraftPR

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
GLVSKiriti: I almost reduced the code now and let me know if any more changes should be done

GLVSKiriti:
> Did you model your code after another code in Jaeger? If so, please point to the existing code.

Do you mean should I merge the latest changes from "main" into this branch like by updating the branch

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -184,67 +181,6 @@ export default class TraceStatistics extends Component<Props, State> {
...
-      }
-      for (let j = 0; j < tempArray.length; j++) {
-        sortArray2.push(tempArray[j]);
-      }
-    }
-    return sortArray2;
-  };
-
   /**
    * Opern the popup button.
```
yashrsharma44, line 185:
```suggestion
   * Open the popup button.
```

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -392,79 +263,168 @@ export default class TraceStatistics extends Component<Props, State> {
...
+              return cell+'ms'
+        },
+        onCell: (record) => onCellFunction(record)
+      },
+      {
+        title: 'ST Max',
+        dataIndex: 'selfMax',
+        sorter: (a, b) => sorterFunction({"type": a.type,"index": a.selfMax},{"type": b.type,"index": b.selfMax}),
+        render: (cell: string) => {
+              return cell+'ms'
```
yashrsharma44, line 385:
```suggestion
              return `${cell} ms`
```

You can switch the other string concatenation with the above format.

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -479,12 +439,32 @@ export default class TraceStatistics extends Component<Props, State> {
...
           <PopupSQL closePopup={this.togglePopup} popupContent={this.state.popupContent} />
         ) : null}
-        <table className="test1893">
+        {/* <table className="test1893">
           <tbody className="DetailTraceTableTbody--TraceStatistics">
             {this.renderTableHead()}
             {this.renderTableData()}
           </tbody>
-        </table>
+        </table> */}
```
yashrsharma44, line 447: You can remove this if the section is not used.

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -479,12 +439,32 @@ export default class TraceStatistics extends Component<Props, State> {
...
+            pagination={{
+              total: isNoDetail.length,
+              pageSizeOptions: ['10', '20', '50', '100'],
+              showSizeChanger: true,
+              showQuickJumper: true,
+            }}
+            rowClassName={(record,index)=>{
+              if(record.type === 'undefined'){
+                return "undefClass"
+              }
```
yashrsharma44, line 461: Is this needed? Can we simplify the code by removing this?

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -392,79 +263,168 @@ export default class TraceStatistics extends Component<Props, State> {
...
+        render: (cell: string) => {
+              return cell+'%'
+        },
+        onCell: (record) => onCellFunction(record)
+      },
+    ];
+    let isDetailArray: ITableSpan[] = [];
+    let isNoDetail: ITableSpan[] = [];
+  
+    const findTablesAccToSelectors = (tableValue: ITableSpan[]) => {
```
yashrsharma44, line 402: Can we refactor `findTablesAccToSelectors` to return `isDetailArray` and `isNoDetail` instead? This would make it easy to write unit tests for `findTablesAccToSelectors`

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -392,79 +263,168 @@ export default class TraceStatistics extends Component<Props, State> {
...
+      for (let i = 0; i < tableValue.length; i++) {
+        if (tableValue[i].isDetail) {
+          isDetailArray.push(tableValue[i]);
+        } else {
+          isNoDetail.push(tableValue[i]);
+        }
+      }
```
yashrsharma44, line 409: Let's reduce this to a simplified version - 
```suggestion
      isDetailArray = tableValue.filter((val : ITableSpan) => val.isDetail) 
      isNoDetail = tableValue.filter((val : ITableSpan) => !val.isDetail) 
```

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -392,79 +263,168 @@ export default class TraceStatistics extends Component<Props, State> {
...
+    const findTablesAccToSelectors = (tableValue: ITableSpan[]) => {
+      for (let i = 0; i < tableValue.length; i++) {
+        if (tableValue[i].isDetail) {
+          isDetailArray.push(tableValue[i]);
+        } else {
+          isNoDetail.push(tableValue[i]);
+        }
+      } 
+      for(let i = 0; i < isNoDetail.length; i++){
+          let newArr = isDetailArray.filter((value)=>{return ((value.parentElement) === (isNoDetail[i].name))})
```
yashrsharma44, line 411:
```suggestion
          let newArr = isDetailArray.filter((value)=>((value.parentElement) === (isNoDetail[i].name)))
```

JS lambda functions consider the expression as the return type if you don't enclose the expression in braces.

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -392,79 +263,168 @@ export default class TraceStatistics extends Component<Props, State> {
...
+          newArr = newArr.map((value,index)=>{
+            const x = i.toString()+ index.toString();
+            const keyo = {
+              key: x
+            }
```
yashrsharma44, line 416:
```suggestion
            const _key = {
              key: `${i}-${index}`
            }
```
Your logic to use `const x = i.toString()+ index.toString();` as a *unique* key won't work when we have `i=1, index=23` and `i=12, index=3`. Let's use a separator like `-` for creating a unique key.

nit: let's use a `_` prefix for variables which are internally used.

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -392,79 +263,168 @@ export default class TraceStatistics extends Component<Props, State> {
...
-          />
-        );
+  render() {
+
+    const onClickOption = (type:string,name:string) => {
+      if(this.state.valueNameSelector1 === 'sql.query' && type !== 'undefined')
+              this.togglePopup(name)
+    }
+
+    const sorterFunction = (a:any,b:any)=>{
```
yashrsharma44, line 273: Let's define a type for `a` and `b`. We can use [this](https://github.com/jaegertracing/jaeger-ui/blob/d8318132071a174083f15a64c300ebd63b7e90e9/packages/jaeger-ui/src/components/TracePage/TraceStatistics/types.tsx) for declaring an interface for the data type of `a`

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -392,79 +263,168 @@ export default class TraceStatistics extends Component<Props, State> {
...
-      );
-    });
-  }
+      if(b.type === 'undefined'){
+        return -1
+      }
+      if(typeof a.index === 'number')
+      return a.index - b.index
+      else if(typeof a.index === 'string')
+      return a.index.localeCompare(b.index)
```
yashrsharma44, line 283: If we use an interface/strict type for `a`, we can assume that the `typeof` for `a.index` would be a string. In that case, you can make a simple return
```suggestion
      return a.index.localeCompare(b.index)
```

yashrsharma44 (changes requested):
I have made some requests about the implementation. Can you take a look at the [`CONTRIBUTING.md`](https://github.com/jaegertracing/jaeger-ui/blob/main/CONTRIBUTING.md#style-guide) and run the linter to make sure the variables/code are properly formatted?

yashrsharma44:
> > Did you model your code after another code in Jaeger? If so, please point to the existing code.
> 
> Do you mean should I merge the latest changes from "main" into this branch, like by updating the branch

You can link the inspiration/source of truth for the structure of the `Table` that you're using. I can see that we are deriving the changes, framed after the [`TraceSpan`](https://github.com/jaegertracing/jaeger-ui/blob/d8318132071a174083f15a64c300ebd63b7e90e9/packages/jaeger-ui/src/components/TracePage/TraceSpanView/index.tsx) component.

yashrsharma44:
Thanks for raising this PR 🙂 

I have some nits for raising a PR in general. 
* You can make the commit message a bit more descriptive, and make sure that the message depicts the changes made.

For example - 
In this [commit](https://github.com/jaegertracing/jaeger-ui/pull/1500/commits/e2d9be31806087d689e14edba4da9dab2a765b28), you had introduced a sorting function - `sorterFunction`, so make sure the git commit message describes the changes that you have made. *Currently, I can see the same commit message - `Look of TraceStatistics and Tracespan tables should be similar
`, which makes it difficult for someone reviewing the changes to understand at a glance, what changes did you make*. You can [check this out](https://github.blog/2022-06-30-write-better-commits-build-better-projects/) for learning more about writing good commit messages :)

* Check the contributing/readme for running the validation checks - e.g. linter/formatter for making sure, the code changes are compliant to the standards we have set.

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
GLVSKiriti: Yes this returns a styling to the row whose type is undefined

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -392,79 +188,183 @@ export default class TraceStatistics extends Component<Props, State> {
...
+      let isDetailArray: ITableSpan[] = [];
+      let isNoDetail: ITableSpan[] = [];
+
+      isDetailArray = tableValue.filter((val: ITableSpan) => val.isDetail);
+      isNoDetail = tableValue.filter((val: ITableSpan) => !val.isDetail);
+      for (let i = 0; i < isNoDetail.length; i++) {
+        let newArr = isDetailArray.filter(value => value.parentElement === isNoDetail[i].name);
+        newArr = newArr.map((value, index) => {
+          const _key = {
+            key: `$${i}-${index}`,
```
yashrsharma44, line 354: I think this is a typo
```suggestion
            key: `${i}-${index}`,
```

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -392,79 +188,183 @@ export default class TraceStatistics extends Component<Props, State> {
...
+        onCell: record => onCellFunction(record),
+      },
+    ];
+
+    const findTablesAccToSelectors = (tableValue: ITableSpan[]): ITableSpan[] => {
+      let isDetailArray: ITableSpan[] = [];
+      let isNoDetail: ITableSpan[] = [];
+
+      isDetailArray = tableValue.filter((val: ITableSpan) => val.isDetail);
+      isNoDetail = tableValue.filter((val: ITableSpan) => !val.isDetail);
```
yashrsharma44, line 349:
```suggestion
      let isDetailArray: ITableSpan[] = tableValue.filter((val: ITableSpan) => val.isDetail);
      let isNoDetail: ITableSpan[] = tableValue.filter((val: ITableSpan) => !val.isDetail);
      
```

yashrsharma44 (changes requested):
Thanks, the changes look much better at this point. Some small changes, and once they are addressed, it looks good to me :)

yashrsharma44:
You can mark the PR ready for review.

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -392,79 +188,180 @@ export default class TraceStatistics extends Component<Props, State> {
...
+                  borderLeft: `4px solid transparent`,
+                  padding: '7px 0px 7px 10px',
+                  cursor: 'default',
+                }}
+              >
+                {name}
+              </span>
+            );
+          }
+          return (
```
yurishkuro, line 238: This return is nearly identical to the one above, can we merge into one and just handle color inline?

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -392,79 +188,180 @@ export default class TraceStatistics extends Component<Props, State> {
...
+        style: { background: record.colorToPercent, borderColor: record.colorToPercent },
+      };
+    };
 
-  render() {
+    const columns: ColumnProps<ITableSpan>[] = [
+      {
+        title: 'Name',
+        dataIndex: 'name',
+        sorter: (a, b) => sorterFunction({ type: a.type, index: a.name }, { type: b.type, index: b.name }),
```
yurishkuro, line 222: Can the sorterFunction actually return the function, so you could simplify this?
```suggestion
        sorter: sorterFunction({ type: a.type, index: a.name }, { type: b.type, index: b.name }),
```

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -392,79 +188,180 @@ export default class TraceStatistics extends Component<Props, State> {
...
+          return (
+            <span
+              onClick={() => onClickOption(row.type, row.name)}
+              style={{ borderLeft: `4px solid ${row.color}`, padding: '7px 0px 7px 10px', cursor: 'default' }}
+            >
+              {name}
+            </span>
+          );
+        },
+        onCell: record => onCellFunction(record),
```
yurishkuro, line 247: This looks identical on every column, is it even needed in the data? At minimum it can be assigned in a loop.

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -479,12 +376,26 @@ export default class TraceStatistics extends Component<Props, State> {
...
+          columns={columns}
+          dataSource={isNoDetail}
+          pagination={{
+            total: isNoDetail.length,
+            pageSizeOptions: ['10', '20', '50', '100'],
+            showSizeChanger: true,
+            showQuickJumper: true,
+          }}
+          rowClassName={record => {
+            if (record.type === 'undefined') {
```
yurishkuro, line 390: I can understand column.type, but what is record.type?

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -392,79 +188,180 @@ export default class TraceStatistics extends Component<Props, State> {
...
+        render: (cell: string) => {
+          return `${cell} %`;
+        },
+        onCell: record => onCellFunction(record),
+      },
+    ];
+
+    const findTablesAccToSelectors = (tableValue: ITableSpan[]): ITableSpan[] => {
+      const isDetailArray: ITableSpan[] = tableValue.filter((val: ITableSpan) => val.isDetail);
+      const isNoDetail: ITableSpan[]  = tableValue.filter((val: ITableSpan) => !val.isDetail);
```
yurishkuro, line 346: the prefix "is" is usually used for Boolean vars, these are arrays. I'd suggest withDetail/withoutDetail

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -392,79 +188,180 @@ export default class TraceStatistics extends Component<Props, State> {
...
+    const onCellFunction = (record: ITableSpan) => {
+      if (this.props.uiFind && record.searchColor !== 'transparent') {
+        return {
+          style: { background: record.searchColor, borderColor: record.searchColor },
+        };
+      }
+      return {
+        style: { background: record.colorToPercent, borderColor: record.colorToPercent },
+      };
+    };
```
yurishkuro, line 216: DRY into one return

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -392,79 +188,180 @@ export default class TraceStatistics extends Component<Props, State> {
...
+          sorterFunction({ type: a.type, index: a.percent }, { type: b.type, index: b.percent }),
+        render: (cell: string) => {
+          return `${cell} %`;
+        },
+        onCell: record => onCellFunction(record),
+      },
+    ];
+
+    const findTablesAccToSelectors = (tableValue: ITableSpan[]): ITableSpan[] => {
+      const isDetailArray: ITableSpan[] = tableValue.filter((val: ITableSpan) => val.isDetail);
```
yurishkuro, line 345: What does isDetail signify?

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
GLVSKiriti: Here sorterFunction returns a boolean value , If (a,b) => is removed we get error because sorter gives a,b as input to the function where a,b are ITableSpan type.

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
GLVSKiriti: yes I think instead of record.type , writing row.type makes more sense

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -392,79 +188,180 @@ export default class TraceStatistics extends Component<Props, State> {
...
+          style: { background: record.searchColor, borderColor: record.searchColor },
+        };
+      }
+      return {
+        style: { background: record.colorToPercent, borderColor: record.colorToPercent },
+      };
+    };
 
-  render() {
+    const columns: ColumnProps<ITableSpan>[] = [
```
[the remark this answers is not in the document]
GLVSKiriti: Because in each row of the table is ITableSpan type, as we already had a interface [ITableSpan](packages/jaeger-ui/src/components/TracePage/TraceStatistics/types.tsx) which has the required attributes to make table. So ITableSpan is used

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
GLVSKiriti: isDetail array means it has all the rows/spans which are belong to the subgroups which we get when we keep valuenameselector2 i.e subgroup attribute in tracestatistics header. So basically if isDetail value of a row is false then it is parent and if isDetail value of a row is true then it is child to the respective parent

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
GLVSKiriti: yes it is needed because the background color should be changed for each row when we search anything in find search box or if valuenameselector3 is used by checking it in traceStatisticsHeader

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
yurishkuro: a function can return another function

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
yurishkuro: row.type is not more clear - what are examples of different row types?

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
GLVSKiriti: I will assign it in a loop

yurishkuro: but the actual assignment is not different across entries, so DRY it by making this assignment in a loop afterwards.

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
GLVSKiriti: then what would you suggest ?? Here each row of the table refers to a span right?

yurishkuro:
> +202 −1,036

I love it that we'll be net-removing ~800 lines of code!

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
GLVSKiriti: yes we can call like isChild but the row which is now in subgroup , If we change valuenameselector1 and valuenameselector2 then the same row can now may be in subgroup or it may be a parent.
Like if isDetail means in my view it tells about whether it is detail info i.e in subgroup of a row or it is a row which is grouped .
what would you suggest to use??

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
GLVSKiriti: Here each row means it is like a span ,

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
yurishkuro: we're getting bogged down in the details. I am looking at the ITableSpan type declaration, and it has `isDetail` field. What does that field mean? Can we write a concise semantic definition for it (and add it as comment to the type)? For example, this is from HotROD app - which rows here are supposed to have isDetail=true?

image

yurishkuro: Does isDetail refer to sub-grouping? E.g. in this pic the 1st row has isDetail=false and 2nd row has isDetail=true? 

image

GLVSKiriti:
Yes it is working like that right? clicking anywhere in the column header causes the sort

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
GLVSKiriti: Ok sure I will add a comment in ITableSpan type declaration about isDetail attribute

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -392,79 +188,159 @@ export default class TraceStatistics extends Component<Props, State> {
...
-            clickColumn={this.clickColumn}
-            colorToPercent={colorToPercent}
-          />
-        );
+  render() {
+    const onClickOption = (type: string, name: string) => {
+      if (this.state.valueNameSelector1 === 'sql.query' && type !== 'undefined') this.togglePopup(name);
+    };
+
+    const sorterFunction = (a: ISorterInput, b: ISorterInput) => {
```
yurishkuro, line 196: I was thinking something like this:
```
function sorterFunction (field: string): CompareFn[ISorterInput, ISorterInput] {
  function sort(a, b: any) {
    // sort based on a[field] / b[field]
  }
}
```

GLVSKiriti: yep 

> I was thinking something like this:
> 
> ```
> function sorterFunction (field: string): CompareFn[ISorterInput, ISorterInput] {
>   function sort(a, b: any) {
>     // sort based on a[field] / b[field]
>   }
>   return sort;
> }
> ```

Yepp!! I modified the code

GLVSKiriti:
Let me know any thing more changes should I do , I think every thing is done now

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -392,79 +188,159 @@ export default class TraceStatistics extends Component<Props, State> {
...
+      {
+        title: 'Name',
+        dataIndex: 'name',
+        sorter: sorterFunction('name'),
+        render: (name: string, row: ITableSpan) => {
+          return (
+            <span
+              onClick={() => onClickOption(row.type, row.name)}
+              style={{
+                borderLeft: `4px solid ${`${row.color}` || `transparent`}`,
```
yurishkuro, line 229:
```suggestion
                borderLeft: `4px solid ${row.color || `transparent`}`,
```
Wouldn't this work?

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -479,12 +355,26 @@ export default class TraceStatistics extends Component<Props, State> {
...
+          columns={columns}
+          dataSource={withoutDetail}
+          pagination={{
+            total: withoutDetail.length,
+            pageSizeOptions: ['10', '20', '50', '100'],
+            showSizeChanger: true,
+            showQuickJumper: true,
+          }}
+          rowClassName={row => {
+            if (row.type === 'undefined') {
```
yurishkuro, line 369: What is the meaning of row.type? What are the possible values?

packages/jaeger-ui/src/components/TracePage/TraceStatistics/types.tsx
```diff
@@ -25,7 +25,7 @@ export interface ITableSpan {
   selfMin: number;
   selfMax: number;
   percent: number;
-  isDetail: boolean;
+  isDetail: boolean;//It is about sub-grouping.It is true if it belongs to subgroup.
```
yurishkuro, line 28:
```suggestion
  isDetail: boolean; // True is the entry represents a subgroup aggregation.
```

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -392,79 +188,159 @@ export default class TraceStatistics extends Component<Props, State> {
...
+          return `${cell} %`;
+        },
+      },
+    ];
+
+    columns = columns.map(val => {
+      return { ...val, onCell: record => onCellFunction(record) };
+    });
+
+    const findTablesAccToSelectors = (tableValue: ITableSpan[]): ITableSpan[] => {
```
yurishkuro, line 323: Can you describe in words what this function is doing? The name seems very non descriptive

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -392,79 +188,159 @@ export default class TraceStatistics extends Component<Props, State> {
...
+        });
+        const child = {
+          key: i.toString(),
+          children: newArr,
+        };
+        withoutDetail[i] = { ...withoutDetail[i], ...child };
+      }
+      return withoutDetail;
+    };
+    const withoutDetail: ITableSpan[] = findTablesAccToSelectors(this.state.tableValue);
```
yurishkuro, line 343: At this point why is withoutDetail an appropriate name?

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.css
```diff
@@ -29,3 +29,18 @@ limitations under the License.
 .test1893 {
   width: 99.999%;
 }
+
```
yurishkuro, line 32: Is test style needed?

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.css
```diff
@@ -29,3 +29,18 @@ limitations under the License.
...
+}
+
+.MainTableData--tr {
+  border-top: 1px solid rgb(255, 255, 255);
+}
+
+.MainTableData--tr:hover {
+  background: rgb(250, 250, 250);
+  color: rgb(0, 0, 0);
+}
```
yurishkuro, line 46: Two styles above are namespaced by a suffix --TraceStatistics. Why wouldn't we use the same naming pattern?

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
GLVSKiriti: yes it works

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
GLVSKiriti: When you subgroup the rows based on (say http.status code here)in that subgroup if there are no spans with that tag then those rows are of type undefined.
let us see in this pic now the row under the mysql row is of type undefined and row under the customer row is of type defined
And possible values are defined and undefined
new1

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
GLVSKiriti: This functions makes the datasource for the antd table according to the valuenameselector values. Like it adds the child attribute for each element in datasource if we give valuenameselector2 (sub-grouping) . So this function finally outputs what are the main rows and what are the sub-group rows of  each main row

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
GLVSKiriti: Like withDetail refers to the inner details rows or subgroups of each main row. So each main row does not have isDetail value we kept it as withoutDetail

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.css
GLVSKiriti: No I will remove it

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
yashrsharma44:
```suggestion
    /**
    * Pre-process the table data into groups and sub-groups
    */
    const findTablesAccToSelectors = (tableValue: ITableSpan[]): ITableSpan[] => {
```

yashrsharma44 (commented):
I have added some comments regarding the description of the function's behaviour. Also have made some suggestions regarding the naming of the functions, let me know if this makes sense.

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
yurishkuro: It would also be helpful to have a comment that defines an invariant about the data shape expectation. Say I have 2 groups, each has 3 subgroups. Does the final output structure has `2 + 2 * 3 = 8` entries, or only 2 entries each having 3 children? I found it very hard to deduce just from reading the code.

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
yurishkuro: Thanks, now it makes sense. Sounds like `type` is a terrible name to describe this concept -- what you have instead is a Boolean indicator telling whether the row has a subgroup label defined or not (the catch-all case), e.g. `hasSubgroupValue`. Can we change this and make a corresponding comment in the type declaration?

packages/jaeger-ui/src/components/TracePage/TraceStatistics/types.tsx
```diff
@@ -25,7 +25,7 @@ export interface ITableSpan {
   selfMin: number;
   selfMax: number;
   percent: number;
-  isDetail: boolean;
+  isDetail: boolean; // True is the entry which represents a subgroup aggregation.
```
yurishkuro, line 28:
```suggestion
  isDetail: boolean; // True when the entry represents a subgroup aggregation.
```

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -42,75 +39,6 @@ type State = {
   valueNameSelector2: string | null;
 };
 
-const columnsArray: any[] = [
```
yurishkuro, line 45: why did we delete this structure? It looks like it contained a lot of semantic details that you could've used to construct the columns automatically, like using `suffix` to decide which render should be used, or `isDecimal` to decide which comparison function to use. It would also make the PR smaller.

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -392,79 +188,162 @@ export default class TraceStatistics extends Component<Props, State> {
...
-            sortClick={this.sortClick}
-            sortAsc={sortAsc}
-          />
-        ))}
-      </tr>
-    );
-  }
+    const sorterFunction = <T extends keyof ITableSpan>(field: T): CompareFn<ITableSpan> => {
+      const sort = (a: ITableSpan, b: ITableSpan) => {
+        if (a.type === 'undefined') {
```
yurishkuro, line 198: Does `type` here have a similar meaning to `row.type`? Except for `Name` all other columns are numeric, so when is this `undefined`?

GLVSKiriti: Yes here also type has same meaning as row.type 
Here a.type is about each row not about sorting column.
Here a.type value is checking because
Basically we want to keep the undefined row at the bottom of subgroup irrespective of sorting so to attain this we are writing this conditions

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
GLVSKiriti: Yes that might be a good idea , I will make those changes in code

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
GLVSKiriti: Yep modified in that way

packages/jaeger-ui/src/components/TracePage/TraceStatistics/types.tsx
```diff
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 export interface ITableSpan {
-  type: string;
+  hasSubgroupValue: boolean; // True is the entry which has the subgroup tag in it.
```
yashrsharma44, line 16:
```suggestion
  hasSubgroupValue: boolean; // True when the entry has the subgroup tag in it.
```

yashrsharma44 (approved):
I think it looks good to me, let's wait for @yurishkuro's review 🙂

yurishkuro (approved):
lgtm

yurishkuro:
not clear why unit tests are failing, did you run them locally?

GLVSKiriti:
> not clear why unit tests are failing, did you run them locally?
Error is causing because I didn't removed the test cases for the deleted functions in index.tsx
Now I fixed them

yurishkuro (approved):

yurishkuro:
CI still failing. I recommend making sure `yarn test` and `yarn lint` are successful locally.

GLVSKiriti:
> CI still failing. I recommend making sure `yarn test` and `yarn lint` are successful locally.

Ok yarn lint also runned locally, Now every thing is working fine locally

yurishkuro (approved):

yurishkuro:
🎉 🎉 🎉 🎉 🎉 🎉

yashrsharma44:
❤️ 🎉

PR #1521, opened 2023-06-27
Add boolean property to config

Binrix:
This pull request resolves the issue #1211 

This property is later used to prevent the popup of a new page. When set to true, it will alter all _blank targets for links to empty. A utils function is provided.

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TracePageHeader/AltViewOptions.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TracePageHeader/AltViewOptions.tsx
@@ -27,6 +27,7 @@ import {
 } from './TracePageHeader.track';
 import prefixUrl from '../../../utils/prefix-url';
 import { ETraceViewType } from '../types';
+import { getTargetBlankOrTop } from '../../../utils/config/get-target';
 
 type Props = {
   onTraceViewChange: (viewType: ETraceViewType) => void;
@@ -88,7 +89,7 @@ export default function AltViewOptions(props: Props) {
           <Link
             to={prefixUrl(`/api/traces/${traceID}?prettyPrint=true`)}
             rel="noopener noreferrer"
-            target="_blank"
+            target={getTargetBlankOrTop()}
             onClick={trackJsonView}
           >
             Trace JSON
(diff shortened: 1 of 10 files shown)


Binrix:
As discussed before, I added a utils function that checks if the boolean property in the config is set to true or false and sets the target of a link to _blank, _top or empty.

packages/jaeger-ui/src/components/TracePage/TracePageHeader/AltViewOptions.tsx
```diff
@@ -88,7 +89,7 @@ export default function AltViewOptions(props: Props) {
           <Link
             to={prefixUrl(`/api/traces/${traceID}?prettyPrint=true`)}
             rel="noopener noreferrer"
-            target="_blank"
+            target={getTargetBlankOrTop()}
```
yurishkuro, line 92: why did you not think EmptyOrBlank would work here?

packages/jaeger-ui/src/utils/config/get-target.test.js
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
yurishkuro, line 1:
```suggestion
// Copyright (c) 2023 The Jaeger Authors
```

packages/jaeger-ui/src/utils/config/get-target.tsx
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
yurishkuro, line 1:
```suggestion
// Copyright (c) 2023 The Jaeger Authors
```

yurishkuro (commented):
lgtm

packages/jaeger-ui/src/components/TracePage/TracePageHeader/AltViewOptions.tsx
Binrix: The problem is that when the target is empty, the json will be shown in the same window and the user cannot return to the page before (at least on the instrument because it has no keyboard or possibility to reach a return button from the browser). This is why I decided to use top: when the target is top the user can still access the navigation bar and go back.

yurishkuro (approved):

PR #1535, opened 2023-07-03
Replaced the copy icons with plain text

### LINKED ISSUE #1506: [Feature]: Replace Copy icons with text buttons
In #1505 we added a second copy icon to allow copying tag value, not always full JSON.

image

It would increase usability if instead of the icons we used plain text buttons "Copy" and "JSON".

Player256:
Signed-off-by: Player256 <dattucodes@gmail.com>

- Resolves #1506 

Jaeger UI - Brave 03-07-2023 21_12_03

### DIFF
--- a/packages/jaeger-ui/src/components/common/CopyIcon.tsx
+++ b/packages/jaeger-ui/src/components/common/CopyIcon.tsx
@@ -27,6 +27,7 @@ type PropsType = {
   icon?: string;
   placement?: TooltipPlacement;
   tooltipTitle: string;
+  buttonText: string;
 };
 
 type StateType = {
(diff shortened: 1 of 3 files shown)


yurishkuro:
do you need to style the button to be more compact? It looks like it's breaking the layout, makes the rows taller.

packages/jaeger-ui/src/components/common/CopyIcon.tsx
```diff
@@ -66,14 +67,11 @@ export default class CopyIcon extends React.PureComponent<PropsType, StateType>
         mouseLeaveDelay={0.1}
         onVisibleChange={this.handleTooltipVisibilityChange}
         placement={this.props.placement}
-        title={this.state.hasCopied ? 'Copied' : this.props.tooltipTitle}
+        title={this.state.hasCopied ? 'Copied' : ''}
```
yurishkuro, line 70: will the empty string make the tooltip not show and only appear on copy?

Player256: yes

Player256:
> do you need to style the button to be more compact? It looks like it's breaking the layout, makes the rows taller.
Jaeger UI - Brave 04-07-2023 17_44_51 (2) 
Is this fine?

yurishkuro:
it still doesn't look right - buttons should not be messing with the row size. Also, why are the two buttons  not aligned vertically?

Player256:
> it still doesn't look right - buttons should not be messing with the row size. Also, why are the two buttons not aligned vertically?

its the transform : translate(-3px) css property that makes it uplift. if not needed I can remove it.

yurishkuro:
resolved by another PR #1547

PR #1547, opened 2023-07-06
Add text label to copy buttons on tags

### LINKED ISSUE #1506: [Feature]: Replace Copy icons with text buttons

Yushmanth-reddy:
-  Resolves #1506 

- Previously there we only have icons to copy text and JSON. But now we have implemented plain text along with the icons.
Screenshot from 2023-07-06 13-23-12

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.tsx
@@ -153,12 +153,14 @@ export default function KeyValuesTable(props: KeyValuesTableProps) {
                     className="KeyValueTable--copyIcon"
                     copyText={row.value}
                     tooltipTitle="Copy value"
+                    buttonText="Copy"
                   />
                   <CopyIcon
-                    className="KeyValueTable--copyIcon"
+                    className="KeyValueTable--copyIcon copyIcon"
                     icon="snippets"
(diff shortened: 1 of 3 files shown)


yurishkuro (dismissed):
looks great!

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.tsx
```diff
@@ -153,12 +153,14 @@ export default function KeyValuesTable(props: KeyValuesTableProps) {
                     className="KeyValueTable--copyIcon"
                     copyText={row.value}
                     tooltipTitle="Copy value"
+                    buttonText="Copy"
                   />
                   <CopyIcon
-                    className="KeyValueTable--copyIcon"
+                    className="KeyValueTable--copyIcon copyIcon"
```
yurishkuro, line 159: `copyIcon` - why is this needed? It's not even defined in the css file.

Yushmanth-reddy: Sorry my bad I thought to use it in css of keyValueTable.css but later realised to use the css of Copyicon.css. Finally it is of no use.Let me make the changes.

yurishkuro (approved):

PR #1565, opened 2023-07-10
[chore]: minimize eslint setting overrides

### LINKED ISSUE #1258: [chore]: minimize eslint setting overrides
### Requirement

To keep the code in the best forward-compatible shape with linters, it's best to minimize the number of lint rule overrides.

### Problem

https://github.com/jaegertracing/jaeger-ui/pull/1256#discussion_r1133295150

### Proposal

make the root `eslintrc` also extend the recommended config from `@typescript-eslint`, because that config already does this toggling for us. This is what `plexus` does already. 

### Open questions

_No response_

anikdhabal:
Resolves #1258 

made the root eslintrc also extend the recommended config from @typescript-eslint.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -47,6 +47,7 @@ module.exports = {
         ],
 
         // Disable ESLint core rules for which @typescript-eslint provides TypeScript-specific equivalents.
+        '@typescript-eslint/no-this-alias': 0,
         'no-unused-vars': 0,
         '@typescript-eslint/no-unused-vars': 1,
         'no-use-before-define': 0,
@@ -79,10 +80,10 @@ module.exports = {
     'prefer-destructuring': 0,
     
     /* tsx */
-    "@typescript-eslint/no-var-requires": 1,
-    "@typescript-eslint/no-empty-function": 1,
-    "@typescript-eslint/ban-types": 1,
-    "@typescript-eslint/ban-ts-comment": 1,
+    "@typescript-eslint/no-var-requires": "warn",
+    "@typescript-eslint/no-empty-function": "warn",
+    "@typescript-eslint/ban-types": "warn",
+    "@typescript-eslint/ban-ts-comment": "warn",
 
     /* jsx */
     'jsx-a11y/anchor-is-valid': 0,
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -77,6 +77,12 @@ module.exports = {
     'no-self-compare': 0,
     'no-underscore-dangle': 0,
     'prefer-destructuring': 0,
+    
+    /* tsx */
+    "@typescript-eslint/no-var-requires": 1,
+    "@typescript-eslint/no-empty-function": 1,
+    "@typescript-eslint/ban-types": 1,
+    "@typescript-eslint/ban-ts-comment": 1,
 
     /* jsx */
     'jsx-a11y/anchor-is-valid': 0,
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 module.exports = {
+  root:true,
   env: {
     browser: true,
     jest: true,
@@ -25,20 +26,8 @@ module.exports = {
       },
     },
   },
-  extends: ['airbnb', 'prettier'],
+  extends: ['airbnb', 'prettier','eslint:recommended','plugin:@typescript-eslint/recommended'],
   overrides: [
-    {
-      files: ['*.jsx', '*.test.js'],
-      parser: '@babel/eslint-parser',
-      parserOptions: {
-        requireConfigFile: false,
-        babelOptions: {
-          babelrc: false,
-          configFile: false,
-          presets: ["@babel/preset-react"],
-        },
-      }
-    },
     {
       files: ['*.ts', '*.tsx'],
       parser: '@typescript-eslint/parser',
@@ -73,13 +62,6 @@ module.exports = {
         'react/require-default-props': 0,
         'react/default-props-match-prop-types': 0,
         'react/no-unused-prop-types': 0,
-      },
-    },
-    {
-      files: ['*.test.js'],
-      rules: {
-        // Used for Jest module mocking.
-        'no-import-assign': 0,
       },
     },
   ],
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -48,8 +48,6 @@ module.exports = {
 
         // Disable ESLint core rules for which @typescript-eslint provides TypeScript-specific equivalents.
         '@typescript-eslint/no-this-alias': 0,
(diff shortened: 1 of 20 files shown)


anikdhabal:
> What could be the testing methodology for this PR? I.e., just having a green CI seems insufficient as it doesn't prove that a group of files are not getting excluded from linting after the change.

I've tested it locally using the "eslint", but I don't think it is relevant. Any suggestion?

yurishkuro:
maybe introducing some errors in the files and observing that linting errors are reported

anikdhabal:
> maybe introducing some errors in the files and observing that linting errors are reported

Yes, I've debugging and it is inducing "linting issue" in many files. Due to typescript-eslint/recommended.

anikdhabal:
> maybe introducing some errors in the files and observing that linting errors are reported

@yurishkuro Instead of extend the recommended config from @typescript-eslint if we only minimize the setting overrides and also make root eslintrc , then it would be fine. Due to @typescript-eslint/recomended several linting has been occurred in many files.

anikdhabal:
@yurishkuro can you review it once?

.eslintrc.js
```diff
@@ -95,6 +77,12 @@ module.exports = {
     'no-self-compare': 0,
     'no-underscore-dangle': 0,
     'prefer-destructuring': 0,
+    
+    /* tsx */
+    "@typescript-eslint/no-var-requires": 1,
+    "@typescript-eslint/no-empty-function": 1,
+    "@typescript-eslint/ban-types": 1,
+    "@typescript-eslint/ban-ts-comment": 1,
```
yurishkuro, line 85: please explain the purpose of these settings. Are they different from some defaults?

I am all for enabling more rues, but it seems this section was previously used to _disabled_ some of them, so I am not clear why we need to enable others, and why specifically these 4 and not more.

packages/jaeger-ui/src/components/DeepDependencies/Graph/DdgNodeContent/calc-positioning.tsx
```diff
@@ -102,7 +102,7 @@ export function _initOpSpan() {
  *     recursive call stack.
  * @returns {number} - Width of narrowest rectangle that contains given words across given lines.
  */
-function calcWidth(lengths: number[], lines: number, longestThusFar: number = 0): number {
+function calcWidth(lengths: number[], lines: number, longestThusFar = 0): number {
```
yurishkuro, line 105: what is the reason for this change? Is a new linter rule complaining about it? It seems like a normal syntax.

anikdhabal: By default it's giving error, called no-inferrable-types. It disallow explicit type declarations for variables or parameters initialized to a number, string, or boolean.

.eslintrc.js
anikdhabal: Because, the new lint is giving error of these 4 types. So, I added these and convert the error to warning. Eslint by default doesn't handle these settings, it's always giving error if we do not handle it.

packages/jaeger-ui/src/constants/index.tsx
```diff
@@ -12,14 +12,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-export const TOP_NAV_HEIGHT = 46 as 46;
+export const TOP_NAV_HEIGHT = 46 as const;
```
yurishkuro, line 15: what's the difference if this simply said `export const TOP_NAV_HEIGHT = 46;`, without `as`?

packages/jaeger-ui/src/model/ddg/PathElem.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-this-alias */
```
yurishkuro, line 1: No a fan of disabling - can we remove the aliasing instead? And if we must disable, it's should be localized only to where needed, not the whole file.

anikdhabal: Ok Understood

packages/jaeger-ui/src/constants/index.tsx
anikdhabal: nothing, I just fixed it.

yurishkuro (dismissed):
thanks!

yurishkuro:
looks like linting is failing https://github.com/jaegertracing/jaeger-ui/actions/runs/5836530756/job/15830302191?pr=1565

anikdhabal:
> looks like linting is failing https://github.com/jaegertracing/jaeger-ui/actions/runs/5836530756/job/15830302191?pr=1565

Yes, trying to fix it.

packages/jaeger-ui/src/components/DeepDependencies/index.tsx
```diff
@@ -173,7 +173,7 @@ export class DeepDependencyGraphPageImpl extends React.PureComponent<TProps, TSt
     const { graphState } = this.props;
     const { visEncoding } = this.props.urlState;
 
-    if (graphState && graphState.state === fetchedState.DONE) {
+    if (graphState && graphState.state === fetchedState.DONE && 'model' in graphState) {
```
yurishkuro, line 176: this doesn't feel like the right solution to me. Is the reason for linting errors due to missing or incomplete type declaration? Based on the existing code it looks like the `model` field is always expected under these conditions, so why do we need to test for it at runtime, instead of enforce it via type checking?

anikdhabal: Yes,this is the reason for linting errors. Yeah,the is approach in not intuitive. Let me fix it.

anikdhabal: The problem is [TS2339: Property does not exist on type {}]

yurishkuro: why can't we add it to the type declaration?

anikdhabal: Yes, I've added it to the type declaration, now it's fine. Let me fix the PR.

anikdhabal:
@yurishkuro fixed

yurishkuro (approved):

anshgoyalevil:
286 eslint warnings were introduced with PR
https://github.com/jaegertracing/jaeger-ui/actions/runs/5907286634/job/16024953128#step:6:636
Was this the expected behavior?

yurishkuro:
> 286 eslint warnings were introduced with PR ... Was this the expected behavior?

Not ideal. Perhaps we can fine tune the settings, on case by case basis.

PR #1576, opened 2023-07-13
[feat]: Support Google Analytics 4 properties

### LINKED ISSUE #1332: [Feature]: Support Google Analytics 4 properties
### Requirement

```
```

### Problem

### Proposal

### Open questions

_No response_

anikdhabal:
Resolves #1332 

-


packages/jaeger-ui/index.html
```diff
@@ -14,6 +14,18 @@
...
+    <script
+      async
+      src="https://www.googletagmanager.com/gtag/js?id=TAG_ID" >
+    </script>
+    <script>
+      window.dataLayer = window.dataLayer || [];
+      function gtag(){dataLayer.push(arguments);}
+      gtag('js', new Date());
+      gtag('config', 'TAG_ID')
+    </script>
```
yurishkuro, line 28: this is now how we use GA, please see the discussion in the ticket.

anikdhabal: In this way GA4 analytics support without any 3rd party libraries.

yurishkuro: yes, for tracking different pages of a static website, but not of a Reach app that is a single page.

PR #1582, opened 2023-07-14
[feature] Visualize Critical Path of a trace

### LINKED ISSUE #1288: [Feature]: Visualize Critical Path of a trace
This is filed as an intern project for the LFX Mentorship. Please refer to  https://github.com/jaegertracing/jaeger/issues/4453. [Apply here](https://mentorship.lfx.linuxfoundation.org/project/0fc6c44b-5ddf-467f-8016-72cc35b4e3ff).

### Requirement

Have an ability to view critical path for a given trace.

### Problem

A call graph can have multiple spans, but not all spans really contribute to the latency computation. Critical path is the longest chain of dependent tasks in a microservice dependency graph( i.e.,specific sequence of spans from the root to a leaf) . Reducing the critical path length is necessary to reduce the end-to-end latency of a request. Hence, a latency optimization effort benefits by prioritizing the services that are on the critical path. A tool to help identify the critical path of any trace I choose to investigate is  beneficial for developers.

Tools like "Lightstep observability" are already having this feature. This effort will be a feature parity for jaeger with other tools.

### Proposal

Jaeger already supports a bunch of different analysis views of a trace. It would be useful if Jaeger UI would allow a developer to view the critical path of the trace in the same context. 

We propose adding a new option under that dropdown labeled “Critical path”. A critical path is a specific sequence of spans from the root to a leaf. The [algorithm](https://github.com/uber-research/CRISP) to compute the critical path can be run in real-time even in the browser. Once the critical path is computed, it needs to be presented to the developer. 

The critical path could be rendered in either (or both) of the formats - tabular and flamegraph.
For efficiency, we may want to re-use the tabular and flame graph trace analysis views already available in Jaeger.

### Open questions

_No response_

GLVSKiriti:
This PR is to add support(UI) for critical path analysis to Jaeger UI.
Resolves #1288 

Respective Algorithm PR is #1532 
@yurishkuro @yashrsharma44


GLVSKiriti:
Now present critical path is shown like this (Any suggestions to change the UI further??)
cp1
cp2

yurishkuro:
have you tried how this [would look](https://twitter.com/mathias/status/1237393102635012101?s=20) for people with color-blindness? I personally would prefer red line for CP, if it gives enough contrast.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBar.css
```diff
@@ -44,6 +44,13 @@ limitations under the License.
   z-index: 1;
 }
 
+.SpanBar--cp {
```
yurishkuro, line 47:
```suggestion
.SpanBar--criticalPath {
```
nit: don't be cryptic

yurishkuro:
how does this look when client-server spans are collapsed? Do we display dual color overlay?

image

GLVSKiriti:
> have you tried how this [would look](https://twitter.com/mathias/status/1237393102635012101?s=20) for people with color-blindness? I personally would prefer red line for CP, if it gives enough contrast.

Yepp I tried how the view looks for the people with color-blindness, CP path is differentiable from the spans color. 
And when I use red line for CP the view is like this , For those spans which has colors like orange or brown , I think it doesn't gives enough contrast. So what will you prefer using red or black?
cp3

GLVSKiriti:
> how does this look when client-server spans are collapsed? Do we display dual color overlay?
> image

Present it works as before only. CP line is not showed when spans are collapsed. CP path is only shown on span for which is not in collapsed state. Now the doubts I have are
1.) Should I also want to show CP line on collapsed span ??
2.) If yes, how can we differentiate the CP path(i.e CP line refers to parent span or child span)?

yurishkuro:
> So what will you prefer using red or black?

let's keep black.

> how can we differentiate the CP path(i.e CP line refers to parent span or child span)?

We don't! We show combined CP. I actually made this point in my book. Imagine you have a parent `t=(1-10)` with one child `t=(4-6)` in the middle.  This gives `CP={(1-4, P), (4-6, C), (6-10, P)}`. This is what your code does currently. Now imaging the following two scenarios:
  1. child was never instrumented and we don't have a child span. The CP becomes `(1-10, P)`, right?
  1. child span is present, but we collapsed the parent span so that none of its children are visible.

What should be the CP in the second scenario? Since we do not display the child span, should we only highlight `CP={(1-4, P), (6-10, P)}`? That doesn't make sense, what happens during the middle part? I would argue that when the parent is collapsed, situations 1 and 2 are identical, and we should display `CP=(1-10, P)`.

This approach incidentally solves the question you asked: if parent and child are combined in one bar with two colors, now to display CP? Answer: we take all the CP segments of the parent and all of its children and we merge them into as few segments as possible (whenever they join), and draw it on the combined parent/child span. When the collapsing does not result in merged spans (they only merge for client/server pairs), _we do the same thing_, merge CP segments and show combined CP on the parent, that represents everything that's" hidden" by collapsing the parent.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBar.css
```diff
@@ -44,6 +44,13 @@ limitations under the License.
   z-index: 1;
 }
 
+.SpanBar--criticalPath {
+  position: absolute;
+  top: 46%;
+  height: 10%;
+  z-index: 1;
```
yashrsharma44, line 51: If we want to overlay critical path on top of spans, let's increase the z-index to more than 1.
```suggestion
  z-index: 2;
```

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBar.tsx
```diff
@@ -59,8 +60,13 @@ function toPercent(value: number) {
   return `${(value * 100).toFixed(1)}%`;
 }
 
+function toPercentInDecimal(value: number) {
+  return `${value * 100}%`;
```
yashrsharma44, line 64: Can we cap the value to be a positive number?

yashrsharma44:
To explain @yurishkuro's idea with an visual example - 
image

This looks fair to me, especially when we have a deeply nested span contain child spans more than 3-4 levels deep. 

Aggregating the critical path to the parent span(say Span A) helps us identify which parent span contains the critical path among the others, as we can always expand on the specific parent span with child spans, and adjust our rendering of critical path as we go deep into the call stack.

yashrsharma44:
Some questions I had regarding the UI, feel free to let me know if this is out of scope for this problem - 

* Do we want to enable critical path algorithm feature for all the spans? Like having an option to enable on demand would be a better option, as we might consumed CPU/Memory resources on the user's side for running the algorithm every-time we load the page. If we want to enable this on-demand, let's have a dropdown/toggle button to enable this?
* Let's add some test-cases for testing the rendering time for the UI. What happens when we have a large number of Spans(parent span with 1 child for example), and we want to run the critical path algorithm?
* For huge number of spans, we should be able to scroll-down to the critical path span automatically, rather than having to visually find out where the span is located. Having a marker/visual cue to highlight the critical path, is something that would be a good bonus, as users can quickly get confused once we have multiple number of spans.

The rest of the changes looks good to me :)

yurishkuro:
> For huge number of spans, we should be able to scroll-down to the critical path span automatically

that's a good callout, if you have a span with critical path like this `[    ======   ]`, you may want a quick way to navigate to the next segment of the CP, which may not be visible on the same page. Not sure how to implement it visually.

yurishkuro:
> Do we want to enable critical path algorithm feature for all the spans? Like having an option to enable on demand

we discussed this already elsewhere - unless there is an explicit performance concern, I would not bother with a toggle and always display it.

GLVSKiriti:
@yurishkuro @yashrsharma44 This PR is also ready to merge. 
Kindly Review it

packages/jaeger-ui/src/components/TracePage/CriticalPath/utils/sanitizeOverFlowingChildren.tsx
```diff
@@ -0,0 +1,110 @@
...
+ * to ensure they fit within the time range of their parent span.
+ * @param spanMap - A Map where span IDs are keys and the corresponding spans are values.
+ * @returns - A sanitized span Map.
+ */
+const sanitizeOverFlowingChildren = (spanMap: Map<string, Span>): Map<string, Span> => {
+  let spanIds: string[] = [...spanMap.keys()];
+
+  spanIds.forEach(spanId => {
+    const span = spanMap.get(spanId)!;
+    if (span.references.length) {
```
yashrsharma44, line 30:
```suggestion
    if (span && span.references.length) {
```

yashrsharma44:
I took a look at the PR, and while we are addressing the changes, let's gate this feature(Critical Path Algorithm Component) using a config value - https://github.com/jaegertracing/jaeger-ui/blob/fabd2e756113b8bfdc2103c657ee2915b35fa56d/packages/jaeger-ui/src/constants/default-config.tsx

We can disable this feature to start with, and add the option to enable this, if people want to try this feature. We can merge this PR while keeping this feature gated, and add some follow up items to improve the space complexity of this algorithm. I can suggest more review items, but I think this would make this PR more lengthier than I would like.

packages/jaeger-ui/src/components/TracePage/CriticalPath/index.test.js
```diff
@@ -0,0 +1,43 @@
...
+import getChildOfSpans from './utils/getChildOfSpans';
+import sanitizeOverFlowingChildren from './utils/sanitizeOverFlowingChildren';
+import test6 from './testCases/test6';
+import test7 from './testCases/test7';
+import test5 from './testCases/test5';
+
+describe.each([[test1], [test2], [test3], [test4], [test5], [test6], [test7]])('Happy Path', testProps => {
+  it('Should find criticalPathSections correctly', () => {
+    const rootSpanId = testProps.trace.spans[0].spanID;
+    const SpanMap = testProps.trace.spans.reduce((map, span) => {
```
yashrsharma44, line 29: Can we change this to `spanMap`? Capitalised variables are generally used for class names, and camelCase for variables.
```suggestion
    const spanMap = testProps.trace.spans.reduce((map, span) => {
```

GLVSKiriti:
I don't know why the unit tests are failing. Unit tests are passing in my local machine

yurishkuro (approved):

bobrik:
I tried out the jaeger v1.49.0 and it doesn't seem to show what I would expect to see:

image

I would expect to see `request` to be a part of the critical path here.

bobrik:
Manually adjusting duration in a json file doesn't help:

image

yurishkuro:
is this a follows-from span kind? in the first s/s it was after the parent span so not on the critical path of THAT span, and CP is defined recursively starting from the root span, at least in the current version.

yurishkuro:
btw, glad to see you're trying it out, bug reports are very welcome (maybe start another ticket).

PR #1583, opened 2023-07-14
[feat] Support Google Analytics 4 properties

### LINKED ISSUE #1332: [Feature]: Support Google Analytics 4 properties

anikdhabal:
 Resolves #1332 

- 
Support Google Analytics 4 properties


anikdhabal:
@yurishkuro can you review this PR?

package.json
```diff
@@ -63,5 +63,8 @@
     "hooks": {
       "pre-commit": "npm-run-all -ln --parallel lint test"
     }
+  },
+  "dependencies": {
+    "react-ga4": "^2.1.0"
```
yurishkuro, line 68: is it needed here? already added at the lower-level package.json

packages/jaeger-ui/src/utils/tracking/ga.tsx
```diff
@@ -56,7 +56,7 @@ const GA: IWebAnalyticsFunc = (config: Config, versionShort: string, versionLong
       msg = `jaeger/${msg}`;
     }
     msg = msg.slice(0, 149);
-    ReactGA.exception({ description: msg, fatal: false });
```
yurishkuro, line 59: does the new module not support a distinction between exceptions and other events?  The `send` in L155 specifies `hitType`, should this one also have `hitType=exception`?

yurishkuro (dismissed):
please remove `package-lock.json` from PR

packages/jaeger-ui/src/utils/tracking/ga.tsx
anikdhabal: Ahh My bad this should be hitType=exception . Thank you.

package.json
anikdhabal: ok ok

yurishkuro (approved):
lgtm. Did you test it with a real GA ID?

anikdhabal:
> lgtm. Did you test it with a real GA ID?

Yes, it is fine.

yurishkuro:
> Yes, it is fine.

can you include some screenshots?

anikdhabal:
> > Yes, it is fine.
> 
> can you include some screenshots?

Okk I will share some result.

yurishkuro:
@anikdhabal this PR does not include any configuration changes, so the same GA ID config will be used. Will it work if someone has the old GA ID in their config, which is not GA4?

anikdhabal:
> @anikdhabal this PR does not include any configuration changes, so the same GA ID config will be used. Will it work if someone has the old GA ID in their config, which is not GA4?

No, it doesn't work for the old GA ID; it only works for GA4. I haven't found a straightforward way to support the old GA configuration. Let me do a bit more research about it or talk to the maintainer to include the configuration.

yurishkuro:
> it only works for GA4. I haven't found a straightforward way to support the old GA configuration

Should we keep using both libraries and switch depending on the version of GA ID? Do you know what the timeline for Google to completely phase out old IDs?

anshgoyalevil:
According to [Google](https://support.google.com/analytics/answer/10089681?hl=en&ref_topic=12154439,12153943,2986333,&sjid=17505237808164032051-AP&visit_id=638285326746816164-286965898&rd=1),

> Starting July 1, 2023, standard UA properties [will stop processing data](https://support.google.com/analytics/answer/11583528)

yurishkuro:
Yeah, then I think it's fine to drop support for UA, and only support GA4

anikdhabal:
> Yeah, then I think it's fine to drop support for UA, and only support GA4

Yeah it's fine.

yurishkuro:
CI is still failing on this pr

anikdhabal:
> CI is still failing on this pr

I am fixing it now.

yurishkuro:
Fixed in #2071

PR #1584, opened 2023-07-14
Replace the usage of archived package `recompose` with plain hooks

zhuje:
- N/A

- The package `recompose` ([repo](https://github.com/react-recompose/react-recompose)) has been archived as of September 10, 2022.
- It is was recommended we replace with package with plain React Hooks. 
 
Screenshot 2023-07-14 at 11 27 30 AM


yurishkuro:
based on the description in `recompose`, its author recommends using react hooks. So I am not sure why we want to replace the usage of this lib with a similar lib. Also, there are just two places in the code that use recompose
```
./packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBar.tsx:18:import { onlyUpdateForKeys, compose, withState, withProps } from 'recompose';
./packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ScatterPlot.jsx:19:import { compose, withState, withProps } from 'recompose';
```

Is it possible to just change those to something else, most standard?

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ScatterPlot.jsx
```diff
@@ -16,20 +16,29 @@ import React, { useRef, useState, useLayoutEffect } from 'react';
...
-  const { data, onValueClick, overValue, onValueOver, onValueOut, calculateContainerWidth } = props;
+export default function ScatterPlot(props) {
+  const { data, onValueClick, calculateContainerWidth } = props;
 
   const containerRef = useRef(null);
   const [containerWidth, setContainerWidth] = useState(0);
 
+  const [overValue, setValueOver] = useState(null);
+
+  const onValueOver = value => {
```
yurishkuro, line 34: Is there a way to simulate this event in unit tests?

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBar.tsx
```diff
@@ -59,20 +52,19 @@ function toPercent(value: number) {
...
   const {
     viewEnd,
     viewStart,
     getViewedBounds,
     color,
-    label,
     hintSide,
     onClick,
-    setLongLabel,
-    setShortLabel,
```
yurishkuro, line 72: So previously these functions could've been supplied externally?  Is it safe to remove them?

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBar.tsx
```diff
@@ -81,12 +73,22 @@ function SpanBar(props: TInnerProps) {
...
+  };
+
   return (
     <div
       className="SpanBar--wrapper"
       onClick={onClick}
-      onMouseOut={setShortLabel}
-      onMouseOver={setLongLabel}
+      onMouseOut={handleMouseOut}
+      onMouseOver={handleMouseOver}
```
yurishkuro, line 91: nit: can you keep the same names setShortLabel/setLongLabel? In fact, you're breaking the separation of concerns somewhat, `handleMouseOut` has nothing to do with mouse out in its function, it's just setting the label, and where it's _used_ does not need to be reflected in its name.

yurishkuro (commented):
lgtm! a couple minor questions, and it would be good if the callbacks could be exercised in unit tests

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBar.tsx
```diff
@@ -59,20 +52,19 @@ function toPercent(value: number) {
   return `${(value * 100).toFixed(1)}%`;
 }
 
-function SpanBar(props: TInnerProps) {
```
yurishkuro, line 62: what was the purpose of InnerProps before? Can we have everything merged into one shape (since the distinction between Outer and Common seems no longer relevant, after removing Inner)

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBar.tsx
yurishkuro: were they defined to be used from tests?

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ScatterPlot.jsx
zhuje: I updated `ScatterPlot.test.js` to simulate this event in the unit tests.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBar.tsx
zhuje: The `SpanBar` component is referenced in `SpanBarRow.tsx` and `SpanBar.test.js` but it doesn't look like the props `setLongLabel` or `setShortLabel` were used. 


image

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBar.tsx
zhuje: Updated in latest push.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBar.tsx
zhuje: Updated in latest push.

yurishkuro (approved):
Thanks!

PR #1586, opened 2023-07-16
react-icons upgrade from 2.2.7 to 4.10.1

kirtanchandak:
In this PR I have attempted to upgrade the react-icons version from 2.2.7 to 4.10.1

### DIFF
--- a/packages/jaeger-ui/package.json
+++ b/packages/jaeger-ui/package.json
@@ -92,7 +92,7 @@
     "react-dom": "^18.2.0",
     "react-ga": "^3.3.1",
     "react-helmet": "^6.1.0",
-    "react-icons": "2.2.7",
+    "react-icons": "^4.10.1",
     "react-metrics": "^2.3.2",
     "react-redux": "^5.0.6",
     "react-router-dom": "4.3.1",
--- a/packages/jaeger-ui/src/components/common/NewWindowIcon.tsx
+++ b/packages/jaeger-ui/src/components/common/NewWindowIcon.tsx
@@ -14,7 +14,7 @@
 
 import React from 'react';
 import cx from 'classnames';
-import IoAndroidOpen from 'react-icons/lib/io/android-open';
+import { IoLogoAndroid } from 'react-icons/io5';
 
 import './NewWindowIcon.css';
 
@@ -25,7 +25,7 @@ type Props = {
 export default function NewWindowIcon(props: Props) {
   const { isLarge, ...rest } = props;
   const cls = cx('NewWindowIcon', { 'is-large': isLarge });
-  return <IoAndroidOpen className={cls} {...rest} />;
+  return <IoLogoAndroid className={cls} {...rest} />;
 }
 
 NewWindowIcon.defaultProps = {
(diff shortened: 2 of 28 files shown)


kirtanchandak:
Hey @yurishkuro, I've not touched above logic in the unit tests and I think those are warnings, can we merge this request or do I need to do some more changes?

packages/jaeger-ui/package.json
```diff
@@ -92,7 +92,8 @@
     "react-dom": "^18.2.0",
     "react-ga": "^3.3.1",
     "react-helmet": "^6.1.0",
-    "react-icons": "2.2.7",
+    "react-icons": "^4.10.1",
+    "react-metrics": "^2.3.2",
```
yurishkuro, line 96: not needed?

packages/jaeger-ui/src/components/common/NewWindowIcon.tsx
```diff
@@ -14,7 +14,7 @@
 
 import React from 'react';
 import cx from 'classnames';
-import IoAndroidOpen from 'react-icons/lib/io/android-open';
+import { IoLogoAndroid } from 'react-icons/io5';
```
yurishkuro, line 17: What is your methodology when changing the icons? I just looked at IoLogoAndroid, and it's an android logo, which I don't recall using in Jaeger, certainly not as New Window Icon.

We're using Ant Design as the primary components library. Can use its icons components directly, instead of react-icons (which also has a section for Ant Design, but I am not sure what that means)?

kirtanchandak: ### So, should I change all the react-icons with ant-icons ??
Screenshot 2023-07-17 093141

yurishkuro: It would be good to keep a visual record (e.g. google doc with screenshots) of which icons are being changed for which.

kirtanchandak: So finally should I use ant-icons or react-icons?
And yes I will maintain a google docs file with ss to record the changes!

yurishkuro: so let me ask you, since I am not an expert in this area - `react-icons` seems like an umbrella project that includes icons from many design systems. You can use `ant-icons` directly or via `react-icons/ant` - what would be the difference if you do? If we use `react-icons/ant`, does it make it easier to swap to `react-icons/smth-else`?

kirtanchandak: using ```react-icons/ant``` would be a better option because we can use icons from both react-icons and ant design system and nad yes it would be easier to swap - ```react-icons/smth-else```

kirtanchandak: I think as antd is the primary design system, we should use ```@ant-design/icons``` instead of all the ```react-icons```

yurishkuro: ok, I', fine with that.

yurishkuro:
done in #1721

PR #1632, opened 2023-08-01
Fix flaky test causing spurious drops in code coverage

### LINKED ISSUE #1489: [bug] Flaky test causing spurious drops in code coverage
The test coverage keeps changing from run to run, indicating that there is some non-determinism in the test that's causing different code paths to be exercised. Ideally we should have tests that exercise both code paths.

image

GLVSKiriti:
- Resolves #1489 

- To remove the non determinism in the code

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBarRow.test.js
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBarRow.test.js
@@ -44,7 +44,8 @@ describe('<SpanBarRow>', () => {
     showErrorIcon: false,
     getViewedBounds: () => ({ start: 0, end: 1 }),
     span: {
-      duration: 'test-duration',
+      startTime: 10,
+      duration: 50,
       hasChildren: true,
       process: {
         serviceName: 'service-name',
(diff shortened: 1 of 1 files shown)


yurishkuro:
could you please explain why this change works?

GLVSKiriti:
> could you please explain why this change works?
https://github.com/jaegertracing/jaeger-ui/blob/b4a49a70739bf2db9e7654652cb6e4aaf67b6afc/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBarRow.test.js#L45-L48

Here I am trying to execute at least  once if code path is executed

https://github.com/jaegertracing/jaeger-ui/blob/b4a49a70739bf2db9e7654652cb6e4aaf67b6afc/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBarRow.tsx#L117-L122

yurishkuro:
two different mock returns are good if the function is guaranteed to be called multiple times in the test. Is it?

GLVSKiriti:
> two different mock returns are good if the function is guaranteed to be called multiple times in the test. Is it?

Yes we are calling multiple times in test file
https://github.com/jaegertracing/jaeger-ui/blob/b4a49a70739bf2db9e7654652cb6e4aaf67b6afc/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBarRow.test.js#L111-L133

Indirectly for every shallow render getViewBounds function is called

yurishkuro (approved):
thanks!

PR #1637, opened 2023-08-04
[deps] Remove @types/moment, @types/history and @types/classnames

anshgoyalevil:
- This PR remove the @types/moment, @types/history and @types/classnames packages since these are stub types definitions. Their respective packages provide their own type definitions, so we do not need this installed.

## Description of the changes
- This PR remove the @types/moment, @types/history and @types/classnames packages since these are stub types definitions. Their respective packages provide their own type definitions, so we do not need this installed.

- using yarn start and yarn lint

## Checklist


anshgoyalevil:
//cc @yurishkuro 🚀

yurishkuro (approved):
thanks!

PR #1643, opened 2023-08-08
Init Migration from Enzyme to React Testing Library

anshgoyalevil:
- This PR migrates a test `ScoreCard.test.js` from Enzyme to React Testing Library
- Enzyme is no longer supported in React 18

## Description of the changes
- It uses the custom shallow function to render the components

- It was tested by running the command `yarn test -- -u` to re-generate snapshots of the changed tests
- After that, `yarn test-ci ScoreCard.test.js --coverage` was used to check if all tests pass and coverage is 100%

## Checklist


packages/jaeger-ui/src/components/QualityMetrics/ScoreCard.test.js
```diff
@@ -24,47 +23,44 @@ describe('ScoreCard', () => {
...
-          link={link}
-          score={{
-            label,
-            max,
-            value,
-          }}
-        />
-      )
-    ).toMatchSnapshot();
+    const { container } = render(
```
yurishkuro, line 26: couldn't you just change shallow->render and keep the rest the same, to minimize the changes?

packages/jaeger-ui/package.json
```diff
@@ -47,6 +47,8 @@
   "dependencies": {
     "@jaegertracing/plexus": "0.2.0",
     "@pyroscope/flamegraph": "0.21.4",
+    "@testing-library/jest-dom": "^5.17.0",
+    "@testing-library/react": "^14.0.0",
```
yurishkuro, line 51: can these not be devDeps?

packages/jaeger-ui/src/components/QualityMetrics/__snapshots__/ScoreCard.test.js.snap
```diff
@@ -1,98 +1,235 @@
...
-    />
+    <span
+      class="ScoreCard--TitleHeader"
+    >
+      Test Score
+    </span>
+    <div
+      class="ScoreCard--CircularProgressbarWrapper"
+    >
+      <svg
```
yurishkuro, line 16: the previous snapshot was a lot more compact, this one contains a lot of junk which is difficult to grok for a human, making it much less valuable as a tool to detect regressions. Perhaps this was the difference in using the `shallow` function - is there an equivalent in the new libs?

packages/jaeger-ui/package.json
anshgoyalevil: Sorry for that. Yes, it can be devDeps

packages/jaeger-ui/src/components/QualityMetrics/__snapshots__/ScoreCard.test.js.snap
anshgoyalevil: Actually, the reason for that is, previously, the snapshots were containing the react-dom, but in the case of RTL, they contain HTML dom

According to official migration guide https://testing-library.com/docs/react-testing-library/migrate-from-enzyme/
The shallow function needs to be replaced with render, which seems like the only solution.

packages/jaeger-ui/src/components/QualityMetrics/ScoreCard.test.js
anshgoyalevil: As I have discovered till now, the render function is not a direct replacement for shallow. As per docs, 
> Enzyme's shallow renderer doesn't render sub-components, so React Testing Library's render method is more similar to Enzyme's mount method.

This is also the reason why snapshot file is modified with a lot of code additions.

anshgoyalevil: It can be inlined, but we need to use container out of render function to match the snapshot, as it doesn't provide that utility itself.

packages/jaeger-ui/src/components/QualityMetrics/__snapshots__/ScoreCard.test.js.snap
anshgoyalevil: There is a [solution on the internet](https://thesametech.com/snapshot-testing-in-rtl/#:~:text=Snapshot%20testing%20is%20often%20used,operate%20with%20the%20real%20DOM.) that suggests changing the jest snapshot serializer to output smaller snapshots.

That could break all of the tests, and thus tests would need a huge code change.

anshgoyalevil:
@yurishkuro Which path should we go for in migration?
- Create a helper function `shallowRender` using the library `react-shallow-renderer` and modify each test to use that instead of mount from enzyme. It would not affect snapshots much.
- Use `render` from RTL, which would then make the snapshots bulky, but tests won't have to be modified much.

yurishkuro:
Using render from RTL is unacceptabl. I see that it creates huge snapshots which are very difficult to use for detecting regressions, and based on other information I read are not even stable.

Ideal solution would be to avoid using snapshot testing altogether, given many reasonable arguments against snapshots. But that means we have to migrate 175 call sites, a pretty large lift. So using `react-shallow-renderer` seems like a reasonable compromise, and we can try to enforce the policy of no new snapshot testing going forward (and perhaps eventually phase out the existing ones in favor of proper user-facing behavior testing).

yurishkuro:
NB: please do not create a single PR that updates all 175 call sites. We should do it module by module, so that the changes to the snapshots can be actually reviewed.

anshgoyalevil:
Thanks for the inputs 🚀 
I would try to keep the scope of this PR to a single test file only.

anshgoyalevil:
@yurishkuro I have made the changes, and as I am seeing, the new helper function just reverted the snapshot changes. That's amazing 🚀 
That means we can safely migrate from enzyme, without much input in code reviewing.

packages/jaeger-ui/src/components/QualityMetrics/ScoreCard.test.js
```diff
@@ -13,9 +13,8 @@
 // limitations under the License.
 
 import React from 'react';
-import { shallow } from 'enzyme';
-
 import ScoreCard from './ScoreCard';
+import ReactShallowRenderer from '../../utils/helpers/ReactShallowRenderer';
```
yurishkuro, line 17: why not define a replacement `shallow` function there?

anshgoyalevil: Done. Boiled the code to even simpler usage.
Previously, I was using the code as described in this [guide](https://thesametech.com/snapshot-testing-in-rtl/#:~:text=Snapshot%20testing%20is%20often%20used,operate%20with%20the%20real%20DOM.).

Now, only the shallow function's import statement needs to be changed.

anshgoyalevil:
@yurishkuro I think, this PR can be merged safely. 

The codecov tests are failing because the new helper function doesn't have tests written for it. Should I add it to codecov ignore list?

packages/jaeger-ui/package.json
```diff
@@ -17,6 +17,8 @@
     "@babel/preset-typescript": "^7.21.0",
     "@svgr/babel-plugin-transform-svg-component": "^8.0.0",
     "@svgr/babel-preset": "^8.0.0",
+    "@testing-library/jest-dom": "^5.17.0",
```
yurishkuro, line 20: why is this explicitly added to deps? It's not used in the code, looks like it's a transitive dep that will be included in the lock file

anshgoyalevil:
Closing due to merge conflicts

packages/jaeger-ui/package.json
anshgoyalevil: This would need to be installed when migrating from `mount` function (in Enzyme) to RTL, according to the[ official docs.](https://testing-library.com/docs/react-testing-library/migrate-from-enzyme/)

packages/jaeger-ui/package.json
```diff
@@ -36,6 +38,8 @@
     "jest-environment-jsdom": "^29.5.0",
     "jest-junit": "^16.0.0",
     "less": "3.13.1",
+    "react-is": "^18.2.0",
+    "react-shallow-renderer": "^16.15.0",
```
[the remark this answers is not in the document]
anshgoyalevil: This are the dependencies for React Test Renderer, which is already installed inside jaeger-ui

PR #1648, opened 2023-08-08
Fix json logs escape sequence

### LINKED ISSUE #1570: [Bug]: UI not showing escaped JSON correctly
### What happened?

An event with escaped JSON gets unescaped when displaying in the UI.

I have a log value such as:

`
{
    "name" : "Marcus",
    "value": "{ \"height\" : 193 }"
}`

The value property above, as you can see, is a JSON that got serialized and escaped.

When the trace is built, it then serializes the value property again:

This becomes the following string on my trace:

`
"{\"name\":\"Marcus\",\"value\":\"{\\\"height\\\":139}\"}"
`

Then, when Jager UI displays this JSON, it deserializes it multiple times creating a non-valid JSON value on the ui

### Steps to reproduce

Use the following file on JSON import of Jaeger UI

[json test.txt](https://github.com/jaegertracing/jaeger-ui/files/12017156/json.test.txt)

Please rename it to .json

### Expected behavior

I expect the interface to properly deserialize the JSON and producing a valid output

### Relevant log output

_No response_

### Screenshot

Looks correct when minimized:

image

But when I expand:

image

### Additional context

_No response_

### Jaeger backend version

jaegertracing/jaeger-collector:1.36

### SDK

_No response_

### Pipeline

_No response_

### Stogage backend

_No response_

### Operating system

Mac OS

### Deployment model

Kybernetes / Terraform

### Deployment configs

_No response_

Xaid-vfx:
- Resolves #1570 

## Description of the changes
- This PR resolves the invalid JSON value that was appearing in UI. An extra replace function is added to handle the escape sequences if there is an object inside string value.

- yarn start
- json.test.txt file that was provided with the issue
Screenshot 2023-08-08 at 8 16 42 PM

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/jsonMarkup.js
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/jsonMarkup.js
@@ -42,7 +42,7 @@ function type(doc) {
 }
 
 function escape(str) {
-  return str.replace(/</g, '&lt;').replace(/>/g, '&gt;');
+  return str.replace(/</g, '&lt;').replace(/>/g, '&gt;').replace('{', '{\\').replace('":', '\\":');
 }
 
 export function jsonMarkup(doc, styleFile) {


Xaid-vfx:
@yurishkuro although, I have fixed the issue as mentioned, I'm not sure if this is the right way to fix this issue. 
Created a draft PR as of now, will make necessary changes after your input.

yurishkuro:
jsonMarkup has been replaced with a different dependency in #1724

Issue #1608, opened 2023-07-27
[chore]: Fix eslint warnings


punithnayak:
### What happened?

There some lint warnings shown

### Steps to reproduce

1. Go to the root dir of jaeger-UI
2. run `yarn lint`

### Expected behavior

There should be no eslint warnings

### Relevant log output

```shell
`[eslint       ] 
[eslint       ] /jaeger-ui/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.test.js
[eslint       ]   178:23  warning  Unexpected unnamed function  func-names
[eslint       ] 
[eslint       ] /jaeger-ui/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.tsx
[eslint       ]   117:18  warning  'createBlob' was used before it was defined  @typescript-eslint/no-use-before-define
[eslint       ] 
[eslint       ] /jaeger-ui/packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
[eslint       ]   284:13  warning  Avoid non-native interactive elements. If using native HTML is not possible, add an appropriate role and support for tabbing, mouse, keyboard, and touch inputs to an interactive content element  jsx-a11y/no-static-element-interactions
[eslint       ] 
[eslint       ] /jaeger-ui/packages/jaeger-ui/src/selectors/span.js
[eslint       ]   15:1  warning  Prefer default export on a file with single export  import/prefer-default-export
[eslint       ] 
[eslint       ] ✖ 4 problems (0 errors, 4 warnings)
[eslint       ]
```

### Screenshot

![image](https://github.com/jaegertracing/jaeger-ui/assets/110713318/a7b1eac8-4d7a-4dc8-8e6c-b9872fbf9406)

punithnayak:
I am also planning to add lint check during ci/cd to prevent further lint warnings/errors in the code

yurishkuro:
Some of the warnings are due to outdated APIs.

shuence:
@yurishkuro Anyone Working on this? if not can you assign this to me?

yurishkuro:
you can see linked issues/PRs for parts that are being work on. We do not assign issues.

DerenToy:
Dear @yurishkuro ,

I'm new here and have tried to make the relevant changes by looking at previous PRs as best as I could. If there's a mistake or something I've overlooked, I'd greatly appreciate any suggestions for corrections.

Best regards

PR #1652, opened 2023-08-09
Fix lint issues when bumping eslint-plugin-import

### LINKED ISSUE #1613: [deps] Help with upgrading eslint-plugin-import
@dependabot tried to upgrade https://github.com/jaegertracing/jaeger-ui/pull/1609 but it broke the CI, probably requires some minor code changes to satisfy the new linter version

Kavinjsir:
- Resolves #1613
- Part of #1608

## Description of the changes
- Update code lines to resolve lint check warnings

- `yarn run esilnt`

## Checklist


packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.test.js
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2017 Uber Technologies, Inc.
+// Copyright (c) 2023 Uber Technologies, Inc.
```
yurishkuro, line 1: why is this changed?

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.tsx
```diff
@@ -93,6 +93,11 @@ const SelectSort = reduxForm({
 
 export const sortFormSelector = formValueSelector('traceResultsSort');
 
+// export for tests
+export function createBlob(rawTraces: TraceData[]) {
+  return new Blob([`{"data":${JSON.stringify(rawTraces)}}`], { type: 'application/json' });
+}
```
yurishkuro, line 99: this looks just like a move, what is the purpose?

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.test.js
Kavinjsir: I am wondering if we need to keep the year up to date if we have the code lines in the file get updated?
Or do we keep them as origin?

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.tsx
Kavinjsir: So the update of the code lines is an attempt to resolve the lint warnings.
For this one, before the update, running `yarn run eslint` will result in:
```
/Users/jintony/pro/tmp/jaeger-ui/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.tsx
  117:18  warning  'createBlob' was used before it was defined  @typescript-eslint/no-use-before-define
```

That's why I lift the definition up.

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.test.js
yurishkuro: Uber's was the original copyright, new contributions should use "The Jaeger Authors", which you can add as a line above with the current year if you want.

We tend to keep the original year a specific (c) by-line was introduced. I checked with legal once, there wasn't a lot of value in updating the year.

Kavinjsir:
I think I might miss the point of ci error before. Just realized that `unit-test` in ci failed by running `yarn lint`:
https://github.com/jaegertracing/jaeger-ui/actions/runs/5815079038/job/15766674298

It seems we can reproduce the issue locally by removing the *--cache* flag when running *eslint*. Let me try resolving them.

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.test.js
Kavinjsir: Sounds good! I'm going to roll back the corresponding change as it is not necessary. Thx for letting me know:)

packages/jaeger-ui/src/demo/trace-generators.js
```diff
@@ -14,7 +14,7 @@
 
 import Chance from 'chance';
 
-import { getSpanId } from '../selectors/span';
+import getSpanId from '../selectors/span';
```
yurishkuro, line 17: this is an exceedingly silly selector, is it actually used, or can we delete it?

packages/jaeger-ui/src/selectors/trace.test.js
```diff
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 import { followsFromRef } from './trace.fixture';
-import { getSpanId } from './span';
+import getSpanId from './span';
```
yurishkuro, line 16: if this is the only place, let's just replace it with span.ID and delete the selector

yurishkuro:
your description links to issues, but does not put the word Fixes or Resolves in front of them. Please update, this way when the PR is merged the issues will be closed (unless it does not resolve them completely, in which case we typically use the format `Part of #xxxx`

package.json
```diff
@@ -14,7 +14,7 @@
     "eslint": "8.46.0",
     "eslint-config-airbnb": "19.0.4",
     "eslint-config-prettier": "9.0.0",
-    "eslint-plugin-import": "2.27.5",
+    "eslint-plugin-import": "2.28.0",
```
yurishkuro, line 17: once the CI is green, I would prefer to remove the version upgrade and only leave the code changes, then let the dependabot to redo its PR for the actual upgrade.

yurishkuro (dismissed):

yurishkuro:
@Kavinjsir I removed the upgrade to keep the scope to code changes only, and it fails the CI, can you please check why?

packages/jaeger-ui/src/selectors/trace.test.js
Kavinjsir: Sounds good!

yurishkuro (approved):

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.test.js
```diff
@@ -175,8 +175,11 @@ describe('<SearchResults>', () => {
 
       it('when click on DownloadResults then call download function', () => {
         const originalBlob = global.Blob;
-        global.Blob = function (text, options) {
-          return { text, options };
+        global.Blob = class {
+          constructor(text, options) {
+            this.text = text;
+            this.options = options;
+          }
```
Kavinjsir, line 182: @yurishkuro From my local, it seems like what really brought the `unit-test` failure is:
Screenshot 2023-08-10 at 12 47 34 PM

(Not sure why it was missed in the CI logs)

For this error above, I think I made a mistake to convert `Blob` to an arrow function that is not able to call by `new`. 
 
So I change such `global.Blob` to an "explicit" class. Assuming this can:
1. Still resolve the origin lint warning
2. Attempt to make it clear so satisfy calling by keyword `new`.

How do you think of this?

yurishkuro:
Thanks, @Kavinjsir, the dependabot PR #1609 was successfully merged after these changes.

PR #1671, opened 2023-08-13
Remove unused source-map-explorer package

anshgoyalevil:
- This PR removes the `source-map-explorer` package because: 
   - It is not being used anywhere inside the project
   - According to its npm desc., it needs to be installed globally, so anyways, not needed inside the project directly. 

## Checklist

### DIFF
--- a/packages/jaeger-ui/package.json
+++ b/packages/jaeger-ui/package.json
@@ -41,7 +41,6 @@
     "react-test-renderer": "^18.2.0",
     "rollup-plugin-visualizer": "^5.9.0",
     "sinon": "7.3.2",
-    "source-map-explorer": "^2.5.3",
     "terser": "^5.16.5",
     "vite": "^4.1.4",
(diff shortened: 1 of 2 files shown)


yurishkuro:
what are we doing for source maps if we're not using this module?

anshgoyalevil:
I don't think this library is being used anywhere. Where do we have source-maps inside the repository?

yurishkuro:
I don't know, but I am able to see __some__ source code in the Inspector:
image

yurishkuro:
but I can also see the same after applying your PR, so probably fine to merge. It's possible source maps are provided by babel or vite directly.

yurishkuro (approved):

anshgoyalevil:
Yes. I can see a webpack config for that which uses `source-map`. (It's different from `source-map-explorer`)
```js
output: {
      path: join(__dirname, 'build'),
      publicPath: '/',
      filename: 'assets/[name].js',
    },
```

PR #1673, opened 2023-08-13
Remove unused src parameter in process method

wthrajat:
- The PR removes unused `src` parameter used in the `process` method. The `src` parameter is declared but remains unused within the transformation logic.
- Purpose: just a quick fix

## Checklist

### DIFF
--- a/packages/jaeger-ui/test/generic-file-transform.js
+++ b/packages/jaeger-ui/test/generic-file-transform.js
@@ -16,7 +16,7 @@ const path = require('path');
 
 // Simple Jest transform that stubs out CSS files and returns the file name as the mock content for other file types.
 module.exports = {
-  process(src, filename) {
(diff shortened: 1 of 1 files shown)


yurishkuro:
I ran the CI just to make sure, it failed as I expected. Just because a param is not used inside a function doesn't mean you can just remove it and change the function signature.

wthrajat:
@yurishkuro I see.
Strange that both `yarn lint` and `yarn test` passed while making the changes and if I now run `yarn test`, my memory usage increases to the point it crashed the system (before even failing or passing the tests)

```bash
$ journalctl --since=today
...
Aug 14 03:21:21 arch kernel: extended physical RAM map:
Aug 14 03:21:21 arch kernel: reserve setup_data: [mem 0x0000000000>
```
I'll just close this PR

PR #1675, opened 2023-08-14
Remove unused packages

anshgoyalevil:
Fixes a part of #1672

## What changes is this PR making?
- This PR removes the unused packages from packages/jaeger-ui
- The removed packages are:
   - `@svgr/babel-plugin-transform-svg-component`
   - `@svgr/babel-preset`
   - `@types/testing-library__jest-dom`
   - `@types/react-copy-to-clipboard`
   - `@types/recompose`
   - `tar`

## Checklist
- [x] I have read https://github.com/jaegertracing/jaeger/blob/master/CONTRIBUTING_GUIDELINES.md
- [x] I have signed all commits
- [x] I have added unit tests for the new functionality
- [x] I have run lint and test steps successfully
  - for `jaeger`: `make lint test`
  - for `jaeger-ui`: `yarn lint` and `yarn test`

PR #1677, opened 2023-08-14
Add depcheck for CI step to check for unused dependencies

### LINKED ISSUE #1672: [Feature]: Add CI step to check dependencies
Recently we discovered several dependencies that were listed in the package.json but were not actually used (e.g. #1669, #1616, #1637). It would be good to have a GH Action that runs this checks and fails the CI when unused or missing dependencies are discovered.

I don't know if `depcheck` module is the best solution for this, but when I ran it in packages/jaeger-ui, it printed:
```
# ...jaeger-ui/packages/jaeger-ui 
$ depcheck
Unused dependencies
* @types/react-copy-to-clipboard
* @types/recompose
* d3-scale
* redux-async-middleware
* tar
Unused devDependencies
* @babel/preset-env
* @babel/preset-react
* @babel/preset-typescript
* @svgr/babel-plugin-transform-svg-component
* @svgr/babel-preset
* @types/react-dom
* @types/react-router-redux
* babel-plugin-inline-react-svg
* jest-environment-jsdom
* react-test-renderer
* source-map-explorer
Missing dependencies
* react-select: ./src/components/common/VirtSelect.tsx
* react-virtualized: ./src/components/common/VirtSelect.tsx
```

I don't think it's completely accurate, especially some of devDependencies may be used in the build configurations and are a false positive. Maybe it can be additionally tuned/configured, or perhaps we can add a dummy "tools.js" file that will import these modules so that they will not be flagged.

The GH Action should:
  * run the check in both jaeger-ui and plexus workspaces
  * fail if missing or unused dependencies are found
  * before failing, print a recommendation on how to add or remove them, e.g. `cd packages/jaeger-ui; yarn remove $pkg`
  * maybe it doesn't need to be a separate GH Action, but run as part of `yarn lint`

anshgoyalevil:
- Fixes #1672 

## Description of the changes
- This PR adds a step in the unit-test workflow to check for unused dependencies using the depcheck package.
- Certain dependencies which are installed but still being given in the output are put into `.depcheckrc.json` file to be ignored.
- The `--skip-missing` flag is also added to ignore the missing deps warning

- The `ignores` array without any dep in `.depcheckrc.json` file was pushed to the branch for testing, and it was observed that the CI failed, which is the required behaviour: Workflow Run: https://github.com/anshgoyalevil/jaeger-ui/actions/runs/5854765474/job/15871258991
- The `ignores` array with all unused deps was pushed for CI testing, and the CI passed the Depcheck flow. Workflow Run: https://github.com/anshgoyalevil/jaeger-ui/actions/runs/5854816286/job/15871414318#step:6:10 (Ignore the Lint fail error, as it is resolved in this PR, but workflow run is not available because target push branch was changed to `main` in that commit, and thus workflow run is not available to be shared)

## Notes:
Before merging this PR, all the unused deps which are neither added to the `ignores` array, not have been removed need to be removed. Thus PR #1675 needs to be merged before this one.
After that, this PR's branch needs to be updated with that code so that the Depcheck doesn't fail in CI for this or for upcoming PRs.

## Checklist

### DIFF
--- a/scripts/generateDepcheckrcJaegerUI.js
+++ b/scripts/generateDepcheckrcJaegerUI.js
@@ -39,7 +39,6 @@ const depcheckrcContent = {
 const outputFile = process.argv[2];
 
 if (!outputFile) {
-  console.error('Usage: node generateDepcheckrcJaegerUI.js <outputFile>');
   process.exit(1);
 }
 
--- a/scripts/generateDepcheckrcPlexus.js
+++ b/scripts/generateDepcheckrcPlexus.js
@@ -49,7 +49,6 @@ const depcheckrcContent = {
 const outputFile = process.argv[2];
 
 if (!outputFile) {
-  console.error('Usage: node generateDepcheckrcJaegerUI.js <outputFile>');
   process.exit(1);
 }
 


packages/jaeger-ui/.depcheckrc.json
```diff
@@ -0,0 +1,12 @@
+{
+  "ignores": [
+    "@babel/preset-env",
+    "@babel/preset-react",
+    "@babel/preset-typescript",
+    "@types/react-dom",
+    "@types/react-router-redux",
```
yurishkuro, line 7: these are just type declarations, why are they needed if no code imports them?

packages/jaeger-ui/.depcheckrc.json
```diff
@@ -0,0 +1,12 @@
...
+    "babel-plugin-inline-react-svg",
+    "jest-environment-jsdom"
+  ],
+  "skip-missing": true
```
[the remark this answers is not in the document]
anshgoyalevil: Missing deps would anyway be caught while linting because of no imports ts error.

Currently, we have the following as missing deps
```js
Missing dependencies
* react-is: ./src/utils/ReactShallowRenderer.test.js // Already installed as it is a dep for react-test-renderer (The warning suggests adding it as an explicit package inside package.json
* react-select: ./src/components/common/VirtSelect.tsx // same reason
* react-virtualized: ./src/components/common/VirtSelect.tsx // same reason
* _: ./build/static/index-21cea8b1.js // it is a build file
```

packages/jaeger-ui/.depcheckrc.json
anshgoyalevil: Type declarations are never imported inside the codebase. They are supplemented so that typescript doesn't throw type errors for the custom functions or components that are used inside the project.

Other type declarations are also not imported anywhere but are essential because of typescript compatibility.

`@types/react-dom` is for `react` and `@types/react-router-redux` is for `react-router-redux` and both of these packages are being used inside the project, so having their type declarations are necessary IMO.

packages/jaeger-ui/.depcheckrc.json
```diff
@@ -0,0 +1,12 @@
+{
+  "ignores": [
+    "@babel/preset-env",
```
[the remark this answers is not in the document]
anshgoyalevil: In that case, we don't need this package altogether.

But, creating a custom script to analyze the whole codebase, and checking for deps in the codebase against the deps inside package.json is what I wouldn't advocate for, though it can be done, but would be more like reinventing the wheel.

In case we aren't sure of the list (given to us by depcheck), we can go ahead with another dep checker, i.e., [npm-check](https://www.npmjs.com/package/npm-check).
I manually checked with it also, and the list is same as below:
<details>
<summary>Output</summary>

```
tar                                          😕  NOTUSED?  Still using tar?
                                                          Depcheck did not find code similar to require('tar') or import from 'tar'.
                                                          Check your code before removing as depcheck isn't able to foresee all ways dependencies can be used.
                                                          Use rc file options to remove unused check, but still monitor for outdated version:
                                                              .npmcheckrc {"depcheck": {"ignoreMatches": ["tar"]}}
                                                          Use --skip-unused to skip this check.
                                                          To remove this package: yarn remove tar 

@babel/preset-env                            😕  NOTUSED?  Still using @babel/preset-env?
                                                          Depcheck did not find code similar to require('@babel/preset-env') or import from '@babel/preset-env'.
                                                          Check your code before removing as depcheck isn't able to foresee all ways dependencies can be used.
                                                          Use rc file options to remove unused check, but still monitor for outdated version:
                                                              .npmcheckrc {"depcheck": {"ignoreMatches": ["@babel/preset-env"]}}
                                                          Use --skip-unused to skip this check.
                                                          To remove this package: yarn remove @babel/preset-env --dev

@babel/preset-react                          😕  NOTUSED?  Still using @babel/preset-react?
                                                          Depcheck did not find code similar to require('@babel/preset-react') or import from '@babel/preset-react'.
                                                          Check your code before removing as depcheck isn't able to foresee all ways dependencies can be used.
                                                          Use rc file options to remove unused check, but still monitor for outdated version:
                                                              .npmcheckrc {"depcheck": {"ignoreMatches": ["@babel/preset-react"]}}
                                                          Use --skip-unused to skip this check.
                                                          To remove this package: yarn remove @babel/preset-react --dev

@babel/preset-typescript                     😕  NOTUSED?  Still using @babel/preset-typescript?
                                                          Depcheck did not find code similar to require('@babel/preset-typescript') or import from '@babel/preset-typescript'.
                                                          Check your code before removing as depcheck isn't able to foresee all ways dependencies can be used.
                                                          Use rc file options to remove unused check, but still monitor for outdated version:
                                                              .npmcheckrc {"depcheck": {"ignoreMatches": ["@babel/preset-typescript"]}}
                                                          Use --skip-unused to skip this check.
                                                          To remove this package: yarn remove @babel/preset-typescript --dev

@svgr/babel-plugin-transform-svg-component   😟  MISSING!  Not installed.
                                             😕  NOTUSED?  Still using @svgr/babel-plugin-transform-svg-component?
                                                          Depcheck did not find code similar to require('@svgr/babel-plugin-transform-svg-component') or import from '@svgr/babel-plugin-transform-svg-component'.
                                                          Check your code before removing as depcheck isn't able to foresee all ways dependencies can be used.
                                                          Use rc file options to remove unused check, but still monitor for outdated version:
                                                              .npmcheckrc {"depcheck": {"ignoreMatches": ["@svgr/babel-plugin-transform-svg-component"]}}
                                                          Use --skip-unused to skip this check.
                                                          To remove this package: yarn remove @svgr/babel-plugin-transform-svg-component --dev

@svgr/babel-preset                           😟  MISSING!  Not installed.
                                             😕  NOTUSED?  Still using @svgr/babel-preset?
                                                          Depcheck did not find code similar to require('@svgr/babel-preset') or import from '@svgr/babel-preset'.
                                                          Check your code before removing as depcheck isn't able to foresee all ways dependencies can be used.
                                                          Use rc file options to remove unused check, but still monitor for outdated version:
                                                              .npmcheckrc {"depcheck": {"ignoreMatches": ["@svgr/babel-preset"]}}
                                                          Use --skip-unused to skip this check.
                                                          To remove this package: yarn remove @svgr/babel-preset --dev

@testing-library/jest-dom                    😎  MAJOR UP  Major update available. https://github.com/testing-library/jest-dom#readme
                                                          yarn add @testing-library/jest-dom@6.0.0 --dev to go from 5.17.0 to 6.0.0

@types/match-sorter                          😎  MAJOR UP  Major update available.
                                                          yarn add @types/match-sorter@6.0.0 --dev to go from 2.3.0 to 6.0.0

@types/react                                 😎  MAJOR UP  Major update available. https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/react
                                                          yarn add @types/react@18.2.20 --dev to go from 16.8.7 to 18.2.20

@types/react-dom                             😎  PATCH UP  Patch update available. https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/react-dom
                                                          yarn add @types/react-dom@18.2.7 --dev to go from 18.2.5 to 18.2.7

@types/react-router-redux                    😍  UPDATE!   Your local install is out of date. https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/react-router-redux
                                                          yarn add @types/react-router-redux@5.0.22 --dev to go from 5.0.21 to 5.0.22

jest                                         😎  MAJOR UP  Major update available. https://jestjs.io/
                                                          yarn add jest@29.6.2 --dev to go from 28.1.3 to 29.6.2

jest-environment-jsdom                       😍  UPDATE!   Your local install is out of date. https://github.com/facebook/jest#readme
                                                          yarn add jest-environment-jsdom@29.6.2 --dev to go from 29.6.1 to 29.6.2
                                             😕  NOTUSED?  Still using jest-environment-jsdom?
                                                          Depcheck did not find code similar to require('jest-environment-jsdom') or import from 'jest-environment-jsdom'.
                                                          Check your code before removing as depcheck isn't able to foresee all ways dependencies can be used.
                                                          Use rc file options to remove unused check, but still monitor for outdated version:
                                                              .npmcheckrc {"depcheck": {"ignoreMatches": ["jest-environment-jsdom"]}}
                                                          Use --skip-unused to skip this check.
                                                          To remove this package: yarn remove jest-environment-jsdom --dev

less                                         😎  MAJOR UP  Major update available. http://lesscss.org
                                                          yarn add less@4.2.0 --dev to go from 3.13.1 to 4.2.0

sinon                                        😎  MAJOR UP  Major update available. https://sinonjs.org/
                                                          yarn add sinon@15.2.0 --dev to go from 7.3.2 to 15.2.0

terser                                       😍  UPDATE!   Your local install is out of date. https://terser.org
                                                          yarn add terser@5.19.2 --dev to go from 5.19.0 to 5.19.2

react-is                                     😟  PKG ERR!  Not in the package.json. Found in: /src/utils/ReactShallowRenderer.test.js
                                             😎  MAJOR UP  Major update available. https://reactjs.org/
                                                          yarn add react-is@18.2.0  to go from 16.13.1 to 18.2.0

react-select                                 😟  PKG ERR!  Not in the package.json. Found in: /src/components/common/VirtSelect.tsx
                                             😎  MAJOR UP  Major update available. https://github.com/JedWatson/react-select/tree/master#readme
                                                          yarn add react-select@5.7.4  to go from 1.3.0 to 5.7.4

react-virtualized                            😟  PKG ERR!  Not in the package.json. Found in: /src/components/common/VirtSelect.tsx
                                             😎  MINOR UP  Minor update available. https://github.com/bvaughn/react-virtualized
                                                          yarn add react-virtualized@9.22.5  to go from 9.21.0 to 9.22.5

Use npm-check -u for interactive update.
```

</details>

The good thing about this tool is, it didn't gave a false +ve on @types deps, but downside is, it is very slow.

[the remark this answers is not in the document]
yurishkuro: > But, creating a custom script to analyze the whole codebase

This is now that I was suggesting. Let's say there is a list of plugins mentioned in `.babelrc` or something similar. If depcheck does not verify that file (a reasonable behavior, it cannot know about all possible libraries out there and their bespoke configuration), we can still write a smalls script that will read `.babelrc` or simiar and output a temporary `.depcheckrc.json`. This way everything stays consistent - if we remove a plugin from babel, the CI will fail if we it's not removed from dependencies as well.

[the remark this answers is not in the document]
anshgoyalevil: I got the scenario here, but the issue is with what would we compare those list of dependencies with, which we parse from the files like `.babelrc`

For example:

`{A, B}` are parsed from `.babelrc`, `{C, D}` is parsed from `vite.config.js`.

Now, we have `package.json` with `{A, B, C, D, E, F, . . . . n}`.
Here, we would compare the parsed result (`{A, B, C, D}`) with `A, B, C, D`, but what about `E, F, . . . n`
We don't know if they also contain any unused dep, since that part is managed by Depcheck.

There are some other dependencies that are not explicitly required but need to be installed.
For example: `jest-environment-jsdom`. 
We cannot parse it from any config file, though it is needed if we use the following configuration in our `package.json`
```
"testEnvironment": "jsdom"
```
which is the case for us.

We can also configure our parser to add the dep `jest-environment-jsdom` to ignore list in case it encounters `"testEnvironment": jsdom`

packages/jaeger-ui/.depcheckrc.json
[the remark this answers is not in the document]
anshgoyalevil: Agreed! Would install them explicitly. But, would still need to skip the `build` directory for checking the deps (As in case of `* _: ./build/static/index-21cea8b1.js`)

packages/jaeger-ui/.depcheckrc.json
anshgoyalevil: I couldn't find an exact tool for that, but there's another tool `npm-check` that uses `depcheck` behind the scenes.

I checked the results of `npm-check` and it didn't show any of `@types/xxx` dep as unused. Though, all other results were same. Maybe its parser is more correct.

packages/jaeger-ui/.depcheckrc.json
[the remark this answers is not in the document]
yurishkuro: > what would we compare those list of dependencies with

Why do we need to compare? We can parse those known config files, extract plugin names, and write them out into the ignore section of .depcheckrc, then just run depcheck

anshgoyalevil:
@yurishkuro 

I have done the following change by now:

- Added a `generateDepcheck.js` script to generate a `.depcheckrc.json` file on the fly
- The file would parse the babel options from `babel-transform.js` file and would use it in ignores array.
- Added the 3 missing deps `react-is`, `react-select`, `react-virtualized`

.github/workflows/unit-tests.yml
```diff
@@ -17,7 +17,21 @@ jobs:
...
+        working-directory: ./packages/jaeger-ui
+        run: |
+          depcheck
+          DEPCHECK_STATUS=$?
+          if [ $DEPCHECK_STATUS -ne 0 ]; then
+            echo "Unused dependencies found by depcheck."
+            exit 1  # Exit with a failure status
+          fi
+      - name: Remove .depcheckrc.js
+        run: rm packages/jaeger-ui/.depcheckrc.json
```
yurishkuro, line 34: This is not a good way to do this, i.e. have the logic only in GH workflow. As this is a linter step we should be able to run it locally via `yarn depcheck`, so can we move this all into package.json/scripts?

scripts/generateDepcheckrc.js
```diff
@@ -0,0 +1,48 @@
...
+  ignores: [...packageNames, ...otherPackages],
+  'ignore-dirs': ['build'],
+};
+
+// Path to depcheckrc.json file
+const depcheckrcFilePath = 'packages/jaeger-ui/.depcheckrc.json';
+
+// Format the content using Prettier
+(async () => {
+  const formattedContent = await prettier.format(JSON.stringify(depcheckrcContent, null, 2), {
```
yurishkuro, line 44: this is a temp file, why do we need to run prettier on it?

yurishkuro:
CI is failing with 
```
Unused devDependencies
* @svgr/babel-plugin-transform-svg-component
* @svgr/babel-preset
* @types/react-dom
* @types/react-router-redux
* ```

File: scripts/generateDepcheckrc.js (hunk 5 above)
Comment (anshgoyalevil) [reply scripts/generateDepcheckrc.js:44]: I tried it without prettier on it, but since we are running the lint check after depcheck, we would have to apply prettier on it so lint check doesn't fail.
But you are right, we are anyway deleting it in CI.

Should I remove prettier output? I think having it would be good though

Comment (anshgoyalevil) [reply scripts/generateDepcheckrc.js:44]: Okay. Got it

File: .github/workflows/unit-tests.yml (hunk 4 above)
Comment (anshgoyalevil) [reply .github/workflows/unit-tests.yml]: moved to package.json to work as a yarn script

**Comment by anshgoyalevil** [comment]
@yurishkuro CI is passing the checks now.

File: package.json [hunk 6]
```diff
@@ -40,6 +40,7 @@
     "build": "lerna run build",
     "check-license": "./scripts/check-license.sh",
     "coverage": "lerna run coverage",
+    "depcheck": "(node scripts/generateDepcheckrc.js && depcheck packages/jaeger-ui && rm -f packages/jaeger-ui/.depcheckrc.json;) || (EXIT_CODE=$?; rm -f packages/jaeger-ui/.depcheckrc.json; exit $EXIT_CODE)",
```
yurishkuro, line 43: `depcheck packages/jaeger-ui` - why only this package and not plexus?

anshgoyalevil, line 43: This would store the exit code (255) in case there would be some unused dependency, then would delete the file, and finally exits the CI, stopping further jobs.

yurishkuro: if depcheck fails, the `rm` command will not be run because of `&&`

yurishkuro: can depcheck be invoked programmatically as a module? Then you could call it from the custom script you added

anshgoyalevil: > if depcheck fails, the rm command will not be run because of &&

I have included an `||` condition in case it fails, the other remove command would remove it

anshgoyalevil: > can depcheck be invoked programmatically as a module

Yes, it can be invoked. In that case, we wouldn't have to handle write or delete ops.

anshgoyalevil:
@yurishkuro I have added CI check for the plexus also. Checked everything manually, like deletion in case of exits, etc.

I also tried the Depcheck as a module, but it is not working because of its distribution problems.

.github/workflows/unit-tests.yml
```diff
@@ -17,7 +17,10 @@ jobs:
         with:
           cache: yarn
           node-version: '16'
+      - run: yarn global add depcheck
       - run: yarn install --frozen-lockfile
+      - name: Run depcheck
+        run: yarn run depcheck
```
yurishkuro, line 23:
```suggestion
      - run: yarn install --frozen-lockfile
      - name: Run depcheck
        run: |
          yarn global add depcheck
          yarn run depcheck
```
no need to separate

package.json
```diff
@@ -40,6 +40,7 @@
     "build": "lerna run build",
     "check-license": "./scripts/check-license.sh",
     "coverage": "lerna run coverage",
+    "depcheck": "((node scripts/generateDepcheckrc.js --package 'jaeger-ui' && depcheck packages/jaeger-ui && rm -f packages/jaeger-ui/.depcheckrc.json;) || (EXIT_CODE=$?; rm -f packages/jaeger-ui/.depcheckrc.json; exit $EXIT_CODE)); ((node scripts/generateDepcheckrc.js --package 'plexus' && depcheck packages/plexus && rm -f packages/plexus/.depcheckrc.json;) || (EXIT_CODE=$?; rm -f packages/plexus/.depcheckrc.json; exit $EXIT_CODE));",
```
yurishkuro, line 43: could we just put this into a script, similar to ./scripts/check-license.sh? I think trying to fit this into a single line makes it harder than it needs to be, and the syntax is not portable anyway (e.g. for Windows) so why not a bash script.

anshgoyalevil: Yes, we can move it to a shell script too. Working on it.

> Syntax is not portable anyway (e.g. for Windows)
This would most probably work on git bash (on Windows, which is a UNIX-based terminal.

scripts/generateDepcheckrc.js
```diff
@@ -0,0 +1,89 @@
...
+};
+
+const pathToJaegerUI = 'packages/jaeger-ui/.depcheckrc.json';
+const pathToPlexus = 'packages/plexus/.depcheckrc.json';
+
+if (targetPackage === 'plexus') {
+  fs.writeFileSync(pathToPlexus, JSON.stringify(depcheckrcContent, null, 2));
+} else {
+  fs.writeFileSync(pathToJaegerUI, JSON.stringify(depcheckrcContent, null, 2));
+}
```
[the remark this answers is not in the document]
anshgoyalevil: Sure. Moving the logics to different files for each of these packages.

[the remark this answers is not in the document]
anshgoyalevil: done 🚀

package.json
anshgoyalevil: done 🚀

package.json
```diff
@@ -38,8 +38,9 @@
   },
   "scripts": {
     "build": "lerna run build",
-    "check-license": "./scripts/check-license.sh",
+    "check-license": "chmod +x scripts/check-license.sh && ./scripts/check-license.sh",
```
yurishkuro, line 41: changing source files in a script is not a good practice. Aren't permissions already stored in Git?

If not, you could do `bash <script>`

anshgoyalevil: Didn't knew that before. Thanks.
Reverting this change.

scripts/run-depcheck.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh
+# This script runs depcheck on two different packages: 'jaeger-ui' and 'plexus'
+
+# Run depcheck for the 'jaeger-ui' package
```
yurishkuro, line 4: I think this is more complicated than it needs to be:
  * depcheck takes a `--config` arg so I assume the rc file does not need to be in the local directory or specifically named
  * if so, you can use `tempfile()` to create two config files and not worry about deleting them (will happen automatically)
  * then you can use `set -ex` so that if depcheck returns a non-zero exit code the script will abort

with these changes I think your whole script can be just a few lines

scripts/generateDepcheckrc.js
```diff
@@ -0,0 +1,89 @@
+// Copyright (c) 2023 The Jaeger Authors.
```
yurishkuro, line 1: so is this file needed or is it replaced by the two subpackage-specific files below?

anshgoyalevil: It's now replaces by two subpackage specific files now.

scripts/run-depcheck.sh
anshgoyalevil: Thanks for this solution. Things are much simpler now 😄

scripts/run-depcheck.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+set -ex
+
+# Create a temporary depcheckrc file for 'jaeger-ui'
+tempfile_jaeger=$(mktemp /tmp/depcheckrc.XXXXXX.json)
+node scripts/generateDepcheckrcJaegerUI.js "$tempfile_jaeger"
```
yurishkuro, line 6: you could've done this more "unixy" by writing to stdout from the Node script and redirecting here to the temp file. This way there would be fewer dependencies between the steps, the Node script is easier to debug, and it would be simpler (no writing to files).

yurishkuro (approved):

scripts/run-depcheck.sh
anshgoyalevil: I was earlier thinking of the same to redirect the output using `>` operator in Unix.
Would create a sister PR to rectify the same.

anshgoyalevil:
Thanks @yurishkuro for guiding throughout and reviewing at each step 🚀

Issue #1668, opened 2023-08-12
Remove dependency on enzyme


yurishkuro:
Enzyme is not compatible with React 18, so we need to phase it out.

Preferred approach is to migrate to `react-testing-library`. In some cases we may use a replacement `shallow` function #1653, but only when absolutely necessary, it's better to move away from snapshot testing.
  * https://testing-library.com/docs/react-testing-library/migrate-from-enzyme
  * https://medium.com/zipyai/migrating-from-enzyme-to-react-testing-library-b1fb607d81ab

For examples of this being partially done, see merged PRs in this ticket below (e.g. #1659 #1653 #1643).

The scope of this is sizeable, so the **recommendation is to limit the changes to specific modules**, a couple test files at a time.

Current enzyme references (July 19 2025):
```
$ grep -rl 'import.*enzyme' packages/jaeger-ui/src
packages/jaeger-ui/src/components/TracePage/index.test.js
packages/jaeger-ui/src/components/SearchTracePage/SearchForm.test.js
```

Sangam-ghimire:
I want to work on this @yurishkuro. Can you assign it to me please? Will be a very good start for me

yurishkuro:
We do not assign issues. There are multiple packages that are affected, don't try to change all in one PR.

Sangam-ghimire:
@yurishkuro Any previous discussion that  I should not miss, execpt the mentioned ones.

AbhishekTiwari23:
I also want to work on this. There are several test files, so I think we can split the task between us, @Sangam-ghimire. Please let me know if you want to handle all the files alone because if we both work on same files, it will waste time for both of us.
Thanks

Sangam-ghimire:
Sure @AbhishekTiwari23 that's a great spirit. Let's split and work together. I might need some help too.

anshgoyalevil:
@Sangam-ghimire @AbhishekTiwari23 Please have a look at the discussion among the PRs: https://github.com/jaegertracing/jaeger-ui/pull/1659 https://github.com/jaegertracing/jaeger-ui/pull/1653 https://github.com/jaegertracing/jaeger-ui/pull/1643
It would help you understand the issue's landscape.

anikdhabal:
https://testing-library.com/docs/react-testing-library/migrate-from-enzyme you can refer this @Sangam-ghimire @AbhishekTiwari23

AbhishekTiwari23:
Thanks, working on it

[...]

PR #1691, opened 2023-08-16
[tests] Migrate TraceIDSearchInput.test from Enzyme to RTL

anshgoyalevil:
- This PR migrates the `TraceIDSearchInput.test.js` from Enzyme to RTL
- Snapshot testing is replaced with specific expect statement
- Part of: https://github.com/jaegertracing/jaeger-ui/issues/1668

- Using yarn lint && yarn test-ci TraceIDSearchInput --coverage

## Checklist


packages/jaeger-ui/src/components/App/TraceIDSearchInput.tsx
```diff
@@ -36,7 +36,12 @@ class TraceIDSearchInput extends React.PureComponent<Props> {
 
   render() {
     return (
-      <Form layout="horizontal" onSubmit={this.goToTrace} className="TraceIDSearchInput--form">
+      <Form
+        aria-label="form"
```
yurishkuro, line 40: Form as HTML element is generally invisible to user, how does accessibility label help? I think it would make more sense on the `<input>` with label "trace ID"

yurishkuro: I don't think testing needs is a good reason to add (permanent) attributes to HTML if they don't improve the HTML itself. I would suggest finding other ways to locate elements for tests.

yurishkuro: Also, because form itself is invisible for the user, if some test is written that requires access to the form then the test probably does not pass the principles of testing established by RTL, "test from user perspective".

anshgoyalevil: > does not pass the principles of testing established by RTL, "test from user perspective".

Yes, for that reason only, E2E tests work best in frontend testing.

packages/jaeger-ui/src/components/App/TraceIDSearchInput.tsx
```diff
@@ -36,8 +36,18 @@ class TraceIDSearchInput extends React.PureComponent<Props> {
...
-        <Input name="idInput" placeholder="Lookup by Trace ID..." prefix={<Icon type="search" />} />
+      <Form
+        data-testid="TraceIDSearchInput--form"
+        layout="horizontal"
+        onSubmit={this.goToTrace}
+        className="TraceIDSearchInput--form"
+      >
+        <Input
+          data-testid="idInput"
+          name="idInput"
```
yurishkuro, line 47: can we not search by idInput name?

packages/jaeger-ui/src/components/App/TraceIDSearchInput.tsx
```diff
@@ -36,8 +36,18 @@ class TraceIDSearchInput extends React.PureComponent<Props> {
 
   render() {
     return (
-      <Form layout="horizontal" onSubmit={this.goToTrace} className="TraceIDSearchInput--form">
-        <Input name="idInput" placeholder="Lookup by Trace ID..." prefix={<Icon type="search" />} />
+      <Form
+        data-testid="TraceIDSearchInput--form"
```
yurishkuro, line 40: is `data-testid` preserved in the production output or stipped out?

anshgoyalevil: I just read about it, and yes it ships to the production builds.

The official [FAQs](https://testing-library.com/docs/dom-testing-library/faq/) of RTL suggests using this [babel plugin](https://www.npmjs.com/package/babel-plugin-react-remove-properties) for stripping out that in production, if that's an issue

packages/jaeger-ui/src/components/App/TraceIDSearchInput.tsx
anshgoyalevil: No. Currently RTL doesn't support that method

packages/jaeger-ui/src/components/App/TraceIDSearchInput.tsx
yurishkuro: let's use that plugin then, it's weird to have useless test properties in prod

anshgoyalevil: I agree. Never realized production code needs this much sanity.

yurishkuro: So I think this specific component is not as critical since there's only one of it, but I want to establish a _sustainable pattern_ that can be used for _all tests_, as some of the components could be as granular as a single span in a trace (and a trace may have 10s of thousands of spans), so having these test properties on those components will have noticeable impact on prod HTML.

anshgoyalevil: Absolutely. It might affect performance, as you stated, because each chunk of extra data matters.

I have added the Babel plugin to strip that off.

I manually analyzed the build files, and we were correct, it was having that `"data-testid":"TraceIDSearchInput--form"` attribute.
Took multiple iterations of `yarn build` to finally make the plugin work with vite tooling.

scripts/generateDepcheckrcJaegerUI.js
```diff
@@ -27,7 +27,7 @@ const packageNames = [
   ...babelConfiguration.plugins,
 ];
 
-const otherPackages = ['jest-environment-jsdom'];
+const otherPackages = ['jest-environment-jsdom', 'babel-plugin-react-remove-properties'];
```
yurishkuro, line 30: wait, why is it not auto-discovered from babelrc? I don't see babelrc change, isn't it necessary to make babel actually use the plugin, not just install it as dep?

anshgoyalevil: We are not using Babel plugins for production builds.

I was also confused earlier, since we have a `babel-transform.js` file. I tried adding the plugin to that file, but it didn't worked.
Later realised that it is only being used for tests, and not for builds.

For builds, we are using `vite.config.js` and depcheck can't find that out that we are using babel plugin inside vite.config.js

anshgoyalevil: Parsing it seems tricky. I applied another workaround, since depcheck by default supports .babelrc files.

Issue #1592, opened 2023-07-18
[Bug]: Negative Self time in Trace Statistics view


TheOxFromOutOfTheBox:
### What happened?

I have a sample C++ program where in  I have a main span that lasts for 10 seconds.
Span 1 - 0-10 seconds
Now I have 5 children of Span 1
Span 2-6 - 2-8 seconds (duration of 6 seconds)

So, actual self time should be 4 seconds, however I get a self time of -14 seconds.

### Steps to reproduce

Create 1 span with 4-5 child spans with each having a duration just smaller than the parent span.

### Expected behavior

Self time should be positive, however it is not.

### Jaeger backend version

1.45

### SDK

OpenTelemetry C++ SDK

### Pipeline

Otel SDK -> Otel Collector -> Jaeger Collector & Query -> Cassandra DB

### Storage backend

Cassandra v4

### Operating system

Windows

### Deployment model

binaries on bare metal

yurishkuro:
moving to UI repo

yurishkuro:
This probably requires some tweak to the algorithm that may be too simplistic right now. This also overlaps with the Critical Path work #1288 by @GLVSKiriti

GLVSKiriti:
@yurishkuro Please clarify this doubt Can a `CHiILD_OF` type span overflow from its parent span?

yurishkuro:
what do you mean by overflow - end later than parent? Yes it can.

GLVSKiriti:
> what do you mean by overflow - end later than parent? Yes it can.

Then what is difference between follow_from and child_of

like generally child_of span for sync process and follow_from span for async process
if parent not waits for child_of span to complete then how it is different from follow_from

yurishkuro:
Parent is expected to wait for child to complete, but the finish time of the span is not necessarily when the response was sent to the parent. I've seen in Java instrumentation children spans finishing significantly later due to thread suspension, even though the successful response was already sent to the parent. There are also cases when parent times out waiting on a child.

However, these are usually cases with _remote child_. The relationship between parent server-span and children client-spans is more robust.

GLVSKiriti:
What is selfTime in this cases.
(1.) child span ends later
![a1](https://github.com/jaegertracing/jaeger-ui/assets/116095646/bb724ad8-76b7-45c9-ab26-5002541180ef)
Here is selfTime of A = 2-1 = 1 and selfTime of B = 11-2 = 9
(2.) child starts later
[...]

PR #1700, opened 2023-08-19
[Part of] Self time Calculation to be unified across multiple screens

### LINKED ISSUE #1263: [Enhancement]: Self time Calculation to be unified across multiple screens
### Requirement

Enhancements to Trace statistics view 

- Similar Self Time as Trace Graph page. 
- Page to load faster for trace containing > 10K spans

### Problem

Trace Statistics View is unresponsive for traces > 10K spans if second Group By with high cardinality like tag.db@statement are selected.

Currently, there are 2 views in which self time is computed. 
- Trace Graph
- Trace Statistics

Both of them follow different logic. 
Key difference being Trace Graph [ignores](https://github.com/jaegertracing/jaeger-ui/blob/v1.27.4/packages/jaeger-ui/src/components/TracePage/TraceGraph/calculateTraceDagEV.tsx#L59) the follows-from relationship since the producer span doesn't have any control over when will the consumer span start whereas the Trace Statistics view computes the self time factoring in the [time spent between the start of the producer span and start of the consumer span](https://github.com/jaegertracing/jaeger-ui/blob/v1.27.4/packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx#L214-L220). 

Moreover the logic used in trace graph is much simpler and easier to understand whereas the trace statistics view logic is very complex and page is unresponsive when computing self time for traces containing > 10K spans.

### Proposal

Use the similar logic in both Trace Graph & Trace Statistics views to compute the self time. If it's alright, I can raise a PR for the same. 

### Open questions

_No response_

GLVSKiriti:
## Description of the changes
- Added getChilOfSpans function which ignores both followFrom spans and its descendants in selfTime Calculation

## Checklist
- [x] Improve the selfTime logic in TraceGraph view
- [ ]  To add unit tests to the getChildOfSpans function
- [ ] Implement similar selfTime logic in TraceStatistics view
    - Also solves #1592

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceGraph/calculateTraceDagEV.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceGraph/calculateTraceDagEV.tsx
@@ -21,8 +21,7 @@ import TDagNode from '../../../model/trace-dag/types/TDagNode';
 import { TDenseSpanMembers } from '../../../model/trace-dag/types';
 import { Trace, Span, KeyValuePair } from '../../../types/trace';
 import { TSumSpan, TEv } from './types';
-
-let parentChildOfMap: Record<string, Span[]>;
+import getChildOfSpans from '../../../utils/getChildOfSpans';
 
 export function isError(tags: Array<KeyValuePair>) {
   if (tags) {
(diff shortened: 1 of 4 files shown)


packages/jaeger-ui/src/components/TracePage/TraceGraph/calculateTraceDagEV.tsx
```diff
@@ -50,43 +49,35 @@ function mapFollowsFrom(
...
+      childSpan.startTime + (childSpan.duration <= 0 ? 0 : childSpan.duration - 1)
+    );
   });
   return childrenDrange;
 }
 
 export function calculateTraceDag(trace: Trace): TraceDag<TSumSpan & TDenseSpanMembers> {
   const baseDag = TraceDag.newFromTrace(trace);
   const dag = new TraceDag<TSumSpan & TDenseSpanMembers>();
+  const SpanMap = trace.spans.reduce((map, span) => {
```
yurishkuro, line 68:
```suggestion
  const spanMap = trace.spans.reduce((map, span) => {
```

packages/jaeger-ui/src/utils/getChildOfSpans.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const followFromSpanIds: string[] = [];
+  const followFromSpansDescendantIds: string[] = [];
+
+  // First find all FOLLOWS_FROM refType spans
+  spanMap.forEach(each => {
+    if (each.references[0]?.refType === 'FOLLOWS_FROM') {
+      followFromSpanIds.push(each.spanID);
+      // Remove the spanId from childSpanIds array of its parentSpan
+      const parentSpan = spanMap.get(each.references[0].spanID)!;
+      parentSpan.childSpanIds = parentSpan.childSpanIds.filter(a => a !== each.spanID);
```
yurishkuro, line 31: this is unexpected side effect. Functions should not mutate data structure passed to them.

yurishkuro (commented):
Sorry, I don't follow the objective. Say you have this graph:
```
[span A] <-(follows-from)<- [span B] <-(child-of)<- [span C]
```
it is my understanding that this new function `getChildOfSpans` would remove B and C from the list and only leave A. However, when span B is displayed in a graph or table and needs to show its own self-time, it still needs to know about its child C.

PR #1710, opened 2023-08-21
[CI]: Add Concurrency Grouping to GitHub Workflows

### LINKED ISSUE #1709: [CI]: Add Concurrency Grouping to GitHub Workflows
### Requirement

Whenever we push commits to a PR, workflows are triggered for that commit. After that, if we push additional commits to this PR, workflows in Github Actions run on both commit

### Problem

We need to cancel the previous run and run only on the most recent pushed commit. This would help save some GitHub Action Minutes and unexpected workflow failures on subsequent commits.

### Proposal

_No response_

### Open questions

_No response_

anshgoyalevil:
- fixes #1709 

## Description of the changes
- Whenever we push commits to a PR, workflows are triggered for that commit. After that, if we push additional commits to this PR, workflows in Github Actions run on both commits.
- We need to cancel the previous run and run only on the most recent pushed commit. This would help save some GitHub Action Minutes and unexpected workflow failures on subsequent commits.

- It was checked in forked repo, and is working fine there. Following is the run, which was cancelled due to subsequent commit
- https://github.com/anshgoyalevil/jaeger-ui/actions/runs/5930202449

## Checklist

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -9,7 +9,7 @@ on:
     branches: [main]
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref || github.run_id }}
+  group: ${{ github.workflow }}-${{ (github.event.pull_request && github.event.pull_request.number) || github.ref || github.run_id }}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -2,10 +2,15 @@ name: "CodeQL"
 
 on:
   push:
-    branches: [main]
+    branches-ignore:
+      - '*'
 
   pull_request:
     branches: [main]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref || github.run_id }}
+  cancel-in-progress: true
 
 jobs:
   codeql-analyze:
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -2,8 +2,8 @@ name: "CodeQL"
 
 on:
   push:
-    branches-ignore:
-      - '*'
+    branches:
+      - '**'
 
   pull_request:
     branches: [main]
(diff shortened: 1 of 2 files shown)


.github/workflows/codeql.yml
```diff
@@ -2,11 +2,16 @@ name: "CodeQL"
...
   push:
-    branches: [main]
+    branches:
+      - '**'
 
   pull_request:
     branches: [main]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref || github.run_id }}
```
yurishkuro, line 12: what happens when it's not a pull request? `github.event.pull_request` is null, will this NPE trying to access `.number`?

yurishkuro: I understand OR, my concern is whether it will throw an exception on non-PR because it cannot evaluate the first token (null pointer exception)

anshgoyalevil: No. It won't give an exception like that. I have tested it with push events also, where the `github.event.pull_request.number` is expected to be null or undefined.

yurishkuro: to put this differently, if this was javascript running on non-PR, `github.event.pull_request.number` would bomb, but `(github.event.pull_request && github.event.pull_request.number)` would always execute.

anshgoyalevil: Official docs explain it better: https://docs.github.com/en/actions/using-jobs/using-concurrency#example-using-a-fallback-value

anshgoyalevil: > to put this differently, if this was javascript running on non-PR, github.event.pull_request.number would bomb, but (github.event.pull_request && github.event.pull_request.number) would always execute.

Got your concern, but it is different from JS, though your syntax is also correct, and can be used to be double sure

anshgoyalevil: Updated the syntax 🚀

yurishkuro (approved):

yurishkuro:
@anshgoyalevil something isn't right with your workflow change, we have [duplicate checks running now](https://github.com/jaegertracing/jaeger-ui/pull/1713)

image

anshgoyalevil:
That seems a strange behaviour. Would have to debug that, since having wildcard in branches shouldn't affect its behaviour as per my knowledge

PR #1721, opened 2023-08-22
Upgrade react-icons from 2.2.7 to 4.10.1

anshgoyalevil:
- This PR upgrades the react-icons from 2.2.7 to 4.10.1
- It also removes the types/react-icons as it is now supplied with the package itself
- The visual changes are tracked using https://gist.github.com/anshgoyalevil/aa6d32dce103c7c9bea7865d0346b2e9

## Checklist


anshgoyalevil:
@yurishkuro I have included the icon changes.

The previous icons were not available to look up somewhere (yes, 2nd option was to navigate pages manually in jaeger-ui), so I dug up their raw SVG path from node_modules and then took a snip of their rendered version using [SVG Viewer](https://codebeautify.org/svg-viewer)

Feel free to suggest any alternate icon, in case you feel like they are changed a lot. You may use this to surf all of [react icons](https://react-icons.github.io/react-icons).

yurishkuro:
not exactly the scope of this PR, but I dislike that we use both antd/icons and react-icons, is there a strong reason to do that and not converge onto one?

anshgoyalevil:
As per my analysis so far, we can use react-icons in place of antd icons, but not vice versa.
Since react-icons offers the best of both worlds, and we too are using some icons which, I haven't looked exactly but I think are not available in antd icons, or atleast react-icone/ai

yurishkuro (approved):

yurishkuro:
> As per my analysis so far, we can use react-icons in place of antd icons, but not vice versa.

it's possible, although I am skeptical if we can't find close equivalents.

What I don't like about react-icons is that it's all over the place with all kinds of namespaces, like md, io, fa. As I understand, those are competing icon sets, presumably created with some design consistency principles, but we keep mixing them randomly.

anshgoyalevil:
I agree.
Maybe in future, would have to put up some extra work in finding the alternatives.
Like we can stick to react-icons/ `fa` (font awesome) icons. As far as I know, they offer a wide range of icons.

yurishkuro:
I suggest we book a ticket and include a simple grep/awk output that lists all distinct icons we're using. I doubt it's more than a couple dozen.

yurishkuro:
https://github.com/jaegertracing/jaeger-ui/issues/1723

anshgoyalevil:
Thanks for creating the ticket.
I would try finding the icons and prepare a gist.
In the meantime, someone from community can pick it up

PR #1722, opened 2023-08-22
Replace react-vis with Recharts

### LINKED ISSUE #1597: Replace react-vis library for drawing charts
This library (https://github.com/uber/react-vis) is on the verge of being unsupported (it was recently archived, but then resurrected,  apparently with a single maintainer). We do not use a lot of features from it, mostly for the charts:
```
$ grep -rl react-vis ./packages/jaeger-ui/src | grep -v DependencyGraph
./packages/jaeger-ui/src/components/Monitor/ServicesView/serviceGraph.tsx
./packages/jaeger-ui/src/components/Monitor/ServicesView/operationDetailsTable/opsGraph.tsx
./packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ScatterPlot.jsx
./packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ScatterPlot.test.js
```
I think it should not be too difficult to replace it with something more modern an actively supported / used. I don't have a specific candidate in mind, doing a bit of research and pros/cons is part of this ticket (e.g. https://wpdatatables.com/highcharts-alternative/).

NB: I intentionally excluded DependencyGraph from this, since that could be part of another issue:
  * https://github.com/jaegertracing/jaeger-ui/issues/1466

anikdhabal:
Resolves #1597 

Screenshot 2023-08-22 053025

## Checklist


packages/jaeger-ui/src/components/Monitor/ServicesView/index.test.js
```diff
@@ -394,6 +394,6 @@ describe('yAxisTickFormat', () => {
   it('value is 2 second', () => {
     const timeInMs = 1000;
     const displayTimeUnit = 'seconds';
-    expect(yAxisTickFormat(timeInMs, displayTimeUnit)).toBe(1);
+    expect(yAxisTickFormat(timeInMs, displayTimeUnit)).toBe('1');
```
yurishkuro, line 397: The snapshot above, packages/jaeger-ui/src/components/Monitor/ServicesView/__snapshots__/serviceGraph.test.js.snap, is very large and not useful. Can we rethink the unit tests that produces it and replace it with more explicit tests for user-facing features instead of a snapshot?

yurishkuro (commented):
@anshgoyalevil mind reviewing this?

anshgoyalevil:
On it 🚀

package.json
```diff
@@ -63,5 +63,8 @@
     "hooks": {
       "pre-commit": "npm-run-all -ln --parallel lint test"
     }
+  },
+  "dependencies": {
+    "recharts": "^2.7.3"
```
anshgoyalevil, line 68: Isn't it supposed to be a dependency for packages/jaeger-ui?

packages/jaeger-ui/src/components/Monitor/ServicesView/index.tsx
```diff
@@ -123,8 +123,10 @@ const calcDisplayTimeUnit = (serviceLatencies: ServiceMetricsObject | ServiceMet
 };
 
 // export for tests
-export const yAxisTickFormat = (timeInMS: number, displayTimeUnit: string) =>
-  convertToTimeUnit(timeInMS * 1000, displayTimeUnit);
+export const yAxisTickFormat = (timeInMS: number, displayTimeUnit: string) => {
+  const formattedValue = convertToTimeUnit(timeInMS * 1000, displayTimeUnit);
+  return formattedValue.toString();
+};
```
anshgoyalevil, line 129: I think we can omit the use of `return formattedValue.toString();` and simply return `formattedValue`.
Is there any specific reason for returning a string value?

anikdhabal: > I think we can omit the use of `return formattedValue.toString();` and simply return `formattedValue`. Is there any specific reason for returning a string value?

Because in recharts the tickFormatter function expect sting | undefined. BaseAxisProps.tickFormatter?: ((value: any, index: number) => string) | undefined

packages/jaeger-ui/src/components/Monitor/ServicesView/operationDetailsTable/opsGraph.tsx
```diff
@@ -60,25 +55,22 @@ export class OperationsGraph extends React.PureComponent<TProps> {
...
+              top: 2,
+            }}
             data={dataPoints}
-            {...dynProps}
-          />
-          <LineSeries className="area-elevated-line-series" color={color} data={dataPoints} {...dynProps} />
-        </XYPlot>
+          >
+            <XAxis dataKey="x" hide />
+            <YAxis hide />
```
anshgoyalevil, line 69: Why are we using `hide` attribute here? Can't these components be just omitted?

package.json
anikdhabal: Ahh yes. Silly mistake.
Thank you.

packages/jaeger-ui/src/components/Monitor/ServicesView/serviceGraph.test.js
```diff
@@ -108,35 +108,17 @@ describe('<ServiceGraph>', () => {
...
-  it('AreaSeries onNearestX, XYPlot onMouseLeave', () => {
-    wrapper.setProps({
-      ...props,
-      loading: false,
-      metricsData: serviceMetrics.service_call_rate,
     });
-    wrapper.find('AreaSeries').at(0).prop('onNearestX')({ x: 1, y: 2 }, { index: 7 });
-    expect(wrapper.state().crosshairValues).toEqual([{ label: 0.95 }]);
-    wrapper.find('XYPlot').prop('onMouseLeave')();
-    expect(wrapper.state().crosshairValues).toEqual([]);
```
anshgoyalevil, line 139: I think removing these tests defeats the purpose of test-driven development. Can we add some similar tests for tooltip hover test to verify the functionality?

packages/jaeger-ui/src/components/Monitor/ServicesView/serviceGraph.tsx
```diff
@@ -40,16 +40,11 @@ type TProps = {
   yDomain?: number[];
   color?: string;
   marginClassName?: string;
-  yAxisTickFormat?: (v: number) => number;
+  yAxisTickFormat?: (v: number) => string;
```
anshgoyalevil, line 43: Can't we return a number here?

packages/jaeger-ui/src/components/Monitor/ServicesView/serviceGraph.tsx
```diff
@@ -87,30 +79,20 @@ export class ServiceGraphImpl extends React.PureComponent<TProps> {
...
-              });
-            }}
-            opacity={0.1}
-            color={[color || this.colors[idx]]}
+            type="monotone"
+            dataKey="y"
+            connectNulls
+            fillOpacity={0.1}
+            stroke={color || this.colors[idx]}
+            fill={color || this.colors[idx]}
```
anshgoyalevil, line 89: Aren't we passing on the data prop here?

packages/jaeger-ui/src/components/Monitor/ServicesView/serviceGraph.tsx
```diff
@@ -87,30 +79,20 @@ export class ServiceGraphImpl extends React.PureComponent<TProps> {
...
           />
         );
         graphs.push(
-          <LineSeries
-            getNull={(d: Points) => d.y !== null}
-            key={i++}
-            data={line.metricPoints ? line.metricPoints : []}
-            color={[color || this.colors[idx]]}
-          />
+          <Line connectNulls key={i++} type="monotone" dataKey="y" stroke={color || this.colors[idx]} />
```
anshgoyalevil, line 93: Aren't we passing on the data prop here?

packages/jaeger-ui/src/components/Monitor/ServicesView/serviceGraph.tsx
```diff
@@ -87,30 +79,20 @@ export class ServiceGraphImpl extends React.PureComponent<TProps> {
...
         );
+        data.push(line.metricPoints ? line.metricPoints : []);
```
anshgoyalevil, line 95: That isn't the right way of doing this. 
How would this data be supplied to the components without a rerender?

packages/jaeger-ui/src/components/Monitor/ServicesView/index.test.js
anikdhabal: Yes, instead of relying on snapshots, we can write individual test cases that cover specific scenarios and assertions. Thank you.

packages/jaeger-ui/src/components/Monitor/ServicesView/operationDetailsTable/opsGraph.tsx
anikdhabal: Yes, we can omit these components.

packages/jaeger-ui/src/components/Monitor/ServicesView/serviceGraph.tsx
anikdhabal: Nope, as I previously described that timeFormatter function only accept string.

packages/jaeger-ui/src/components/Monitor/ServicesView/serviceGraph.tsx
anikdhabal: There are no props to pass the data here. Also, we don't need to pass the data props here; the data props is passed to the parent component.

packages/jaeger-ui/src/components/Monitor/ServicesView/serviceGraph.tsx
anikdhabal: No need.

packages/jaeger-ui/src/components/Monitor/ServicesView/serviceGraph.test.js
anikdhabal: Yes, on it.

packages/jaeger-ui/src/components/Monitor/ServicesView/serviceGraph.tsx
anikdhabal: That's a valid point. I'm replacing it with state.

packages/jaeger-ui/src/components/Monitor/ServicesView/index.test.js
anikdhabal: I've replaced the snapshot test with more explicit test and also added test for tooltip formatter. Can you check it once?

packages/jaeger-ui/src/components/Monitor/ServicesView/serviceGraph.test.js
```diff
@@ -39,61 +39,66 @@ describe('<ServiceGraph>', () => {
     expect(wrapper.length).toBe(1);
   });
 
-  it('Loading indicator is displayed', () => {
-    expect(wrapper).toMatchSnapshot();
+  it('displays loading indicator when loading', () => {
+    expect(wrapper.find('LoadingIndicator').exists()).toBe(true);
```
yurishkuro, line 43: `exists()).toBe(true)` - is there a simpler form like isNotNull?

yurishkuro:
Q: in the screenshot all dots are the same size, but the original chart used number of spans to control the size of the trace. Is such functionality supported in recharts?

packages/jaeger-ui/src/components/Monitor/ServicesView/serviceGraph.test.js
anikdhabal: I don't think. But there are similar from like ***`toBeTruthy()`*** or **`toBeTrue()`**

anikdhabal:
> Q: in the screenshot all dots are the same size, but the original chart used number of spans to control the size of the trace. Is such functionality supported in recharts?

Apart from the X and Y axes, Recharts also provides a Z-axis. With the assistance of the range property of Z axis, we can achieve this desired functionality. (similar to sizeRange in react-vis)

packages/jaeger-ui/src/components/Monitor/ServicesView/serviceGraph.test.js
yurishkuro: https://github.com/jaegertracing/jaeger-ui/blob/60093a4ce45c6f14ace930386d1f19d37d8f9a36/packages/jaeger-ui/src/components/Monitor/ServicesView/index.test.js#L289

yurishkuro:
> Apart from the X and Y axes, Recharts also provides a Z-axis. With the assistance of the range property of Z axis, we can achieve this desired functionality. (similar to sizeRange in react-vis)

Are you planning to add that? Otherwise you're removing exiting functionality

anikdhabal:
> > Apart from the X and Y axes, Recharts also provides a Z-axis. With the assistance of the range property of Z axis, we can achieve this desired functionality. (similar to sizeRange in react-vis)
> 
> Are you planning to add that? Otherwise you're removing exiting functionality

Yes, I'm going to add that.

packages/jaeger-ui/src/components/Monitor/ServicesView/serviceGraph.test.js
anikdhabal: Ahh OKK

anikdhabal:
@yurishkuro unable to trace why the CI is failed.

> > Apart from the X and Y axes, Recharts also provides a Z-axis. With the assistance of the range property of Z axis, we can achieve this desired functionality. (similar to sizeRange in react-vis)
> 
> Are you planning to add that? Otherwise you're removing exiting functionality

I've added Z-axis for the size range and set the data key to 'size' as it carries the span length.

anikdhabal:
> > Apart from the X and Y axes, Recharts also provides a Z-axis. With the assistance of the range property of Z axis, we can achieve this desired functionality. (similar to sizeRange in react-vis)
> 
> Are you planning to add that? Otherwise you're removing exiting functionality

I've added Z-axis for the size range and set the data key to 'size' as it carries the span length.
Screenshot 2023-08-29 000542

anikdhabal:
> when I run `yarn test` locally I am getting various errors related to your changes, like this:
> 
> ```
>  FAIL  src/components/SearchTracePage/SearchResults/ScatterPlot.test.js
>   ● Console
> 
>     console.error
>       Warning: Failed prop type: Invalid prop `data[0].traceID` of type `number` supplied to `ScatterPlot`, expected `string`.
>           at data (/Users/ysh/dev/jaegertracing/jaeger-ui/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ScatterPlot.jsx:33:11)
> 
>       55 | it('<ScatterPlot /> should render base case correctly', () => {
>       56 |   const wrapper = shallow(
>     > 57 |     <ScatterPlot
>          |     ^
>       58 |       data={[
>       59 |         { x: Date.now() - 3000, y: 1, traceID: 1 },
>       60 |         { x: Date.now() - 2000, y: 2, traceID: 2 },
> ```
> 
> (for some reason they are not included in the CI output).

Okay, let me go through it and try to fix it.

anshgoyalevil:
> for some reason they are not included in the CI output

I also noticed the same. There are a lot of warnings because of the Deprecated Enzyme API, and due to that, the useful errors cannot be seen in CI output.

A quick fix here can be to use the `--silent` flag in `test-ci` script.
This would help suppressing the warnings and console logs, after which the errors can hopefully be seen in CI output.

anikdhabal:
> > for some reason they are not included in the CI output
> 
> I also noticed the same. There are a lot of warnings because of the Deprecated Enzyme API, and due to that, the useful errors cannot be seen in CI output.
> 
> A quick fix here can be to use the `--silent` flag in `test-ci` script. This would help suppressing the warnings and console logs, after which the errors can hopefully be seen in CI output.
> 
> Though ignoring the warnings by suppressing them is also wrong. In future, we may fix all of the warnings (hopefully), and fail the CI on warning too.

Thanks for your review. I will definitely try it.

yurishkuro:
the two CI builds now very clearly show the breaks, both from linter and compiler

anikdhabal:
> the two CI builds now very clearly show the breaks, both from linter and compiler

Yes, let me fix it.

anikdhabal:
> the two CI builds now very clearly show the breaks, both from linter and compiler

It gives this error every time: 'TypeError: window.ResizeObserver is not a constructor.

anshgoyalevil:
@anikdhabal 

```js
global.ResizeObserver = jest.fn().mockImplementation(() => ({
    observe: jest.fn(),
    unobserve: jest.fn(),
    disconnect: jest.fn(),
}))
```

Add the above code to `packages/jaeger-ui/test/jest-per-test-setup.js` or `jest.global-setup.js`.

It should then work.

Ref: https://stackoverflow.com/questions/64558062/how-to-mock-resizeobserver-to-work-in-unit-tests-using-react-testing-library

priyanshu-kun:
@anikdhabal are you still working on this issue?

anikdhabal:
> @anikdhabal are you still working on this issue?

Yeah, I was inactive due to my exam, but I have fixed that issue now.

MarkPollmann:
Can we help to get this shipped?

yurishkuro:
help is welcome, this needs a rebase

MarkPollmann:
Okay, I'm on it

yurishkuro:
closing for inactivity

PR #1724, opened 2023-08-23
Add JSON Folding Support in Logs

### LINKED ISSUE #1415: [Feature]: log json folding
### Requirement

As a user of Jaeger UI 
I want ability to fold long json blocks in span log entries
so that simplify search of particular value.

### Problem

Current JSON view only highlights keys, giving long hard-to find stuff view, example:
Снимок экрана 2023-05-12 в 13 26 44

Currently we use long log messages of hundreds of lines, very hard to quickly navigate to important parts.

### Proposal

Add folding interactions to JSON view, like in many JSON editors:
Снимок экрана 2023-05-12 в 13 30 29

Possible options:
* fold by default to 1st level of nesting
* expand/collapse all
* fold to 2/3/4 level 

See more possible ideas here:
https://www.jetbrains.com/help/rider/Code_Folding.html

But the basic: just ability to fold not-interesting region -- we absolutely need it.

### Open questions

_No response_

prathamesh-mutkure:

Resolves #1415 

## Description of the changes
- I've added `react-json-view-lite` to render json in tree view
- The old `jsonMarkup.js` file has been removed
- The styling has been matched to old design

- The test cases passed
- The old and new approaches followed same design

Before -

Screenshot 2023-08-23 at 1 11 03 PM

Now -

Screenshot 2023-08-23 at 1 11 30 PM

## Checklist


packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/index.css
```diff
@@ -22,21 +22,53 @@ limitations under the License.
   line-height: 17px;
   font-size: 13px;
   font-family: monospace;
+
   white-space: pre-wrap;
+  white-space: -moz-pre-wrap;
+  white-space: -pre-wrap;
+  white-space: -o-pre-wrap;
```
yurishkuro, line 29: how does this work? You're redefining the same property, isn't it going to only keep the latest value?

prathamesh-mutkure: The `react-json-view-lite` library had a default styling in their [code](https://github.com/AnyRoad/react-json-view-lite/blob/release/src/styles.module.css#L3)

I copied the base CSS from there and then overridden some properties based on our design

yurishkuro: let's remove it. I don't know what it means. Vendor prefixes are usually used with attribute names, I've never seen them used in values.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.tsx
```diff
@@ -79,7 +71,26 @@ function formatValue(key: string, value: any) {
   } else if (Array.isArray(parsed) && shouldDisplayAsStringList(key)) {
     content = stringListMarkup(parsed);
   } else {
-    content = _jsonMarkup(parsed);
+    content = (
+      <JsonView
+        data={parsed}
+        shouldInitiallyExpand={allExpanded}
```
yurishkuro, line 77: could this be conditional on the size of the output, e.g. if the JSON object has more than 10 keys then don't start in expanded form.

Also, please post a screenshot of how it looks in the collapsed form.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/index.css
prathamesh-mutkure: Sure

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.tsx
prathamesh-mutkure: Here's how it looks when collapsed at the root

Screenshot 2023-08-24 at 12 55 08 PM

prathamesh-mutkure: The JSON tree will now only expand if it has 10 or less keys, nested objects will remain collapsed otherwise

yurishkuro (approved):

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.tsx
```diff
@@ -79,7 +71,28 @@ function formatValue(key: string, value: any) {
   } else if (Array.isArray(parsed) && shouldDisplayAsStringList(key)) {
     content = stringListMarkup(parsed);
   } else {
-    content = _jsonMarkup(parsed);
+    const shouldJsonTreeExpand = Object.keys(parsed).length <= 10;
```
yurishkuro, line 74: I think this is good as a first approximation, but the right solution would be estimating the total height to be >10, not just the number of first-level children.

prathamesh-mutkure: Oh, I guess we could work on some algorithm, It could be something like 1 property = 8 px, and then we could just go on calculating the height of the first X elements (recursively) and then decide to expand or collapse the tree

This task could be included in the upcoming LFX term

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.tsx
```diff
@@ -79,7 +71,28 @@ function formatValue(key: string, value: any) {
...
+          ...defaultStyles,
+          container: 'json-markup',
+          label: 'json-markup-key',
+          stringValue: 'json-markup-string',
+          numberValue: 'json-markup-number',
+          booleanValue: 'json-markup-bool',
+          nullValue: 'json-markup-null',
+          undefinedValue: 'json-markup-undefined',
+          expander: 'json-markup-expander',
+          basicChildStyle: 'json-markup-child',
```
bobrik, line 90: This doesn't look like a desirable outcome:

image

yurishkuro: please elaborate

bobrik: I made https://github.com/jaegertracing/jaeger-ui/pull/2168

PR #1734, opened 2023-08-25
Upgrade antd to 4.24.13

anshgoyalevil:
- This PR attempts to upgrade `antd` to `v4.24.13 (Latest as of 28-08-2023`

## Description of the changes
- AntD v4 requires a newer version of React typings, which also requires an explicit Yarn resolution to avoid pulling in other versions of the typings.
- React refs for AntD Input components need a typing update.
- The NumberInput component now guarantees to return number | null, so some custom type-checking code related to that can be removed.
- Typing changed around the Table component around row selection and sorting logic, requiring appropriate updates. In particular, sorting by multiple sortable columns now reports all columns that are active in sorting, so the event handling code must also accept an array.
- `window.matchMedia` API is not provided by jsdom, so [custom mocking logic](https://jestjs.io/docs/manual-mocks#mocking-methods-which-are-not-implemented-in-jsdom) needs to be in place
- `dataIndex` now requires an array of keys to work with tables with nested keys.
- `antd/compatible` library is used to make forms work with the existing libraries
- Tests are updated
- Visual changes are verified

Most of the suggestions have been taken up from the PR https://github.com/jaegertracing/jaeger-ui/pull/1261, thanks to @mszabo-wikia 

## Checklist


packages/jaeger-ui/src/components/Monitor/ServicesView/operationDetailsTable/__snapshots__/index.test.js.snap
anshgoyalevil, line 1: This snapshot file is very very huge and is technically useless. I would try to remove snapshot testing from this, to introduce some expect statements

yurishkuro (commented):
Q: are any of these changes possible with antd 3.x, i.e. before the actual upgrade?

anshgoyalevil:
> Q: are any of these changes possible with antd 3.x, i.e. before the actual upgrade?

Not really, because most of these changes include type updates, which are supplied by antd 4.x.
If we talk about migration from snapshot testing, that could be done before this PR, but I am planning of doing it in other PRs, as it would anyway not affect the antd upgrade, though some changes are needed to the failing tests.

anshgoyalevil:
@yurishkuro 
Could you please verify the visual changes which are introduced with `antd v4`. Though, I have checked every inch of the UI, but your view would be appreciated.

anshgoyalevil:
There are some known significant visual changes in v4, which I haven't modified, like the Header Dropdown for `About Jaeger` is now replaced with three dots, and then the dropdown
Should I fix it too?

yurishkuro:
> There are some known significant visual changes in v4 ...

This is why I don't like this lock step upgrade that requires a huge PR. It would be easier if we could do changes component by component.

If you know the "known" changes, let's list then and decide one by one.

anshgoyalevil:
> This is why I don't like this lock step upgrade that requires a huge PR. It would be easier if we could do changes component by component.

I understand that, but since we are upgrading a library which causes a regression in the UI of the product, some UI part wasn't even displaying the data because of `dataIndex` changes.

Doing that component by component and merging the incomplete PR would have left those broken components at a worse stage.

> If you know the "known" changes, let's list then and decide one by one.

There are not many, just 1 or 2 are there. I am creating a gist, showing each page side by side, of old and new UI.

yurishkuro:
Are you certain that there were no intermediate versions of antd where both old and new style components were available? Usually framework give a grace period for upgrades where old way is deprecated (and logs warnings) but is still working, precisely to allow for this kind of situation where a big bang upgrade is too large to do in a single change.

anshgoyalevil:
Yes. Most probably, I checked that by manually installing the initial few versions of antd v4, ranging from `4.0.0-rc.0` to 4.12.3

Here's that quick analysis:

4.12.3 issue persists all over
4.2.0 issue persists all over
4.1.0 issue persists in the grid, but header is good
4.0.0 issue persists in the grid, but header is good
4.0.0-rc.0 issue persists in the grid, but header is good

The header thing looked easy to tackle so, I changed that too in the v4.23.13, as, a few other things were breaking in the v4.1.0

anshgoyalevil:
@yurishkuro 
I have framed the old vs. new table. Could you please check it out?

https://gist.github.com/anshgoyalevil/6081a14efbc9c7defa436a3a96263c0d

yurishkuro:
> 4.12.3 issue persists all over

Do you mean compile errors?

If there are minimal errors in earlier versions, could we split this PR and merge in phases?

yurishkuro:
cc @mmorel-35 I remember you mentioned incremental upgrades of some deps to have more narrow scope of changes, was it for antd or other deps?

anshgoyalevil:
> > 4.12.3 issue persists all over
> 
> Do you mean compile errors?
> 
> If there are minimal errors in earlier versions, could we split this PR and merge in phases?

That, we could do that too. There are some compile errors, as well as some UI changes, like Monitor page's grid view.

Would have to start all over, to determine the exact changes, to split into multiple PRs. Should I close this PR and create a narrow PR? As it would be somewhat difficult to go back in these changes and check.

yurishkuro:
How big of a change would be needed to go to antd/v4.0.0 as the first step? How many components would not compile or would not work correctly?

anshgoyalevil:
Can't exactly say that currently. Will update here asap.

anshgoyalevil:
I just checked the whole scenario at v4.0.0, and here's an overview:

- Monitor page's UI got messed up
- Trace Page -> Trace Spans Graph Page and Trace Statistics Page have some UI data errors (due to `dataIndex` change) | Same with compare page's Cohort Table
- Lookup by Trace ID Search Form is not working
- Search Page's form is not working
- Some Icons on Trace Page Header are not rendering

### Which components are same as per the main branch?
- Header (TovNav)
- Header (Dropdown)
- Form (Selectors) on Trace Spans Table page

anshgoyalevil:
@yurishkuro Any views upon this? https://github.com/jaegertracing/jaeger-ui/pull/1734#issuecomment-1695139859

yurishkuro:
It seems the basic prep they suggest is to move to the latest 3.x version and make sure that no deprecation warnings are showing. I assume they may have added new things in 3.x already while supporting the old behaviors. If that's the case then we can probably make some incremental changes before 4.x upgrade?

anshgoyalevil:
So here's what I got so far:

- Type Changes:
   - React refs for AntD `Input` components need a typing update. `(Only introduced in v4.x)`
   - The `NumberInput` component now guarantees to return `number | null`, so some custom type-checking code related to that can be removed. `(Only introduced in v4.x)`
   - Typing changed around the `Table` component around row selection and sorting logic, requiring appropriate updates. In particular, sorting by multiple sortable columns now reports all columns that are active in sorting, so event handling code must also accept an array. `(Only introduced in v4.x)`
   - The trace search form needs a workaround to function with redux-form, as AntD Forms no longer expose an onSubmit event that this library could hook into. For now, this is done by attaching the submit callback directly to the submit button instead as a click handler (which, somehow, still catches Enter keypresses on the search form). `(Can be done before this PR, using antd/compatible library)`
- Since there is no deprecation warning, the component and styling changes are introduced in v4.x

------------------------------------
What can we do before this PR?

- Use `antd/compatible` library
- Make some changes related to the grid.

Here's how I would be going:
- PR for adding antd/compatible lib and making changes to grid
- PR for upgrading antd to v4.0.0, so header won't need any change
- Upgrade to latest version of antd 4.x with header changes.

anshgoyalevil:
Completed in parts with #1748 #1750 #1751

PR #1737, opened 2023-08-27
Replace react-virtualized-select with Ant's Select

### LINKED ISSUE #1735: Remove `react-select` and `react-virtualized-select` for Ant's Select component
### Requirement

We currently have two packages for the Select component, `react-select` and `react-virtualized-select`, these must be replaced with Ant's [Select](https://ant.design/components/select/) component 

### Problem

- `react-virtualized-select` has been deprecated and no longer maintained
- `react-select` is not being used anywhere in the project except it's CSS styles, see [here](https://github.com/jaegertracing/jaeger-ui/blob/main/packages/jaeger-ui/src/components/common/VirtSelect.tsx#L22)
- There's a `TODO` to replace `react-virtualized-select` with `react-select` in the same [file](https://github.com/jaegertracing/jaeger-ui/blob/main/packages/jaeger-ui/src/components/common/VirtSelect.tsx#L62)

So we have to remove anyway `react-virtualized-select`, either with `react-select` or ant design component

Features of `react-select` 

- Animated Components
- Async support for loading options
- Customizable
- Creatable and Multi options 

Features of Ant's Select component

- All the above features
- Ability to group options
- Coordinated selections, the options of one select component can be controlled based on the value of other select component
- The sizing is more uniform with other UI components using Ant
- Better searching and sorting with features like Debouncing and remote search
- Automatic tokenization

Considering the above things, I guess it's just much better to use Ant Design's Select component as it's much more mature, has a uniform design with other components, and no overhead of additional packages.

More discussion on #1692 

### Proposal

- Remove `react-virtualized-select` and `react-select`
- Replace them with Ant's Select
- Adjust the design and CSS styles
- Fix failing tests (if any)

### Open questions

_No response_

prathamesh-mutkure:
- Resolves #1735 
- Also resolves [PR 1692](https://github.com/jaegertracing/jaeger-ui/pull/1692)

## Description of the changes
- I've removed the `react-virtualized-select`, `react-virtualized` and `react-select` package
- The first two were replaced with `antd`'s `<Select />` component
- The last component wasn't used anywhere in jaeger-ui
- The old components were styled to match the theme of Ant Design used by other components
- Thus, I removed the CSS files associated with old components as they were no longer needed
- Updated the test snapshots

- The existing test cases passed
- And the snapshots were updated
- The design of new components matched the old ones

Screenshot of changes

Screenshot 2023-08-28 at 4 08 31 AM

Screenshot 2023-08-28 at 4 10 34 AM

## Checklist


packages/jaeger-ui/src/components/Monitor/ServicesView/index.tsx
```diff
@@ -274,17 +269,21 @@ export class MonitorATMServicesViewImpl extends React.PureComponent<TPropsWithIn
...
-                component={AdaptedVirtualSelect}
+                component={reduxFormFieldAdapter({ AntInputComponent: Select })}
                 placeholder="Select A Service"
                 props={{
                   className: 'select-a-service-input',
                   value: this.getSelectedService(),
                   disabled: metrics.operationMetricsLoading,
-                  clearable: false,
-                  options: services.map((s: string) => ({ label: s, value: s })),
-                  required: true,
```
prathamesh-mutkure, line 285: We no longer require these as

- `clearable` is the default behaviour 
- `options` are being passed as children
- `required` prop is not present 

This applies to all the places

prathamesh-mutkure (commented):
I've added explanation for removing some props

packages/jaeger-ui/src/components/Monitor/ServicesView/index.tsx
```diff
@@ -77,11 +76,7 @@ type TDispatchProps = {
 const trackSearchOperationDebounced = _debounce(searchQuery => trackSearchOperation(searchQuery), 1000);
 
 const Search = Input.Search;
-
-const AdaptedVirtualSelect = reduxFormFieldAdapter({
-  AntInputComponent: VirtSelect,
-  onChangeAdapter: option => (option ? (option as any).value : null),
```
yurishkuro, line 83: why is this property no longer defined when you use `reduxFormFieldAdapter`?

packages/jaeger-ui/src/components/Monitor/ServicesView/index.tsx
yurishkuro: what happens to `required` now? Previously the form could not be submitted without required selections.

packages/jaeger-ui/src/components/Monitor/ServicesView/index.tsx
prathamesh-mutkure: As per my understanding, that adapter was for components other than Ant, where `onChange` was handled differently 

But since we're using Ant's Select, I don't think we need to override the `onChange` callback with `onChangeAdapter` prop

This might be the reason `onChangeAdapter` was omitted in [SearchResults](https://github.com/jaegertracing/jaeger-ui/blob/main/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.tsx#L76) too while using it with Ant's Select

packages/jaeger-ui/src/components/Monitor/ServicesView/index.tsx
prathamesh-mutkure: Actually, the implementation of the `Select` element was such that we don't really require the `required` prop

In the `Monitor/ServicesView`, the `Select` component isn't really wrapped inside a form and thus is not being submitted, it's not getting submitted anywhere, we just change the data in charts based on the value of Select

The Select just chooses the first option as the default value, so it's not possible for it to remain empty unless there's no data or service running, in that case, it is not even visible to the user

prathamesh-mutkure: In cases where we are actually submitting the form, for example in the `SearchForm` component, the UI looks something like this when there is no data or services running - 

Screenshot 2023-08-28 at 4 42 10 AM

We just disable the Select as well as the Submit button, and if there were services running, it would just pick the first one as default

packages/jaeger-ui/src/components/SearchTracePage/SearchForm.jsx
```diff
@@ -287,15 +283,18 @@ export class SearchFormImpl extends React.PureComponent {
         >
           <Field
             name="service"
-            component={AdaptedVirtualSelect}
+            component={AdaptedSelect}
```
yurishkuro, line 286: any idea why Service / Operation dropdowns allow only 7 entries simultaneously visible, but Lookback allows 8?

yurishkuro (approved):

packages/jaeger-ui/src/components/SearchTracePage/SearchForm.jsx
prathamesh-mutkure: That's interesting, I'll check this out

yurishkuro:
@prathamesh-mutkure could you please check #1784, if there is a way to re-enable type-to-search capability in Ant Select?

yurishkuro:
cc @prathamesh-mutkure

PR #1738, opened 2023-08-28
Moment.js replaced with Dayjs

### LINKED ISSUE #1736: [deps] Replace the use of moment.js with actively maintained library
Moment.js is "a legacy project in maintenance mode", according to its official website https://momentjs.com/docs/#/-project-status/.

* [ ] do an inventory of features / APIs we are using from moment.js
* [ ] do analysis which replacement is most suitable (the fewer dependencies the better)
* [ ] make the corresponding changes. It's probably worth consolidating all usage of any new 3rd party library into `utils/date.tsx`

```
$ grx "from 'moment'" packages/jaeger-ui/src
packages/jaeger-ui/src/utils/date.tsx:15:import moment, { unitOfTime } from 'moment';
packages/jaeger-ui/src/components/common/RelativeDate.tsx:15:import moment from 'moment';
packages/jaeger-ui/src/components/TracePage/TraceSpanView/index.tsx:17:import moment from 'moment';
packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ScatterPlot.jsx:16:import moment from 'moment';
packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ResultItem.tsx:21:import moment from 'moment';
packages/jaeger-ui/src/components/SearchTracePage/SearchForm.jsx:20:import moment from 'moment';
packages/jaeger-ui/src/components/SearchTracePage/SearchForm.test.js:20:import moment from 'moment';
packages/jaeger-ui/src/api/jaeger.js:16:import moment from 'moment';

$ grx "from 'moment'" packages/jaeger-ui/src packages/plexus/src
packages/jaeger-ui/src/utils/date.tsx:15:import moment, { unitOfTime } from 'moment';
packages/jaeger-ui/src/components/common/RelativeDate.tsx:15:import moment from 'moment';
packages/jaeger-ui/src/components/TracePage/TraceSpanView/index.tsx:17:import moment from 'moment';
packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ScatterPlot.jsx:16:import moment from 'moment';
packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ResultItem.tsx:21:import moment from 'moment';
packages/jaeger-ui/src/components/SearchTracePage/SearchForm.jsx:20:import moment from 'moment';
packages/jaeger-ui/src/components/SearchTracePage/SearchForm.test.js:20:import moment from 'moment';
packages/jaeger-ui/src/api/jaeger.js:16:import moment from 'moment';
```

prathamesh-mutkure:
- Resolves #1736 

## Description of the changes
- I've removed `moment.js` and replaced it with `dayjs` packages
- I've fixed all the failing tests
- Tested all dates and time formatting

- All the test cases are passing
- All the pages that got impacted were tested visually by me
- Including the `Monitor` tab, `Span` and `Trace` view, `Search` page, all the charts, etc.

The UI was checked between `jaeger-all-in-one` and changes of this commit

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/SearchTracePage/SearchForm.test.js
+++ b/packages/jaeger-ui/src/components/SearchTracePage/SearchForm.test.js
@@ -87,19 +87,16 @@ describe('conversion utils', () => {
 
   describe('convertQueryParamsToFormDates()', () => {
     it('converts correctly', () => {
-      const startDayjs = dayjs().subtract(1, 'day');
-      const endDayjs = dayjs();
-      const params = {
-        start: `${startDayjs.valueOf()}000`,
-        end: `${endDayjs.valueOf()}000`,
-      };
-
       const { queryStartDate, queryStartDateTime, queryEndDate, queryEndDateTime } =
-        convertQueryParamsToFormDates(params);
-      expect(queryStartDate).toBe(startDayjs.format(DATE_FORMAT));
-      expect(queryStartDateTime).toBe(startDayjs.format(TIME_FORMAT));
-      expect(queryEndDate).toBe(endDayjs.format(DATE_FORMAT));
-      expect(queryEndDateTime).toBe(endDayjs.format(TIME_FORMAT));
+        convertQueryParamsToFormDates({
+          start: '946720800000000', // Jan 1, 2000 10:00 AM
+          end: '946807200000000', // Jan 2, 2000 10:00 AM
+        });
+
+      expect(queryStartDate).toBe('2000-01-01');
+      expect(queryStartDateTime).toBe('10:00');
+      expect(queryEndDate).toBe('2000-01-02');
+      expect(queryEndDateTime).toBe('10:00');
     });
   });
 
--- a/packages/jaeger-ui/src/components/SearchTracePage/SearchForm.test.js
+++ b/packages/jaeger-ui/src/components/SearchTracePage/SearchForm.test.js
@@ -17,7 +17,7 @@ jest.mock('store');
 
 import React from 'react';
 import { shallow } from 'enzyme';
-import moment from 'moment';
+import dayjs from 'dayjs';
 import queryString from 'query-string';
 import store from 'store';
 
@@ -87,19 +87,19 @@ describe('conversion utils', () => {
 
   describe('convertQueryParamsToFormDates()', () => {
     it('converts correctly', () => {
-      const startMoment = moment().subtract(1, 'day');
-      const endMoment = moment();
+      const startDayjs = dayjs().subtract(1, 'day');
+      const endDayjs = dayjs();
       const params = {
-        start: `${startMoment.valueOf()}000`,
-        end: `${endMoment.valueOf()}000`,
+        start: `${startDayjs.valueOf()}000`,
+        end: `${endDayjs.valueOf()}000`,
       };
 
       const { queryStartDate, queryStartDateTime, queryEndDate, queryEndDateTime } =
         convertQueryParamsToFormDates(params);
-      expect(queryStartDate).toBe(startMoment.format(DATE_FORMAT));
-      expect(queryStartDateTime).toBe(startMoment.format(TIME_FORMAT));
-      expect(queryEndDate).toBe(endMoment.format(DATE_FORMAT));
-      expect(queryEndDateTime).toBe(endMoment.format(TIME_FORMAT));
+      expect(queryStartDate).toBe(startDayjs.format(DATE_FORMAT));
+      expect(queryStartDateTime).toBe(startDayjs.format(TIME_FORMAT));
+      expect(queryEndDate).toBe(endDayjs.format(DATE_FORMAT));
+      expect(queryEndDateTime).toBe(endDayjs.format(TIME_FORMAT));
     });
   });
 
@@ -298,7 +298,7 @@ describe('submitForm()', () => {
     function getCalledDuration(mock) {
       const { start, end } = mock.calls[0][0];
       const diffMs = (Number(end) - Number(start)) / 1000;
-      return moment.duration(diffMs);
+      return dayjs.duration(diffMs);
     }
 
     it('subtracts `lookback` from `fields.end`', () => {
--- a/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ResultItem.tsx
+++ b/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ResultItem.tsx
@@ -60,7 +60,7 @@ export default class ResultItem extends React.PureComponent<Props, State> {
     super(props, state);
     const { startTime, spans } = props.trace;
 
-    const dayjsDate = dayjs(startTime / 1000);
+    const startTimeDayjs = dayjs(startTime / 1000);
 
     const erroredServices: Set<string> = new Set<string>();
 
@@ -74,8 +74,8 @@ export default class ResultItem extends React.PureComponent<Props, State> {
 
     this.state = {
       numSpans: spans.length,
-      timeStr: dayjsDate.format('h:mm:ss a'),
-      fromNow: dayjsDate.fromNow(),
+      timeStr: startTimeDayjs.format('h:mm:ss a'),
+      fromNow: startTimeDayjs.fromNow(),
       numErredSpans,
       erroredServices,
     };
--- a/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ResultItem.tsx
+++ b/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ResultItem.tsx
@@ -18,7 +18,8 @@ import { LocationDescriptor } from 'history';
 import { Link } from 'react-router-dom';
 
 import { sortBy } from 'lodash';
-import moment from 'moment';
+import dayjs from 'dayjs';
+import relativeTime from 'dayjs/plugin/relativeTime';
 
 import { IoAlert } from 'react-icons/io5';
 
@@ -31,6 +32,8 @@ import { formatRelativeDate } from '../../../utils/date';
 import { KeyValuePair, Trace } from '../../../types/trace';
 
 import './ResultItem.css';
+
+dayjs.extend(relativeTime);
 
 type Props = {
   durationPercent: number;
@@ -57,7 +60,7 @@ export default class ResultItem extends React.PureComponent<Props, State> {
     super(props, state);
     const { startTime, spans } = props.trace;
 
-    const mDate = moment(startTime / 1000);
+    const dayjsDate = dayjs(startTime / 1000);
 
     const erroredServices: Set<string> = new Set<string>();
 
@@ -71,8 +74,8 @@ export default class ResultItem extends React.PureComponent<Props, State> {
 
     this.state = {
       numSpans: spans.length,
-      timeStr: mDate.format('h:mm:ss a'),
-      fromNow: mDate.fromNow(),
+      timeStr: dayjsDate.format('h:mm:ss a'),
+      fromNow: dayjsDate.fromNow(),
       numErredSpans,
       erroredServices,
     };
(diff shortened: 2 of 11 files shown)


packages/jaeger-ui/src/utils/date.tsx
```diff
@@ -158,12 +162,12 @@ export const getSuitableTimeUnit = (microseconds: number): string => {
     return 'microseconds';
   }
 
-  const duration = moment.duration(microseconds / 1000, 'ms');
+  const duration = dayjs.duration(microseconds / 1000, 'ms');
 
   return Object.keys(timeUnitToShortTermMapper)
     .reverse()
```
yurishkuro, line 168: `timeUnitToShortTermMapper` keys are just unit names, why is `reverse` meaningful for them? I expect the order of object keys not to be guaranteed in JavaScript. If we need strong order then it should be an array, not a dict.

yurishkuro (commented):
can you add more tests to raise util/date coverage to 100%?

mmorel-35:
I was on antd migration guide from v3 to v4 and I noticed, they provided [a guide](https://4x.ant.design/docs/react/replace-moment) to migrate from moment to dayjs. Have you used it to achieve this ?

packages/jaeger-ui/src/components/SearchTracePage/SearchForm.test.js
```diff
@@ -87,19 +87,19 @@ describe('conversion utils', () => {
...
     it('converts correctly', () => {
-      const startMoment = moment().subtract(1, 'day');
-      const endMoment = moment();
+      const startDayjs = dayjs().subtract(1, 'day');
+      const endDayjs = dayjs();
       const params = {
-        start: `${startMoment.valueOf()}000`,
-        end: `${endMoment.valueOf()}000`,
+        start: `${startDayjs.valueOf()}000`,
+        end: `${endDayjs.valueOf()}000`,
```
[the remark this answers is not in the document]
prathamesh-mutkure: Yes, ideally that's how it should have been, I'll re-write the tests

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ResultItem.tsx
```diff
@@ -57,7 +60,7 @@ export default class ResultItem extends React.PureComponent<Props, State> {
     super(props, state);
     const { startTime, spans } = props.trace;
 
-    const mDate = moment(startTime / 1000);
+    const dayjsDate = dayjs(startTime / 1000);
```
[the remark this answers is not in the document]
prathamesh-mutkure: Sure, I'll fix that

prathamesh-mutkure:
> I was on antd migration guide from v3 to v4 and I noticed, they provided [a guide](https://4x.ant.design/docs/react/replace-moment) to migrate from moment to dayjs. Have you used it to achieve this ?

Hi @mmorel-35, the guide you mentioned is specifically for Ant Design component, this is not specific to moment to dayjs guide and thus not relevant to our scenario

packages/jaeger-ui/src/utils/date.tsx
prathamesh-mutkure: Yes, even I got confused due to this, I'll see which approach is better

packages/jaeger-ui/src/utils/date.tsx
```diff
@@ -158,12 +162,12 @@ export const getSuitableTimeUnit = (microseconds: number): string => {
...
     .find(timeUnit => {
-      const durationInTimeUnit = duration.as(timeUnit as unitOfTime.Base);
+      const durationInTimeUnit = duration.as(timeUnit as DurationUnitType);
```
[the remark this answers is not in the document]
prathamesh-mutkure: Yes, that will be much safer code

packages/jaeger-ui/src/utils/date.tsx
```diff
@@ -100,7 +103,7 @@ export function formatMillisecondTime(duration: number) {
  */
 export function formatSecondTime(duration: number) {
```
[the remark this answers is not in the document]
prathamesh-mutkure: Doesn't typescript automatically infer type in this case, as we are returning a string at the end? Do we need explicitly mentioned return type?

prathamesh-mutkure:
I forgot to add more unit tests, let me do it real quick

packages/jaeger-ui/src/utils/date.tsx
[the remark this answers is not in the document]
yurishkuro: declaring types greatly improves readability, especially when you mouse over the call to this function and only see a signature in the IDE tooltip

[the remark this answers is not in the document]
prathamesh-mutkure: Makes sense, I've added explicit return types for all functions in date utils

packages/jaeger-ui/src/utils/date.tsx
```diff
@@ -40,67 +43,73 @@ const UNIT_STEPS: { unit: string; microseconds: number; ofPrevious: number }[] =
...
+export function formatDatetime(duration: number): string {
+  return dayjs(duration / ONE_MILLISECOND).format(STANDARD_DATETIME_FORMAT);
 }
 
 /**
  * @param {number} duration (in microseconds)
  * @return {string} formatted, unit-labelled string with time in milliseconds
  */
-export function formatMillisecondTime(duration: number) {
+export function formatMillisecondTime(duration: number): string {
```
yurishkuro, line 101: Please add tests.

Side note: Many of these functions feel poorly conceived to me. Like `formatDate()` on L77 that takes `duration` as input - I don't really understand that that means, is it interpreting duration as offset from epoch to calculate the date and then format it? Comments could use improvement by showing examples.

yurishkuro:
@prathamesh-mutkure please add missing tests to have patch coverage 100%

packages/jaeger-ui/src/utils/date.tsx
```diff
@@ -40,67 +43,100 @@ const UNIT_STEPS: { unit: string; microseconds: number; ofPrevious: number }[] =
...
   return (duration / totalDuration) * 100;
 }
 
-const quantizeDuration = (duration: number, floatPrecision: number, conversionFactor: number) =>
+const quantizeDuration = (duration: number, floatPrecision: number, conversionFactor: number): number =>
   toFloatPrecision(duration / conversionFactor, floatPrecision) * conversionFactor;
 
 /**
- * @param {number} duration (in microseconds)
+ * @param {number} duration - number of microseconds elapsed since midnight, January 1, 1970 Universal Coordinated Time (UTC).
```
yashrsharma44, line 74: Some nits - 
You can also replace the comment with unix-time, as it is a well established standard for time.

packages/jaeger-ui/src/components/SearchTracePage/SearchForm.test.js
```diff
@@ -87,19 +87,16 @@ describe('conversion utils', () => {
...
-
       const { queryStartDate, queryStartDateTime, queryEndDate, queryEndDateTime } =
-        convertQueryParamsToFormDates(params);
-      expect(queryStartDate).toBe(startMoment.format(DATE_FORMAT));
-      expect(queryStartDateTime).toBe(startMoment.format(TIME_FORMAT));
-      expect(queryEndDate).toBe(endMoment.format(DATE_FORMAT));
-      expect(queryEndDateTime).toBe(endMoment.format(TIME_FORMAT));
+        convertQueryParamsToFormDates({
+          start: '946720800000000', // Jan 1, 2000 10:00 AM
+          end: '946807200000000', // Jan 2, 2000 10:00 AM
```
yashrsharma44, line 93: +1 on the comment description, helps people in understanding the test-case we're using.

packages/jaeger-ui/src/utils/date.test.js
```diff
@@ -218,3 +227,33 @@ describe('formatRelativeDate', () => {
...
+
+  it('formatDateTime formats microseconds to standard date format', () => {
+    const dateInMicroseconds = dateInMilliseconds * ONE_MILLISECOND;
+    expect(formatDatetime(dateInMicroseconds)).toBe('January 1 2000, 10:00:00.000');
+  });
+
+  it('formatMillisecondTime formats microseconds to milliseconds', () => {
+    const durationInMicroseconds = 1000 * ONE_MILLISECOND;
+    expect(formatMillisecondTime(durationInMicroseconds)).toBe('1000ms');
+  });
```
yashrsharma44, line 253:
```suggestion
  it('formatMillisecondTime formats microseconds to milliseconds', () => {
    expect(formatMillisecondTime(ONE_MICROSECOND)).toBe('1000ms');
  });
```
nit : let's define a constant for the microsecond?

yashrsharma44 (approved):
LGTM!

yashrsharma44:
Can you rebase your changes with the latest main branch?

packages/jaeger-ui/src/utils/date.test.js
prathamesh-mutkure: Actually, microsecond is the base unit we're using, so `ONE_MICROSECOND` will essentially be just `1`, I think it should increase code readability, adding it

packages/jaeger-ui/src/utils/date.tsx
```diff
@@ -27,7 +27,8 @@ const YESTERDAY = 'Yesterday';
 export const STANDARD_DATE_FORMAT = 'YYYY-MM-DD';
 export const STANDARD_TIME_FORMAT = 'HH:mm';
 export const STANDARD_DATETIME_FORMAT = 'MMMM D YYYY, HH:mm:ss.SSS';
-export const ONE_MILLISECOND = 1000;
+export const ONE_MICROSECOND = 1;
+export const ONE_MILLISECOND = 1000 * ONE_MICROSECOND;
```
yurishkuro, line 31: tbh, I don't find these constants intuitive at all. In contrast, in Go stdlib there are similar constants which have a clear semantic meaning because Go counts all times relative to 1ns granularity, so `const Millisecond=1_000_000` makes sense. Here I don't know what these constants mean, because the baseline is only implied. E.g.:
```
ScatterPlot.jsx:21:import { ONE_MILLISECOND } from '../../../utils/date';
```
wut?

yurishkuro: Separately, ONE_MICROSECOND is not used anywhere in the codebase, so why export it?

yurishkuro, line 31:
```suggestion
// 1ms as the number of microseconds, which is the precision of Jaeger timestamps
export const ONE_MILLISECOND = 1000 * ONE_MICROSECOND;
```

prathamesh-mutkure: The base unit for the time in jaeger-ui is microsecond, those constants represent the time unit expressed in microsecond

ONE_MICROSECOND = 1 microsecond
ONE_MILLISECOND = 1000 microsecond 
ONE_SECOND = 1000_000 microsecond and so on

I've exported the ONE_MICROSECOND constant just in case we need it in the future, using `ONE_MICROSECOND` is more readable and concise than just writing `1` as a literal

prathamesh-mutkure: Added JSDoc comment for all constants

yurishkuro: ONE_MICROSECOND is not more readable, because it's meaningless without context. The context being "this is the value that represents 1ms as the amount of _default units_", where default units are not specified and instead implied. For example, say you see a code using a constant ONE_APPLE - what does it tell you? It could be weight in grams, it could be amount of calories - we don't know. It's the same story with these constants.

But even if they were meaningful, that's no justification for adding code and exporting constants that are not elsewhere used.  If/when they are used, we can add them.

prathamesh-mutkure: Oh got it, should I better remove it for now?

yurishkuro: yes

Issue #1703, opened 2023-08-20
Upgrade Antd to v4


yurishkuro:
Creating a tracking issue since this is not a straightforward upgrade.

ChaudharyRaman:
Hi @yurishkuro, I want to work on this issue. Can you provide me more details.

yurishkuro:
See https://github.com/jaegertracing/jaeger-ui/pull/1636 for more discussion

PR #1748, opened 2023-09-01
Upgrade antd to v4.15.6

anshgoyalevil:
- part of: https://github.com/jaegertracing/jaeger-ui/pull/1734
- This PR is a part of upgrading antd to latest version of 4.x

## Description of the changes
- Disabled the NavBar hover effect which was un-necessarily introduced
- NavBar's height was increased internally. Fixed it to match previous height (48px)
- The Monitor -> EmptyPage was using the offsets to align content to the center, but it is not required now, as it is a default behavior
- `SorterResult` API is removed, so used lodash in place of that
- `CompareFn` typedef is no longer required
- Mocked the `window.matchMedia`, since some tests were using it
- `dataIndex` now requires an array instead of nested string. Ex: `"abc.def.ghi"` -> `["abc", "def", "ghi"]`
- Tests are updated according to these changes
- Snapshots are updated according to internal changes in components

- Manually, and using automated testing

## What's next?
- In the next PRs, the antd would be upgraded to 4.20.0 (or higher), to minimise the changes in a single PR.
- New type errors would be introduced thereby, like NumberInput lookback changes, etc.

## Checklist


packages/jaeger-ui/src/components/Monitor/EmptyState/index.tsx
```diff
@@ -32,7 +32,7 @@ export default class MonitorATMEmptyState extends React.PureComponent {
     return (
       <Col>
         <Row justify="center">
-          <Col span={8} offset={8}>
```
yurishkuro, line 35: doesn't removing `offset` change the position of the column? When I see `<Col span={8} offset={8}>` I interpret it "start from column 8 and span columns 8-15"

packages/jaeger-ui/src/components/Monitor/ServicesView/operationDetailsTable/index.tsx
```diff
@@ -235,12 +242,18 @@ export class OperationTableDetails extends React.PureComponent<TProps, TState> {
...
-          onChange={(pagination, filters, { columnKey, order }) => {
-            if (!isEqual({ columnKey, order }, this.state.tableSorting)) {
-              const clickedColumn = tableTitles.get(columnKey || this.state.tableSorting.columnKey);
+          onChange={(pagination, filters, sorter) => {
+            const activeSorters = isArray(sorter) ? sorter : [sorter];
+            const { tableSorting } = this.state;
+
+            if (!isEqual(activeSorters, tableSorting)) {
+              const lastColumn =
+                activeSorters[activeSorters.length - 1] ?? tableSorting[tableSorting.length - 1];
```
yurishkuro, line 251: if both arrays are empty, isn't it possible to end up with `lastColumn = undefined`?

packages/jaeger-ui/src/components/Monitor/ServicesView/operationDetailsTable/index.tsx
```diff
@@ -235,12 +242,18 @@ export class OperationTableDetails extends React.PureComponent<TProps, TState> {
               },
             };
           }}
-          onChange={(pagination, filters, { columnKey, order }) => {
-            if (!isEqual({ columnKey, order }, this.state.tableSorting)) {
-              const clickedColumn = tableTitles.get(columnKey || this.state.tableSorting.columnKey);
+          onChange={(pagination, filters, sorter) => {
```
yurishkuro, line 245: what specifically is changing in antd that requires these modifications, which look more complicated than before?

packages/jaeger-ui/src/components/SearchTracePage/FileLoader.tsx
```diff
@@ -24,7 +24,11 @@ type FileLoaderProps = {
 
 export default function FileLoader(props: FileLoaderProps) {
   return (
-    <Dragger accept=".json" customRequest={props.loadJsonTraces} multiple>
+    <Dragger
+      accept=".json"
+      beforeUpload={(file, fileList) => fileList.forEach(file => props.loadJsonTraces({ file }))}
```
yurishkuro, line 29: maybe move this to a function and add unit test just for it?

packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageSearchBar.test.js
```diff
@@ -61,16 +61,16 @@ describe('<TracePageSearchBar>', () => {
...
+      expect(wrapper.find('[data-testid="DownOutlined"]').prop('onClick')).toBe(defaultProps.nextResult);
+      expect(wrapper.find('[data-testid="CloseOutlined"]').prop('onClick')).toBe(defaultProps.clearSearch);
     });
 
     it('hides navigation buttons when not navigable', () => {
       wrapper.setProps({ navigable: false });
       const button = wrapper.find('Button');
       expect(button.length).toBe(1);
-      expect(button.prop('icon')).toBe('close');
+      expect(button.getElement(CloseOutlined)).toBeDefined();
```
yurishkuro, line 73: can you not look it up by data-testid to avoid explicit dependency on ant icons?

packages/jaeger-ui/vite.config.ts
```diff
@@ -83,7 +83,7 @@ export default defineConfig({
           '@layout-body-background': '#fff',
           '@layout-header-background': '#404040',
           '@layout-footer-background': '@layout-body-background',
-          '@layout-header-height': '64px',
+          '@layout-header-height': '48px',
```
yurishkuro, line 86: please add comments to all css changes to explain why they are necessary, so that we have a record. Otherwise it's very difficult to judge if the change is necessary or not.

packages/jaeger-ui/src/components/TracePage/TraceSpanView/index.tsx
```diff
@@ -133,7 +133,7 @@ export default class TraceSpanView extends Component<Props, State> {
     const columns: ColumnProps<Span>[] = [
       {
         title: 'Service Name',
-        dataIndex: 'process.serviceName',
+        dataIndex: ['process', 'serviceName'],
```
yurishkuro, line 136: is this our own notation or something that antd component supports, like a nested indexing path into a data structure?

packages/jaeger-ui/src/components/TracePage/TraceSpanView/index.tsx
```diff
@@ -205,7 +205,11 @@ export default class TraceSpanView extends Component<Props, State> {
                 }}
               >
                 {this.state.serviceNamesList.map(name => {
-                  return <Option key={name}>{name} </Option>;
+                  return (
+                    <Option value={name} key={name}>
+                      {name}{' '}
```
yurishkuro, line 210: just curious, what's the need for the extra space in the display string?

packages/jaeger-ui/src/components/TracePage/TraceSpanView/index.test.js
```diff
@@ -38,7 +38,7 @@ describe('<TraceSpanView>', () => {
   it('does not explode', () => {
     expect(wrapper).toBeDefined();
     expect(wrapper.find('.title--TraceSpanView').length).toBe(1);
-    expect(wrapper.find('.span-view-table').length).toBe(3);
+    expect(wrapper.find('.span-view-table').length).toBe(2);
```
yurishkuro, line 41: why is this count changing?

packages/jaeger-ui/src/components/Monitor/EmptyState/index.tsx
anshgoyalevil: Previously (in antd 3x), the `<Col span={8}>` was interpreted as, "The width of column should be 8 boxes out of the 24 boxes in a grid".
That `offset` property defines that those 8 boxes must start from the 9th box out of those 24 boxes so that basically centers the col.

In antd 4x, this behavior is controlled by `<Row justify="center">`. Don't know why it was not working previously in 3x, even though it existed before too.

Here's what the page looks like now:

image

packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageSearchBar.test.js
anshgoyalevil: Yes, maybe it could be done that way too. I will try and update.

packages/jaeger-ui/src/components/TracePage/TraceSpanView/index.tsx
anshgoyalevil: antd `dataIndex` definition changes from `'xxx.yyy' to ['xxx', 'yyy']`.
So basically, antd now want us to supply dataIndex path in the form of an array for nested indexing pathnames.

Here's a link to that: https://4x.ant.design/docs/react/migration-v4#:~:text=Nest%20dataIndex%20definition%20changes%20from%20%27xxx.yyy%27%20to%20%5B%27xxx%27%2C%20%27yyy%27%5D

packages/jaeger-ui/src/components/TracePage/TraceSpanView/index.tsx
anshgoyalevil: Previously it was `return <Option key={name}>{name} </Option>;` (Notice an extra space after `{name}`).
So, on running `yarn fmt`, prettier detected that maybe that space is needed, and formatted it that way (`{' '}`)

packages/jaeger-ui/src/components/Monitor/ServicesView/operationDetailsTable/index.tsx
anshgoyalevil: In antd 4x, sorting by multiple sortable columns now reports all columns that are active in sorting, so the event handling code must also accept an array.

packages/jaeger-ui/src/components/Monitor/ServicesView/operationDetailsTable/index.tsx
anshgoyalevil: But, I think, in that case, the onChange function wouldn't be called.
Whenever someone clicks the table header for sorting, the tableSorting or activeSorters would always return an array of size greater than or equal to 1.

packages/jaeger-ui/src/components/SearchTracePage/FileLoader.tsx
anshgoyalevil: Sure. Would try adding a test for it, if possible

packages/jaeger-ui/src/components/App/Page.css
```diff
@@ -34,3 +34,7 @@ limitations under the License.
 .Page--content--no-embedded {
   top: var(--nav-height);
 }
+
+.ant-menu-item:hover {
+  background-color: transparent !important;
+}
```
anshgoyalevil, line 40: In antd 4x, the TopNav contains the hover effect by default, which is not what we had earlier.
This CSS property disables that behavior.

packages/jaeger-ui/src/components/Monitor/ServicesView/index.css
```diff
@@ -33,8 +33,6 @@ limitations under the License.
 .timeframe-selector {
   display: inline-flex;
   justify-content: flex-end;
-  position: absolute;
-  bottom: 14px;
```
anshgoyalevil, line 37: The position of the Timeframe Selector needs to be relative to the Col beside it.

Due to `position: absolute`, the Selector was going to the bottom left corner, which is not the desired behavior.

`bottom: 14px` is removed because it is already at an appropriate distance from the col below it.

In Antd 3x:


With this PR:
image

packages/jaeger-ui/src/components/TracePage/TracePageHeader/KeyboardShortcutsHelp.css
```diff
@@ -16,7 +16,7 @@ limitations under the License.
 
 .KeyboardShortcutsHelp--cta {
   font-size: 1.2rem;
-  margin: 0 -7px;
+  margin: -3px -7px;
```
anshgoyalevil, line 19: This is to center the Keyboard shortcut panel icon vertically.

packages/jaeger-ui/vite.config.ts
anshgoyalevil, line 86: The Header's height was increased with 4.x upgrade. The resulting NavBar wasn't the one we had earlier. To revert this internal change, the height has to be decreased here.

packages/jaeger-ui/src/components/TracePage/TracePageHeader/KeyboardShortcutsHelp.css
anshgoyalevil: Without this:
image


With this:
image

packages/jaeger-ui/src/components/SearchTracePage/FileLoader.tsx
anshgoyalevil: Done!

packages/jaeger-ui/src/components/TracePage/TraceSpanView/index.test.js
anshgoyalevil: Since the re-write of the Table in 4.x, there doesn't exist an extra `withStore(Table)` dom element with the class name `span-view-table`.

yurishkuro:
Code coverage drop is due to this:
image
Is this a flaky test or the result of the upgrade?

packages/jaeger-ui/src/components/Monitor/EmptyState/index.tsx
```diff
@@ -59,7 +59,7 @@ export default class MonitorATMEmptyState extends React.PureComponent {
             )}
             {this.config.alert && (
               <Row justify="center">
```
yurishkuro, line 61: is this correct that we have a row inside a row?

anshgoyalevil: As far as I am seeing, it's giving the expected result. Not sure 100% though, as there is not a strict guide about nested row and cols.

yurishkuro (approved):

anshgoyalevil:
> Code coverage drop is due to this: image Is this a flaky test or the result of the upgrade?

Don't know why but its giving 100 % coverage locally.

image

The issue's is not with the test, but the codecov. I have seen this behaviour in many places.

yurishkuro:
🥳 🥳 🥳 🥳 🥳 🥳 🥳 🥳 🥳 🥳 🥳 🥳 🥳 🥳 🥳 🥳 🥳 🥳 🥳 🥳 🥳

yurishkuro:
I rebased https://github.com/jaegertracing/jaeger-ui/pull/1636, which bumps antd even further - may fail on other issues

packages/jaeger-ui/src/components/App/Page.css
yurishkuro: don't understand this one. The About Jaeger menu also self-activates on hover.

anshgoyalevil:
> I rebased #1636, which bumps antd even further - may fail on other issues

Yes, antd requires another manual bump. The dependabot's PR would fail for now.

packages/jaeger-ui/src/components/App/Page.css
anshgoyalevil: > The About Jaeger menu also self-activates on hover.

I just checked the live build on the previous version (3.x) without any change. It showed the same behavior there too.
If that was a bug, I would fix it in next PR.

About that CSS change, here's how it behaves without it:

[screen-capture.webm](https://github.com/jaegertracing/jaeger-ui/assets/94157520/1dd49920-195a-4433-8d68-122d71fd5990)

yurishkuro: lgtm

yurishkuro:
part of #1703

PR #1750, opened 2023-09-03
Upgrade antd to v4.16.0 from v4.15.6

anshgoyalevil:
- part of the upgrade process of ant-design
- related to: https://github.com/jaegertracing/jaeger-ui/pull/1734

## Description of the changes
- This PR fixes NavBar's dropdown "About Jaeger" which was changed due to some internal changes in antd v4.16.0

- Manually, and automated testing

## What's Next?
- antd would be upgraded to latest version of v4.x

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/App/TopNav.tsx
+++ b/packages/jaeger-ui/src/components/App/TopNav.tsx
@@ -96,12 +96,7 @@ function getItem(item: ConfigMenuItem) {
 function CustomNavDropdown({ label, items }: ConfigMenuGroup) {
   const menuItems = <Menu>{items.map(getItem)}</Menu>;
   return (
-    <Dropdown
-      trigger={['click']}
-      overlay={menuItems}
-      overlayStyle={{ paddingRight: '20px' }}
-      placement="bottomRight"
-    >
+    <Dropdown overlay={menuItems} overlayStyle={{ paddingRight: '20px' }} placement="bottomRight">
       <a style={{ color: 'white' }}>
         {label} <DownOutlined style={{ paddingRight: '30px' }} />
       </a>
--- a/packages/jaeger-ui/src/components/App/TopNav.tsx
+++ b/packages/jaeger-ui/src/components/App/TopNav.tsx
@@ -96,9 +96,14 @@ function getItem(item: ConfigMenuItem) {
 function CustomNavDropdown({ label, items }: ConfigMenuGroup) {
   const menuItems = <Menu>{items.map(getItem)}</Menu>;
   return (
-    <Dropdown overlay={menuItems} placement="bottomRight">
-      <a>
-        {label} <DownOutlined />
+    <Dropdown
+      trigger={['click']}
+      overlay={menuItems}
+      overlayStyle={{ paddingRight: '20px' }}
+      placement="bottomRight"
+    >
+      <a style={{ color: 'white' }}>
+        {label} <DownOutlined style={{ paddingRight: '30px' }} />
       </a>
     </Dropdown>
   );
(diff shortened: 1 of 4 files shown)


packages/jaeger-ui/src/components/App/TopNav.tsx
```diff
@@ -96,9 +96,14 @@ function getItem(item: ConfigMenuItem) {
 function CustomNavDropdown({ label, items }: ConfigMenuGroup) {
   const menuItems = <Menu>{items.map(getItem)}</Menu>;
   return (
-    <Dropdown overlay={menuItems} placement="bottomRight">
-      <a>
-        {label} <DownOutlined />
+    <Dropdown
+      trigger={['click']}
```
anshgoyalevil, line 100: Previously, "About Jaeger" dropdown was used to trigger automatically on hover. This attribute changes that behavior to trigger on click.

packages/jaeger-ui/src/components/App/TopNav.tsx
```diff
@@ -96,9 +96,14 @@ function getItem(item: ConfigMenuItem) {
...
+      overlay={menuItems}
+      overlayStyle={{ paddingRight: '20px' }}
```
anshgoyalevil, line 102: This gives the overlay of dropdown (The panel that opens when someone clicks the About Jaeger Menu item), a padding. It was changed after the upgrade to v4.16.0

packages/jaeger-ui/src/components/App/TopNav.tsx
```diff
@@ -96,9 +96,14 @@ function getItem(item: ConfigMenuItem) {
...
+      placement="bottomRight"
+    >
+      <a style={{ color: 'white' }}>
```
anshgoyalevil, line 105: To change the color of `About Jaeger` from default green to white

packages/jaeger-ui/src/components/App/TopNav.tsx
```diff
@@ -96,9 +96,14 @@ function getItem(item: ConfigMenuItem) {
...
+        {label} <DownOutlined style={{ paddingRight: '30px' }} />
```
anshgoyalevil, line 106: To give some space after the down arrow icon

packages/jaeger-ui/src/components/App/TopNav.tsx
```diff
@@ -116,18 +121,14 @@ export function TopNavImpl(props: Props) {
   return (
     <div>
       <Menu theme="dark" mode="horizontal" selectable={false} className="ub-right" selectedKeys={[pathname]}>
-        <Menu.Item>
+        <Menu.Item style={{ paddingRight: '40px' }}>
```
anshgoyalevil, line 124: To give some space after the search form, since it was changed after the upgrade to v4.16.0

anshgoyalevil:
Previous:
prev

Now:
now

packages/jaeger-ui/src/components/App/TopNav.tsx
yurishkuro: Why do we want to change it like that?

anshgoyalevil: Should I revert this `trigger` change? I think we had a conversation before about this behavior.

> The About Jaeger menu also self-activates on hover.
https://github.com/jaegertracing/jaeger-ui/pull/1748#discussion_r1313883548

yurishkuro: We should always try to keep the existing behavior. It's not just the menu, the trace view selector also activates on hover, so it's a pattern.

anshgoyalevil: Got it 🚀

packages/jaeger-ui/src/components/Monitor/ServicesView/operationDetailsTable/__snapshots__/index.test.js.snap
anshgoyalevil, line 1: antd community is changing things rapidly with every upgrade. Seems like we need to stop dependabot from its automatic bumps.

yurishkuro: I'd rather say we need to fix our tests so that they don't depend on implementation details. Depebdabot bumps are a very good way to see breaking changes.

packages/jaeger-ui/src/components/App/TopNav.tsx
yurishkuro: I don't think it's a good idea to override styles inline, it makes it impossible to provide skinning capability in the future (not that we need fully customizable skinning, but at least light mode / dark mode is a common feature these days).

yurishkuro (approved):

packages/jaeger-ui/src/components/Monitor/ServicesView/operationDetailsTable/__snapshots__/index.test.js.snap
anshgoyalevil: Agree with that.

yurishkuro:
part of #1703

PR #1756, opened 2023-09-04
Split lint and test CI and suppress warnings

yurishkuro:
- Sometimes we get the `coverage` step failures that are very hard to diagnose because the logs are spammed by warnings, while the real errors are not showing (for some reason local test runs look differently)
- We also run linters and coverage sequentially, instead of in parallel, to speed things up

## Description of the changes
- split the workflow into lint-build and unit-tests
- use `jest --silent` in CI to suppress warnings and not obfuscate the error output

- CI

### DIFF
--- a/.github/workflows/lint-build.yml
+++ b/.github/workflows/lint-build.yml
@@ -1,4 +1,4 @@
-name: "Unit Tests"
+name: "Lint and Build"
 
 on:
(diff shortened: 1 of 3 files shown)


yurishkuro:
unit tests ran in 7min instead of 12min in the past

PR #1780, opened 2023-09-12
Handled the case in which child span starts before and ends after parent span.

### LINKED ISSUE #1779: [Bug]: Prevent Critical Path code from throwing exceptions
### What happened?

Critical Path code occasionally throws Range exception here: https://github.com/jaegertracing/jaeger-ui/blob/17eb2916047f7c3d1f5691957e29bbf6784b2e22/packages/jaeger-ui/src/components/TracePage/CriticalPath/utils/sanitizeOverFlowingChildren.tsx#L87

### Steps to reproduce

Unknown, the failures have been happening sporadically in CI

### Expected behavior

Code should handle all input trace data, fall back gracefully to not returning CP if impossible to calculate.

### Relevant log output

_No response_

### Screenshot

_No response_

### Additional context

We need to rethink the [switch statement][1]. The relative positioning of parent/child spans has a fixed number of permutations, they should be covered by the appropriate if-conditions, ideally in a hierarchical manner, e.g. 
```
if child starts after parent
  if child ends before parent
    etc.
```

The function should also be simplified to do early returns. For example, instead of introducing a nesting level with
```
   // nested logic
}
```
replace it with
```
if (!(span && span.references.length)) {
   return
}
```

[1]: https://github.com/jaegertracing/jaeger-ui/blob/17eb2916047f7c3d1f5691957e29bbf6784b2e22/packages/jaeger-ui/src/components/TracePage/CriticalPath/utils/sanitizeOverFlowingChildren.tsx#L36

### Jaeger backend version

_No response_

### SDK

_No response_

### Pipeline

_No response_

### Stogage backend

_No response_

### Operating system

_No response_

### Deployment model

_No response_

### Deployment configs

_No response_

GLVSKiriti:
|----parent----|
             |------------child------------|

The above case is not handled in previous code
Resolves #1779

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/CriticalPath/index.test.js
+++ b/packages/jaeger-ui/src/components/TracePage/CriticalPath/index.test.js
@@ -17,8 +17,6 @@ import test1 from './testCases/test1';
 import test2 from './testCases/test2';
 import test3 from './testCases/test3';
 import test4 from './testCases/test4';
-import getChildOfSpans from './utils/getChildOfSpans';
-import sanitizeOverFlowingChildren from './utils/sanitizeOverFlowingChildren';
 import test6 from './testCases/test6';
 import test7 from './testCases/test7';
 import test5 from './testCases/test5';
--- a/packages/jaeger-ui/src/components/TracePage/CriticalPath/index.test.js
+++ b/packages/jaeger-ui/src/components/TracePage/CriticalPath/index.test.js
@@ -27,7 +27,7 @@ import test8 from './testCases/test8';
 describe.each([[test1], [test2], [test3], [test4], [test5], [test6], [test7], [test8]])(
   'Happy Path',
   testProps => {
-    it('Should find criticalPathSections correctly', () => {
+    it('should find criticalPathSections correctly', () => {
       const rootSpanId = testProps.trace.spans[0].spanID;
       const spanMap = testProps.trace.spans.reduce((map, span) => {
         map.set(span.spanID, span);
--- a/packages/jaeger-ui/src/components/TracePage/CriticalPath/index.test.js
+++ b/packages/jaeger-ui/src/components/TracePage/CriticalPath/index.test.js
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import TraceCriticalPath, { computeCriticalPath } from './index';
+import TraceCriticalPath from './index';
 import test1 from './testCases/test1';
 import test2 from './testCases/test2';
 import test3 from './testCases/test3';
@@ -27,7 +27,7 @@ import test8 from './testCases/test8';
 describe.each([[test1], [test2], [test3], [test4], [test5], [test6], [test7], [test8]])(
   'Happy Path',
   testProps => {
-    it('Critical path sections', () => {
+    it('should find criticalPathSections correctly', () => {
       const criticalPath = TraceCriticalPath(testProps.trace);
       expect(criticalPath).toStrictEqual(testProps.criticalPathSections);
     });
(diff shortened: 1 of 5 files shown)


packages/jaeger-ui/src/components/TracePage/CriticalPath/index.test.js
```diff
@@ -22,22 +22,26 @@ import sanitizeOverFlowingChildren from './utils/sanitizeOverFlowingChildren';
...
+      const criticalPath = computeCriticalPath(sanitizedSpanMap, rootSpanId, []);
+      expect(criticalPath).toStrictEqual(testProps.criticalPathSections);
+    });
 
-  it('Critical path sections', () => {
-    const criticalPath = TraceCriticalPath(testProps.trace);
-    expect(criticalPath).toStrictEqual(testProps.criticalPathSections);
-  });
-});
+    it('Critical path sections', () => {
```
yurishkuro, line 42: these are meant to be read as a sentence, "it should find critical...". What do you test here?

yurishkuro (dismissed):

packages/jaeger-ui/src/components/TracePage/CriticalPath/index.test.js
GLVSKiriti: Here I test whether the calculated critical path sections of a trace is same as expected critical path sections or not

packages/jaeger-ui/src/components/TracePage/CriticalPath/index.test.js
```diff
@@ -22,22 +22,26 @@ import sanitizeOverFlowingChildren from './utils/sanitizeOverFlowingChildren';
...
-    }, new Map());
-    const refinedSpanMap = getChildOfSpans(spanMap);
-    const sanitizedSpanMap = sanitizeOverFlowingChildren(refinedSpanMap);
-    const criticalPath = computeCriticalPath(sanitizedSpanMap, rootSpanId, []);
-    expect(criticalPath).toStrictEqual(testProps.criticalPathSections);
-  });
+describe.each([[test1], [test2], [test3], [test4], [test5], [test6], [test7], [test8]])(
+  'Happy Path',
+  testProps => {
+    it('Should find criticalPathSections correctly', () => {
```
yurishkuro, line 30:
```suggestion
    it('should find criticalPathSections correctly', () => {
```

packages/jaeger-ui/src/components/TracePage/CriticalPath/index.test.js
yurishkuro: why do we even need two test functions here? The one above does a lot more work that is already done inside the business functions.

GLVSKiriti:
Basically the first test is to test computeCriticalPath function.
And second test for the whole criticalPath function i.e, memoizedTraceCriticalPath

But memoizedTraceCriticalPath returns the same as computeCriticalPath function. Then should I remove the 2nd test case??

yurishkuro:
I would remove the first one, it's calling internal implementation details which are irrelevant for the purpose of the test.

packages/jaeger-ui/src/components/TracePage/CriticalPath/index.test.js
```diff
@@ -22,22 +22,14 @@ import sanitizeOverFlowingChildren from './utils/sanitizeOverFlowingChildren';
 import test6 from './testCases/test6';
```
yurishkuro, line 22: these imports are not needed?

GLVSKiriti: But we want to test them right.  Without importing how can we do that?

yurishkuro: not the tests, but the two implementation functions that you're no longer calling

yurishkuro: I couldn't comment on the previous line

GLVSKiriti: Okk I got it!

yurishkuro (approved):

PR #1785, opened 2023-09-13
Refactor Critical Path code to prevent it from ever throwing exceptions

### LINKED ISSUE #1779: [Bug]: Prevent Critical Path code from throwing exceptions

GLVSKiriti:
Resolves #1779 

In this PR switch statement is refactored to if-else block and covered all of our decision space.

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/CriticalPath/utils/sanitizeOverFlowingChildren.test.js
+++ b/packages/jaeger-ui/src/components/TracePage/CriticalPath/utils/sanitizeOverFlowingChildren.test.js
@@ -35,8 +35,8 @@ function getExpectedSanitizedData(spans, test) {
 }
 
 describe.each([
-  [test3, new Map().set(test3.trace.spans[0].spanID, test3.trace.spans[0])],
-  [test4, new Map().set(test4.trace.spans[0].spanID, test4.trace.spans[0])],
+  [test3, new Map().set(test3.trace.spans[0].spanID, { ...test3.trace.spans[0], childSpanIds: [] })],
+  [test4, new Map().set(test4.trace.spans[0].spanID, { ...test4.trace.spans[0], childSpanIds: [] })],
   [test6, getExpectedSanitizedData(test6.trace.spans, 'test6')],
   [test7, getExpectedSanitizedData(test7.trace.spans, 'test7')],
   [test8, getExpectedSanitizedData(test8.trace.spans, 'test8')],
--- a/packages/jaeger-ui/src/components/TracePage/CriticalPath/utils/sanitizeOverFlowingChildren.test.js
+++ b/packages/jaeger-ui/src/components/TracePage/CriticalPath/utils/sanitizeOverFlowingChildren.test.js
@@ -17,6 +17,7 @@ import test4 from '../testCases/test4';
 import test6 from '../testCases/test6';
 import test7 from '../testCases/test7';
 import test8 from '../testCases/test8';
+import test9 from '../testCases/test9';
 import getChildOfSpans from './getChildOfSpans';
 import sanitizeOverFlowingChildren from './sanitizeOverFlowingChildren';
 
@@ -40,6 +41,7 @@ describe.each([
   [test6, getExpectedSanitizedData(test6.trace.spans, 'test6')],
   [test7, getExpectedSanitizedData(test7.trace.spans, 'test7')],
   [test8, getExpectedSanitizedData(test8.trace.spans, 'test8')],
+  [test9, new Map().set(test9.trace.spans[0].spanID, { ...test9.trace.spans[0], childSpanIds: [] })],
 ])('sanitizeOverFlowingChildren', (testProps, expectedSanitizedData) => {
   it('Should sanitize the data(overflowing spans) correctly', () => {
     const refinedSpanData = getChildOfSpans(testProps.trace.spans);
--- a/packages/jaeger-ui/src/components/TracePage/CriticalPath/utils/sanitizeOverFlowingChildren.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/CriticalPath/utils/sanitizeOverFlowingChildren.tsx
@@ -50,7 +50,6 @@ const sanitizeOverFlowingChildren = (spanMap: Map<string, Span>): Map<string, Sp
 
         // Remove the childSpanId from its parent span
         parentSpan.childSpanIds = parentSpan.childSpanIds.filter(id => id !== span.spanID);
-        spanMap.set(parentSpan.spanID, { ...parentSpan });
         return;
       }
       if (childEndTime > parentEndTime) {
@@ -78,7 +77,6 @@ const sanitizeOverFlowingChildren = (spanMap: Map<string, Span>): Map<string, Sp
 
       // Remove the childSpanId from its parent span
       parentSpan.childSpanIds = parentSpan.childSpanIds.filter(id => id !== span.spanID);
-      spanMap.set(parentSpan.spanID, { ...parentSpan });
     } else if (childEndTime <= parentEndTime) {
       // child start before parent, truncate is needed
       //      |----parent----|
--- a/packages/jaeger-ui/src/components/TracePage/CriticalPath/utils/sanitizeOverFlowingChildren.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/CriticalPath/utils/sanitizeOverFlowingChildren.tsx
@@ -32,66 +32,71 @@ const sanitizeOverFlowingChildren = (spanMap: Map<string, Span>): Map<string, Sp
     }
     // parentSpan will be undefined when its parentSpan is dropped previously
     const parentSpan = spanMap.get(span.references[0].spanID);
-    if (parentSpan) {
-      const childEndTime = span.startTime + span.duration;
-      const parentEndTime = parentSpan.startTime + parentSpan.duration;
-      if (span.startTime >= parentSpan.startTime) {
-        if (span.startTime >= parentEndTime) {
-          // child outside of parent range => drop the child span
-          //      |----parent----|
-          //                        |----child--|
-          // Remove the childSpan from spanMap
-          spanMap.delete(span.spanID);
 
-          // Remove the childSpanId from its parent span
-          parentSpan.childSpanIds = parentSpan.childSpanIds.filter(id => id === span.spanID);
-          spanMap.set(parentSpan.spanID, { ...parentSpan });
-        } else if (childEndTime > parentEndTime) {
-            // child end after parent, truncate is needed
-            //      |----parent----|
-            //              |----child--|
-            spanMap.set(span.spanID, {
-              ...span,
-              duration: parentEndTime - span.startTime,
-            });
-        }else{
-          // everything looks good
-          // |----parent----|
-          //   |----child--|
-        }
-      } else if (childEndTime <= parentSpan.startTime) {
-          // child outside of parent range => drop the child span
-          //                      |----parent----|
-          //       |----child--|
-
-          // Remove the childSpan from spanMap
-          spanMap.delete(span.spanID);
-
-          // Remove the childSpanId from its parent span
-          parentSpan.childSpanIds = parentSpan.childSpanIds.filter(id => id === span.spanID);
-          spanMap.set(parentSpan.spanID, { ...parentSpan });
-        } else if (childEndTime <= parentEndTime) {
-            // child start before parent, truncate is needed
-            //      |----parent----|
-            //   |----child--|
-            spanMap.set(span.spanID, {
-              ...span,
-              startTime: parentSpan.startTime,
-              duration: childEndTime - parentSpan.startTime,
-            });
-          } else {
-            // child start before parent and end after parent, truncate is needed
-            //      |----parent----|
-            //  |---------child---------|
-            spanMap.set(span.spanID, {
-              ...span,
-              startTime: parentSpan.startTime,
-              duration: parentEndTime - parentSpan.startTime,
-            });
-          }
-    } else {
+    if (!parentSpan) {
       // Drop the child spans of dropped parent span
       spanMap.delete(span.spanID);
+      return;
+    }
+    const childEndTime = span.startTime + span.duration;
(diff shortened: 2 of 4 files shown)


packages/jaeger-ui/src/components/TracePage/CriticalPath/utils/sanitizeOverFlowingChildren.tsx
```diff
@@ -27,80 +27,71 @@ const sanitizeOverFlowingChildren = (spanMap: Map<string, Span>): Map<string, Sp
...
-            // case 1: everything looks good
-            // |----parent----|
-            //   |----child--|
-            break;
+    if (!(span && span.references.length)) {
+      return;
+    }
+    // parentSpan will be undefined when its parentSpan is dropped previously
+    const parentSpan = spanMap.get(span.references[0].spanID);
+    if (parentSpan) {
```
yurishkuro, line 35: reduce nesting further
```
if (!parentSpan) {
    spanMap.delete(span.spanID);
    return;
}
```

packages/jaeger-ui/src/components/TracePage/CriticalPath/utils/sanitizeOverFlowingChildren.tsx
```diff
@@ -27,80 +27,71 @@ const sanitizeOverFlowingChildren = (spanMap: Map<string, Span>): Map<string, Sp
...
+          spanMap.delete(span.spanID);
 
-          case span.startTime < parentSpan.startTime &&
-            childEndTime <= parentEndTime &&
-            childEndTime > parentSpan.startTime:
-            // case 2: child start before parent, truncate is needed
+          // Remove the childSpanId from its parent span
+          parentSpan.childSpanIds = parentSpan.childSpanIds.filter(id => id === span.spanID);
+          spanMap.set(parentSpan.spanID, { ...parentSpan });
+        } else if (childEndTime > parentEndTime) {
```
yurishkuro, line 49:
```suggestion
          spanMap.set(parentSpan.spanID, { ...parentSpan });
          return
        }
        if (childEndTime > parentEndTime) {
```

packages/jaeger-ui/src/components/TracePage/CriticalPath/utils/sanitizeOverFlowingChildren.tsx
```diff
@@ -27,80 +27,71 @@ const sanitizeOverFlowingChildren = (spanMap: Map<string, Span>): Map<string, Sp
...
+          //                        |----child--|
+          // Remove the childSpan from spanMap
+          spanMap.delete(span.spanID);
 
-          case span.startTime < parentSpan.startTime &&
-            childEndTime <= parentEndTime &&
-            childEndTime > parentSpan.startTime:
-            // case 2: child start before parent, truncate is needed
+          // Remove the childSpanId from its parent span
+          parentSpan.childSpanIds = parentSpan.childSpanIds.filter(id => id === span.spanID);
```
yurishkuro, line 47: `filter` function keeps the items if the condition is true. The result of this line will be `childSpanIds == [span.spanOD]`, is that what you wanted?

packages/jaeger-ui/src/components/TracePage/CriticalPath/utils/sanitizeOverFlowingChildren.tsx
```diff
@@ -27,80 +27,71 @@ const sanitizeOverFlowingChildren = (spanMap: Map<string, Span>): Map<string, Sp
...
+        } else if (childEndTime > parentEndTime) {
+            // child end after parent, truncate is needed
             //      |----parent----|
-            //   |----child--|
+            //              |----child--|
             spanMap.set(span.spanID, {
               ...span,
-              startTime: parentSpan.startTime,
-              duration: childEndTime - parentSpan.startTime,
+              duration: parentEndTime - span.startTime,
```
yurishkuro, line 55: this is a serious bug: you are overriding the source data of the trace, literally changing how it looks on the screen, not just decorating it. I think you can easily account for the difference in CP code by using some min/max functions.

packages/jaeger-ui/src/components/TracePage/CriticalPath/utils/sanitizeOverFlowingChildren.tsx
```diff
@@ -27,80 +27,71 @@ const sanitizeOverFlowingChildren = (spanMap: Map<string, Span>): Map<string, Sp
...
             });
-            break;
+        }else{
+          // everything looks good
+          // |----parent----|
+          //   |----child--|
+        }
```
yurishkuro, line 61:
```suggestion
          return;
        }
        // everything looks good
        // |----parent----|
        //   |----child--|
```

packages/jaeger-ui/src/components/TracePage/CriticalPath/utils/sanitizeOverFlowingChildren.tsx
GLVSKiriti: Yeah I think I did a typo here. I need result to be all child spanIds except span.spanID

yurishkuro: so is this a bug? should it be caught by a unit test?

packages/jaeger-ui/src/components/TracePage/CriticalPath/utils/sanitizeOverFlowingChildren.tsx
```diff
@@ -27,80 +27,76 @@ const sanitizeOverFlowingChildren = (spanMap: Map<string, Span>): Map<string, Sp
...
+        //      |----parent----|
+        //                        |----child--|
+        // Remove the childSpan from spanMap
+        spanMap.delete(span.spanID);
 
-            // Remove the childSpan from spanMap
-            spanMap.delete(span.spanID);
+        // Remove the childSpanId from its parent span
+        parentSpan.childSpanIds = parentSpan.childSpanIds.filter(id => id !== span.spanID);
+        spanMap.set(parentSpan.spanID, { ...parentSpan });
```
yurishkuro, line 53: why is this needed? parentSpan is already in the map.

yurishkuro (approved):

packages/jaeger-ui/src/components/TracePage/CriticalPath/utils/sanitizeOverFlowingChildren.tsx
```diff
@@ -27,80 +27,76 @@ const sanitizeOverFlowingChildren = (spanMap: Map<string, Span>): Map<string, Sp
...
-            throw RangeError(`Error while computing Critical Path Algorithm.`);
-        }
-      } else {
-        // Drop the child spans of dropped parent span
-        spanMap.delete(span.spanID);
-      }
+      // Remove the childSpanId from its parent span
+      parentSpan.childSpanIds = parentSpan.childSpanIds.filter(id => id !== span.spanID);
+      spanMap.set(parentSpan.spanID, { ...parentSpan });
+    } else if (childEndTime <= parentEndTime) {
```
yurishkuro, line 82: codecov says this branch (child before parent) is not covered by unit tests, is that correct?

packages/jaeger-ui/src/components/TracePage/CriticalPath/utils/sanitizeOverFlowingChildren.tsx
GLVSKiriti: Here we are just updating the childSpanIds of parent span if any child span is removed.

yurishkuro: map stores a reference to parentSpan object. You already updated its children in L52. What is the need to update the map (it should have no effect)?

packages/jaeger-ui/src/components/TracePage/CriticalPath/utils/sanitizeOverFlowingChildren.tsx
GLVSKiriti: Yeah, I will to add a new test case. Previously (child before parent )and (child after parent) are in same case block so that both cases are covered as there is already a test case written for (child after parent)

packages/jaeger-ui/src/components/TracePage/CriticalPath/utils/sanitizeOverFlowingChildren.tsx
GLVSKiriti: But we are just working on the copy of the spanMap which is received as parameter to the function sanitizeOverFlowingChildren function and We are also not modifying the original span data.

packages/jaeger-ui/src/components/TracePage/CriticalPath/utils/sanitizeOverFlowingChildren.tsx
GLVSKiriti: Added a new test case !!

packages/jaeger-ui/src/components/TracePage/CriticalPath/utils/sanitizeOverFlowingChildren.tsx
GLVSKiriti: Or instead can we send a copy of the trace data to the memoizedTraceCriticalPath function instead of sending data directly?? So that original data will not be effected

https://github.com/GLVSKiriti/jaeger-ui/blob/9da89cc3633442dabec033dbfec4eed2d8c74c30/packages/jaeger-ui/src/components/TracePage/index.tsx#L394

Here can we send memoizedTraceCriticalPath({...data}) ?? Will the problem solved then??

packages/jaeger-ui/src/components/TracePage/CriticalPath/utils/sanitizeOverFlowingChildren.test.js
```diff
@@ -35,11 +36,12 @@ function getExpectedSanitizedData(spans, test) {
...
 
 describe.each([
-  [test3, new Map().set(test3.trace.spans[0].spanID, test3.trace.spans[0])],
-  [test4, new Map().set(test4.trace.spans[0].spanID, test4.trace.spans[0])],
+  [test3, new Map().set(test3.trace.spans[0].spanID, { ...test3.trace.spans[0], childSpanIds: [] })],
+  [test4, new Map().set(test4.trace.spans[0].spanID, { ...test4.trace.spans[0], childSpanIds: [] })],
   [test6, getExpectedSanitizedData(test6.trace.spans, 'test6')],
   [test7, getExpectedSanitizedData(test7.trace.spans, 'test7')],
   [test8, getExpectedSanitizedData(test8.trace.spans, 'test8')],
+  [test9, new Map().set(test9.trace.spans[0].spanID, { ...test9.trace.spans[0], childSpanIds: [] })],
```
yurishkuro, line 44: doesn't need to be part of this PR, but why L31-34 call getExpectedSanitizedData in the test setup? What it means is the test is effectively asserting `f(x) == f(x)`, which does not test anything, only increases coverage, but not in a meaningful way.

yurishkuro:
Thanks!

packages/jaeger-ui/src/components/TracePage/CriticalPath/utils/sanitizeOverFlowingChildren.tsx
GLVSKiriti: @yurishkuro can you clarify my doubt here!!

packages/jaeger-ui/src/components/TracePage/CriticalPath/utils/sanitizeOverFlowingChildren.test.js
GLVSKiriti: I don't get your point, can you elaborate it. test6,test7,test8 are needed to test the cases like when child overflows from parent(like child end after parent or start before parent etc..)

packages/jaeger-ui/src/components/TracePage/CriticalPath/utils/sanitizeOverFlowingChildren.tsx
yurishkuro: I think it's fine, I misread it as changing in place, but it's creating a clone of the span, so no issue.

packages/jaeger-ui/src/components/TracePage/CriticalPath/utils/sanitizeOverFlowingChildren.test.js
yurishkuro:
```
[test8, getExpectedSanitizedData(test8.trace.spans, 'test8')],
            ^^^^^^^^^^^^^^^^
```
I am referring to this part

PR #1789, opened 2023-09-15
Upgrade query-string to v8.1.0

anshgoyalevil:
- This PR upgrades the query-string dep from v6.3.0 to v8.1.0

## Description of the changes
- In query-string v6.4.0, `null` type was introduced. https://github.com/sindresorhus/query-string/commit/49d2203965f911940fb1b43c8263c8aaea5f95fd#diff-7aa4473ede4abd9ec099e87fec67fd57afafaf39e05d493ab4533acc38547eb8R36
- Further, in later versions, `string[]` type is changed to `(string | null)[]`

- locally

## Checklist

### DIFF
--- a/packages/jaeger-ui/package.json
+++ b/packages/jaeger-ui/package.json
@@ -75,7 +75,7 @@
     "moment": "^2.18.1",
     "object-hash": "^3.0.0",
     "prop-types": "^15.5.10",
-    "query-string": "^6.3.0",
+    "query-string": "^8.1.0",
     "raven-js": "^3.22.1",
     "react": "^18.2.0",
     "react-circular-progressbar": "^2.1.0",
--- a/packages/jaeger-ui/src/components/DeepDependencies/url.tsx
+++ b/packages/jaeger-ui/src/components/DeepDependencies/url.tsx
@@ -42,13 +42,13 @@ export function getUrl(args?: { [key: string]: unknown; showOp?: boolean }, base
   return baseUrl;
 }
 
-function firstParam(arg: string | string[]): string {
+function firstParam(arg: string | (string | null)[] | null): string {
   if (Array.isArray(arg)) {
     const returnVal = arg[0];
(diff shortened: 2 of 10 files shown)


yurishkuro (commented):
These changes seem problematic to me. Just because some library returns a different result, you changes a lot of signatures in our code, expanding them to more states. Why do those states make sense on all those places? It reduces strong typing.

I would rather add a util function that wraps the library but returns the same [] signature as before. It doesn't make sense se introducing null everywhere.

anshgoyalevil:
Totally agree. I didn't think of the wrapper function approach at first. Thanks for that. Making the changes.

However, doesn't the current approach seem more readable as the query-string's function signature matches with what we are expecting our function's parameters to be?

yurishkuro:
there is a principle "be liberal in what you accept and conservative in what you send". Loosening the type of fields within structs breaks that principle because other places in the code that read the struct now must deal with bigger surface. 

On the other hand, loosening the type of function parameters seems to align with this principle, but it's not good either, especially in the existing code base. By loosening function signature when it was already tight and the existing code was already confirming to stricter type you are giving permission to that other code to be more sloppy about the typing.

Implementation-wise, I don't have a strong opinion, it could be a helper function, or just `x ?? []` when the library is called.

anshgoyalevil:
> there is a principle "be liberal in what you accept and conservative in what you send". Loosening the type of fields within structs breaks that principle because other places in the code that read the struct now must deal with bigger surface.
> 
> On the other hand, loosening the type of function parameters seems to align with this principle, but it's not good either, especially in the existing code base. By loosening function signature when it was already tight and the existing code was already confirming to stricter type you are giving permission to that other code to be more sloppy about the typing.
> 
> Implementation-wise, I don't have a strong opinion, it could be a helper function, or just `x ?? []` when the library is called.

Learning more daily 🚀  Thanks for the useful info.
Have tried this wrapper approach to modify the function's typecasting for the first time.

anshgoyalevil:
Superseded by https://github.com/jaegertracing/jaeger-ui/pull/1794

PR #1794, opened 2023-09-15
Upgrade query-string to v8.1.0 | Add Wrapper function for queryString.parse()

anshgoyalevil:
- This PR upgrades the query-string dep from v6.3.0 to v8.1.0
- improved version of: https://github.com/jaegertracing/jaeger-ui/pull/1789

## Description of the changes
- In query-string v6.4.0, null type was introduced. https://github.com/sindresorhus/query-string/commit/49d2203965f911940fb1b43c8263c8aaea5f95fd#diff-7aa4473ede4abd9ec099e87fec67fd57afafaf39e05d493ab4533acc38547eb8R36
- Further, in later versions, string[] type is changed to (string | null)[]
- This PR adds a wrapper function for `queryString.parse()` function to modify its type as per our codebase.

- locally

## Checklist


packages/jaeger-ui/src/components/SearchTracePage/url.tsx
```diff
@@ -20,6 +20,7 @@ import prefixUrl from '../../utils/prefix-url';
 import { MAX_LENGTH } from '../DeepDependencies/Graph/DdgNodeContent/constants';
 
 import { SearchQuery } from '../../types/search';
+import parseQuery from '../../utils/parseQuery.test';
```
yurishkuro, line 23: why are we importing test module into prod module?

packages/jaeger-ui/src/utils/parseQuery.test.ts
```diff
@@ -0,0 +1,39 @@
...
+      result[key] = (parsed[key] as string[]).map((item: string | null) => item || '');
+    } else {
+      result[key] = parsed[key] as string | string[];
+    }
+  });
+
+  return result;
+}
+
+parseQuery.wrapper = parseQuery;
```
yurishkuro, line 37: what is the purpose of `wrapper`?

packages/jaeger-ui/src/components/SearchTracePage/url.tsx
anshgoyalevil: I think that way we won't have to worry about its code coverage. Changing it back to normal .ts file

packages/jaeger-ui/src/utils/parseQuery.test.ts
anshgoyalevil: This needed to be done because the way jest works is, it needs an object out of which it could spyOn the function.

So, something like `jest.spyOn(func)` won't work. Instead, it has to be `jest.spyOn(obj.func)` notation.

Thus, the wrapper thing is make it work with jest. I tried the former approach also, without `wrapper`, but it didn't work with jest.

packages/jaeger-ui/tsconfig.lint.json
```diff
@@ -51,6 +51,7 @@
     "src/reducers/trace.js",
     "src/selectors/dependencies.js",
     "src/utils/configure-store.js",
+    "src/utils/parseQuery.ts",
```
yurishkuro, line 54: Why is this needed? Don't we pick up `*ts` files for linting automatically? If we don't it's a very sad state - linting should be on by default.

packages/jaeger-ui/src/utils/parseQuery.ts
```diff
@@ -0,0 +1,39 @@
...
+      result[key] = (parsed[key] as string[]).map((item: string | null) => item || '');
+    } else {
+      result[key] = parsed[key] as string | string[];
+    }
+  });
+
+  return result;
+}
+
+parseQuery.wrapper = parseQuery;
```
yurishkuro, line 37: what is the point of this `.wrapper`? Can't we just export the function?

anshgoyalevil: We can export is without `.wrapper` also. But, then it won't work with `jest.spyOn`

packages/jaeger-ui/tsconfig.lint.json
anshgoyalevil: Not actually. We aren't picking up *ts files automatically.

Without this rule:
image

packages/jaeger-ui/src/utils/parseQuery.ts
anshgoyalevil: Done! Found a better workaround 🚀

packages/jaeger-ui/src/components/DeepDependencies/url.tsx
```diff
@@ -14,8 +14,9 @@
 
 import _isEmpty from 'lodash/isEmpty';
 import memoizeOne from 'memoize-one';
-import queryString from 'query-string';
 import { matchPath } from 'react-router-dom';
+import queryString from 'query-string';
```
yurishkuro, line 18: is it still used?

packages/jaeger-ui/tsconfig.lint.json
yurishkuro: let's rename the file to tsx then, for now. I think we need to fix this separately.

packages/jaeger-ui/src/components/DeepDependencies/url.tsx
anshgoyalevil: Yes. In `queryString.stringify` function

packages/jaeger-ui/tsconfig.lint.json
anshgoyalevil: Done!

anshgoyalevil: Though, it can also be configured in tsconfig.json

```diff
{
    "extends": "../../tsconfig",
-   "include": ["src/**/*.tsx", "typings"]
+  "include": ["src/**/*.tsx", "src/**/*.ts", "typings"]
}

File: packages/jaeger-ui/src/utils/parseQuery.tsx [hunk 6]
```diff
@@ -0,0 +1,37 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import queryString, { ParseOptions } from 'query-string';
+
+interface IParsedQuery {
+  [key: string]: string | null | (string | null)[];
+}
+
+function parseQuery(query: string, options?: ParseOptions): { [key: string]: string | string[] } {
```
yurishkuro, line 21: 1. can we through in a unit test to maintain coverage?
2. under what conditions does this return a plain array? An HTTP query string is fundamentally a dictionary.

anshgoyalevil: Yes, we can have a unit test for it. Adding it in.

anshgoyalevil: Done 🚀

yurishkuro (approved):

Issue #1797, opened 2023-09-15
[deps] Upgrade react-json-view-lite


yurishkuro:
PR #1774 is failing because of breaking changes in 1.0, needs a fix.

This PR in the backend may be helpful for testing JSON rendering: https://github.com/jaegertracing/jaeger/pull/4765

anurag-dhamala:
I'd like to work on this issue. Thanks.

duckling69:
@yurishkuro I'd like to take this issue

PR #1798, opened 2023-09-15
fix: upgrade react-json-view-lite

MeenuyD:
fix the failing because of breaking changes in 1.0, needs a fix.Property shouldInitiallyExpand has different name shouldExpandNode change the name.
#1797


yurishkuro (commented):
this is not enough, the `expander` style is also gone and without it the expand/collapse shevrons are not getting displayed

MeenuyD:
Ok I got it I will make those changes also

yurishkuro:
fixed by #1808

PR #1805, opened 2023-09-19
Restore ability to search in Select fields

### LINKED ISSUE #1784: [Bug]: Can not use search in service or Operation
### What happened?

i get the new ui with 1.33, i can not search anything at service or Operation.

roll back 1.32,it works.

### Steps to reproduce

just input anything  at service or Operation,you will find without anything input

### Expected behavior

no

### Relevant log output

_No response_

### Screenshot

_No response_

### Additional context

_No response_

### Jaeger backend version

v1.49.0

### SDK

_No response_

### Pipeline

_No response_

### Stogage backend

_No response_

### Operating system

_No response_

### Deployment model

_No response_

### Deployment configs

_No response_

prathamesh-mutkure:
- Example: Resolves #1784 

## Description of the changes
- All the Ant Design's Select fields are now searchable
- Added new util to override default behaviour of searching by value instead of label

- The test was done manually
- I'll add tests if this approach seems okay

## Screenshot 

Screenshot 2023-09-19 at 2 05 21 PM

## Checklist


prathamesh-mutkure:
I'll add tests if this approach seems okay

packages/jaeger-ui/src/utils/ant-design.tsx
```diff
@@ -0,0 +1,30 @@
...
+ * Pass this to the `filterOption` prop with `showSearch` prop set to `true`
+ *
+ * @example
+ * ```
+ * <Select {...props} filterOption={antSelectSearchByLabel} showSearch={true} />
+ * ```
+ */
+// eslint-disable-next-line import/prefer-default-export
+export function antSelectSearchByLabel(input: string, option?: DefaultOptionType) {
+  return (option?.children?.toString() ?? '').toLowerCase().includes(input.toLowerCase());
```
yurishkuro, line 29: I am not following this logic. What is the expectation of `filterOption` function? Isn't it to return true/false if a given option matches the search? If so, shouldn't the checking be done against `option.value` rather than `option.children`?

packages/jaeger-ui/src/utils/ant-design.tsx
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) 2020 The Jaeger Authors
```
yurishkuro, line 1: 2023

packages/jaeger-ui/src/utils/ant-design.tsx
```diff
@@ -0,0 +1,30 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { DefaultOptionType } from 'antd/lib/select';
+
+/**
```
yurishkuro, line 17: new functions should have unit tests

packages/jaeger-ui/src/utils/ant-design.tsx
```diff
@@ -0,0 +1,30 @@
...
+ * This enables search by label in Ant Design's Select component (v4.x)
+ *
+ * Pass this to the `filterOption` prop with `showSearch` prop set to `true`
+ *
+ * @example
+ * ```
+ * <Select {...props} filterOption={antSelectSearchByLabel} showSearch={true} />
```
yurishkuro, line 24: instead of providing this util function for search, could we instead write a function that decorates the Ant Select component in a way that it's always searchable, i.e. somewhat similar to `reduxFormFieldAdapter` function?

packages/jaeger-ui/src/utils/ant-design.tsx
```diff
@@ -0,0 +1,30 @@
...
+ * ```
+ */
+// eslint-disable-next-line import/prefer-default-export
```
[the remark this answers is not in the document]
prathamesh-mutkure: > // eslint-disable-next-line import/prefer-default-export

This linter is configured to prefer default exports for files with a single function, I can avoid the linter warning by using default export for the function but it'd be kind of difficult to refactor the code later when we add more functions to the file, as it won't be a default function after that

packages/jaeger-ui/src/utils/ant-design.tsx
prathamesh-mutkure: That's how the default behaviour works, it searches based on `option.value` but more often than not `option.value` is different from what is displayed on the UI through the `option.label` value

For e.g. In the `SearchForm`, lookback dropdown have the labels in human-readable form like `5 Minutes` while internally it used the value `5m`, the user doesn't know about the internal value, they would search by labels displayed to them

Screenshot 2023-09-20 at 2 16 05 AM

The `DefaultOptionType` doesn't have the `label` directly accessible, it rather has `children` property that contains the label, this is the reason I combined the string literal `'Last'` and `label` value so that it becomes a single child instead of both being different children, slight performance improvement
 
Screenshot 2023-09-20 at 2 25 19 AM

packages/jaeger-ui/src/utils/ant-design.tsx
prathamesh-mutkure: Yes, that'd be better I guess, what should I name it? `CustomSelect` or `SearchableSelect` 

Also, if I create a separate wrapper, do I still keep the util file or just move everything to the wrapper?

yurishkuro: A very important principle (also related to the discussion in #1738) is this: never make functionality public unless absolutely necessary (aka "premature generalization"). You can always make something public later if needed, but the reverse is not true. So right now the only need we have is a new SearchableSelect component. The fact that it would use a filter function like this one is an implementation detail, which should be kept private to the module, until a strong use case appears that really needs it (highly doubtful considering we're talking about one liner).

prathamesh-mutkure: That makes quite a lot of sense, "premature generalization" - this is something new that I learned, thanks 🚀

yurishkuro:
* missing sign-off on some commits
* need to regenerate snapshot (`yarn test -u` in packages/jaeger-ui)

packages/jaeger-ui/src/components/common/SearchableSelect.test.js
```diff
@@ -0,0 +1,100 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import React from 'react';
+import { shallow } from 'enzyme';
+
+import { Select } from 'antd';
+import SearchableSelect, { filterOptionsByLabel } from './SearchableSelect';
+
+describe('SearchableSelect', () => {
```
yurishkuro, line 21: You're missing the test of actually searching when typing. Instead of testing the internal findBy... function, you could've made those same tests from the perspective of the end user, who is typing into the field, not calling a function.

prathamesh-mutkure: Yes, that's how we initially planned to do it but isn't the part for showing matching options on the UI handled by Ant Design that is already tested? I can add an explicit UI test if that makes more sense

yurishkuro: Recently we broke the ability to search and no unit tests failed. If something changes in ant design regarding HOW filtering is supposed to be hooked up, no tests will fail again. So (a) we need to test for that, and (b) why restrict the input permission tests that you have to test inner function if we could test them end to end to the user facing behavior?

prathamesh-mutkure: Sure, that makes more sense

yurishkuro:
I am going to merge this so that it can make it into the next release, we can alter the tests in another PR.

PR #1814, opened 2023-09-22
Priyanshu kun/clean any

priyanshu-kun:
- Cleaned unnecessary `any` keyword.

## Description of the changes
- In this PR I cleaned unnecessary `any` keyword and I'm doing it gradually.

- I successfully ran linting and unit tests.

## Checklist


packages/jaeger-ui/src/components/common/RelativeDate.tsx
```diff
@@ -12,22 +12,23 @@
...
   fullMonthName: boolean | undefined | null;
   includeTime: boolean | undefined | null;
-  value: number | Date | any;
+  value: number | Date;
 };
 
 // TODO typescript doesn't understand text or null as react nodes
 // https://github.com/Microsoft/TypeScript/issues/21699
-export default function RelativeDate(props: Props): any {
+export default function RelativeDate(props: Props): JSX.Element {
```
yurishkuro, line 28: why can't this be a string instead of element?

packages/jaeger-ui/src/utils/readJsonFile.tsx
```diff
@@ -22,8 +22,10 @@ export default function readJsonFile(fileList: { file: File }) {
       }
       try {
         resolve(JSON.parse(reader.result));
-      } catch (error: any) {
-        reject(new Error(`Error parsing JSON: ${error.message}`));
+      } catch (error: unknown) {
```
yurishkuro, line 25: can we not just do this?
```suggestion
      } catch (error: Error) {
```

packages/jaeger-ui/src/utils/ts/pluckTruthy.tsx
```diff
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 // Because TypeScript doesn't believe in Array#filter(Boolean)
-export default function pluckTruthy<T>(values: (T | any)[]): T[] {
+export default function pluckTruthy<T>(values: (T | undefined)[]): T[] {
```
yurishkuro, line 16: this change doesn't make sense, `any` actually seems reasonable here.

packages/jaeger-ui/src/components/common/RelativeDate.tsx
priyanshu-kun: Because RelativeDate is used as a JSX element not like a string for ex: `<RelativeDate props />` so when I use string as a return type it throw an error.

packages/jaeger-ui/src/utils/readJsonFile.tsx
priyanshu-kun: I already try to use `Error` instead `unknown` but It gives `Catch clause variable type annotation must be 'any' or 'unknown' if specified.` error.

packages/jaeger-ui/src/utils/ts/pluckTruthy.tsx
priyanshu-kun: ok, I will revert back it to any.

packages/jaeger-ui/src/utils/readJsonFile.tsx
```diff
@@ -22,8 +22,10 @@ export default function readJsonFile(fileList: { file: File }) {
...
+        if (error instanceof Error) {
+          reject(new Error(`Error parsing JSON: ${error.message}`));
+        }
```
yurishkuro, line 28: how about this then? If the type is already Error, its toString will return the message, but if it's some other object not extending error, then we get `[Object object]`. But at least in all cases we do call `reject`, which your code may not call.
```suggestion
      } catch (error: object) {
          reject(new Error(`Error parsing JSON: ${error}`));
      }
```

packages/jaeger-ui/src/utils/ts/pluckTruthy.tsx
yurishkuro: just add linter-ignore for that line

packages/jaeger-ui/src/utils/readJsonFile.tsx
priyanshu-kun: Ok, but we cannot use `error: object` as well it gives same error that I mention above. So, in this case we have revert it back to any.

yurishkuro: Ok then go with any

yurishkuro: And a lint ignore line

yurishkuro (approved):

PR #1818, opened 2023-09-24
Speed up tests locally by upto 45%

anshgoyalevil:
# Which problem is this PR solving?
- This PR aims to add an additional flag to the jest script which would make running tests locally even faster than ever.

## Description of the changes
- It uses `--maxWorkers=50%` flag to do so.

- It was tested with several flags as below;

```bash
-----------------
yarn test

Time:        96.903 s
-----------------
yarn test --runInBand

Time:        131.721 s, estimated 606 s
-----------------
yarn test --maxWorkers=50%

Time:        53.295 s
-----------------
```

Operating System Used to Benchmark:
`Ubuntu 22`

## Checklist


yurishkuro (approved):

packages/jaeger-ui/package.json
```diff
@@ -104,7 +104,7 @@
     "coverage": "yarn run test-ci --coverage",
     "start:ga-debug": "REACT_APP_GA_DEBUG=1 REACT_APP_VSN_STATE=$(../../scripts/get-tracking-version.js) vite",
     "start": "vite",
-    "test": "jest",
+    "test": "jest --maxWorkers=50%",
```
yurishkuro, line 107: does `%` refer to the number of available CPU cores?

anshgoyalevil: Yes. It's the CPU Cores present in the current run environment. Most of the articles on internet recommends this setting.

yurishkuro: Makes sense for local testing, so as not to bring the whole system down during the test. But if this %-based addressing scheme is available, why does the line below have `--maxWorkers=4`? We don't really know how many CPU cores are available in a CI running, but it would make sense to utilize all of them.

anshgoyalevil: Every CI environment has their own workers limit, and the setting which we are currently using for test-ci is adapted to GitHub Actions only.

Moreover, reason for choosing a concrete value (4) instead of % over here is because in CI, we know we won't be performing multi-tasking, so all threads can be used for tests. (Minus the thread for main workflow task)

In future, if we have any plans to move to another CI environment like CircleCI, I think we would again have to figure out this max number of workers. (I guess, it's 8 for CircleCI).

yurishkuro: But why can't we say 100% for GitHub? When the tests are running, no other activity is higher priority.

anshgoyalevil: Theoretically that seems right. Don't know why everyone suggests using a number value instead of percentage in CI for jest.

I would have to try that thing.

PR #1824, opened 2023-09-25
Fixed all controversial icons .

### LINKED ISSUE #1723: [deps] Choose between antd/icons and react-icons
Follow up on this discussion https://github.com/jaegertracing/jaeger-ui/pull/1721#issuecomment-1688653824

We are probably using about 10-20 different icons in the project, pulled rather randomly from antd and react-icons (and within react-icons from completely different sets).

The proposal is to consolidate onto a single set, for example react-icons/fa (assuming it can accommodate our existing usage).

* [ ] run grep/awk to come up with an inventory of all icons in use
* [ ] compare and select a single set that can represent them with equivalent icons
* [ ] make code changes
* [ ] remove unused dependencies

priyanshu-kun:
- Resolves #1723 

## Description of the changes
- In this PR, I replace all ant-design icons with [lucide](https://lucide.dev/) icons set as well icons which I missed in last PR.

## Checklist


packages/jaeger-ui/package.json
```diff
@@ -69,6 +68,7 @@
     "lodash": "^4.17.19",
     "logfmt": "^1.2.0",
     "lru-memoize": "^1.1.0",
+    "lucide-react": "^0.279.0",
```
yurishkuro, line 71: why are we adding another module? I though the whole point was to restrict ourselves to react-icons.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow/TimelineCollapser.tsx
```diff
@@ -14,8 +14,9 @@
 
 import React from 'react';
 import { Tooltip } from 'antd';
-import { DoubleRightOutlined } from '@ant-design/icons';
-import { IoChevronForward } from 'react-icons/io5';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { ChevronsRight, ChevronsDown } from 'lucide-react';
```
yurishkuro, line 18: why can't this be `react-icons/lu`?

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow/TimelineCollapser.tsx
```diff
@@ -14,8 +14,9 @@
...
+import { IoChevronDown, IoChevronForward } from 'react-icons/io5';
```
yurishkuro, line 19: I would stick to a single family for related components - use LuChevronDown

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow/TimelineCollapser.tsx
```diff
@@ -52,16 +53,16 @@ export default class TimelineCollapser extends React.PureComponent<CollapserProp
...
         <Tooltip title={getTitle('Collapse +1')} getPopupContainer={this.getContainer}>
-          <IoChevronForward
+          <IoChevronDown
             onClick={onCollapseOne}
             className="TimelineCollapser--btn TimelineCollapser--btn-size"
           />
         </Tooltip>
         <Tooltip title={getTitle('Expand All')} getPopupContainer={this.getContainer}>
-          <DoubleRightOutlined onClick={onExpandAll} className="TimelineCollapser--btn-expand" />
+          <ChevronsDown onClick={onExpandAll} className="TimelineCollapser--btn-expand" />
```
yurishkuro, line 62: why is down equivalent to dbl-right? We can reuse ChevronsDown but rotate by 90deg, but I don't see this in the css

priyanshu-kun:
@yurishkuro  My mistake, I thought I can find icons set from outside react-icons and I didn't check if lucide set preset in react-icons or not. But anyway I fixed all for your concerns=:)

yurishkuro (approved):

yurishkuro:
🥳

Issue #1825, opened 2023-09-26
[Tracker]: Upgrade react-router-dom to latest version


anshgoyalevil:
### What happened?

.

### Steps to reproduce

.

### Expected behavior

.

anshgoyalevil:
Yes, we would have to upgrade it sequentially version by version, keeping the project intact

priyanshu-kun:
@anshgoyalevil  do we have to migrate it to v5 before v6?

Piyushrathoree:
@yurishkuro is this still unresolved ? if not I would love to work on it.

parshipcy:
done #3313

PR #1826, opened 2023-09-26
Replace react-router-redux with redux-first-history

anshgoyalevil:
- related to: https://github.com/jaegertracing/jaeger-ui/issues/1825
- The `react-router-redux` is a dead project and is not compatible with `react-router-dom` `v5` and `v6`.
- The whole react-router community suggested using `redux-first-history` which is compatible with `v4`, `v5` and `v6` both.

- Manually along with redux dev toolkit and automated testing

## Checklist


yurishkuro:
> Manually along with redux dev toolkit and automated testing

Can you elaborate on these tests? It's impossible to understand the impact from the change alone.

anshgoyalevil:
> > Manually along with redux dev toolkit and automated testing
> 
> Can you elaborate on these tests? It's impossible to understand the impact of the change alone.

Yes, Since the change is related to react-router-redux, it affects the UiFindInput Component directly, since this consumes the `state.router.location` from the Redux Store.

Now, if something is broken on the `redux-first-history` part, the `Find Span By ID` feature won't work unless we reload the page.

How do I verify?
- Visit any TracePage in Jaeger UI
- Copy a span ID, and paste it into the search box at the TracePage header.
- If the copies span id's span gets highlighted, redux-first-history is working behind the scenes.

Now, try creating a new `history` object instead of using the one imported from `configure-store`.
Repeat the above steps, the UiFindInput feature won't highlight the span anymore.

- Test using Redux DevTools
Using the Redux DevTools, we can see that every action is being dispatched. Like the specified above TracePage's search bar input, on entering a value, the redux-first-history dispatches an action of saving the value to history and then uses the location from the history of the redux store to highlight the span.

image

yurishkuro (approved):

PR #1829, opened 2023-09-27
Remove unused icon

### LINKED ISSUE #1723: [deps] Choose between antd/icons and react-icons

priyanshu-kun:
Resolves #1723 

## Description of the changes
- In this PR, I refactor an unused icon which I missed in last PR.


yurishkuro (approved):

.gitignore
```diff
@@ -21,3 +21,4 @@ npm-debug.log
 yarn-error.log
 lerna-debug.log
 .eslintcache
+.vscode
```
yurishkuro, line 24: I recommend you add this to your global .gitignore

PR #1837, opened 2023-09-30
replace match.params with useParams hook

anshgoyalevil:
- part of: #1825 

## Description of the changes
- In rrd v6, we can no longer use `match` prop. Since the introduction of hooks, we need to use `useParams()` hook for accessing that.
- rrd v5 to v5.1 guide recommends using `<Route children ..` instead of `<Route component ..> since it would be deprecated in v6.
- This is the last PR for the v6 preparation. The next PR would finally upgrade rrd to v6.

- Manually, by visiting each page.

[screen-capture (4).webm](https://github.com/jaegertracing/jaeger-ui/assets/94157520/42ead7db-26bd-47c9-a16d-a4500b76cd3f)

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/App/index.jsx
+++ b/packages/jaeger-ui/src/components/App/index.jsx
@@ -56,19 +56,35 @@ export default class JaegerUIApp extends Component {
           <Router history={history}>
             <Page>
               <Switch>
-                <Route path={searchPath} component={SearchTracePage} />
-                <Route path={traceDiffPath} component={TraceDiff} />
-                <Route path={tracePath} component={TracePage} />
-                <Route path={dependenciesPath} component={DependencyGraph} />
-                <Route path={deepDependenciesPath} component={DeepDependencies} />
-                <Route path={qualityMetricsPath} component={QualityMetrics} />
-                <Route path={monitorATMPath} component={MonitorATMPage} />
+                <Route path={searchPath}>
+                  <SearchTracePage />
+                </Route>
+                <Route path={traceDiffPath}>
+                  <TraceDiff />
+                </Route>
+                <Route path={tracePath}>
+                  <TracePage />
+                </Route>
+                <Route path={dependenciesPath}>
+                  <DependencyGraph />
+                </Route>
+                <Route path={deepDependenciesPath}>
+                  <DeepDependencies />
+                </Route>
+                <Route path={qualityMetricsPath}>
+                  <QualityMetrics />
+                </Route>
+                <Route path={monitorATMPath}>
+                  <MonitorATMPage />
+                </Route>
 
-                <Redirect exact path="/" to={searchPath} />
-                <Redirect exact path={prefixUrl()} to={searchPath} />
-                <Redirect exact path={prefixUrl('/')} to={searchPath} />
+                <Route exact path="/" render={() => <Redirect to={searchPath} />} />
+                <Route exact path={prefixUrl()} render={() => <Redirect to={searchPath} />} />
+                <Route exact path={prefixUrl('/')} render={() => <Redirect to={searchPath} />} />
 
-                <Route component={NotFound} />
+                <Route>
+                  <NotFound />
+                </Route>
               </Switch>
             </Page>
           </Router>
--- a/packages/jaeger-ui/src/components/App/index.jsx
+++ b/packages/jaeger-ui/src/components/App/index.jsx
@@ -78,9 +78,15 @@ export default class JaegerUIApp extends Component {
                   <MonitorATMPage />
                 </Route>
 
-                <Route exact path="/" render={() => <Redirect to={searchPath} />} />
-                <Route exact path={prefixUrl()} render={() => <Redirect to={searchPath} />} />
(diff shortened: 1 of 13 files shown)


packages/jaeger-ui/src/components/SearchTracePage/index.jsx
```diff
@@ -309,4 +310,4 @@ function mapDispatchToProps(dispatch) {
   };
 }
 
-export default connect(mapStateToProps, mapDispatchToProps)(SearchTracePageImpl);
+export default withRouteProps(connect(mapStateToProps, mapDispatchToProps)(SearchTracePageImpl));
```
yurishkuro, line 313: Curious, why wasn't this change needed in the previous PR where withRouteProps was introduced? Does it work both ways?

packages/jaeger-ui/src/components/TraceDiff/TraceDiff.tsx
```diff
@@ -187,7 +187,9 @@ export function mapDispatchToProps(dispatch: Dispatch<any>) {
   return { fetchMultipleTraces, forceState };
 }
 
-export default connect<TStateProps, TDispatchProps, TOwnProps, ReduxState>(
-  mapStateToProps,
-  mapDispatchToProps
-)(TraceDiffImpl);
+export default withRouteProps(
```
yurishkuro, line 190: could you please add documentation to this function (should've asked in the prev PR)?

https://github.com/jaegertracing/jaeger-ui/blob/bcee38a2157d6ac1fe573e463d17239810684773/packages/jaeger-ui/src/utils/withRouteProps.tsx#L28

packages/jaeger-ui/src/components/TraceDiff/TraceDiff.tsx
```diff
@@ -46,7 +46,7 @@ type TDispatchProps = {
 
 type TOwnProps = {
   history: RouterHistory;
-  match: match<TDiffRouteParams>;
+  params: TDiffRouteParams;
```
yurishkuro, line 49: Does this change have any impart on the shape of the URL? IIUC the state is included in the URL.

packages/jaeger-ui/src/components/SearchTracePage/index.jsx
anshgoyalevil: No, it doesn't work both ways.

In the previous PR's code, it was working fine. I have just verified with the code on the main branch.

This file uses: `this.props.history.push` and even in the previous PR, I didn't see it using `withRouter`.
So, where did its `history` prop come from in the main branch? That's strange.

packages/jaeger-ui/src/components/TraceDiff/TraceDiff.tsx
anshgoyalevil: Documentation in the form of jsDocs comments added throughout the function? Or is there any dedicated documentation guide for each function used in Jaeger-UI?

packages/jaeger-ui/src/components/TraceDiff/TraceDiff.tsx
anshgoyalevil: No. Absolutely not. 

Previous URL structure: 
```
http://localhost:5173/trace/4e9e0a8990106fdd...196659de1e174bd4?cohort=4e9e0a8990106fdd&cohort=196659de1e174bd4
```

Current:
```
http://localhost:5173/trace/51c935040996787d...0f7faf874d731c2c?cohort=51c935040996787d&cohort=0f7faf874d731c2c
```

It's just the place from where we use the extracted URL parameters.

packages/jaeger-ui/src/components/SearchTracePage/index.jsx
anshgoyalevil: You can too verify this by clicking on a trace spot from this chart:

image

In the main branch's code, it was working.

But in this PR's code, without the use of `withRouteProps`, this doesn't seem to work. I couldn't figure out why.

anshgoyalevil: Just figured it out.

Actually, it is due to the change in how we are using `Route children` and `Route component`.

packages/jaeger-ui/src/components/App/index.jsx
```diff
@@ -56,19 +56,41 @@ export default class JaegerUIApp extends Component {
...
-                <Route path={searchPath} component={SearchTracePage} />
-                <Route path={traceDiffPath} component={TraceDiff} />
-                <Route path={tracePath} component={TracePage} />
-                <Route path={dependenciesPath} component={DependencyGraph} />
-                <Route path={deepDependenciesPath} component={DeepDependencies} />
-                <Route path={qualityMetricsPath} component={QualityMetrics} />
-                <Route path={monitorATMPath} component={MonitorATMPage} />
+                <Route path={searchPath}>
+                  <SearchTracePage />
+                </Route>
```
anshgoyalevil, line 61: With this change, the child components, `SearchTracePage` in this example, would no longer receive the Route props.
So, we need to use the `withRouteProps` HOC to get them.

yurishkuro (approved):

Issue #1846, opened 2023-10-04
[Bug] Unstable test in jaeger-ui/src/api/jaeger.test.js


yurishkuro:
This test recently failed and then succeeded on the rerun. From the stack trace it looks like an issue with randomly generated timestamps.

https://github.com/jaegertracing/jaeger-ui/actions/runs/6406079597/job/17389965009?pr=1845#step:5:65

```
FAIL  src/api/jaeger.test.js
  ● Test suite failed to run

    RangeError: Chance: Min cannot be greater than Max.

      148 |       references: [],
      149 |       startTime,
    > 150 |       duration: chance.integer({ min: 1, max: traceEndTime - startTime }),
          |                        ^
      151 |       tags: chance.tags(),
      152 |       logs: [],
      153 |     };

      at testRange (../../node_modules/chance/chance.js:98:19)
      at Chance.testRange [as integer] (../../node_modules/chance/chance.js:281:9)
      at Chance.integer [as span] (src/demo/trace-generators.js:150:24)
      at Chance.apply [as n] (../../node_modules/chance/chance.js:634:25)
      at Chance.n [as trace] (src/demo/trace-generators.js:106:24)
      at Chance.apply [as n] (../../node_modules/chance/chance.js:634:25)
      at Chance.n [as traces] (src/demo/trace-generators.js:163:19)
      at traces (src/api/jaeger.test.js:117:42)
      at Object.describe (src/api/jaeger.test.js:116:1)
```

MeenuyD:
Hello @yurishkuro I would like to work on this issue can you assign it to me?

PR #1848, opened 2023-10-04
fix: Unstable test in jaeger-ui/src/api/jaeger.test.js

MeenuyD:
fix the test that recently failed and then succeeded on the rerun. From the stack trace it looks like an issue with randomly generated timestamps.
https://github.com/jaegertracing/jaeger-ui/actions/runs/6406079597/job/17389965009?pr=1845#step:5:65
#1846


packages/jaeger-ui/src/demo/trace-generators.js
```diff
@@ -147,7 +147,7 @@ export default chance.mixin({
       operationName: chance.pickone(operations),
       references: [],
       startTime,
-      duration: chance.integer({ min: 1, max: traceEndTime - startTime }),
+      duration: chance.integer({ min: 1, max: Math.max(1, traceEndTime - startTime)}),
```
yurishkuro, line 150: this is a band-aid that hides the underlying issue. If we have the expectation that trace/span duration should be more than 1 (it's 1 microsecond here, iirc), then the data generator should act accordingly and not generate span durations that are too small

yurishkuro: And another issue is that even if the actual span durations are not small, it seems like this specific subtraction can produce a small value, which also indicates some weird timing of the generated span.

yurishkuro: you need to understand why this condition could happen.

yurishkuro:
Solved in #1871

PR #1844, opened 2023-10-04
Prepare release v1.34.0

albertteoh:
Prepares Jaeger UI release v1.34.0

PR #1849, opened 2023-10-05
Prepare release v1.34.0

albertteoh:
Prepares Jaeger UI release v1.34.0.

Includes some quality of life improvements for my fellow maintainers:
- Hopefully, simpler process to generate changelogs (ignores dependabot, etc.).
- DRY up release scripts by downloading those defined in jaeger repo, which should make our releases a bit more consistent.

Supersedes https://github.com/jaegertracing/jaeger-ui/pull/1844.

## Checklist


yurishkuro (approved):

.prettierignore
```diff
@@ -1,2 +1,3 @@
 # Ignore generated files
-packages/jaeger-ui/index.d.ts
\ No newline at end of file
+packages/jaeger-ui/index.d.ts
+scripts
```
yurishkuro, line 3: I think the issue was not with scripts, but with the CHANGELOG that prettier was trying to change (switching between `*` and `-` for bullets)

yurishkuro: It's fine to ignore the downloaded scripts, but then I'd narrow it to
```suggestion
scripts/release-notes.py
scripts/draft-release.py
```

albertteoh: Thanks 😄

scripts/get-tracking-version.js
```diff
@@ -1,139 +0,0 @@
-#!/usr/bin/env node
```
yurishkuro, line 1: @albertteoh did you delete this by mistake?  I don't am not sure what it is supposed to do, but it is being used during the build https://github.com/jaegertracing/jaeger-ui/blob/f391b862ce143ef32d60947053130c6bdee468c5/packages/jaeger-ui/package.json#L103

yurishkuro: https://github.com/jaegertracing/jaeger-ui/pull/1998

PR #1854, opened 2023-10-08
Remove deprecated usages of ant-design v3 props from TopNav component

anshgoyalevil:

## Description of the changes
- This PR removes the deprecated usages of some props from ant-design v4 components.
- The `Menu` no longer supports children, and needs an items array to be passed.
- `DropDown` no longer supports overlay, and needs an items array to be passed. Because of this change, we can no longer test the DropDown items but can verify the number of items.

- Manually, and unit tests.
- Since this PR directly affects the TopNav, here's a demonstration of it working perfectly.

[screen-capture (6).webm](https://github.com/jaegertracing/jaeger-ui/assets/94157520/eeb5284f-7b28-47f8-9b59-c3aee481dffa)

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/App/TopNav.test.js
+++ b/packages/jaeger-ui/src/components/App/TopNav.test.js
@@ -46,7 +46,6 @@ describe('<TopNav>', () => {
   const defaultProps = {
     config: {
       menu: [
-        configMenuGroup,
         {
           label: labelGitHub,
           url: githubUrl,
@@ -56,6 +55,7 @@ describe('<TopNav>', () => {
           label: 'Blog',
           url: blogUrl,
         },
+        configMenuGroup,
       ],
     },
     router: {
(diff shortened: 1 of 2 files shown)


packages/jaeger-ui/src/components/App/TopNav.test.js
```diff
@@ -111,32 +115,6 @@ describe('<TopNav>', () => {
...
-
-      it('renders sub-menu links', () => {
-        dropdownItems.slice(1, 2).forEach(itemConfig => {
-          const item = subMenu.dive().find(`[href="${itemConfig.url}"]`);
-          expect(item.length).toBe(1);
-          expect(item.prop('target')).toBe(itemConfig.anchorTarget || '_blank');
-          expect(item.text()).toBe(itemConfig.label);
-        });
-      });
-    });
```
yurishkuro, line 139: So basically we're only relying on the typing now (`MenuProps`)?

yurishkuro (approved):

packages/jaeger-ui/src/components/App/TopNav.test.js
anshgoyalevil: Yes. But not specifically `MenuProps`. DropDown items use `ConfigMenuItem` type from `src/types/config.tsx`

PR #1855, opened 2023-10-09
Replace deprecated visible prop of ant-design v3 with open prop of ant-design v4

anshgoyalevil:

## Description of the changes
- This PR replaces the deprecated visible prop of ant-design v3 with open prop of ant-design v4
- These props are exactly the same, just the name differs.

- unit tests, and manually

## Checklist


anshgoyalevil:
@yurishkuro Shouldn't we use something like https://github.com/asyncapi/website/pull/2170#issuecomment-1735255198
A check that fails the CI if the PR's title is not properly linted according to [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/).

yurishkuro:
> Shouldn't we use something like https://github.com/asyncapi/website/pull/2170#issuecomment-1735255198

yep, that's the idea behind #4799. I like the list published by that bot.

yurishkuro (approved):

PR #1857, opened 2023-10-09
Replace deprecated usage of ant-design Tabs component from SearchTracePage

anshgoyalevil:

## Description of the changes
- replace deprecated usage of Tabs component from SearchTracePage
- The ant-design v4 introduces a new API `items` which increases the performance by deferring the In-Active components for the first load.
- This is the reason we cannot test for the actual `FileLoader` component to be present, rather, we can [test for the Tab Key to be present](https://github.com/jaegertracing/jaeger-ui/pull/1857/files#diff-aed5cd35151c3219d94e8071c5fe875b963075a0c89df5f070c9b4c34dadf646R178).

- manually, and automated tests

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/SearchTracePage/index.jsx
+++ b/packages/jaeger-ui/src/components/SearchTracePage/index.jsx
@@ -108,13 +108,7 @@ export class SearchTracePageImpl extends Component {
       tabItems.push({
         label: 'Upload',
         key: 'fileLoader',
-        children: (
-          <FileLoader
-            loadJsonTraces={fileList => {
-              loadJsonTraces(fileList);
-            }}
-          />
-        ),
+        children: <FileLoader loadJsonTraces={loadJsonTraces} />,
       });
     }
     return (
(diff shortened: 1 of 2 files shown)


packages/jaeger-ui/src/components/SearchTracePage/index.jsx
```diff
@@ -100,25 +98,31 @@ export class SearchTracePageImpl extends Component {
...
+      tabItems.push({ label: 'Search', key: 'searchForm', children: <LoadingIndicator /> });
+    }
+    if (!disableFileUploadControl) {
+      tabItems.push({
+        label: 'Upload',
+        key: 'fileLoader',
+        children: (
+          <FileLoader
+            loadJsonTraces={fileList => {
+              loadJsonTraces(fileList);
```
yurishkuro, line 114: why is the lambda used here instead of simply referencing the function name? I think it would remove the code coverage warning

anshgoyalevil: Right actually. I just copied this piece of code from the previous one.
Removed lambda function from this part, though, the coverage is because this part has no test for it, previously too.

yurishkuro (approved):

PR #1864, opened 2023-10-11
Replace deprecated ant-design props from ReferenceButton and KeyValuesTable

anshgoyalevil:

## Description of the changes
- This PR replaces the deprecated ant-design props from ReferenceButton and KeyValuesTable
- The `colums.render` prop is no longer required to return children as JSX. Instead, it is by default.

- manually, and unit tests

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/ReferencesButton.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/ReferencesButton.tsx
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 import React from 'react';
-import { Dropdown, Menu, Tooltip } from 'antd';
+import { Dropdown, Tooltip } from 'antd';
 import { TooltipPlacement } from 'antd/lib/tooltip';
 import NewWindowIcon from '../../common/NewWindowIcon';
 import { SpanReference } from '../../../types/trace';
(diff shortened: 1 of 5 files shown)


packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.tsx
```diff
@@ -110,23 +110,21 @@ LinkValue.defaultProps = {
...
+  return dropdownItems;
+};
 
 type KeyValuesTableProps = {
   data: KeyValuePair[];
   linksGetter: ((pairs: KeyValuePair[], index: number) => Link[]) | TNil;
 };
 
+// KeyValuesTable is displayed as a menu at span level.
+// Example: https://github.com/jaegertracing/jaeger-ui/assets/94157520/b518cad9-cb37-4775-a3d6-b667a1235f89
```
anshgoyalevil, line 127: Example

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/ReferencesButton.tsx
```diff
@@ -28,28 +28,30 @@ type TReferencesButtonProps = {
   focusSpan: (spanID: string) => void;
 };
 
+// ReferencesButton is displayed as a menu at the span level.
+// Example: https://github.com/jaegertracing/jaeger-ui/assets/94157520/2b29921a-2225-4a01-9018-1a1952f186ef
```
anshgoyalevil, line 32: Example

(Used some dummy data to visualize this component)

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.tsx
yurishkuro: what happens if there's only one link, do we still show the sandwich menu it outgoing link directly? The latter would be more usable (I think it's unusual to have >1 link on a tag)

anshgoyalevil: In case there's one link only, the link is shown directly, instead of DropDown.

https://github.com/jaegertracing/jaeger-ui/blob/1639eb493937e89120b9b6cdc4e5b128bd10c246/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.tsx#L139-L147

yurishkuro (approved):

PR #1871, opened 2023-10-15
Add tooltip when hovering Critical Path

GLVSKiriti:
- Related to #1582 

## Description of the changes
-  Added Tooltip which is visible on hovering the critical path

@yurishkuro

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBar.css
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBar.css
@@ -56,7 +56,7 @@ limitations under the License.
   content: '';
   position: absolute;
   top: 0;
-  width: 10%; 
+  width: 10%;
   height: 100%;
   visibility: hidden;
   background-color: white;
@@ -104,13 +104,15 @@ limitations under the License.
   background-color: rgba(0, 0, 0, 0.5);
   cursor: pointer;
   height: 60%;
-  min-width: 1px;
+  min-width: 2px;
   position: absolute;
   top: 20%;
 }
 
 .SpanBar--logMarker:hover {
   background-color: #000;
+  transform: scale(1.2);
+  transition: transform 0.3s ease;
 }
 
 .SpanBar--logMarker::before,
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBar.css
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBar.css
@@ -104,7 +104,7 @@ limitations under the License.
   background-color: rgba(0, 0, 0, 0.5);
   cursor: pointer;
   height: 60%;
-  min-width: 2px;
+  min-width: 1px;
   position: absolute;
   top: 20%;
 }
@@ -122,7 +122,7 @@ limitations under the License.
   top: 0;
   bottom: 0;
   right: 0;
-  border: 1px solid transparent;
+  border: 3px solid transparent;
 }
 
 .SpanBar--logMarker::after {
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBar.css
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBar.css
@@ -49,6 +49,32 @@ limitations under the License.
   top: 45%;
   height: 11%;
   z-index: 2;
+  overflow: hidden;
+}
+
+.SpanBar--criticalPath::before {
+  content: '';
+  position: absolute;
+  top: 0;
+  width: 10%; 
+  height: 100%;
+  visibility: hidden;
+  background-color: white;
+  border: 1px solid white;
+}
+
+.SpanBar--criticalPath:hover::before {
+  visibility: visible;
+  animation: runOnce 2s forwards;
+}
+
+@keyframes runOnce {
+  0% {
+    left: 0;
+  }
+  100% {
+    left: 100%;
+  }
 }
 
 .SpanBar--label {
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBar.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBar.tsx
@@ -30,6 +30,7 @@ type TCommonProps = {
   // onClick: (evt: React.MouseEvent<any>) => void;
   onClick?: (evt: React.MouseEvent<any>) => void;
   criticalPath: criticalPathSection[];
+  isChildrenExpanded: boolean;
   viewEnd: number;
   viewStart: number;
   getViewedBounds: ViewedBoundsFunctionType;
@@ -57,6 +58,7 @@ function toPercentInDecimal(value: number) {
 function SpanBar(props: TCommonProps) {
   const {
     criticalPath,
+    isChildrenExpanded,
     viewEnd,
     viewStart,
     getViewedBounds,
@@ -85,6 +87,23 @@ function SpanBar(props: TCommonProps) {
   const setLongLabel = () => {
     setLabel(longLabel);
   };
+
+  let criticalPathWhenParentCollapsed = {
+    ViewStart: Infinity,
+    ViewEnd: -Infinity,
+  };
+  if (!isChildrenExpanded && criticalPath) {
+    criticalPathWhenParentCollapsed = criticalPath.reduce((accumulator, each) => {
+      const critcalPathViewBounds = getViewedBounds(each.section_start, each.section_end);
+      const criticalPathViewStart = critcalPathViewBounds.start;
+      const criticalPathViewEnd = critcalPathViewBounds.end;
+
+      accumulator.ViewStart = Math.min(accumulator.ViewStart, criticalPathViewStart);
+      accumulator.ViewEnd = Math.max(accumulator.ViewEnd, criticalPathViewEnd);
+
+      return accumulator;
+    }, criticalPathWhenParentCollapsed);
+  }
 
   return (
     <div
@@ -124,7 +143,7 @@ function SpanBar(props: TCommonProps) {
             <div
               data-testid="SpanBar--logMarker"
               className="SpanBar--logMarker"
-              style={{ left: positionKey }}
+              style={{ left: positionKey, zIndex: 3 }}
             />
           </Popover>
         ))}
@@ -140,6 +159,7 @@ function SpanBar(props: TCommonProps) {
         />
       )}
       {criticalPath &&
+        (!span.hasChildren || isChildrenExpanded) &&
         criticalPath.map((each, index) => {
           const critcalPathViewBounds = getViewedBounds(each.section_start, each.section_end);
           const criticalPathViewStart = critcalPathViewBounds.start;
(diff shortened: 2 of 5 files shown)


GLVSKiriti:
tooltipCriticalPath

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBar.tsx
```diff
@@ -146,16 +146,18 @@ function SpanBar(props: TCommonProps) {
...
-              key={key}
-              data-testid="SpanBar--criticalPath"
-              className="SpanBar--criticalPath"
-              style={{
-                background: 'black',
-                left: toPercentInDecimal(criticalPathViewStart),
-                width: toPercentInDecimal(criticalPathViewEnd - criticalPathViewStart),
-              }}
-            />
+            <Tooltip placement="top" title="Critical Path">
```
yurishkuro, line 149:
```suggestion
            <Tooltip
              placement="top"
              title={
                <div>
                  A segment on the <em>critical path</em> of the overall trace/request/workflow.
                </div>
              }
            >
```

yurishkuro (commented):
Thanks, looks good. I suggested a change to the wording to be a bit more explicit. An even more advanced form would be to use display similar to logs where we show a full list of CP segments in the span with some of their stats, like start, end, % of total CP. But we can start with just static popup.

There is one issue currently, though. When the span also has log ticks, it's almost impossible to get the logs popup to show if there's an also a CP segment in the same timeframe. Somehow we need to make sure that the logs tooltip takes priority over CP. Perhaps this could be achieved with z-order, I am not sure.

image

cc @anshgoyalevil

anshgoyalevil:
I just gave it a go, and yes, the `zIndex` solution would work here.

```html
<div
  data-testid="SpanBar--logMarker"
  className="SpanBar--logMarker"
  style={{ left: positionKey, zIndex: 3}}
/>
```
(I tried with the value 3 (zIndex). Anything lower than that isn't working. Would recommend a value greater than it though 3 would also work)

anshgoyalevil:
@yurishkuro 
How about an addition? Color change on hover
20dntwuf

yurishkuro:
> How about an addition? Color change on hover

Not a bad idea, as long as the color is distinct from the span color palette. E.g., don't we already use brown for spans?

I also think it would be nice for the log ticks to enlarge when hovered (like using a bolder line), and maybe have a slightly larger clickable bounding box. Right now it takes very fine pointing to get a popup from a log tick.

anshgoyalevil:
> don't we already use brown for spans?
We already use maybe. Not sure though, I chose brown just for example

> I also think it would be nice for the log ticks to enlarge when hovered (like using a bolder line), and maybe have a slightly larger clickable bounding box. Right now it takes very fine pointing to get a popup from a log tick.
As far as I understand, isn't the size of that point so thin, to accommodate all ticks in a continuous timeframe of milliseconds

yurishkuro:
> so thin, to accommodate all ticks in a continuous timeframe of milliseconds

Yes, but when you have too many logs clustered together no matter what you do they will not be distinguishable without zooming in on the timeline. For comparison, Zipkin UI used to use oval pills to represent event (much thicker than our ticks):

image

anshgoyalevil:
Ah yes, I forgot that the end users can zoom in on the timeline.

GLVSKiriti:
> Not a bad idea, as long as the color is distinct from the span color palette. E.g., don't we already use brown for spans?

What about color White?
cpOnHover

yurishkuro:
We have many light colors like yellow in the same pic, I'm not sure white will be contrasting enough on those. May we could do animation of white spot running left to right once?

GLVSKiriti:
> May we could do animation of white spot running left to right once?

We can adjust the spot width and time 
singleSpan

But if child spans are collapsed then it would be like this
paretnSpan

Is this ok?

yurishkuro:
The first s/s is awesome!  

When children are collapsed, wouldn't the CP segments have identical edge timestamps, like [t1, t2], [t2, t3], ...? If so, why don't we merge them? It would help this running spot, and in general it's meaningless to have adjacent CP segments without a gap between them.

GLVSKiriti:
> When children are collapsed, wouldn't the CP segments have identical edge timestamps, like [t1, t2], [t2, t3], ...? If so, why don't we merge them? It would help this running spot, and in general it's meaningless to have adjacent CP segments without a gap between them.

Yep I merged the CP segments and the result would be like this!
Screencast from 17-10-23 11 44 56 AM IST

> I also think it would be nice for the log ticks to enlarge when hovered (like using a bolder line),

And I also increased size of log ticks !

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBar.tsx
```diff
@@ -86,6 +88,23 @@ function SpanBar(props: TCommonProps) {
...
+      const critcalPathViewBounds = getViewedBounds(each.section_start, each.section_end);
+      const criticalPathViewStart = critcalPathViewBounds.start;
+      const criticalPathViewEnd = critcalPathViewBounds.end;
+
+      accumulator.ViewStart = Math.min(accumulator.ViewStart, criticalPathViewStart);
+      accumulator.ViewEnd = Math.max(accumulator.ViewEnd, criticalPathViewEnd);
+
+      return accumulator;
+    }, criticalPathWhenParentCollapsed);
+  }
```
yurishkuro, line 106: I feel like this is overly complicated and mixes view & data concerns. This is how I would do it:
  * if children are collapsed, re-compute the `criticalPath` array by merging adjacent segments (and btw I am not seeing where you check that they are in fact adjacent, so you're making an assumption that it's a contiguous area, an assumption that is not necessary in this piece of code)
  * no changes / bifurcation would be required in the view code below, it still works off a single `criticalPath` array

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBar.css
```diff
@@ -78,13 +104,15 @@ limitations under the License.
   background-color: rgba(0, 0, 0, 0.5);
   cursor: pointer;
   height: 60%;
-  min-width: 1px;
+  min-width: 2px;
```
yurishkuro, line 107: is it possible to make the bounding box of the ticks larger, instead of ticks themselves? My concern was that it was difficult to move the mouse precisely to cause a popup, going from 1px to 2px does not really help. But I did not have a problem with the visual being 1px.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
```diff
@@ -358,24 +358,39 @@ export class VirtualizedTraceViewImpl extends React.Component<VirtualizedTraceVi
...
-            const currentChildSpan = trace.spans.find(a => a.spanID === eachId)!;
-            if (currentChildSpan.hasChildren) {
-              allChildSpanIds.push(...currentChildSpan.childSpanIds);
-              findAllDescendants(currentChildSpan.childSpanIds);
-            }
-          });
-        };
-        findAllDescendants(childSpanIds);
-        return allChildSpanIds.includes(each.spanId);
+
```
yurishkuro, line 361: this is a large chunk of code, please move to a function

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
```diff
@@ -358,24 +358,39 @@ export class VirtualizedTraceViewImpl extends React.Component<VirtualizedTraceVi
...
+    // Define an array to store the IDs of the span and its descendants (if the span is collapsed)
+    const allRequiredSpanIds = [spanID];
+
+    // If the span is collapsed, recursively find all of its descendants.
+    if (isCollapsed) {
```
yurishkuro, line 366: Q: if any of this logic required when the span is NOT collapsed? If not, please refactor to:
```
const criticalPathToRender = isCollapsed ? mergeChildrenCP() : criticalPath;
```

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
```diff
@@ -358,24 +358,39 @@ export class VirtualizedTraceViewImpl extends React.Component<VirtualizedTraceVi
...
+          if (currentChildSpan.hasChildren) {
+            allRequiredSpanIds.push(...currentChildSpan.childSpanIds);
+            findAllDescendants(currentChildSpan.childSpanIds);
+          }
+        });
+      };
+      findAllDescendants(allRequiredSpanIds);
+    }
+    const criticalPathSections: criticalPathSection[] = [];
+    // Make a deep copy of the criticalpath to avoid reference errors
```
yurishkuro, line 379: can you elaborate? what reference errors?

I think it's better to create new segments on the fly and not risk changing the original, but not actually clone the original as a whole.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
```diff
@@ -358,24 +358,39 @@ export class VirtualizedTraceViewImpl extends React.Component<VirtualizedTraceVi
...
+    const criticalPathCopy: criticalPathSection[] = JSON.parse(JSON.stringify(criticalPath));
+    criticalPathCopy.forEach(each => {
+      if (allRequiredSpanIds.includes(each.spanId)) {
```
yurishkuro, line 382: this is confusing - at this point doesn't `criticalPathCopy` contain only segments of the current span? How would segments children spans get there to make this condition true?

GLVSKiriti: `criticalPathCopy` means it is just a deep copy of all critical paths.

but anyways I removed it !

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
```diff
@@ -358,24 +393,9 @@ export class VirtualizedTraceViewImpl extends React.Component<VirtualizedTraceVi
...
-          });
-        };
-        findAllDescendants(childSpanIds);
-        return allChildSpanIds.includes(each.spanId);
-      }
-      return each.spanId === spanID;
-    });
+    const criticalPathSections = isCollapsed
+      ? mergeChildrenCriticalPath(trace, spanID, criticalPath)
+      : criticalPath.filter(each => each.spanId === spanID);
```
yurishkuro, line 398: just to confirm: `criticalPath` input parameter here is the _full CP of the trace_, not just the current span, correct?

GLVSKiriti: Yes!

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
```diff
@@ -158,6 +158,41 @@ function getCssClasses(currentViewRange: [number, number]) {
...
+  spanID: string,
+  criticalPath: criticalPathSection[]
+): criticalPathSection[] {
+  // Define an array to store the IDs of the span and its descendants (if the span is collapsed)
+  const allRequiredSpanIds = [spanID];
+
+  // If the span is collapsed, recursively find all of its descendants.
+  const findAllDescendants = (currentChildSpanIds: string[]) => {
+    currentChildSpanIds.forEach(eachId => {
+      const currentChildSpan = trace.spans.find(a => a.spanID === eachId)!;
```
yurishkuro, line 172: We should really just add a helper method to the Trace object: `trace.findSpan(spanID)`

yurishkuro (approved):

yurishkuro:
thanks!

GLVSKiriti:
what about adding a helper method `trace.findSpan(spanID)`

yurishkuro:
Well, to be honest, and I raised this during the project, this whole searching should go away. We build a proper tree in one place, we should just keep it as part of the trace and always navigate it, instead of doing the same work over and over. All these searches are O(N).

bobrik:
This might be pretty taxing on CPU usage. We're seeing very slow tag expansion in spans in large traces.

Here's me clicking to expand the tags and then to collapse them back:

image

image

yurishkuro:
@bobrik please open a new ticket. You can disable critical path for all traces via config, but there's certainly room for perf optimization.

bobrik:
I filed #2179 for this.

PR #1901, opened 2023-10-21
[Refactor]: Enhancement to selfTime calculation logic in TraceStatistics view

### LINKED ISSUE #877: Unstable unit test on TraceStatistics/tableValues.tsx
Based on #876 CI run, the code coverage came up with a change, even though no source code changes were made.

There as places with coverage changes in packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx:

image
image
image

### LINKED ISSUE #1263: [Enhancement]: Self time Calculation to be unified across multiple screens

### LINKED ISSUE #1592: [Bug]: Negative Self time in Trace Statistics view

GLVSKiriti:
- Resolves #1592  & Resolves #1263 
- Indirectly Resolves #877 

## Description of the changes
-  When calculating selfTime of a span  
- selfTime is the total time spent in the span when it was not waiting on children.
  - So we skip the `FOLLOWS_FROM` type child spans 
- selfTime = duration - (total time consumed by child spans between span startTime and endTime)
- This is implemented by using `DRange` similar to the logic in TraceGraph   
@yurishkuro


GLVSKiriti:
unitTests
These are three tests which are failing

GLVSKiriti:
1. [traceWithOverlappingChildrenLongerThanParent](https://github.com/jaegertracing/jaeger-ui/blob/main/packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValuesTestTrace/traceWithOverlappingChildrenLongerThanParent.json)
a10
and final value of selfTotal is calculated from this 
https://github.com/jaegertracing/jaeger-ui/blob/1f63ee05ac987487d9207f7dba4626f50cceadf7/packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx#L254
so the final selfTotal = 0.03 but I don't know why it is 10.33 in unit test 

2. [traceWithTwoNonOverlappingChildren](https://github.com/jaegertracing/jaeger-ui/blob/main/packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValuesTestTrace/traceWithTwoNonOverlappingChildren.json)
a11
similarly selfTotal = 0.03

3. [traceWithSingleChildSpanLongerThanParent](https://github.com/jaegertracing/jaeger-ui/blob/main/packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValuesTestTrace/traceWithSingleChildSpanLongerThanParent.json)
a12
similarly selfTotal = 0.04

GLVSKiriti:
Is this calculations are correct Or am I doing wrong ?

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
```diff
@@ -13,84 +13,14 @@
...
-  for (let i = 0; i < noOverlay.length; i++) {
-    duration += noOverlay[i].duration;
-  }
-  tempSelfChange += span.duration - duration;
-
-  return tempSelfChange;
-}
-
 /**
  * Used to calculated the content.
```
yurishkuro, line 25: can we improve the function name / comment? What does "calculate content" even mean?

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
```diff
@@ -106,125 +36,25 @@ function calculateContent(trace: Trace, span: Span, allSpans: Span[], resultValu
...
-              allChildren[i].relativeStartTime + allChildren[i].duration >= allChildren[j].relativeStartTime
-            ) {
-              kinderSchneiden = true;
-              allOverlay.push(allChildren[i]);
-              allOverlay.push(allChildren[j]);
-            }
-          }
-        }
+    const spanDRange = new DRange(span.startTime, span.startTime + span.duration);
+    const childrenDrange = new DRange();
```
yurishkuro, line 41: I don't know why you need to complicate this with a separate range. I think this should work just fine:
```
const spanRange = new DRange(span.startTime, span.startTime + span.duration);
const children = ... // find real children
children.forEach(child => {
    spanRange.subtract(child.startTime, child.startTime + child.duration);
});
selfTime = spanRange.map(r => r.high - r.low).reduce((sum, v) => sum + v, 0);
```

One thing you need to be careful is off-by-one errors. From my understanding drange builds intervals as whole numbers, it does not understand inclusive/exclusive boundaries. For example, when you define `DRange(start, start+duration)`, you're really defining half-open interval `[start, end)` with `selfTime==end-start`, but it's not how drange works. For example:
```
var drange = require("drange");
let r = new drange(1, 10);
r.add(20, 30);
console.log(r.toString()); // "[ 1-10, 20-30 ]"
r.subtract(8, 25);
console.log(r.toString()); // "[ 1-7, 26-30 ]"
```
Notice that even though I logically subtracted `[8, 25)`, the first remaining segment is `1-7`. even though we really want `[1, 8)` half-open interval. So you can compute ST=7-1=6 or ST=8-1=7 - off-by-one (7 is the correct one in this situation).

yurishkuro:
Regarding the tests, I would recommend replacing the times / durations in the fixture files with smaller numbers that are easier to grok without a calculator, like start=100000, duration=100 (pick values to retain the overall magnitude, but make the numbers round)

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
```diff
@@ -106,125 +36,25 @@ function calculateContent(trace: Trace, span: Span, allSpans: Span[], resultValu
...
+    span.childSpanIds.forEach(eachId => {
+      const childSpan = allSpans.find(a => a.spanID === eachId && a.references[0].refType === 'CHILD_OF');
```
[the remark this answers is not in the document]
GLVSKiriti: > you need to check that the reference is actually to the right parent span (it could be pointing to a different trace altogether).

Why to check again, explicitly span.childSpanIds have span ids of its children only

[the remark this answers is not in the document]
yurishkuro: yes, but span.References can point upwards to many "parent" spans (not really parent, but happened-before spans)

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
GLVSKiriti: https://github.com/jaegertracing/jaeger-ui/blob/3db728fcac1795c003c37f77118cc4ed05154513/packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx#L49-L52
I am very bad in writing comments 😅

> range builds intervals as whole numbers, it does not understand inclusive/exclusive boundaries.

As you explained above

`let r = new drange(1, 10)`
means r = {1,2,3,4,5,6,7,8,9,10} which we can consider as `1<= t <= 10`

`r.subtract(2, 5)`

means r - {2,3,4,5} = {1,6,7,8,9,10} which we can consider as `t=1 and 5<=t<=10`  (Which is wrong)

but we do +1 and -1 to avoid those errors
`r.subtract(2+1, 5-1)`

means r - {3,4} = {1,2,5,6,7,8,9,10} which we can consider as `1<=t<=2 and 5<=t<=10` (Which is correct)

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
GLVSKiriti: calculateContent function helps in calculating column values of particular row 

As there may be more than one span with same name or group which is explained by`count` column.

So the calculateContent finds the selfTime of the span and changes table column values of that particular row (i.e ST total,ST max,MIn,Max etc...).

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.test.js
```diff
@@ -506,12 +506,12 @@ describe('check self time', () => {
       'Service Name',
       transformedTraceWithOverlappingChildrenLongerThanParent
     );
-    expect(serviceOne.selfTotal).toBe(10.33);
+    expect(serviceOne.selfTotal).toBe(0.03);
```
yurishkuro, line 509: If all the timestamps are integers, how do we end up with these fractional values?

GLVSKiriti: https://github.com/jaegertracing/jaeger-ui/blob/1f63ee05ac987487d9207f7dba4626f50cceadf7/packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx#L245-L260

becuase column values are calculated from this function finally.

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
```diff
@@ -106,125 +36,26 @@ function calculateContent(trace: Trace, span: Span, allSpans: Span[], resultValu
...
+      each =>
+        span.childSpanIds.includes(each.spanID) &&
+        each.references[0].spanID === span.spanID &&
+        each.references[0].refType === 'CHILD_OF'
+    );
+    children.forEach(child => {
+      // A DRange(start,end) builds interval as whole numbers.
+      // So when subtarcting a DRange directly do (+1 , -1) to (start,end) otherwise
+      // for each subtaracting of a DRange it will take extra 2 microseconds
+      spanRange.subtract(child.startTime + 1, child.startTime + child.duration - 1);
```
yurishkuro, line 52: what if `child.duration == 1`?  It will be invalid drange

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.test.js
yurishkuro: Do you mean that this is a value divided by 1000? If so that's testing the display behavior, but we should have unit tests that validate calculations themselves, in the original units (which would be integers)

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
GLVSKiriti: yes, I think it is an exceptional case.
May be we should write an exceptional code in if block `if(span.duaration <=1)`
```
const n;
if(child.duration <=1) n += 1;
else{
   spanRange.subtract(child.startTime + 1, child.startTime + child.duration - 1);
}
```

finally we should subtract n from selfTime
may be at here temself = tempself + selfTime - n
https://github.com/jaegertracing/jaeger-ui/blob/53dd0df12b554313956876d0e9aec690e24990e1/packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx#L58

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.test.js
GLVSKiriti: Yes but every test in tableValues.test.js is testing display behaviour i.e, each column value we see in TraceStatistics view
https://github.com/jaegertracing/jaeger-ui/blob/1f63ee05ac987487d9207f7dba4626f50cceadf7/packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.test.js#L80-L100

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
yurishkuro: > finally we should subtract n from selfTime

this would only be correct if no other child overlaps with that d=1 child.

Maybe another way to do this is to fool drange to support open intervals. For example, if we multiply all values by 10 and create the ranges as `[10*startTime, 10 * (startTime+duration) - 1]`. So what would be `[10, 20]` in the current code would become `[100, 199]`. Then subtracting ranges would always result in the upper bound of left remaining segment to end with 9 and the lower bound of the next segment to be multiple of 10, i.e. we'd never have improper overlaps.

GLVSKiriti: Yes that works!!

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
```diff
@@ -106,125 +36,23 @@ function calculateContent(trace: Trace, span: Span, allSpans: Span[], resultValu
...
+        each.references[0].spanID === span.spanID &&
+        each.references[0].refType === 'CHILD_OF'
+    );
+    children.forEach(child => {
+      spanRange.subtract(10 * child.startTime, 10 * (child.startTime + child.duration) - 1);
+    });
+    const selfTime = spanRange
+      .subranges()
+      .map(r => r.high - r.low)
+      .reduce((sum, v) => sum + v, 0);
```
yurishkuro, line 54: This simpler form might work now:
```suggestion
    const selfTime = spanRange.length / 10;
```

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValuesTestTrace/traceWithSingleChildSpanLongerThanParent.json
```diff
@@ -32,8 +32,8 @@
           "spanID": "006c3cf93508f205"
         }
       ],
-      "startTime": 1679437737491529,
-      "duration": 79182,
+      "startTime": 200,
+      "duration": 300,
```
yurishkuro, line 36: file name doesn't quite match - the child is not only longer, but starts way outside of parent range.

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValuesTestTrace/traceWithTwoNonOverlappingChildren.json
```diff
@@ -32,7 +32,7 @@
           "spanID": "006c3cf93508f205"
```
yurishkuro, line 32: Similar to timestamps, these test files should really be using manually defined span IDs, like `parentSpan`, `childSpanA`, ...

If I were writing these tests, I would've created the spans directly in the code, with required values.

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
GLVSKiriti: in DRange , length = high - low +1;

if spanRange has more than 10 subranges then there will additional 10 micro if we do spanRange.length.
Then we get extra 1 micro for selfTime.

May be we should keep the code same or we can do like this

`
const selfTime = (spanRange.length - spanRange.subranges().length)/10;`

yurishkuro: > in DRange , length = high - low +1;

but that's exactly the formula that we want. If we start with a span [10, 20] (duration = 10), convert to drange as [100, 199], then length will be 199 - 100 + 1 = 100, 100/10 = 10.

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValuesTestTrace/traceWithTwoNonOverlappingChildren.json
GLVSKiriti: > I would've created the spans directly in the code, with required values.

How will you do this?? is there any function to create the span directly in code?

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValuesTestTrace/traceWithSingleChildSpanLongerThanParent.json
GLVSKiriti: `traceWithSingleChildLongerThanParentAndStartsAfterParent`

May be this would be good! But too long

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
GLVSKiriti: Yeah! got it

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValuesTestTrace/traceWithTwoNonOverlappingChildren.json
yurishkuro: There may be functions somewhere, but the span structure is pretty simple, it's easy to define a local helper function that takes only the arguments that need to vary across the tests, like span id, parent id, start / end, maybe operation name if group-by is being tested

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
```diff
@@ -106,125 +33,30 @@ function calculateContent(trace: Trace, span: Span, allSpans: Span[], resultValu
...
-            tempSelf -= allChildren[i].duration;
-          }
-        }
-      } else {
-        // Overlay
-        tempSelf = onlyOverlay(allOverlay, allChildren, tempSelf, span);
-      }
-    }
-    // no children
+    // DRange treats all intervals as inclusive, even for subtract operations.
```
yurishkuro, line 37:
```suggestion
```
my mistake

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
```diff
@@ -106,125 +33,30 @@ function calculateContent(trace: Trace, span: Span, allSpans: Span[], resultValu
...
+    // We want to represent spans as half-open intervals like [startTime, startTime + duration).
+    // This way the subtraction preserves the right boundaries. However, DRange treats all
+    // intervals as exclusive. For example,
```
yurishkuro, line 40:
```suggestion
    // intervals as inclusive. For example,
```

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
```diff
@@ -106,125 +33,30 @@ function calculateContent(trace: Trace, span: Span, allSpans: Span[], resultValu
...
+      each =>
+        span.childSpanIds.includes(each.spanID) &&
+        each.references[0].spanID === span.spanID &&
+        each.references[0].refType === 'CHILD_OF'
+    );
+    children.forEach(child => {
+      spanRange.subtract(10 * child.startTime, 10 * (child.startTime + child.duration) - 1);
+    });
+    const selfTime = spanRange.length;
+    tempSelf += Math.round(selfTime / 10);
```
yurishkuro, line 59:
```suggestion
    tempSelf += Math.round(spanRange.length / 10);
```
Semantically the first line is incorrect. Also, it's not clear to me why this is using `tempSelf +=` - isn't there only one pass through the code? Why accumulation?

I would suggest moving this out to a separate function computeSelfTime.

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
```diff
@@ -13,88 +13,42 @@
...
-      lowestStartTime = temp.lowestStartTime;
-      duration = temp.duration;
-    }
-  } while (resultGetDuration.allOverlay.length > 1);
-  duration = resultGetDuration.duration + totalDuration;
-  // no cut is observed
-  for (let i = 0; i < noOverlay.length; i++) {
-    duration += noOverlay[i].duration;
+function computeSelfTime(span: Span, allSpans: Span[]): number {
+  if (span.hasChildren) {
```
yurishkuro, line 25: nit: avoid extra nesting
```suggestion
  if (!span.hasChildren) {
      return span.duration;
  }
```

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
```diff
@@ -13,88 +13,42 @@
...
+    // This way the subtraction preserves the right boundaries. However, DRange treats all
+    // intervals as inclusive. For example,
+    //       range(1, 10).subtract(4, 8) => range([1, 3], [9-10])
+    //       length=(3-1)+(10-9)=2+1=3
+    // In other words, we took an interval of length=10-1=9 and subtracted length=8-4=4.
+    // We should've ended up with length 9-4=5, but we got 3.
+    // To work around that, we multiply start/end times by 10 and subtract one from the end.
+    // So instead of [1-10] we get [10-99]. This makes the intervals work like half-open.
+    const spanRange = new DRange(10 * span.startTime, 10 * (span.startTime + span.duration) - 1);
+    // Filter for CHILD_OF we don't want to calculate FOLLOWS_FROM
```
yurishkuro, line 36:
```suggestion
    // Only keep CHILD_OF spans. FOLLOWS_FROM spans do not block the parent.
```

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
```diff
@@ -315,18 +147,18 @@ function valueFirstDropdown(selectedTagKey: string, trace: Trace) {
     for (let j = 0; j < allSpans.length; j++) {
```
yurishkuro, line 147: Just to illustrate my earlier point about linear searches: with this loop in place we're doing O(N^2) work.

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
```diff
@@ -13,88 +13,40 @@
...
-  } while (resultGetDuration.allOverlay.length > 1);
-  duration = resultGetDuration.duration + totalDuration;
-  // no cut is observed
-  for (let i = 0; i < noOverlay.length; i++) {
-    duration += noOverlay[i].duration;
-  }
-  tempSelfChange += span.duration - duration;
-
-  return tempSelfChange;
+  return Math.round(spanRange.length / 10);
```
yurishkuro, line 46: btw, I don't think rounding is necessary, length should always be divisible by 10.

yurishkuro (approved):

yurishkuro:
Thanks!

yurishkuro:
@maxgaponov you may be right

PR #1902, opened 2023-10-21
Added tooltip when hovered on header column of TraceStatistics view

### LINKED ISSUE #962: Create Trace Statistics Tooltips / Help Documentation

The UI should be user friendly to new users of Jaeger and describe all the details of what's shown on the page. Particularly, the trace statistics breakdown view has lots of info on it, but not a lot of details which describe what each column means.
image

An acronym "ST" is used in many places, but no place is it the acronym defined.

Jaeger distinguishes "Total" from "ST Total" stats in the stats breakdown table, but it is very hard to find any documentation that explains the difference. I'm sure it is simple to explain, but finding that info as a beginner has been difficult.

A few approaches:

- Attach hover-over tooltips / helpers to the columns in trace statistics which explain what each column is for
- Provide a web page in documentation that explains in detail each of the trace stat columns and link to that with a help button on the stats breakdown page

GLVSKiriti:
- Resolves #962 

## Description of the changes
- Attach hover-over tooltips  to the columns in trace statistics which explain what each column is for


GLVSKiriti:
Result would be like this!
Screencast from 21-10-23 11 51 39 PM IST(2)

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -44,56 +45,68 @@ const columnsArray: any[] = [
     title: 'Name',
     attribute: 'name',
     suffix: '',
+    titleDescription: 'Service Name of a span',
```
yurishkuro, line 48: I think this is not always the case, since the table allows grouping by different criteria.

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -44,56 +45,68 @@ const columnsArray: any[] = [
...
     title: 'Avg',
     attribute: 'avg',
     suffix: 'ms',
+    titleDescription: 'Average duaration of all spans',
   },
   {
     title: 'Min',
     attribute: 'min',
     suffix: 'ms',
+    titleDescription: 'Minimum duaration from the duaration of all spans',
```
yurishkuro, line 72:
```suggestion
    titleDescription: 'Minimum duration across all spans',
```

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -44,56 +45,68 @@ const columnsArray: any[] = [
...
     attribute: 'max',
     suffix: 'ms',
+    titleDescription: 'Maximum duaration from the duaration of all spans',
   },
   {
     title: 'ST Total',
     attribute: 'selfTotal',
     suffix: 'ms',
+    titleDescription:
+      'Sum of Self time(Total time spent in a span when it was not waiting on children.) of all spans',
```
yurishkuro, line 85:
```suggestion
      'Sum of Self Time (time spent in a span when it was not waiting on children)',
```

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -44,56 +45,68 @@ const columnsArray: any[] = [
...
   },
   {
     title: 'ST Avg',
     attribute: 'selfAvg',
     suffix: 'ms',
+    titleDescription: 'Average value of self time of all spans',
```
yurishkuro, line 91:
```suggestion
    titleDescription: 'Average Self Time (time spent in a span when it was not waiting on children)',
```

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -44,56 +45,68 @@ const columnsArray: any[] = [
...
   },
   {
     title: 'ST Min',
     attribute: 'selfMin',
     suffix: 'ms',
+    titleDescription: 'Minimum value from self time of all spans',
```
yurishkuro, line 97:
```suggestion
    titleDescription: 'Minimum Self Time (time spent in a span when it was not waiting on children)',
```

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -44,56 +45,68 @@ const columnsArray: any[] = [
...
   },
   {
     title: 'ST Max',
     attribute: 'selfMax',
     suffix: 'ms',
+    titleDescription: 'Maximum value from self time of all spans',
```
yurishkuro, line 103:
```suggestion
    titleDescription: 'Maximum Self Time (time spent in a span when it was not waiting on children)',
```

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -44,56 +45,68 @@ const columnsArray: any[] = [
...
   },
   {
     title: 'ST in Duration',
     attribute: 'percent',
     suffix: '%',
+    titleDescription: 'Percentage of self time in the total duration of all spans',
```
yurishkuro, line 109:
```suggestion
    titleDescription: 'Percentage of ST Total vs. Total',
```

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -33,6 +33,7 @@ type State = {
   sortIndex: number;
   sortAsc: boolean;
   showPopup: boolean;
+  showSorterTooltip: boolean;
```
yurishkuro, line 36: why is it needed to be in the state?

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -44,56 +45,68 @@ const columnsArray: any[] = [
...
     title: 'Count',
     attribute: 'count',
     suffix: '',
+    titleDescription: 'Number of spans',
   },
   {
     title: 'Total',
     attribute: 'total',
     suffix: 'ms',
+    titleDescription: 'Total duaration of all spans',
```
yurishkuro, line 60:
```suggestion
    titleDescription: 'Total duration of all spans',
```

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -44,56 +45,68 @@ const columnsArray: any[] = [
...
   },
   {
     title: 'Avg',
     attribute: 'avg',
     suffix: 'ms',
+    titleDescription: 'Average duaration of all spans',
```
yurishkuro, line 66:
```suggestion
    titleDescription: 'Average duration of all spans',
```

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -44,56 +45,68 @@ const columnsArray: any[] = [
...
     title: 'Min',
     attribute: 'min',
     suffix: 'ms',
+    titleDescription: 'Minimum duaration from the duaration of all spans',
   },
   {
     title: 'Max',
     attribute: 'max',
     suffix: 'ms',
+    titleDescription: 'Maximum duaration from the duaration of all spans',
```
yurishkuro, line 78:
```suggestion
    titleDescription: 'Maximum duration across all spans',
```

yurishkuro (commented):
I find it odd that the click tooltip shows on a large portion of the header. It would make sense to show the main tooltip across most of the header, and only show click tooltip when hovering over the up/down icons.

Also, I think click tooltip isn't really needed, it's pretty standard behavior to sort by clicking.

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
GLVSKiriti: previously when I hover on header cell then both tootlips are displayed (our custom tooltip, and antd click tooltip)
So to not display antd click tooltip when I hover on title I used a state.
> Also, I think click tooltip isn't really needed, it's pretty standard behavior to sort by clicking.

So now updated code only shows custom tooltip on hover

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
GLVSKiriti: `Tag name by which data is grouped and sub-grouped` how would be this?

yurishkuro: I suggest not adding a tooltip for this one, and changing `title: 'Group'`

yurishkuro (approved):
Thanks!

PR #1907, opened 2023-10-24
Upgrade ant-design to v5.x

anshgoyalevil:
- fixes https://github.com/jaegertracing/jaeger-ui/issues/1703

## Description of the changes
- The Size of this PR may look big, but there were no breaking changes except the following:
  - ant-design v5.0.0 no longer supports less variables. That means we have to use the `seedToken` and `aliasToken` in place of less variables.
  - With this change, we also needed to use the default styling of ant-design v4. For that, ant-design team provides a `compatible` package.
  - The notification.`close` and `warn` have been renamed `destory` and `warning` respectively.
- PR: https://github.com/jaegertracing/jaeger-ui/pull/1839/files is reversed in this PR, because we are not really on @types/react v18. We are locked on v16 due to yarn resolution. Having v18 written in package.json was creating conflict with the installation of antd v5

[screen-capture (5).webm](https://github.com/jaegertracing/jaeger-ui/assets/94157520/550a79b7-b333-497b-9b73-70d0e6a1a748)

- Manually, and unit tests

## Checklist


anshgoyalevil:
@yurishkuro The size of this PR is quite big, but I wasn't able to distill it down to multiple PRs, because all the breaking changes were introduced in v5.0.0. v5.latest didn't bring any other breaking changes, except the styling changes, which we are already on v4.

The size grew this big due to snapshot changes 📷

packages/jaeger-ui/src/components/App/TopNav.tsx
```diff
@@ -146,6 +146,7 @@ export function TopNavImpl(props: Props) {
         disabledOverflow
         selectedKeys={[pathname]}
         items={itemsGlobalRight}
+        style={{ color: '#fff' }}
```
yurishkuro, line 149: Shouldn't this be in css somewhere?

packages/jaeger-ui/vite.config.ts
```diff
@@ -64,33 +64,6 @@ export default defineConfig({
       less: {
         math: 'always',
         javascriptEnabled: true,
-        modifyVars: {
-          // Supply appropriate overrides to the Ant Design System.
```
yurishkuro, line 68: Can you explain what's happening here?

packages/jaeger-ui/src/components/App/index.jsx
```diff
@@ -38,10 +40,43 @@ import prefixUrl from '../../utils/prefix-url';
 
 import '../common/vars.css';
 import '../common/utils.css';
+import 'antd/dist/reset.css';
 import './index.css';
 import { history, store } from '../../utils/configure-store';
 import { HistoryProvider } from '../../utils/useHistory';
 
+const jaegerTheme = {
```
yurishkuro, line 48: This feels like a possibility to support light/dark themes in the future. Since you were working on this now, do you have an understanding how that can be done? If so, could you book a ticket with an overview of the approach?

packages/jaeger-ui/src/components/App/TopNav.tsx
anshgoyalevil: Yes, but, since the change was a one-liner, I thought of adding an inline style. Moving it to the CSS file.

packages/jaeger-ui/vite.config.ts
anshgoyalevil: Here, we are basically removing the replacement variables.

Before v5, we had an option to replace less variables in ant-design's core files with the ones defined here.
This `modifyVars` object was containing all such variables. But since v5, antd doesn't support less variables, so, this modification thing is of no use.

Instead, ant-design introduced seedTokens and aliasTokens which can be supplied to the theme directly.

packages/jaeger-ui/src/components/App/index.jsx
anshgoyalevil: Sure. Creating a ticket 🚀

packages/jaeger-ui/vite.config.ts
yurishkuro: is there documentation about that? The names seedToken don't tell me much, especially why they would be a replacement for less variables.

On a high level, what is the approach now? My understanding was that less variables allowed us to reuse settings across different css files. Do we have to replicate those now in many places?

anshgoyalevil: [ant-design v4 to v5 migration guide](https://ant.design/docs/react/migration-v5#less-migration:~:text=Remove%20less%2C%20adopt,antd%20specification%20style.) explaining why they removed the support for less variables.

[This document](https://ant.design/docs/react/customize-theme#life-of-design-token) explains in deep how to migrate from less variables to tokens and what exactly is all that seedToken, and other such token thing.

Since we were replacing those variables via the `modifyVars` object, the same can be (is achieved) using the design tokens supplied to the theme configuration of ant-design.

less variables for sure achieves the purpose of reusing them across files, but since the change is internal to ant-design library, we aren't affected in other areas, as we aren't using any antd's less file in our code.

yurishkuro (approved):

yurishkuro:
🚀 🚀 🚀

PR #1916, opened 2023-10-28
Fix linting warnings for Function, Boolean and String types

DerenToy:
- Resolves part of no-unused-vars warnings for #1608 

## Description of the changes
-  Fixed ESlint warnings for Function, Boolean and String types

- By running "yarn lint" command

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/Monitor/ServicesView/operationDetailsTable/index.tsx
+++ b/packages/jaeger-ui/src/components/Monitor/ServicesView/operationDetailsTable/index.tsx
@@ -112,9 +112,7 @@ export class OperationTableDetails extends React.PureComponent<TProps, TState> {
               error={error.opsLatencies}
             />
             <div className="table-graph-avg">
-              {row.dataPoints.service_operation_latencies.length > 0
-                ? formatTimeValue(value * 1000)
-                : ''}
+              {row.dataPoints.service_operation_latencies.length > 0 ? formatTimeValue(value * 1000) : ''}
             </div>
           </div>
         ),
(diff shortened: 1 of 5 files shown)


packages/jaeger-ui/src/components/Monitor/ServicesView/operationDetailsTable/index.tsx
```diff
@@ -112,9 +112,7 @@ export class OperationTableDetails extends React.PureComponent<TProps, TState> {
               error={error.opsLatencies}
             />
             <div className="table-graph-avg">
-              {typeof value === 'number' && row.dataPoints.service_operation_latencies.length > 0
-                ? formatTimeValue(value * 1000)
-                : ''}
+              {row.dataPoints.service_operation_latencies.length > 0 ? formatTimeValue(value * 1000) : ''}
```
yurishkuro, line 115: This replacement is not completely equivalent. What would be a situation when `value` would not be a number?

yurishkuro (commented):
lgtm, except the question about the first file

packages/jaeger-ui/src/components/Monitor/ServicesView/operationDetailsTable/index.tsx
DerenToy: Hello,

In this case, the value of 'value' is 'ServiceOpsMetrics['latency']', and 'latency' is defined as a number type.
 
 export type ServiceOpsMetrics = {
  dataPoints: OpsDataPoints;
  errRates: number;
  impact: number;
  latency: number;
  name: string;
  requests: number;
  key: number;
};

Therefore, I thought that there is no need for additional type checking because 'value' is always of type number. Am I mistaken?

yurishkuro (approved):

yurishkuro:
Thanks!

DerenToy:
@yurishkuro 🙏🏻🙏🏻🙏🏻

PR #1920, opened 2023-10-30
Fix time calculation in Trace Graph view

### LINKED ISSUE #1918: [Bug]: Duration of overlapping operations is calculating incorrectly in TraceGraph view
### What happened?

Currently the duration of an operation in Trace Graph view is calculating as a sum of the invocations of this operation. It works incorrectly in the case of overlapping spans and leads to the situation where the duration of a child can be greater than the duration of its parent. My proposal is to calculate the duration of the union of the spans like we do for child nodes.

### Steps to reproduce

1. Open [example_trace.json](https://github.com/jaegertracing/jaeger-ui/files/13204774/example_trace.json) in Jaeger UI
2. Navigate to Trace Graph view

### Expected behavior

Expected that the duration of a node is not greater than the duration of its parent.
The total duration for overlapping spans should be the duration of their union, but not the sum.

### Relevant log output

_No response_

### Screenshot

image

### Additional context

_No response_

### Jaeger backend version

_No response_

### SDK

_No response_

### Pipeline

_No response_

### Stogage backend

_No response_

### Operating system

_No response_

### Deployment model

_No response_

### Deployment configs

_No response_

maxgaponov:
- Resolves #1918

## Description of the changes
- Now the total duration is calculating as the duration of the union of spans.

- Modified existing tests

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceGraph/calculateTraceDagEV.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceGraph/calculateTraceDagEV.tsx
@@ -81,7 +81,12 @@ export function calculateTraceDag(trace: Trace): TraceDag<TSumSpan & TDenseSpanM
   const dag = new TraceDag<TSumSpan & TDenseSpanMembers>();
 
   baseDag.nodesMap.forEach(node => {
-    const ntime = node.members.reduce((p, m) => p + m.span.duration, 0);
+    const ntime = node.members.reduce(
+      (mdr, m) =>
+        mdr.add(m.span.startTime, m.span.startTime + (m.span.duration <= 0 ? 0 : m.span.duration - 1)),
+      new DRange()
+    ).length;
+    const totalTime = node.members.reduce((p, m) => p + m.span.duration, 0);
     const numErrors = node.members.reduce((p, m) => (p + isError(m.span.tags) ? 1 : 0), 0);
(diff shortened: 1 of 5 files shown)


packages/jaeger-ui/src/components/TracePage/TraceGraph/calculateTraceDagEV.tsx
```diff
@@ -81,7 +81,12 @@ export function calculateTraceDag(trace: Trace): TraceDag<TSumSpan & TDenseSpanM
   const dag = new TraceDag<TSumSpan & TDenseSpanMembers>();
 
   baseDag.nodesMap.forEach(node => {
-    const ntime = node.members.reduce((p, m) => p + m.span.duration, 0);
+    const ntime = node.members.reduce(
+      (mdr, m) =>
+        mdr.add(m.span.startTime, m.span.startTime + (m.span.duration <= 0 ? 0 : m.span.duration - 1)),
+      new DRange()
+    ).length;
+    const totalTime = node.members.reduce((p, m) => p + m.span.duration, 0);
```
yurishkuro, line 89: @GLVSKiriti this sounds similar to the other fix you made in another view. Is there a way to consolidate these calculations?

GLVSKiriti: I don't know exactly but I will take a look on this!

PR #1921, opened 2023-10-31
DAG View Migration to Plexus

prathamesh-mutkure:

This is a draft PR for migrating the `DAG` view from Cytoscape to Plexus

## Description of the changes
- I've added two new tabs to `System Architecture` page, `DAG_DirectedGraph` and `DAG_Diagraph`
- The DAG_DirectedGraph used Plexus `DirectedGraph` component for migration
- The DAG_Diagraph used Plexus `Digraph` component for migration

### Why migration using two components?

- I managed to render the tree using DirectedGraph component but it didn't allow me to override the arrows/edges
- The previous implementation used arrow to display `callCount`
- Thus, I tried implementing the same graph using the Digraph component
- The Diagraph migration is not done yet, thus the canvas is blank

I'll try my best to migrate the previous implementation, we'll go with the component that gives the best results

Screenshot 2023-10-31 at 3 14 25 PM

NA

## Checklist


yurishkuro:
@prathamesh-mutkure so reading through some of the comments, I think `DirectedGraph` is an older component and we probably should avoid using it. All other `jaeger-ui` sources only import `Digraph`, which seems newer. There's also this comment in the readme:
```
// TODO(joe): Update import after killing `DirectedGraph`
import Digraph from 'plexus/Digraph';
```

packages/jaeger-ui/src/components/App/index.jsx
```diff
@@ -92,6 +93,9 @@ export default class JaegerUIApp extends Component {
             <Router history={history}>
               <Page>
                 <Switch>
+                  <Route path="/demo">
+                    <Demo />
+                  </Route>
```
yurishkuro, line 98: make sure to clear up this before final PR

packages/jaeger-ui/src/components/DependencyGraph/_DAG_DirectedGraph.tsx
```diff
@@ -0,0 +1,149 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import React from 'react';
+
+import { LayoutManager, DirectedGraph } from '@jaegertracing/plexus';
```
yurishkuro, line 17: can this file be removed? We don't want to use deprecated `DirectedGraph`

packages/plexus/src/Digraph/SvgEdge.tsx
```diff
@@ -62,15 +64,29 @@ export default class SvgEdge<U = {}> extends React.PureComponent<TProps<U>> {
...
+        <path
+          d={d}
+          fill="none"
+          vectorEffect="non-scaling-stroke"
+          markerEnd={markerEnd}
+          markerStart={markerStart}
+          {...customProps}
+        />
+
+        <text x={labelX - xOffset} y={labelY} fill="#000" fontSize="1rem" fontWeight="bold">
```
yurishkuro, line 86: > fontSize="1rem" fontWeight="bold"

is this something that can be expressed via stylesheet? How do we style the content of the nodes? Maybe `label` should also be a DOM content, not a plain text?

If too much work, ok to keep as plain text in the first iteration, but I'd still like to know the answer to the styling question.

packages/plexus/src/Digraph/SvgEdge.tsx
```diff
@@ -62,15 +64,29 @@ export default class SvgEdge<U = {}> extends React.PureComponent<TProps<U>> {
...
     return (
-      <path
-        d={d}
-        fill="none"
-        vectorEffect="non-scaling-stroke"
-        markerEnd={markerEnd}
-        markerStart={markerStart}
-        {...customProps}
-      />
+      <g>
```
yurishkuro, line 76: what does `<g>` do?

packages/plexus/src/Digraph/SvgEdge.tsx
```diff
@@ -62,15 +64,29 @@ export default class SvgEdge<U = {}> extends React.PureComponent<TProps<U>> {
...
+          d={d}
+          fill="none"
+          vectorEffect="non-scaling-stroke"
+          markerEnd={markerEnd}
+          markerStart={markerStart}
+          {...customProps}
+        />
+
+        <text x={labelX - xOffset} y={labelY} fill="#000" fontSize="1rem" fontWeight="bold">
+          {label}
```
yurishkuro, line 87: I assume inserting this `<text>` element should be conditional on `if (label)` (i.e. non-empty value)

packages/plexus/src/DirectedGraph/builtins/PureEdges.tsx
```diff
@@ -31,6 +31,8 @@ export default class PureEdges extends React.PureComponent<TProps> {
     return layoutEdges.map(edge => (
       <EdgePath
         key={`${edge.edge.from}\v${edge.edge.to}`}
+        id={`${edge.edge.from}-${edge.edge.to}`}
```
yurishkuro, line 34: can you add a comment to this class? What is it used for, what are "pure edges"?

packages/plexus/src/DirectedGraph/builtins/EdgePath.tsx
```diff
@@ -15,6 +15,8 @@
 import * as React from 'react';
 
 type TProps = Record<string, any> & {
+  id: string;
```
yurishkuro, line 18: do we need to change anything under deprecated `DirectedGraph/..`? Can your PR work without that?

packages/plexus/src/types/index.tsx
```diff
@@ -39,6 +39,7 @@ export type TEdge<T = {}> = T & {
   from: TVertexKey;
   to: TVertexKey;
   isBidirectional?: boolean;
+  edgeLabel?: string;
```
yurishkuro, line 42: the prefix `edge` is redundant since the type is `TEdge`

yarn.lock
```diff
@@ -2647,19 +2647,12 @@
   resolved "https://registry.yarnpkg.com/@types/range-parser/-/range-parser-1.2.4.tgz#cd667bcfdd025213aafb7ca5915a932590acdcdc"
```
yurishkuro, line 2647: why are there changes in the lock file? Can you stay on the same lock as the main branch?

packages/jaeger-ui/src/components/App/index.jsx
prathamesh-mutkure: Yeah, this PR is too cluttered and only made for you and @yashrsharma44 to view progress, I'll make another PR with only required changes

packages/jaeger-ui/src/components/DependencyGraph/_DAG_DirectedGraph.tsx
prathamesh-mutkure: Yes it'll be removed in final PR

packages/plexus/src/Digraph/SvgEdge.tsx
prathamesh-mutkure: The code here is for the edge/arrow, nodes can be customised using `renderNode` property which is already being used for this graph, if you're talking about similar customization for edges then the [documentation](https://github.com/jaegertracing/jaeger-ui/tree/main/packages/plexus#edges-layer) says it can be expensive

Screenshot 2023-11-14 at 3 33 00 PM

We can still try a `renderEdge` method and see how it affects the performance

packages/plexus/src/Digraph/SvgEdge.tsx
prathamesh-mutkure: It's there to group `<path>` and `<text>`, just for semantics, similar to `<> ... </>`

packages/plexus/src/Digraph/SvgEdge.tsx
prathamesh-mutkure: Yeah, that should be conditionally rendered

packages/plexus/src/DirectedGraph/builtins/EdgePath.tsx
prathamesh-mutkure: Yes, that was for when I tried `DirectedGraph`, similar changes have been made to `Diagraph`, so we can skip these

packages/plexus/src/types/index.tsx
prathamesh-mutkure: Sure, will make it just `edge`

yarn.lock
prathamesh-mutkure: That was an accident, let me remove it

yurishkuro:
superseded by #1981

PR #1941, opened 2023-11-03
Speed up Trace Statistics view calculation

### LINKED ISSUE #1925: [Feature]: Speed up Trace Statistics view calculation
### Requirement

I would like to view Statistics for traces with a big amount of spans.

### Problem

Currently Trace Statistics view is calculating very slowly which makes it impossible to use for large traces. 

### Proposal

_No response_

### Open questions

_No response_

maxgaponov:
- Resolves #1925

## Description of the changes
- Speed up Trace Statistics view calculation. Now we precompute child spans for each span to calculate statistics. Previously we filtered all spans to find children for each span.

- Existing tests. Ran manually on a large trace.

## Checklist


packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
```diff
@@ -32,18 +62,11 @@ function computeSelfTime(span: Span, allSpans: Span[]): number {
...
   // So instead of [1-10] we get [10-99]. This makes the intervals work like half-open.
-  const spanRange = new DRange(10 * span.startTime, 10 * (span.startTime + span.duration) - 1);
-  // Only keep CHILD_OF spans. FOLLOWS_FROM spans do not block the parent.
-  const children = allSpans.filter(
-    each =>
-      span.childSpanIds.includes(each.spanID) &&
-      each.references[0].spanID === span.spanID &&
-      each.references[0].refType === 'CHILD_OF'
+  if (!span.hasChildren) return span.duration;
+  const spanRange = new DRange(span.startTime, span.startTime + span.duration - 1).subtract(
```
yurishkuro, line 66: I don't think we can change the logic away from `*10`. DRange does not work correctly on the raw timestamps (see the comments above).

Side note: we need a test that illustrates that.

yurishkuro: I don't understand why the calculations need to change - the speed-up is coming from building a map ones, so why not change just that part and leave the calculations intact?

yurishkuro: That still doesn't answer my question why calculations need to change, it a PR that is about _efficiency_ of the code. 

If you want to dig down to the exact reasons for *10 logic, you can see the comment as well as the discussion on the original PR that introduced it. TLDR is that DRange does not understand open/close intervals, and subtraction works incorrectly for our purposes.

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
```diff
@@ -32,17 +62,10 @@ function computeSelfTime(span: Span, allSpans: Span[]): number {
...
-  const spanRange = new DRange(10 * span.startTime, 10 * (span.startTime + span.duration) - 1);
-  // Only keep CHILD_OF spans. FOLLOWS_FROM spans do not block the parent.
-  const children = allSpans.filter(
-    each =>
-      span.childSpanIds.includes(each.spanID) &&
-      each.references[0].spanID === span.spanID &&
-      each.references[0].refType === 'CHILD_OF'
+  if (!span.hasChildren) return span.duration;
+  const spanRange = new DRange(10 * span.startTime, 10 * (span.startTime + span.duration) - 1).subtract(
+    getChildOfDrange(span.spanID, allSpans)
```
yurishkuro, line 67: also unnecessary change, can keep the original logic
```suggestion
  const spanRange = new DRange(10 * span.startTime, 10 * (span.startTime + span.duration) - 1);
  const children = getChildOfSpans(span.spanID, allSpans);
  children.forEach(child => {
    spanRange.subtract(10 * child.startTime, 10 * (child.startTime + child.duration) - 1);
```
it will also like be faster because it will maintain a single array of possible fragmented ranges, whereas in the new code a new fragmented array is created from children and then the parent array is possible transformed to fragmented via subtraction. But mainly I don't like changing working code without reason.

yurishkuro (approved):

PR #1960, opened 2023-11-09
Migrate `TraceTimelineLink.test.js` from enzyme to RTL

EshaanAgg:
Fixes part of #1668 

## Description of the changes
Migrated the `TraceTimelineLink.test.js` to `RTL` using the idiomatic RTL practices

Ran the test suite to ensure all changes passed.

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceTimelineLink.test.js
+++ b/packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceTimelineLink.test.js
@@ -29,7 +29,9 @@ describe('TraceTimelineLink', () => {
   });
 
   it('links to the given trace', () => {
-    expect(screen.getByRole('link').href).toBe(`http://localhost/trace/${traceID}`);
+    const link = screen.getByRole('link');
+    const url = new URL(link.href);
+    expect(url.pathname).toBe(`/trace/${traceID}`);
   });
 
   it('stops event propagation', () => {
--- a/packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceTimelineLink.test.js
+++ b/packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceTimelineLink.test.js
@@ -13,30 +13,31 @@
 // limitations under the License.
 
 import * as React from 'react';
-import { shallow } from 'enzyme';
+import { render, screen, fireEvent, createEvent } from '@testing-library/react';
 
 import TraceTimelineLink from './TraceTimelineLink';
-import NewWindowIcon from '../../common/NewWindowIcon';
 
 describe('TraceTimelineLink', () => {
   const traceID = 'test-trace-id';
-  let wrapper;
 
   beforeEach(() => {
-    wrapper = shallow(<TraceTimelineLink traceID={traceID} />);
+    render(<TraceTimelineLink traceID={traceID} />);
   });
 
   it('renders the NewWindowIcon', () => {
-    expect(wrapper.find(NewWindowIcon).length).toBe(1);
+    expect(screen.getAllByTestId('NewWindowIcon').length).toBe(1);
   });
 
   it('links to the given trace', () => {
-    expect(wrapper.find('a').prop('href')).toBe(`/trace/${traceID}`);
+    expect(screen.getByRole('link').href).toBe(`http://localhost/trace/${traceID}`);
   });
 
   it('stops event propagation', () => {
-    const stopPropagation = jest.fn();
-    wrapper.find('a').simulate('click', { stopPropagation });
-    expect(stopPropagation).toHaveBeenCalled();
+    // Create an event to capture the click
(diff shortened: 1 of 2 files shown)


packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceTimelineLink.test.js
```diff
@@ -13,30 +13,31 @@
...
   });
 
   it('renders the NewWindowIcon', () => {
-    expect(wrapper.find(NewWindowIcon).length).toBe(1);
+    expect(screen.getAllByTestId('NewWindowIcon').length).toBe(1);
   });
 
   it('links to the given trace', () => {
-    expect(wrapper.find('a').prop('href')).toBe(`/trace/${traceID}`);
+    expect(screen.getByRole('link').href).toBe(`http://localhost/trace/${traceID}`);
```
yurishkuro, line 32: interesting, why is this changing from no-host URL to `http://localhost`? That seems counterintuitive - localhost is very dependent on where the code runs

yurishkuro: cc @anshgoyalevil

yurishkuro (commented):
Thanks!

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceTimelineLink.test.js
EshaanAgg: This is because AFAIK while rendering RTL actually converts all the the relative links to the actual complete URLs to which the same anchor is supposed to route to. I get the feeling of it being a bit counterintuitive, but since the tests would always be running in either locally or in workflows (and never in build or production), the same shouldn't be ambiguous or flaky.

yurishkuro (approved):

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceTimelineLink.test.js
yurishkuro: I didn't see any other tests asserting full URL with host name (unless the test provided the value like example.com), so I added URL parsing bit

anshgoyalevil:
```suggestion
    expect(screen.getByRole('link')).toHaveAttribute('href', `/trace/${traceID}`);
```

I think checking for the jsx dom attribute would make more sense in this case rather than the rendered dom attribute.

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceTimelineLink.test.js
```diff
@@ -13,30 +13,31 @@
 // limitations under the License.
 
 import * as React from 'react';
-import { shallow } from 'enzyme';
+import { render, screen, fireEvent, createEvent } from '@testing-library/react';
```
anshgoyalevil, line 16:
```suggestion
import { render, screen, fireEvent, createEvent } from '@testing-library/react';
import '@testing-library/jest-dom';
```

EshaanAgg: Why do we import the `jest-dom` environment here?

anshgoyalevil: Functions like `toHaveAttribute` can be used with that import

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceTimelineLink.test.js
yurishkuro: `expect(...).toHaveAttribute is not a function`

but changing to `toHaveProperty` still results in full URL:
```
    expect(received).toHaveProperty(path, value)

    Expected path: "href"

    Expected value: "/trace/test-trace-id"
    Received value: "http://localhost/trace/test-trace-id"
```

anshgoyalevil: Yes, you need to import `import '@testing-library/jest-dom';` for it to work

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceTimelineLink.test.js
yurishkuro: ah, with this ^ the other suggestion works. But I prefer a more explicit / brute-force `new URL().pathname`.

yurishkuro:
@EshaanAgg thank you!  🎈🎈🎈

PR #1962, opened 2023-11-10
Migrate TraceDiffs/Header from enzyme to RTL

EshaanAgg:
Fixes part of #1668 

## Description of the changes
Migrates `TraceHeader.tsx` from enzyme to RTL. Removes the snapshot tests to relevant assertions with the help of `test-ids`.

Running the test suite locally. 

## Checklist


EshaanAgg:
@yurishkuro Can you please re-run the testing workflow and review this PR?
The failing test in the run seems to be a flaky one, as it gives the following warnings on running locally:
image

yurishkuro:
I don't think it's flaky, I am seeing this error:
```
FAIL  src/components/TracePage/TracePageHeader/TracePageHeader.test.js
  ● Test suite failed to run

    TypeError: Cannot read properties of undefined (reading 'spanID')

      49 | /* this simulates the hierarchy created by CHILD_OF tags */
      50 | function attachReferences(spans, depth, spansPerLevel) {
    > 51 |   let levels = [[spans[0].spanID]];
         |                           ^
      52 |
      53 |   const duplicateLevelFilter = currentLevels => span =>
      54 |     !currentLevels.find(level => level.indexOf(span.spanID) >= 0);

      at spanID (src/demo/trace-generators.js:51:27)
      at Chance.attachReferences [as trace] (src/demo/trace-generators.js:112:13)
      at trace (src/components/TracePage/TracePageHeader/TracePageHeader.test.js:29:51)
      at Object.describe (src/components/TracePage/TracePageHeader/TracePageHeader.test.js:28:1)

```

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceHeader.test.js
```diff
@@ -13,57 +13,77 @@
...
 
 describe('TraceHeader', () => {
-  const props = {
-    duration: 700,
-    error: { errorKey: 'errorValue' },
-    traceID: 'trace-id',
-    traceName: 'trace name',
-  };
-  let wrapper;
+  describe(`TraceHeader`, () => {
```
yurishkuro, line 23: what's the motivation for TraceHeader/TraceHeader nesting?

EshaanAgg:
@yurishkuro But I have only made changes to the TraceHeader component, so how can it affect the TracePageHeader component?

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceHeader.test.js
EshaanAgg: There are three exported components in `TraceHeader`
- The main component `TraceHeader`
- Utility components `Attrs` and `EmptyAttrs`

So I thought it's better to nest all the `describe` block for each individual component under one main `describe`.

yurishkuro: The refactoring makes it much harder to evaluate the actual changes/fixes to the tests, can we split it into another PR?

EshaanAgg: Sure. Will revert this nesting for now.

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceHeader.test.js
```diff
@@ -13,57 +13,75 @@
...
-  it('renders populated attrs component when props.state === fetchedState.DONE', () => {
-    wrapper.setProps({
+  it('renders populated Attrs component when props.state === fetchedState.DONE', () => {
+    renderWithProps({
       startTime: 150,
       totalSpans: 50,
       state: fetchedState.DONE,
     });
-    expect(wrapper).toMatchSnapshot();
+    expect(screen.getAllByTestId('TraceDiffHeader--traceAttr').length).toBe(3);
```
yurishkuro, line 50: this is not very intuitive - why 3? I see each attr has its own test ID like `data-testid="TraceDiffHeader--traceAttr--date"`, so I think it would be more readable and robust test to actually look for each of the 3 attrs by testid explicitly

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceHeader.test.js
```diff
@@ -13,57 +13,75 @@
...
 
   it('renders as expected', () => {
-    expect(wrapper).toMatchSnapshot();
+    renderWithProps();
+    expect(screen.getAllByTestId('TraceDiffHeader--traceHeader').length).toBe(1);
   });
 
-  it('renders populated attrs component when props.state === fetchedState.DONE', () => {
-    wrapper.setProps({
+  it('renders populated Attrs component when props.state === fetchedState.DONE', () => {
```
yurishkuro, line 44: what about when the state is not DONE - attrs should NOT be rendered? Maybe we can combine this into one test with two sub-tests

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceHeader.test.js
```diff
@@ -13,57 +13,75 @@
...
-    expect(wrapper.find('.u-tx-muted').text()).toBe('Select a Trace...');
+    expect(screen.getByText('Select a Trace...'));
   });
 
   describe('EmptyAttrs', () => {
     it('renders as expected', () => {
-      expect(shallow(<EmptyAttrs />)).toMatchSnapshot();
+      render(<EmptyAttrs />);
+      expect(screen.getAllByTestId('TraceDiffHeader--traceAttr').length).toBe(1);
+      expect(screen.getByTestId('TraceDiffHeader--traceAttr').textContent.trim()).toBe('');
```
yurishkuro, line 64: looking for textContext is a bit weird - what if it is legitimately empty? Could we not assert that getByTestId returns an empty list?

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceHeader.test.js
```diff
@@ -13,57 +13,75 @@
...
     });
   });
 
   describe('Attrs', () => {
     it('renders as expected when provided props', () => {
-      expect(shallow(<Attrs duration={700} startTime={150} totalSpans={50} />)).toMatchSnapshot();
+      render(<Attrs duration={700} startTime={150} totalSpans={50} />);
+
+      // Test that the shown values are correctly formatted
+      expect(screen.getByText('January 1, 1970, 12:00:00 am'));
```
yurishkuro, line 73: 1. no assertion
2. where is this exact date coming from? Could we get something more meaningful & matching the inputs?

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceHeader.test.js
```diff
@@ -13,57 +13,75 @@
...
+      expect(screen.getByTestId('TraceDiffHeader--traceAttr--duration').textContent).toBe('700μs');
+      expect(screen.getByTestId('TraceDiffHeader--traceAttr--spans').textContent).toBe('50');
     });
 
     it('Attrs renders as expected when missing props', () => {
-      expect(shallow(<Attrs />)).toMatchSnapshot();
+      render(<Attrs />);
+
+      // Test that the default values are correctly
```
yurishkuro, line 81: interesting - SHOULD they actually be rendered like this? What is the point of "default values" provided by the UI when it displays externally loaded data?

EshaanAgg: It probably shouldn't be. I feel that the same arises from this props definition:
```ts
type Props = {
  duration: number | TNil;
  error?: ApiError;
  startTime: number | TNil;
  state: FetchedState | TNil;
  traceID: string | TNil;
  traceName: string | TNil;
  totalSpans: number | TNil;
};

type AttrsProps = {
  startTime: number | TNil;
  duration: number | TNil;
  totalSpans: number | TNil;
};
```
The props should use conditional type which require `startTime`, `duration` and `totalSpans` to be present if the status is `DONE`, and otherwise they should be `undefined`. And thus the `TNil` should be removed from the `AttrsProps` IMO. Should I work on applying these fixes?

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceHeader.test.js
EshaanAgg: Aight makes sense. I'll try to add an explicit assertion for this in the suite.

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceHeader.test.js
yurishkuro: > getByText is itself an assertion

why, doesn't it return some empty value if it cannot find? or does it throw exception?

> We are providing 150 in the startTime attribute

I believe we typically represent times/durations in microseconds, for timestamps specifically microseconds since epoch. So when you have 150us since epoch rendered as just a datetime with seconds precision you see no difference since the epoch itself. If it's 450us instead of 150us - the output would be the same. So ideally we should use some large enough number, e.g. 1,000,000us, which the will at least change the epoch timestamp to `12:00:01 am`

EshaanAgg: 1. `getBy` queries return exactly one element matching from the DOM, and throw an exception if no element is found or if multiple elements are found.
2. Makes sense. Will incorporate the same in tests and add comments about it.

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceHeader.test.js
EshaanAgg: I think I have handled the same in a more graceful manner in the latest changes. Please let me know what you think about it.

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceHeader.test.js
yurishkuro: I agree, the types are unnecessarily lax, it should be either you have an error or pending state, or you have everything else non-optional. But I think fixing this is a lower priority compared to fixing other enzyme tests

yurishkuro (approved):

yurishkuro:
Thanks!

Issue #1964, opened 2023-11-10
[Bug]: dependabot updates are broken


yurishkuro:
### What happened?

Dependabot is supposed to create new PRs for stale dependencies. I kicked it off manually twice, it failed both times:

Errored with the message "Dependabot encountered an unknown error"
No PRs affected 4 hours ago [view logs](https://github.com/jaegertracing/jaeger-ui/network/updates/748031826)

The failure happens here:
```
Updating @types/react from 18.2.33 to 18.2.37
```

Maybe we can resolve it by doing a manual upgrade. This error is preventing creation of other PRs, like `jest` upgrade that was failing to rebase yesterday.

### Steps to reproduce

There's probably a way to run dependabot locally, but I only ran it via GitHub

### Expected behavior

create PRs for stale deps

### Relevant log output

```shell
Errored with the message "Dependabot encountered an unknown error"

updater | 2023/11/10 12:36:09 INFO <job_748031826> Updating @types/react from 18.2.33 to 18.2.37
  proxy | 2023/11/10 12:36:11 [580] GET https://registry.yarnpkg.com:443/@types%2freact
  proxy | 2023/11/10 12:36:12 [580] 200 https://registry.yarnpkg.com:443/@types%2freact
  proxy | 2023/11/10 12:36:13 [582] GET https://registry.yarnpkg.com:443/lodash
  proxy | 2023/11/10 12:36:13 [582] 200 https://registry.yarnpkg.com:443/lodash
updater | 2023/11/10 12:36:13 INFO <job_748031826> Sending event feebdcd70abe4bebbf035eb56cfe4077 to Sentry
  proxy | 2023/11/10 12:36:14 [584] POST https://sentry.io:443/api/1451818/store/
  proxy | 2023/11/10 12:36:14 [584] 200 https://sentry.io:443/api/1451818/store/
updater | 2023/11/10 12:36:14 ERROR <job_748031826> Error processing @types/react (Dependabot::NpmAndYarn::FileUpdater::NoChangeError)
updater | 2023/11/10 12:36:14 ERROR <job_748031826> No files were updated!
updater | 2023/11/10 12:36:14 ERROR <job_748031826> /home/dependabot/npm_and_yarn/lib/dependabot/npm_and_yarn/file_updater.rb:47:in `updated_dependency_files'
[...]

PR #1968, opened 2023-11-10
Remove yarn resolution for @types/react

anshgoyalevil:
- may fix https://github.com/jaegertracing/jaeger-ui/issues/1964 (Not using the syntax to close the issue, since we have to wait till depandabot again tries the upgrade procedure.)

## Description of the changes
- This PR removes resolution for `types/react`.
- Since the error described in https://github.com/jaegertracing/jaeger-ui/issues/1964 indicates that depandabot tried to upgrade @types/react from 18.2.33 to 18.2.37 but because of no yarn.lock change, it failed.
- Yarn.lock wasn't changed in case of depandabot because of resolution.

- CI

## Checklist

### DIFF
--- a/packages/jaeger-ui/package.json
+++ b/packages/jaeger-ui/package.json
@@ -17,7 +17,7 @@
     "@babel/preset-typescript": "^7.21.0",
     "@testing-library/jest-dom": "^6.0.0",
     "@testing-library/react": "^14.0.0",
-    "@types/react": "^18.2.33",
+    "@types/react": "^18.2.37",
     "@types/react-window": "^1.8.0",
     "@types/redux-form": "^8.3.5",
     "@types/rollup-plugin-visualizer": "^4.2.1",
--- a/packages/jaeger-ui/package.json
+++ b/packages/jaeger-ui/package.json
@@ -17,7 +17,7 @@
     "@babel/preset-typescript": "^7.21.0",
     "@testing-library/jest-dom": "^6.0.0",
     "@testing-library/react": "^14.0.0",
-    "@types/react": "^18.2.37",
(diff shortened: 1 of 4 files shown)


yurishkuro (commented):
I am fine with this PR, but I noticed that you're bumping the lib manually. The latest dependabot PR did bump the lock file for it, but not the explicitly declared version. That was a strange behavior imo, usually it does both, so I am wondering if the resolution is what prevented it from bumping explicit deps. What if we try this PR with removing resolution but without bumping explicit deps?

packages/jaeger-ui/package.json
```diff
@@ -17,7 +17,7 @@
     "@babel/preset-typescript": "^7.21.0",
     "@testing-library/jest-dom": "^6.0.0",
     "@testing-library/react": "^14.0.0",
-    "@types/react": "^18.2.33",
+    "@types/react": "^18.2.37",
```
anshgoyalevil, line 20: This part gets modified too when depandabot tries to upgrade deps even with resolution. Strange why this didn't happen with PR https://github.com/jaegertracing/jaeger-ui/pull/1967

yurishkuro (approved):

anshgoyalevil:
> I am fine with this PR, but I noticed that you're bumping the lib manually. The latest dependabot PR did bump the lock file for it, but not the explicitly declared version. That was a strange behavior imo, usually it does both, so I am wondering if the resolution is what prevented it from bumping explicit deps. What if we try this PR with removing resolution but without bumping explicit deps?

Hmm. I agree with this. We should leave the upgrade on depandabot to see its behavior now

PR #1973, opened 2023-11-13
refac: Migrate `CopyIcon` component from enzyme to RTL

EshaanAgg:
Fixes part of #1668 

## Description of the changes
Migrates `CopyIcon` component from enzyme to RTL
- Introduced the `userEvent` library into the codebase with the standard for testing more complicated user interactions with RTL
- Removed the snapshot tests for `CopyIcon` component
- Decreased the test count of the component as the same when testing the `internal state of the component` which is against the policy of RTL. Instead replaced the same with a single comprehensive test which checks for the nature of the tooltip, and whether the content was copied into the clipboard or not

Running the test suite locally 

## Checklist


packages/jaeger-ui/src/components/common/CopyIcon.test.js
```diff
@@ -28,43 +30,32 @@ describe('<CopyIcon />', () => {
...
+    await user.hover(screen.getByRole('button'));
+    await waitFor(() => expect(screen.getByText('tooltipTitleValue')).toBeInTheDocument());
     expect(copySpy).not.toHaveBeenCalled();
 
-    wrapper.find(Button).simulate('click');
-    expect(wrapper.state().hasCopied).toBe(true);
+    // Simulate the click
+    await user.click(screen.getByRole('button'));
+    await user.hover(screen.getByRole('button'));
+    expect(screen.getByText('Copied')).toBeInTheDocument();
```
yurishkuro, line 58: is this test going to he flaky? Once `user.click` happens, the Copied title only remains visible for a certain time - during CI the execution of this code might be suspended (CPU scheduling etc) and resumed after the timeout when "Copied" is no longer displayed, causing this assertion to fail. Some stateful test would be more reliable here, i.e. validating that a certain function was called (and separately testing that the tooltip changes based on state by _controlling the state_, not the user click behavior)

packages/jaeger-ui/src/components/common/CopyIcon.test.js
```diff
@@ -28,43 +30,32 @@ describe('<CopyIcon />', () => {
...
-    expect(wrapper).toMatchSnapshot();
+    expect(screen.getByRole('button'));
   });
 
-  it('updates state and copies when clicked', () => {
-    expect(wrapper.state().hasCopied).toBe(false);
+  it('updates state and copies when clicked', async () => {
+    // Intially there should be nothing in the clipboard, and thus the tooltip should have the default title
+    await user.hover(screen.getByRole('button'));
+    await waitFor(() => expect(screen.getByText('tooltipTitleValue')).toBeInTheDocument());
```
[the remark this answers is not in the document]
EshaanAgg: This is the title of the tooltip that we are passing in as a `prop` to the component. This would show when we hover over the `CopyIcon` button.

packages/jaeger-ui/src/components/common/CopyIcon.test.js
EshaanAgg: - The test wouldn't be flaky as the tooltip would be visible for time-periods in magnitude of seconds, where as the suspension of the test shouldn't be more than microseconds. We are using `async` functions here as there is some time needed by React to render the elements into the DOM, which is not immediate. This is even true for the functionality in production code, and thus we needn't worry about it.

- I get the sentiment to feel more confident about the component using stateful tests, but the same goes against the philosophy of RTL itself. RTL is so heavily invested on testing user interactions, that it offers no utilities to get access to the `props` or `state` of the component after it's render. All you can do is use `querySelectors` to check for the appearance of elements. styles and classes.

EshaanAgg:
@yurishkuro 
```ts 
handleTooltipVisibilityChange = (visible: boolean) => {
  if (!visible && this.state.hasCopied) {
    this.setState({
      hasCopied: false,
    });
  }
};
```
Can you please let me know what is the use of this function? I can't seem to figure it out.

packages/jaeger-ui/src/components/common/CopyIcon.test.js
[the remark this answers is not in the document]
yurishkuro: I see. Then we should define `const testTooltip = 'tooltipTitleValue'` and use that.

packages/jaeger-ui/src/components/common/CopyIcon.test.js
yurishkuro: 1. It's possible for VM running CI to get suspended for seconds. However, on a second glance, the change in text does not happen on the timer, but on leaving the tooltip hover, which doesn't seem like it would happen in this test, so it's probably ok.
2. I think you can always construct code in a way to work around that. For example, the state change here happens via `handleTooltipVisibilityChange` function, that could be mocked if necessary.

Anyway, I think we're good because of (1) above.

yurishkuro: Actually, you don't seem to test that the text will change back from Copied to Copy, is that something we can do by undoing hover state?

packages/jaeger-ui/src/components/common/CopyIcon.test.js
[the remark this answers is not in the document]
EshaanAgg: Sure will do.

packages/jaeger-ui/src/components/common/CopyIcon.test.js
EshaanAgg: > Actually, you don't seem to test that the text will change back from Copied to Copy, is that something we can do by undoing hover state?

Yes. I was thinking of adding that myself when I saw that I missed the `handleTooltipVisibilityChange` function in coverage. Will update that as well.

EshaanAgg:
@yurishkuro I have added the `mouseLeave` test as you requested, but I can't seem to get it to pass. Can you look at the test and confirm that this is the intended behavior?

PS. I also removed the `userEvent` library for now, as the simple hover interactions can be done with `fireEvent` itself. We may include the library later when more complex interactions arise.

packages/jaeger-ui/src/components/common/CopyIcon.test.js
```diff
@@ -13,58 +13,57 @@
...
   });
 
-  it('updates state and copies when clicked', () => {
-    expect(wrapper.state().hasCopied).toBe(false);
+  it('updates state and copies when clicked', async () => {
+    const triggerElement = screen.getByRole('button');
+
+    // Intially when we hover over the icon, there should be nothing in the clipboard and the tooltip should have the default title
+    fireEvent.mouseEnter(triggerElement);
+    await waitFor(() => expect(screen.getByText(tooltipTitle)).toBeInTheDocument());
```
yurishkuro, line 53: what does waitFor do - is it periodically polling its lambda? Or does expect() always return a future? It's a bit strange that you suddenly start calling function in async which normally was not called this way.

Is `fireEvent.mouseEnter` sync or async?

EshaanAgg: `waitFor` returns a Promise that resolves when the specified condition is met. It retries the condition every `50ms` (by default) until it passes or times out after `1000ms` (by default).

`fireEvent.mouseEnter` is a sync event, but even after the occurrence of the event, `antd` takes some time to insert the markdown for the tooltip into DOM. Thus we need to make use of the `waitFor` utility.

Issue #1943, opened 2023-11-03
[Feature]: Speed up Critical path calculation


maxgaponov:
### Requirement

I would like to view big traces in Jaeger UI.

### Problem

Critical path calculation works slowly on large traces (especially if I press "Collapse all").

yurishkuro:
closing as duplicate of #2179, which has more information

PR #1974, opened 2023-11-13
Exit mergeChildrenCriticalPath early for an empty critical path

maxgaponov:
- Part of #1943 

## Description of the changes
- mergeChildrenCriticalPath shortcuts for an empty critical path

- Existing tests

## Checklist


yurishkuro (approved):

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
```diff
@@ -163,6 +163,9 @@ function mergeChildrenCriticalPath(
   spanID: string,
   criticalPath: criticalPathSection[]
 ): criticalPathSection[] {
+  if (criticalPath.length == 0) {
```
yurishkuro, line 166: linter complains 
`166:27  error    Expected '===' and instead saw '=='   `

yurishkuro: couldn't this be just `if (!criticalPath)` ?

yurishkuro:
tests are always welcome, but from the outside of this function it seems it would be impossible to distinguish, since fully executing the logic still returns the same `[]`, just slower.

PR #1975, opened 2023-11-13
Speed up Critical path calculation

maxgaponov:
- Part of #1943 and #2179

## Description of the changes
- Rewrote computeCriticalPath function to make less recursive calls

- Existing tests

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/CriticalPath/index.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/CriticalPath/index.tsx
@@ -22,64 +22,49 @@ import sanitizeOverFlowingChildren from './utils/sanitizeOverFlowingChildren';
  * Computes the critical path sections of a Jaeger trace.
  * The algorithm begins with the top-level span and iterates through the last finishing children (LFCs).
  * It recursively computes the critical path for each LFC span.
- * Upon return from recursion, the algorithm walks backward and picks another child that
- * finished just before the LFC's start.
  * @param spanMap - A map associating span IDs with spans.
  * @param spanId - The ID of the current span.
  * @param criticalPath - An array of critical path sections.
- * @param returningChildStartTime - Optional parameter representing the span's start time.
- *                    It is provided only during the recursive return phase.
  * @returns - An array of critical path sections for the trace.
  * @example -
  * |-------------spanA--------------|
  *    |--spanB--|    |--spanC--|
  * The LFC of spanA is spanC, as it finishes last among its child spans.
- * After invoking CP recursively on LFC, for spanC there is no LFC, so the algorithm walks backward.
- * At this point, it uses returningChildStartTime (startTime of spanC) to select another child that finished
- * immediately before the LFC's start.
  */
 const computeCriticalPath = (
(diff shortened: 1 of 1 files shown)


packages/jaeger-ui/src/components/TracePage/CriticalPath/index.tsx
```diff
@@ -22,64 +22,49 @@ import sanitizeOverFlowingChildren from './utils/sanitizeOverFlowingChildren';
...
  */
 const computeCriticalPath = (
   spanMap: Map<string, Span>,
   spanId: string,
-  criticalPath: criticalPathSection[],
-  returningChildStartTime?: number
+  criticalPath: criticalPathSection[]
 ): criticalPathSection[] => {
   const currentSpan: Span = spanMap.get(spanId)!;
+  let criticalPathBoundary = currentSpan.startTime + currentSpan.duration;
```
yurishkuro, line 40: confusing name, boundary can be either top or bottom

packages/jaeger-ui/src/components/TracePage/CriticalPath/index.tsx
```diff
@@ -22,64 +22,49 @@ import sanitizeOverFlowingChildren from './utils/sanitizeOverFlowingChildren';
...
-      computeCriticalPath(spanMap, parentSpanId, criticalPath, currentSpan.startTime);
+  for (let i = 0; i < currentSpan.childSpanIds.length; i++) {
+    const child: Span = spanMap.get(currentSpan.childSpanIds[i])!;
+    const childEndTime = child.startTime + child.duration;
+    if (childEndTime <= criticalPathBoundary) {
+      const spanCriticalSection = {
+        spanId: currentSpan.spanID,
+        section_start: childEndTime,
+        section_end: criticalPathBoundary,
+      };
```
yurishkuro, line 50: I don't follow. `criticalPathBoundary` is initially end-of-parent. This says if the child ends before than, then the _difference_ between child.end and parent.end is the critical path contributed by that span?

packages/jaeger-ui/src/components/TracePage/CriticalPath/index.tsx
```diff
@@ -22,64 +22,49 @@ import sanitizeOverFlowingChildren from './utils/sanitizeOverFlowingChildren';
...
-    }
-    // Now as there are no lfc's focus shifts to parent span from startTime of span
-    // return from recursion and walk backwards to one level depth to parent span
-    // provide span's startTime as returningChildStartTime
-    if (currentSpan.references.length) {
-      const parentSpanId: string = currentSpan.references.filter(
-        reference => reference.refType === 'CHILD_OF'
-      )[0].spanID;
-      computeCriticalPath(spanMap, parentSpanId, criticalPath, currentSpan.startTime);
+  for (let i = 0; i < currentSpan.childSpanIds.length; i++) {
```
yurishkuro, line 42: is there an assumption on the order of this array?

yurishkuro (commented):
I recommend writing down the algorithm in words, to understand what you're trying to change. So far it does not seem equivalent to me.

Issue #1466, opened 2023-06-05
[Feature]: combine all graph views into one - LFX Mentorship


yurishkuro:
For the application process please refer to [this issue](https://github.com/jaegertracing/jaeger/issues/4457).

* Mentor: @yurishkuro 
* Mentee: @prathamesh-mutkure

### Requirement

As a user of Jaeger UI I want to have a consistent experience with service graph views.

### Problem

Jaeger currently uses three distinct renderings for service graphs
  * the DAG view, good for small architectures (uses Cytoscape.js)
  * the System view good for large architectures (uses react-viz-force-graph)
  * the internal Plexus lib used for deep dependency graphs  (uses D3 and graphviz rendering engine)
  * 

### Proposal

To reduce the different code paths and provide a consistent user experience we should converge on a single graph visualization.

Plexus is the likely choice, but it needs to be benchmarked on large graphs (the System view can handle several thousand nodes). We can also consider server-side filtering (proposed in #784). 

We may also have to address #1081 (deprecated graphing libs)

### Some Evaluation Criteria

* is there support for stable / deterministic layouts? This is mandatory for graph views of a single trace, where nodes are arranged according to time and depth.
* how "future proof" is the library - actively supported, big contributor community, etc. (including its own dependencies). Are there big projects that use it?
* how many new dependencies it pulls
* can it handle large graphs (say up to 5k nodes)?
* does it allow customization of the graph? Including labels on edges, different viz for notes, interaction with nodes (left/right click)

punithnayak:
Hey @yurishkuro I want to work on this for the upcoming LFX mentorship . Will go through good first Issues and suggest improvements if needed

MdSahil-oss:
Hi @punithnayak, Can you please stop commenting down on all LFX project issues, So that other aspiring mentees can decide whether they want to work on the project or not, And please delete other issue's comment if possible if no longer interested.

punithnayak:
Hey,@MdSahil-oss I am nowhere stopping other candidates from not participating .i am just greeting the org maintainers and letting them know I am interested. So I don't get your point

sivasathyaseeelan:
Hi @yurishkuro, I am very interested and definitely apply for this project in the upcoming LFX mentorship!!

shashwatm1111:
Hey @yurishkuro , would love to contribute to this project under the LFX mentorship . Also going through the first good issues and will work upon them .

yurishkuro:
[...]

PR #1981, opened 2023-11-18
Refactoring: DAG migrated to Plexus's Digraph

prathamesh-mutkure:
- This issue is related to LFX task #1466 
- I can create a more specific sub-issue if that's more suitable

## Description of the changes
- I've removed `cytoscape` and replaced it with `Digraph` from `Plexus`
- Had to add support for labels in `Digraph` to support this change

- The changes were tested visually
- Tested with test data involving around 50 nodes

Before - 

Screenshot 2023-11-18 at 22 06 10

Now -

Screenshot 2023-11-18 at 22 06 22

Tha additional tab visible in above images has been removed

## Checklist


prathamesh-mutkure:
Hey @yurishkuro, the two test cases will fail for `DAG`, I'll fix them ASAP, you can review the changes in the meantime

prathamesh-mutkure:
Hey @yashrsharma44, tagging you for review

The code coverage has dropped, will try to increase it

packages/jaeger-ui/package.json
```diff
@@ -55,9 +55,6 @@
     "classnames": "^2.2.5",
     "combokeys": "^3.0.0",
     "copy-to-clipboard": "^3.1.0",
-    "cytoscape": "^3.2.1",
-    "cytoscape-dagre": "^2.0.0",
-    "dagre": "^0.8.5",
```
yurishkuro, line 60: wow, really, we're even removing dagre?

packages/jaeger-ui/src/components/DependencyGraph/DAG.test.js
```diff
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 import React from 'react';
-import { render, screen } from '@testing-library/react';
+import { shallow } from 'enzyme';
```
yurishkuro, line 16: we are trying to move away from enzyme, please do not reintroduce it here, use testing-library

yurishkuro: especially considering that the tests here are extremely primitive, just a check for top-level element

packages/jaeger-ui/src/components/DependencyGraph/DAG.tsx
```diff
@@ -0,0 +1,145 @@
...
+  }[];
+};
+
+type TState = {
+  nodes: TVertex[];
+  edges: TEdge[];
+};
+
+const renderNode = (vertex: TVertex<TVertex>): ReactNode => {
+  return (
```
yurishkuro, line 37: how is it possible for this function to never be called in the test?

packages/jaeger-ui/src/components/DependencyGraph/DAG.tsx
```diff
@@ -0,0 +1,145 @@
...
+    <div className="DAG--node">
+      <div className="DAG--nodeCircle" />
+
```
yurishkuro, line 40:
```suggestion
```

packages/plexus/src/Digraph/SvgEdge.tsx
```diff
@@ -62,15 +64,31 @@ export default class SvgEdge<U = {}> extends React.PureComponent<TProps<U>> {
       },
       getProps(setOnEdge, layoutEdge, renderUtils)
     );
+
+    const [startX, startY] = pathPoints[0];
+    const [endX, endY] = pathPoints[pathPoints.length - 1];
+
+    const xOffset = (label?.length ?? 0) * 5;
+    const labelX = (startX + endX) / 2;
+    const labelY = (startY + endY) / 2;
```
yurishkuro, line 73:
```suggestion
    const computeLabelCoords = () => {
      const [startX, startY] = pathPoints[0];
      const [endX, endY] = pathPoints[pathPoints.length - 1];
      const xOffset = (label?.length ?? 0) * 5;
      const labelX = (startX + endX) / 2 - xOffset;
      const labelY = (startY + endY) / 2;
      return { labelX, labelY };
    };
    const { labelX, labelY } = computeLabelCoords();
```
this way you do not pollute the scope with temporary variables that are only relevant for calculation

packages/plexus/src/Digraph/SvgEdge.tsx
```diff
@@ -62,15 +64,31 @@ export default class SvgEdge<U = {}> extends React.PureComponent<TProps<U>> {
...
+          d={d}
+          fill="none"
+          vectorEffect="non-scaling-stroke"
+          markerEnd={markerEnd}
+          markerStart={markerStart}
+          {...customProps}
+        />
+
+        {label && (
+          <text x={labelX - xOffset} y={labelY} fill="#000" fontSize="1rem" fontWeight="bold">
```
yurishkuro, line 87:
```suggestion
          <text x={labelX} y={labelY} fill="#000" fontSize="1rem" fontWeight="bold">
```

packages/jaeger-ui/src/components/DependencyGraph/DAG.test.js
```diff
@@ -63,8 +64,10 @@ describe('<DAG>', () => {
         parent: 'parent-id',
       },
     ];
-    render(<DAG serviceCalls={serviceCalls} />);
-    expect(screen.getByTestId('cy')).toBeDefined();
+
+    const wrapper = shallow(<DAG serviceCalls={serviceCalls} />);
+
+    expect(wrapper.exists('.DAG')).toBe(true);
```
yurishkuro, line 70: can we do something smarter, like count how many nodes and edges are displayed?

yurishkuro:
* ping zoom does not seem to work at all, I am getting the whole browser window changing zoom instead of the graph div only, as it was in the old DAG. I know that zoom does work in DDG (but does not seem to work in Trace Diff)
* is dragging of the nodes not supported by Plexus at all? Not the end of the world if it's not.

yurishkuro:
Another odd thing: if I tab to Force Directed Graph and then back to DAG, the minimap goes rather nuts - the light-gray box is supposed to indicate current zoom view in the viewport, and it shows outside of it, even though the main display is fine. Clicking on the reset button returns the box back into full size minimap. I suspect it may be related to the zoom nor working.

image

yurishkuro:
I tried this with graphs with larger number of nodes. With 15 nodes it looks like this:
image
Curious why the nodes / text are so small, there's enough space between them for larger nodes.

The zoom seems to work with these larger graphs.

packages/jaeger-ui/package.json
prathamesh-mutkure: Yes, not used anywhere anymore

packages/jaeger-ui/src/components/DependencyGraph/DAG.test.js
prathamesh-mutkure: There's a comment in the test file that we don't really care about the rendered canvas actually, we still can check for the number of nodes and edges to be more precise

packages/jaeger-ui/src/components/DependencyGraph/DAG.tsx
prathamesh-mutkure: Yeah, this is what caused test coverage to decrease

packages/plexus/src/Digraph/SvgEdge.tsx
prathamesh-mutkure: That's a good suggestion, more memory efficient

packages/jaeger-ui/src/components/DependencyGraph/DAG.test.js
prathamesh-mutkure: Yes, that will be better

prathamesh-mutkure:
> * ping zoom does not seem to work at all, I am getting the whole browser window changing zoom instead of the graph div only, as it was in the old DAG. I know that zoom does work in DDG (but does not seem to work in Trace Diff)

The graph is likely smaller, you seem to got the zoom working with larger graph

> is dragging of the nodes not supported by Plexus at all? Not the end of the world if it's not.

Did not see anything in the documentation and code sadly

> if I tab to Force Directed Graph and then back to DAG, the minimap goes rather nuts

I'll check this one and try to fix it

> I tried this with graphs with a larger number of nodes. With 15 nodes it looks like this:
> Curious why the nodes / text are so small, there's enough space between them for larger nodes

The coordinates are actually returned by Plexus, the size of nodes is static, and we can play around with the size of nodes

Also, can you share data for those 15 nodes?

yurishkuro:
15-node graph https://gist.github.com/yurishkuro/1f7dfbceebc5f192a2e3108c53ad96da

yurishkuro:
> The graph is likely smaller, you seem to got the zoom working with larger graph

that doesn't explain why the zoom doesn't work - with the larger 15-node graph I can zoom it so that the nodes are twice as large as in the default hotrod graph (which doesn't zoom)

packages/jaeger-ui/src/components/DependencyGraph/DAG.test.js
```diff
@@ -13,8 +13,10 @@
 // limitations under the License.
 
 import React from 'react';
+import ShallowRenderer from 'react-test-renderer/shallow';
```
yurishkuro, line 16: why do we need to use `ShallowRenderer` and not the standard `render`? Does `ShallowRenderer` actually render _more_ details?

yurishkuro (approved):

PR #1992, opened 2023-11-21
Change Vite config from CJS to ESM

### LINKED ISSUE #1985: Upgrade to ESM build of Vite
When running `yarn start` the following warning is printed:

> The CJS build of Vite's Node API is deprecated. See https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated for more details.

According to the wiki, the CJS build will be removed in Vite v6.

* [ ] investigate & document what needs to change and what the implications are
* [ ] open PR with the required changes

ShivamMadlani:
This commit fixes the CJS deprication warning.

- Resolves #1985 

## Description of the changes
- Add `type: module` to `package.json`
- Imported visualizer as a function and used as a PluginOption

- `yarn test`

## Checklist


packages/jaeger-ui/vite.config.mts
```diff
@@ -13,11 +13,11 @@
 // limitations under the License.
```
yurishkuro, line 13: why is changing the extension necessary? what is "M" in ".mts"?

yurishkuro, line 13: * are the `*.mts` files included in our linter?
* instead of renaming the file, why not add `"type": "module"` to `./packages/jaeger-ui/package.json`?

ShivamMadlani: >are the *.mts files included in our linter?

no they arent.. forgot to check that. 

yeah you are right that would be better. making the change...

yurishkuro (dismissed):

ShivamMadlani:
hey @yurishkuro some tests are failing as they are using require() instead of import... i think this could be maybe resolved by using .cjs extension for tests in order to use require statements?? or we could rewrite the tests and use import statements instead...

yurishkuro:
can you analyze how many changes would be needed? Fixing `require` would be my preferred path, but if it's too much work then perhaps just renaming vite config to `.mts` is a better fix (we'd need to include it in the linter globs)

ShivamMadlani:
yup currently there are a total of 18 occurances of require in the project. changing around 11 of them located in the `test` folder should pass all the tests. however we could just change all the requires so the whole project gets cleaned up...waiting for your thumbsup

ShivamMadlani:
@yurishkuro made the changes...tests should be able to pass now

yurishkuro:
you can speed up the iterations by running the tests locally (`yarn lint` and `yarn test`) instead of waiting for me to approve the CI run

ShivamMadlani:
hey @yurishkuro i think its best to rename vite.config to mts. changing imports to require is causing few tests to fail which required renaming a bunch of other files that uses CJS which would defeat the purpose of this issue.
however i also added .mts to our formatter and `yarn tests` are passing locally.

yurishkuro (approved):

yurishkuro:
Thanks @ShivamMadlani

Issue #1996, opened 2023-11-22
[Bug]: jaeger-ui fails to compile on s390x due to a dependency on lerna/nx


andreasgerstmayr:
### What happened?

As a maintainer I want to compile jaeger-ui on s390x.

### Steps to reproduce

1. start a s390x machine
2. `yarn install --frozen-lockfile`

### Expected behavior

Build fails with `Error: Unsupported architecture on Linux: s390x`

### Relevant log output

```shell
[root@s390x-kvm-014 jaeger-ui]# yarn install --frozen-lockfile
yarn install v1.22.21
[...]
[4/4] Building fresh packages...
$ lerna run --stream --sort prepublishOnly
lerna notice cli v7.4.2
lerna ERR! Error: Unsupported architecture on Linux: s390x
lerna ERR!     at Object.<anonymous> (/root/jaeger/jaeger-ui/node_modules/nx/src/native/index.js:235:15)
lerna ERR!     at Module._compile (node:internal/modules/cjs/loader:1254:14)
lerna ERR!     at Module._extensions..js (node:internal/modules/cjs/loader:1308:10)
lerna ERR!     at Module.load (node:internal/modules/cjs/loader:1117:32)
lerna ERR!     at Module._load (node:internal/modules/cjs/loader:958:12)
lerna ERR!     at Module.require (node:internal/modules/cjs/loader:1141:19)
lerna ERR!     at require (node:internal/modules/cjs/helpers:110:18)
lerna ERR!     at /root/jaeger/jaeger-ui/node_modules/nx/src/project-graph/utils/retrieve-workspace-files.js:20:45
lerna ERR!     at Generator.next (<anonymous>)
lerna ERR! lerna Unsupported architecture on Linux: s390x
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```
```


### Screenshot

_No response_

### Additional context

_No response_

### Jaeger backend version

_No response_

### SDK

_No response_

### Pipeline

_No response_

### Stogage backend

_No response_

### Operating system

Linux on s390x

### Deployment model

_No response_

### Deployment configs

_No response_

andreasgerstmayr:
The `lerna` dependency depends on `nx`, which doesn't support s390x:
https://github.com/nrwl/nx/blob/16.6.0/packages/nx/src/native/index.js#L161
[...]

Issue #341, opened 2019-03-08
Audit use of lerna


tiffon:
Make sure our use of [lerna](https://github.com/lerna/lerna) is:

- Necessary
- Idiomatic (if it is necessary)

We have ended up relying quite a bit on `yarn` workspaces, which has reduced the relevance of `lerna`.

PR #1997, opened 2023-11-22
Remove lerna dependency

andreasgerstmayr:
lerna/nx does not support s390x, and therefore the jaeger-ui build fails on this platform. This PR removes the lerna dependency from jaeger-ui.

https://github.com/nrwl/nx/blob/16.6.0/packages/nx/src/native/index.js#L161
https://github.com/nrwl/nx/blob/16.6.0/docs/shared/installation/troubleshoot-installation.md#supported-native-module-platforms

Resolves: #1996
Resolves: #341

## Description of the changes
Remove `lerna` and use `yarn workspaces` instead. 

- local

## Checklist


frzifus (approved):

yurishkuro (approved):
Awesome!  🎉

PR #2012, opened 2023-12-02
Add span kind selector

albertteoh:
- Resolves #4035
- Relates to #3898

## Description of the changes
- Add a span kind filter to the Monitor tab.
- As span kinds are an attribute of a span, much like the service name, I thought it made sense to add the dropdown filter next to the service filter.
- I'm not a designer so open to ideas on how to make this better!

- Manually tested as per the following screen recording.
- Update tests.

https://github.com/jaegertracing/jaeger-ui/assets/26584478/2d8ee49f-e30c-4438-9e92-9012814c495d

## Checklist


packages/jaeger-ui/src/components/Monitor/ServicesView/index.tsx
```diff
@@ -94,6 +97,13 @@ export const timeFrameOptions = [
   { label: 'Last 24 hours', value: 24 * oneHourInMilliSeconds },
   { label: 'Last 2 days', value: 48 * oneHourInMilliSeconds },
 ];
+export const spanKindOptions = [
+  { label: 'Client', value: 'client' },
+  { label: 'Server', value: 'server' },
+  { label: 'Internal', value: 'internal' },
+  { label: 'Producer', value: 'producer' },
+  { label: 'Consumer', value: 'consumer' },
```
albertteoh, line 105: Label naming follows convention used by [OpenTelemetry](https://opentelemetry.io/docs/concepts/signals/traces/#span-kind).

packages/jaeger-ui/src/components/Monitor/ServicesView/index.tsx
```diff
@@ -287,6 +304,31 @@ export class MonitorATMServicesViewImpl extends React.PureComponent<TPropsWithIn
                 ))}
               </Field>
             </Col>
+            <Col span={6}>
+              <h2 className="span-kind-selector-header">Choose kind</h2>
```
yurishkuro, line 308:
```suggestion
              <h2 className="span-kind-selector-header">Span Kind</h2>
```
Similarly, `s/Choose service/Service/` in L287, using the word "choose" in the UX is redundant

packages/jaeger-ui/src/components/Monitor/ServicesView/index.tsx
```diff
@@ -287,6 +304,31 @@ export class MonitorATMServicesViewImpl extends React.PureComponent<TPropsWithIn
...
+                name="spanKind"
+                component={reduxFormFieldAdapter({ AntInputComponent: SearchableSelect })}
+                placeholder="Select A Span Kind"
+                onChange={(e, value: string) => {
+                  const { label } = spanKindOptions.find(option => option.value === value)!;
+                  trackSelectSpanKind(label);
+                }}
+                props={{
+                  className: 'span-kind-selector',
+                  defaultValue: spanKindOptions[0],
```
yurishkuro, line 319: client?

packages/jaeger-ui/src/types/metrics.tsx
```diff
@@ -23,14 +23,16 @@ export type AvailableOpsMetrics =
...
+
 export type MetricsAPIQueryParams = {
   quantile: number;
   groupByOperation?: boolean;
   endTs?: number;
   lookback?: number;
   step?: number;
   ratePer?: number;
-  spanKind?: 'unspecified' | 'internal' | 'server' | 'client' | 'producer' | 'consumer';
+  spanKind?: spanKinds;
```
yurishkuro, line 35: lots of optional fields here - are they really optional? Even if they are from the server perspective, wouldn't UI code always populate them?

packages/jaeger-ui/src/components/Monitor/ServicesView/index.tsx
yurishkuro: also, better not to depend on the array order.

packages/jaeger-ui/src/components/Monitor/ServicesView/index.test.js
```diff
@@ -361,6 +361,7 @@ describe('mapStateToProps()', () => {
       metrics: originInitialState,
       services: [],
       selectedService: 's1',
+      selectedSpanKind: 'server',
```
yurishkuro, line 364: do we have behavioral tests? It would be good to have tests simulating a change in span kind

packages/jaeger-ui/src/components/Monitor/ServicesView/index.tsx
albertteoh: 👍🏼 Done in: [46989de](https://github.com/jaegertracing/jaeger-ui/pull/2012/commits/46989deae9944f243b7c267fe197c64e23246dc7)

packages/jaeger-ui/src/components/Monitor/ServicesView/index.tsx
albertteoh: Removed "Choose" in: [07e3c26](https://github.com/jaegertracing/jaeger-ui/pull/2012/commits/07e3c26244e78643a3d194cf38693988d2e7639c)

packages/jaeger-ui/src/types/metrics.tsx
albertteoh: Hmmm.. yeah, not sure why it was done that way initially; removed optionality in: [ccd6cc7](https://github.com/jaegertracing/jaeger-ui/pull/2012/commits/ccd6cc76fc4f4d34178978ce7b3a4f03aaab0182).

packages/jaeger-ui/src/components/Monitor/ServicesView/index.test.js
albertteoh: Yup, I could do better with test coverage; added behavioural tests in: [62a71da](https://github.com/jaegertracing/jaeger-ui/pull/2012/commits/62a71dae01304c159731e6f54661db13c1dfeff9).

albertteoh:
Note that I've retitled this PR to use SPM instead of ATM because that's the acronym used in our [documentation](https://www.jaegertracing.io/docs/1.51/spm/). If you're okay with that, do you think we should also do the same in our codebase?

packages/jaeger-ui/src/components/Monitor/ServicesView/index.tsx
```diff
@@ -210,6 +225,7 @@ export class MonitorATMServicesViewImpl extends React.PureComponent<TPropsWithIn
 
     if (currentService) {
       this.endTime = Date.now();
+      store.set('lastAtmSearchSpanKind', selectedSpanKind);
```
yurishkuro, line 228: When we change this to SPM, a couple more improvements:
- use constants instead of raw strings 
- keep abbreviations in uppercase, SPM not Spm (but constants should probably all be named in uppercase)

yurishkuro (approved):

PR #2041, opened 2023-12-10
Chore: Import only lodash submodules

### LINKED ISSUE #2032: Import only lodash submodules
In order to reduce the bundle size, lodash recommends importing submodules, e.g. `import _get from 'lodash/get'` instead of `import { debounce } from 'lodash'`. Most of our code is already like that, with the exception of just a handful of files:

```
$ grx lodash packages/*/src | grep -v 'lodash/'
packages/jaeger-ui/src/components/DependencyGraph/DependencyForceGraph.jsx:18:import { debounce } from 'lodash';
packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx:16:import * as _ from 'lodash';
packages/jaeger-ui/src/components/TracePage/TraceStatistics/generateDropdownValue.tsx:15:import _ from 'lodash';
packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ResultItem.tsx:20:import { sortBy } from 'lodash';
```

This ticket is to clean up these remaining full imports by replacing them with the needed submodule imports.

It will be interesting to see how much it would affect the bundle size:
image

anshgoyalevil:
- resolves #2032 

## Description of the changes
- Import only lodash submodules

- locally

Super cool that it reduced the size taken by lodash bundle by 70.59% (master branch vs this)

Before:
image
lodash: 773.55 KB

After:
image
lodash: 227.43 KB

## Checklist


packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
```diff
@@ -115,20 +117,22 @@ function valueFirstDropdown(selectedTagKey: string, trace: Trace) {
...
-      .groupBy(x => x.process.serviceName)
-      .map((value, key) => ({ key }))
-      .uniq()
-      .value();
+    const temp = _uniq(
+      _map(
+        _groupBy(allSpans, x => x.process.serviceName),
+        (value, key) => ({ key })
+      )
+    );
```
yurishkuro, line 125: If I understand this, the goal is to get a unique list of service names, right? If so, why use groupBy at all, wouldn't this work (in scala syntax): 

spans.map(_.process.serviceName).uniq()

?

packages/jaeger-ui/src/components/TracePage/TraceStatistics/generateDropdownValue.tsx
```diff
@@ -41,11 +45,9 @@ function getValueTagIsPicked(tableValue: ITableSpan[], trace: Trace, nameSelecto
...
   availableTags = [...new Set(availableTags)];
 
-  const tags = _(availableTags).map('tags').flatten().value();
-  let tagKeys = _(tags).map('key').uniq().value();
-  tagKeys = _.filter(tagKeys, function calc(o) {
-    return o !== nameSelectorTitle;
-  });
+  const tags = _map(availableTags, 'tags');
+  let tagKeys = _uniq(_map(_flatten(tags), 'key'));
+  tagKeys = _filter(tagKeys, o => o !== nameSelectorTitle);
```
yurishkuro, line 50: This may be the equivalent, but could we do better? First, availableTags var is overloaded: starts a set of spans, then set of strings, bad design. And all these lodash ops are revisiting the same data that was already processed in the loop above - could we just calculate what we need right there?

yurishkuro: At least we should rename the var to spansWithFilterTag

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
anshgoyalevil: Yes. I have used a more concise function `uniqBy` here now.

packages/jaeger-ui/src/components/TracePage/TraceStatistics/generateDropdownValue.tsx
anshgoyalevil: I tried playing with the former logic but it is getting a bit complex to accommodate the work we are taking out of lodash functions to the loops.

packages/jaeger-ui/src/components/TracePage/TraceStatistics/generateDropdownValue.tsx
```diff
@@ -41,15 +45,13 @@ function getValueTagIsPicked(tableValue: ITableSpan[], trace: Trace, nameSelecto
...
   availableTags = [...new Set(availableTags)];
 
-  const tags = _(availableTags).map('tags').flatten().value();
-  let tagKeys = _(tags).map('key').uniq().value();
-  tagKeys = _.filter(tagKeys, function calc(o) {
-    return o !== nameSelectorTitle;
-  });
+  const tags = _map(availableTags, 'tags');
+  let spansWithFilterTag = _uniq(_map(_flatten(tags), 'key'));
+  spansWithFilterTag = _filter(spansWithFilterTag, o => o !== nameSelectorTitle);
```
yurishkuro, line 50: this collection does not contain spans, now it's even more confusing. I was referring to L46 and above, where the objects in question are spans, but the collection is called availableTags

packages/jaeger-ui/src/components/TracePage/TraceStatistics/generateDropdownValue.tsx
```diff
@@ -12,19 +12,22 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import _ from 'lodash';
+import _map from 'lodash/map';
```
anshgoyalevil, line 15: We can get rid of this `_map` submodule too, using the built-in JS function. Should I go ahead with that change?

yurishkuro: +1

packages/jaeger-ui/src/components/TracePage/TraceStatistics/generateDropdownValue.tsx
```diff
@@ -33,53 +36,51 @@ function getValueTagIsPicked(tableValue: ITableSpan[], trace: Trace, nameSelecto
...
-    return o !== nameSelectorTitle;
-  });
-  availableTags = [];
-  availableTags.push(serviceName);
-  availableTags.push(operationName);
-  availableTags = availableTags.concat(tagKeys);
+  const tags = spansWithFilterTag.map(o => o.tags);
+  let tagKeys = _uniq(_map(_flatten(tags), 'key'));
+  tagKeys = tagKeys.filter(o => o !== nameSelectorTitle);
+  spansWithFilterTag = [];
```
yurishkuro, line 50: these are not spans anymore, this is just output. Could this not be simplified like this?
```javascript
return [serviceName, operationName, ... tagKeys];
```

packages/jaeger-ui/src/components/TracePage/TraceStatistics/generateDropdownValue.tsx
```diff
@@ -33,53 +36,51 @@ function getValueTagIsPicked(tableValue: ITableSpan[], trace: Trace, nameSelecto
...
 /**
  * Used to get the values if no tag is picked from the first dropdown.
  */
 function getValueNoTagIsPicked(trace: Trace, nameSelectorTitle: string) {
-  let availableTags = [];
+  let spansWithFilterTag = [];
   const allSpans = trace.spans;
   if (nameSelectorTitle === serviceName) {
-    availableTags.push(operationName);
+    spansWithFilterTag.push(operationName);
```
yurishkuro, line 65: these are not spans

packages/jaeger-ui/src/components/TracePage/TraceStatistics/generateDropdownValue.tsx
```diff
@@ -33,53 +36,51 @@ function getValueTagIsPicked(tableValue: ITableSpan[], trace: Trace, nameSelecto
...
+    spansWithFilterTag.push(serviceName);
   }
   for (let i = 0; i < allSpans.length; i++) {
     for (let j = 0; j < allSpans[i].tags.length; j++) {
-      availableTags.push(allSpans[i].tags[j].key);
+      spansWithFilterTag.push(allSpans[i].tags[j].key);
     }
   }
-  availableTags = [...new Set(availableTags)];
+  spansWithFilterTag = [...new Set(spansWithFilterTag)];
```
yurishkuro, line 74: could also do `x = _uniq(x)` here, more explicit than set-to-array transform.

packages/jaeger-ui/src/components/TracePage/TraceStatistics/generateDropdownValue.tsx
yurishkuro: or `_concat()`

anshgoyalevil: Used the Spread syntax, to remove `_concat` completely.

yurishkuro (approved):
🎉

PR #2042, opened 2023-12-10
Replace Reselect with Memoize

### LINKED ISSUE #2031: Remove dependency on reselect
dependabot was trying to upgrade reselect in #2026, so I had a look and found only a single place where it is used
```
$ grx reselect packages/*/src
packages/jaeger-ui/src/selectors/dependencies.js:15:import { createSelector } from 'reselect';
```

It is my understanding that it's not doing anything other than provide memoization. Elsewhere in the code we always use `memoizeOne`, so it seems like we can remove `reselect` as a dependency.

Also, this function is only used in the DependencyGraph, so I would simply move it there instead of keeping under /selectors/.
```
$ grx selectors/dependencies packages/*/src
packages/jaeger-ui/src/components/DependencyGraph/index.jsx:28:import { formatDependenciesAsNodesAndLinks } from '../../selectors/dependencies';
```

We can close #2026 once this is done.

cooleditphoto:
Resolves https://github.com/jaegertracing/jaeger-ui/issues/2031
## Description of the changes
Remove package Reselect
Delete file selectors/dependencies
Add the memoize function to components/DependencyGraph/index.tsx

yarn lint & yarn test
Screenshot 2023-12-10 at 1 13 35 PM

## Checklist


yurishkuro (approved):
Thanks!

PR #2057, opened 2023-12-15
Change dependabot frequency to weekly

yurishkuro:
- Daily frequency in unnecessary and occasionally picks multiple patch releases in a single Jaeger cycle, which is just unnecessary overhead on maintainers

## Description of the changes
- Change run frequency to weekly

- CI


jkowall (approved):
ltgm

PR #2063, opened 2023-12-19
Updated Default Configuration for Deep Dependency Graph

### LINKED ISSUE #600: [Feature] change the default layout selection for deep dependency graph view (view=ddg)

Wise-Wizard:
## Problem

Resolves #600, which reported a concern related to [Feature] change the default layout selection for deep dependency graph view (view=ddg)

## Changes

In order to resolve the mentioned issue, the following changes have been implemented:

- Implemented logic in the componentDidMount lifecycle method to check local storage for a previously selected density, setting it as the default if found.
- Modified the updateDensity method to save the selected density in local storage, ensuring persistence across sessions.
- Enhanced user experience by retaining and applying the last selected density even after page reloads.

## Testing

The changes have been thoroughly tested to ensure their correctness and compatibility. The testing process included:

- Modifying the index.test.js, to also consider a default configuration from local storage, rather than only considering a fixed configuration for testing.

## Checklist

To comply with the project's contribution guidelines, the following checklist items have been addressed:

- ✓ I have carefully read and followed the [contributing guidelines](https://github.com/jaegertracing/jaeger/blob/master/CONTRIBUTING_GUIDELINES.md).
- ✓ I have signed all commits.
- ✓ Unit tests for the new functionality have been added.
- ✓ Linting and testing steps have been successfully executed:
Screenshot 2023-12-19 224644


packages/jaeger-ui/src/components/DeepDependencies/Header/LayoutSettings/index.test.js
```diff
@@ -21,9 +21,11 @@ import * as track from '../../index.track';
 
 import { EDdgDensity } from '../../../../model/ddg/types';
 
+const defaultDensityFromLocalStorage = localStorage.getItem('selectedDensity');
```
yurishkuro, line 24: what's the meaning of reading from local storage in a unit test?

packages/jaeger-ui/src/components/DeepDependencies/Header/LayoutSettings/index.test.js
```diff
@@ -66,15 +68,6 @@ describe('LayoutSettings', () => {
...
   });
 
-  it('no-ops if current density is selected', () => {
-    getWrapper()
-      .find(Radio)
-      .at(densityIdx)
-      .simulate('change', { target: { value: props.density } });
-    expect(props.setDensity).not.toHaveBeenCalled();
-    expect(trackDensityChangeSpy).not.toHaveBeenCalled();
-  });
```
yurishkuro, line 76: why did you remove this test?

packages/jaeger-ui/src/components/DeepDependencies/Header/LayoutSettings/index.tsx
```diff
@@ -73,10 +73,24 @@ export const densityOptions = [
 ];
 
 export default class LayoutSettings extends React.PureComponent<TProps> {
+  componentDidMount() {
+    // Check local storage for previously selected density
+    const storedDensity = localStorage.getItem('selectedDensity');
```
yurishkuro, line 78: * define a constant for the name
* prefix the string value with the component name, e.g. `selectedDensity => ddg.layout.density`

packages/jaeger-ui/src/components/DeepDependencies/Header/LayoutSettings/index.test.js
```diff
@@ -66,15 +68,6 @@ describe('LayoutSettings', () => {
     expect(trackDensityChangeSpy).toHaveBeenCalledWith(props.density, newDensity, densityOptions);
   });
```
yurishkuro, line 69: there should be two new tests:
* one that verifies that if a value is found in the localStorage than that value is used
* another that verifies that when changing the selection in the UI the value is recorded in localStorage

packages/jaeger-ui/src/components/DeepDependencies/Header/LayoutSettings/index.test.js
```diff
@@ -21,9 +21,11 @@ import * as track from '../../index.track';
 
 import { EDdgDensity } from '../../../../model/ddg/types';
 
+const defaultDensityFromLocalStorage = localStorage.getItem('selectedDensity');
+
 describe('LayoutSettings', () => {
   const props = {
-    density: EDdgDensity.PreventPathEntanglement,
+    density: defaultDensityFromLocalStorage || EDdgDensity.PreventPathEntanglement,
```
yurishkuro, line 28: don't see why you need this

packages/jaeger-ui/src/components/DeepDependencies/Header/LayoutSettings/index.test.js
Wise-Wizard: I apologise for my inexperience in writing tests, because I haven't worked on writing tests before. I had thought since the default configuration has changed, the selected node while running the test could be different from the default, hence I though it would be better to use the default selected node when the component mounts for testing rather than just the hard coded EDdgDensity.PreventPathEntanglement node

packages/jaeger-ui/src/components/DeepDependencies/Header/LayoutSettings/index.test.js
Wise-Wizard: Yes, I was wrong it's working fine without this additional configuration. I have resolved it back to it's original state

packages/jaeger-ui/src/components/DeepDependencies/Header/LayoutSettings/index.test.js
Wise-Wizard: Got it, will start working on that!

packages/jaeger-ui/src/components/DeepDependencies/Header/LayoutSettings/index.test.js
Wise-Wizard: I thought because of the new local storage addition for setting a default configuration, this test would be trivial and not necessary because the default configuration will be updated every time

packages/jaeger-ui/src/components/DeepDependencies/Header/LayoutSettings/index.tsx
Wise-Wizard: Can you please elaborate more on the first change you require me to do, I have added the second change => Changed the 'selectedDensity' to 'ddg.layout.density'

packages/jaeger-ui/src/components/DeepDependencies/Header/LayoutSettings/index.test.js
yurishkuro: This test is not about defaults, but about what happens when the current selection is re-selected.

packages/jaeger-ui/src/components/DeepDependencies/Header/LayoutSettings/index.test.js
Wise-Wizard: Update: I have finished writing the tests, should I raise a PR?

packages/jaeger-ui/src/components/DeepDependencies/Header/LayoutSettings/index.test.js
Wise-Wizard: Update: I have rewrittend the tests, which makes sure trackDensityChange is not being called, and it is working fine

Wise-Wizard: Screenshot 2023-12-20 122945
All tests have successfully passed and new tests added and old tests restored.

Wise-Wizard: @yurishkuro I have handled the test case when current selection is reselected.
Should I raise a PR now?

packages/jaeger-ui/src/components/DeepDependencies/Header/LayoutSettings/index.test.js
```diff
@@ -48,6 +48,26 @@ describe('LayoutSettings', () => {
...
+    const newIdx = 2;
+    const newDensity = densityOptions[newIdx].option;
+    getWrapper()
+      .find(Radio)
+      .at(newIdx)
+      .simulate('change', { target: { value: newDensity } });
+    expect(props.setDensity).toHaveBeenCalledWith(newDensity);
+    expect(trackDensityChangeSpy).toHaveBeenCalledWith(props.density, newDensity, densityOptions);
+    // Ensure the value is recorded in localStorage
+    expect(localStorage.getItem(LayoutSettings.DEFAULT_DENSITY)).toBe(newDensity);
```
yurishkuro, line 68: I would recommend setting local storage slot to null at the start of the test, this way you can be sure that it is actually changing and wasn't already holding your desired value. It would make the test self contained.

packages/jaeger-ui/src/components/DeepDependencies/Header/LayoutSettings/index.test.js
```diff
@@ -48,6 +48,26 @@ describe('LayoutSettings', () => {
     jest.clearAllMocks();
   });
 
+  it('uses value found in localStorage', () => {
+    const selectedDensity = props.density;
+    localStorage.setItem(LayoutSettings.DEFAULT_DENSITY, selectedDensity);
+    const radios = getWrapper().find(Radio);
+    expect(Array.from(radios).findIndex(radio => radio.props.checked)).toBe(densityIdx);
+  });
```
yurishkuro, line 56: this test is confusing to me. First, you're setting local storage to value that is already selected. A logical test would be:
  1. set local storage to null, initialize component, validate that it sets the option to some default value
  2. set local storage to another value (different from default), initialize the component, validate that it reused the value from local storage

packages/jaeger-ui/src/components/DeepDependencies/Header/LayoutSettings/index.test.js
```diff
@@ -71,7 +91,7 @@ describe('LayoutSettings', () => {
       .find(Radio)
       .at(densityIdx)
       .simulate('change', { target: { value: props.density } });
-    expect(props.setDensity).not.toHaveBeenCalled();
+    // expect(props.setDensity).not.toHaveBeenCalled();
     expect(trackDensityChangeSpy).not.toHaveBeenCalled();
```
yurishkuro, line 95: Why does this test need to be modified? Your new changes should not be affecting the behavior that this test was testing.

packages/jaeger-ui/src/components/DeepDependencies/Header/LayoutSettings/index.test.js
yurishkuro: which PR?

packages/jaeger-ui/src/components/DeepDependencies/Header/LayoutSettings/index.tsx
```diff
@@ -73,10 +73,27 @@ export const densityOptions = [
...
+  componentDidMount() {
+    // Check local storage for previously selected density
+    const storedDensity = localStorage.getItem(LayoutSettings.DEFAULT_DENSITY);
+    if (storedDensity && densityOptions.some(option => option.option === storedDensity)) {
+      // Set the stored density as the default if it's a valid option
+    this.props.setDensity(storedDensity as EDdgDensity);
+    }
+  }
+
+  private static readonly DEFAULT_DENSITY = 'selectedDensity';
```
yurishkuro, line 86:
```suggestion
  // key used to store density selection in local storage
  private static readonly STORED_DENSITY_KEY = 'ddg.layout.density';
```

yurishkuro: define constant at the top of the class, before methods

packages/jaeger-ui/src/components/DeepDependencies/Header/LayoutSettings/index.test.js
Wise-Wizard: The props.setDensity is being called every time the page renders, because it has to check from localStorage and set the density. Hence, trackDensityChange won't be called but setDensity is necessary everytime.

packages/jaeger-ui/src/components/DeepDependencies/Header/LayoutSettings/index.tsx
```diff
@@ -73,10 +73,28 @@ export const densityOptions = [
 ];
 
 export default class LayoutSettings extends React.PureComponent<TProps> {
+
+  componentDidMount() {
+    // Check local storage for previously selected density
+    const storedDensity = localStorage.getItem(LayoutSettings.DEFAULT_DENSITY);
+    if (storedDensity && densityOptions.some(option => option.option === storedDensity)) {
+      // Set the stored density as the default if it's a valid option
+    this.props.setDensity(storedDensity as EDdgDensity);
```
yurishkuro, line 82:
```suggestion
      this.props.setDensity(storedDensity as EDdgDensity);
```

packages/jaeger-ui/src/components/DeepDependencies/Header/LayoutSettings/index.test.js
yurishkuro: If local storage is empty than setDensity should not be called. You may want to clear local storage in beforeEach so that you don't get cross-test effects.

Wise-Wizard: Ohh Understood, that's brilliant should have noticed that before. Thanks for your input, I have cleared the local storage and added the test back, all the tests are passing now. I have raised a commit for the same!

yurishkuro (dismissed):
Thanks! LGTM

Wise-Wizard:
All the checks except Lint has passed, Is there something that's remaining to be done from my end, or is it common and you shall kindly take a look and merge it manually?

yurishkuro:
error message:
```
*** PLEASE RUN yarn prettier AND RESUBMIT ***
```

Wise-Wizard:
Running yarn prettier, didn't resolve it. Please let me know of the steps I have to take to resolve this. It will be of great help

yurishkuro (approved):

yurishkuro:
I ran `yarn prettier` and it applied the changes:
```
$ gst
	modified:   packages/jaeger-ui/src/components/DeepDependencies/Header/LayoutSettings/index.tsx
```
I just pushed them.

Wise-Wizard:
Yes, even I did the same thing ran yarn prettier but it didn't seem to work for me locally :(

Wise-Wizard:
Thanks a lot for your help, this PR really helped me grow and learn. Appreciate it. Can you please suggest some issues you would like me to take a look at next?

yurishkuro:
I don't have particular suggestions, look for any help-wanted tickets that sound interesting to you.

Wise-Wizard:
Sure, will take up the next one I find interesting:)

PR #2064, opened 2023-12-24
fix: Fix some linting warnings

EshaanAgg:
Fixes part of #1608 

## Description of the changes
- Added the minimal types needed to fix the linting warnings
- Replaced `{}` with the `object` type

- Ran `yarn lint` locally 

## Checklist


packages/jaeger-ui/src/components/TracePage/ScrollManager.tsx
```diff
@@ -87,7 +87,7 @@ function isSpanHidden(span: Span, childrenAreHidden: Set<string>, spansMap: Map<
  */
 export default class ScrollManager {
   _trace: Trace | TNil;
-  _scroller: IScroller;
+  _scroller: IScroller | undefined;
```
yurishkuro, line 90: not a good change - messing with semantics of the type just to appease a linter. The constructor always takes a scroller, so this isn't an optional field

packages/jaeger-ui/src/components/SearchTracePage/SearchForm.track.tsx
```diff
@@ -34,7 +34,7 @@ export const CATEGORY_SERVICE = `${FORM_CATEGORY_BASE}/serviceName`;
 export function trackFormInput(
   resultsLimit: number,
   operation: string,
-  tags: any,
+  tags: string[],
```
yurishkuro, line 37: could you point to where this is shown as array being passed?

yurishkuro (commented):
nice one!

packages/jaeger-ui/src/components/TracePage/ScrollManager.tsx
EshaanAgg: It's this `destroy` function that is setting the same to be `undefined`.
- If this is intended behavior, we should allow it to be `undefined` as we are explicitly setting it to be.
- Otherwise we should remove this line from this function.

https://github.com/jaegertracing/jaeger-ui/blob/ec3aba07a6439800b7477e761a3feabd4e74d47c/packages/jaeger-ui/src/components/TracePage/ScrollManager.tsx#L269-L273

packages/jaeger-ui/src/components/SearchTracePage/SearchForm.track.tsx
EshaanAgg: https://github.com/jaegertracing/jaeger-ui/blob/ec3aba07a6439800b7477e761a3feabd4e74d47c/packages/jaeger-ui/src/components/SearchTracePage/SearchForm.jsx#L246

This was the code that I was able to trace it to. Since all the parameters implicitly have the type `any` here, I couldn't derive any concrete inferences. But since the same is being used in the context of a form, I presume that it must be an array of `string`.

yurishkuro: I think it's just a string that gets parsed later. https://github.com/jaegertracing/jaeger-ui/blob/ec3aba07a6439800b7477e761a3feabd4e74d47c/packages/jaeger-ui/src/components/SearchTracePage/SearchForm.jsx#L255

packages/jaeger-ui/src/components/TracePage/ScrollManager.tsx
yurishkuro: don't really understand the point of this destroy() but ok, let's just keep it consistent with other fields and use TNil?

EshaanAgg: Sounds good!

packages/jaeger-ui/src/components/SearchTracePage/SearchForm.track.tsx
EshaanAgg: Got it!

yurishkuro (approved):
Thanks!

PR #2070, opened 2023-12-25
Refac: Migrate 3 components from Enzyme to RTL

EshaanAgg:
Fixes part of #1668 

## Description of the changes
Migrates the following components from Enzyme to RTL
1. `CircularProgressbar`
2. `ErrorMessage`
3. `TraceName`

- Removed the instances of snapshot testing from the same and migrated to tests checking the functionality associated with them
- Refactored some of the original code of the components to make them more readable

- Running the tests locally

## Checklist


packages/jaeger-ui/src/components/common/ErrorMessage.test.js
```diff
@@ -13,141 +13,124 @@
...
-    wrapper = shallow(<ErrorMessage error={error} />);
-  });
+describe('<ErrorMessage>', () => {
+  const errorMessage = 'some error message';
 
   it('is ok when not passed an error', () => {
-    wrapper.setProps({ error: null });
-    expect(wrapper).toBeDefined();
+    render(<ErrorMessage />);
+    expect(screen).toBeDefined();
```
yurishkuro, line 26: isn't screen always defined?

packages/jaeger-ui/src/components/common/ErrorMessage.test.js
```diff
@@ -13,141 +13,124 @@
...
-    expect(msg.shallow().text()).toMatch(error);
+    render(<ErrorMessage error={errorMessage} />);
+    expect(screen.getByText(errorMessage)).toBeInTheDocument();
   });
 
   it('<Details /> renders empty on string error', () => {
-    wrapper = shallow(<Details error={error} />);
-    expect(wrapper).toMatchSnapshot();
+    render(<Details error={errorMessage} />);
+    expect(screen).toBeDefined();
```
yurishkuro, line 36: is this ever false? why not assert by error message?

packages/jaeger-ui/src/components/common/ErrorMessage.test.js
```diff
@@ -13,141 +13,124 @@
...
   });
 
   it('<Details /> renders wrapper on wrap', () => {
-    error = {
-      message: 'some-http-ish-message',
```
yurishkuro, line 47: why not validate the message too?

packages/jaeger-ui/src/components/common/ErrorMessage.test.js
```diff
@@ -13,141 +13,124 @@
...
+    const error = {
```
yurishkuro, line 40: this struct is repeated many times, maybe pull into a global const?

packages/jaeger-ui/src/components/common/ErrorMessage.test.js
```diff
@@ -13,141 +13,124 @@
...
-      expect(details.length).toBe(1);
-      const detailsWrapper = details.shallow();
-      expect(detailsWrapper).toMatchSnapshot();
+    // All the keys expect httpBody should be present as is
+    Object.keys(error).forEach(key => {
+      if (key !== 'httpBody') expect(screen.getByText(error[key])).toBeInTheDocument();
     });
+
+    // The body should be truncated
+    expect(screen.getByText(`${error.httpBody.slice(0, 1021).trim()}...`)).toBeInTheDocument();
```
yurishkuro, line 119: what us this magic number 1021? If the component has a specific limit, let's expose that as a constant from the component and use here.

packages/jaeger-ui/src/components/common/ErrorMessage.test.js
EshaanAgg: My bad. Will fix.

packages/jaeger-ui/src/components/common/ErrorMessage.test.js
EshaanAgg: Need to do away with the `toBeDefined()` on `screen` as previously pointed. Will fix it.

packages/jaeger-ui/src/components/common/ErrorMessage.test.js
EshaanAgg: This is a test for the `Details` component, which does not have `message` listed in its props and doesn't render the same. `message` is needed when testing the `Message` or `ErrMessage` component.

packages/jaeger-ui/src/components/common/ErrorMessage.test.js
EshaanAgg: Tried doing that it causes more redirection and hard-to-read tests than it reduces duplication. The primary reason for the same being each test changes/deletes just one key of the object. 
If we were to do that, then in each test we would first need to restore the global const from previous test, and then augment it again according to it's need.

packages/jaeger-ui/src/components/common/ErrorMessage.test.js
EshaanAgg: It's the maximum limit `1024` - length of `...`. Will expose it as a constant.

yurishkuro (approved):
Thanks!

PR #2071, opened 2023-12-25
feat: remove `ReactGA` and migrate to `GA4` for tracking

### LINKED ISSUE #1332: [Feature]: Support Google Analytics 4 properties

EshaanAgg:
Fixes #1332 

## Description of the changes
- Remove the `ReactGA` library for analytics and instead load the script ourselves asynchronously
- Manually make calls to the `gtag` function with the appropriate parameters

- NA

## Checklist


packages/jaeger-ui/src/utils/tracking/ga.tsx
```diff
@@ -24,6 +23,40 @@ import { logTrackingCalls } from './utils';
 import { getAppEnvironment, shouldDebugGoogleAnalytics } from '../constants';
 import parseQuery from '../parseQuery';
 
+// Modify the global scope to add the `gtag` function and the `dataLayer` array
```
yurishkuro, line 26: Why do we need to attach gtag function to the window object? Could we not use the same approach as in ReactGA where it had some module-level globals?

packages/jaeger-ui/src/utils/tracking/ga.tsx
```diff
@@ -24,6 +23,40 @@ import { logTrackingCalls } from './utils';
...
+  }
+
+  const script = document.createElement('script');
+  script.async = true;
+  script.src = `${gtagUrl}?id=${GA_MEASUREMENT_ID}`;
+  document.body.appendChild(script);
+
+  window.dataLayer = window.dataLayer || [];
+  window.gtag = (...args) => {
+    window.dataLayer.push(args);
```
yurishkuro, line 48: what is the "dataLayer" and why it's needed? please add comments explaining

packages/jaeger-ui/src/utils/tracking/ga.tsx
yurishkuro: based on https://developers.google.com/tag-platform/gtagjs/install, I think only the dataLayer needs to be public, the gtag function can be private in this module

packages/jaeger-ui/src/utils/tracking/ga.tsx
```diff
@@ -24,6 +23,40 @@ import { logTrackingCalls } from './utils';
 import { getAppEnvironment, shouldDebugGoogleAnalytics } from '../constants';
 import parseQuery from '../parseQuery';
 
+// Modify the global scope to add the `gtag` function and the `dataLayer` array
+declare global {
```
yurishkuro, line 27: What effect does this have on other code? Will the other code start seeing `window` as having a `dataLayer` property? Why do we want that, shouldn't this knowledge be private to this module?

yurishkuro (commented):
I think the approach is great! One less dependency.

packages/jaeger-ui/src/utils/tracking/ga.tsx
EshaanAgg: Ahh yes, nice catch. I will update the implementation! Thanks.

packages/jaeger-ui/src/utils/tracking/ga.tsx
EshaanAgg: I don't think this knowledge should remain private to this module, as the `dataLayer` attribute is globally attached to the `window` object. Keeping it private means that all the other modules can assume there is no property named `dataLayer` on `window`, and thus if needed, they can add that themselves, which will mess with our current implementation.

EshaanAgg: On second thoughts, the most ideal implementation would be for `dataLayer` to have the appropriate types within `ga.ts` and have the type `never` outside it so that no other module tries to access or modify the same.

packages/jaeger-ui/src/utils/tracking/ga.test.js
```diff
@@ -96,33 +97,33 @@ describe('google analytics tracking', () => {
...
     });
   });
 
   describe('trackError', () => {
     it('tracks an error', () => {
       tracking.trackError('a');
-      expect(calls).toEqual([
-        ['send', { hitType: 'exception', exDescription: expect.any(String), exFatal: false }],
+      expect(window.dataLayer).toEqual([
+        ['event', 'exception', { description: expect.any(String), fatal: false }],
```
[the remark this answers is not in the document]
yurishkuro: If there is no external schema and we control the names, changing them means we're introducing breaking changes for people who may already be using the existing integration, so their dashboards could break.

packages/jaeger-ui/src/utils/tracking/utils.tsx
```diff
@@ -12,14 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import ReactGA from 'react-ga';
+import { WindowWithGATracking } from './ga';
+
+declare let window: WindowWithGATracking;
 
 // eslint-disable-next-line import/prefer-default-export
 export const logTrackingCalls = () => {
```
yurishkuro, line 20: can we move this function inside ga.tsx (and tests)?  I don't see it being used outside, so odd to expose as public, especially since the dataLayer contact is for GA only.

packages/jaeger-ui/src/utils/tracking/ga.tsx
```diff
@@ -24,6 +23,40 @@ import { logTrackingCalls } from './utils';
...
+export interface WindowWithGATracking extends Window {
+  dataLayer: (string | object)[][] | undefined;
+}
+
+declare let window: WindowWithGATracking;
+
+// Function to add a new event to the Google Analytics dataLayer
+const gtag = (...args: (string | object)[]) => {
+  if (window !== undefined)
+    if (window.dataLayer !== undefined) {
```
yurishkuro, line 39: Why are you relying on precisely `undefined` here? What about this:
```suggestion
  if (window)
    if (window.dataLayer && isArray(window.dataLayer)) {
```

packages/jaeger-ui/src/utils/tracking/utils.tsx
yurishkuro: in fact, if feels like it could be combined with the gtag() function instead of always calling it explicitly

packages/jaeger-ui/src/vite-env.d.ts
```diff
@@ -14,3 +14,13 @@
...
+
+declare namespace global {
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  interface Window {
+    // dataLayer property is used by Google Tag Manager
+    // It should be only be changed with the help of the custom Interface WindowWithGATracking
+    // defined in utils/tracking/ga.ts
+    dataLayer: never;
+  }
+}
```
yurishkuro, line 26: Could you elaborate why this change is needed?

packages/jaeger-ui/src/utils/tracking/ga.test.js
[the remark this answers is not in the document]
EshaanAgg: There is no definite schema that these calls adhere to. You can find the [API specification here](https://developers.google.com/tag-platform/gtagjs/reference). There are also just some `recommended events` and `recommended parameters` for these events. 

I tried to look at the source of the `ReactGA` and the API specification shared above to figure out the possible actual `tag` calls the library made for us.

packages/jaeger-ui/src/vite-env.d.ts
EshaanAgg: > On second thoughts, the most ideal implementation would be for `dataLayer` to have the appropriate types within `ga.ts` and have the type `never` outside it so that no other module tries to access or modify the same. 

This had been my initial reasoning but I think it would be better to remove it considering the way you have advised to handle the types in the previous comment.

packages/jaeger-ui/src/utils/tracking/utils.tsx
EshaanAgg: Aight. The combining idea sounds nice. I'll check it out.

packages/jaeger-ui/src/utils/tracking/ga.tsx
EshaanAgg: Sign catch!

packages/jaeger-ui/src/utils/tracking/ga.test.js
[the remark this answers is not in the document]
yurishkuro: I an judging by the fact that the unit test is changing, so it seems like you're changing the schema. Can we keep it the same as before?

packages/jaeger-ui/src/utils/tracking/utils.tsx
EshaanAgg: I have combined the logging with the `gtag` function itself in the latest implementation. This `WindowWithGATracking` is now only defined and used in `ga.tsx`

packages/jaeger-ui/src/utils/tracking/ga.test.js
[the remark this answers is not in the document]
EshaanAgg: Earlier, the unit tests tested what calls were being made to the `ReactGA` library. What calls, in turn, the same made to the `gtag` library was completely abstracted away from us.
Now the tests are checking what `gtag` calls are being made. Since we are losing the `ReactGA` library altogether, the tests can't be kept the same.

yurishkuro (approved):
Thanks!

PR #2080, opened 2024-01-02
feat: remove `is-promise` library

### LINKED ISSUE #2074: [Feature]: remove the `is-promise` library
### Requirement

Currently, we are using the `is-promise` package's exported `isPromise` function in some tests to check if the returned payload is a promise. This feels very counterintuitive. 
https://github.com/jaegertracing/jaeger-ui/blob/f7db33e8ecb2e77c64ab7a84b428e8a34630c6e5/packages/jaeger-ui/src/actions/jaeger-api.test.js#L27
https://github.com/jaegertracing/jaeger-ui/blob/f7db33e8ecb2e77c64ab7a84b428e8a34630c6e5/packages/jaeger-ui/src/actions/jaeger-api.test.js#L84-L87

### Problem

Since the codebase uses TypeScript, either of the two things should happen:
- The types of the functions using this function should be updated to give us the confidence that the returned result is always a promise of the appropriate type. (And then these type assertions should be removed from the tests)
- We can make use of the `PromiseLike` type by TS.

Anyhow, I believe the use case is too trivial to add a new library. 

### Proposal
Remove the library and adopt one of the above mentioned approaches.

EshaanAgg:
Fixes #2074 

## Description of the changes
- Removed the `is-promise` library
- Added the type assertions wherever necessary
- Removed the redundant tests
- Made use of the native `Promise.allSettled` function (as described in a comment)

There seems to be an unrelated test in `date.test.js` that fails after these changes. I have fixed the test with what I think the intended testing was supposed to do. Would appreciate a review on the same!

- Running the test suite locally

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/actions/jaeger-api.test.js
+++ b/packages/jaeger-ui/src/actions/jaeger-api.test.js
@@ -23,8 +23,11 @@ jest.mock(
     })
 );
 
-import sinon from 'sinon';
+function isPromise(p) {
+  return p !== null && typeof p === 'object' && typeof p.then === 'function' && typeof p.catch === 'function';
+}
 
+import sinon from 'sinon';
 import * as jaegerApiActions from './jaeger-api';
 import JaegerAPI from '../api/jaeger';
 
@@ -48,6 +51,11 @@ describe('actions/jaeger-api', () => {
     expect(() => mock.verify()).not.toThrow();
   });
 
+  it('@JAEGER_API/FETCH_TRACE should return the promise', () => {
+    const { payload } = jaegerApiActions.fetchTrace(id);
+    expect(isPromise(payload)).toBeTruthy();
+  });
+
   it('@JAEGER_API/FETCH_TRACE should attach the id as meta', () => {
     const { meta } = jaegerApiActions.fetchTrace(id);
     expect(meta.id).toBe(id);
@@ -57,6 +65,11 @@ describe('actions/jaeger-api', () => {
     mock.expects('searchTraces').withExactArgs(sinon.match.has('traceID', ids));
     jaegerApiActions.fetchMultipleTraces(ids);
     expect(() => mock.verify()).not.toThrow();
+  });
+
+  it('@JAEGER_API/FETCH_MULTIPLE_TRACES should return the promise', () => {
+    const { payload } = jaegerApiActions.fetchMultipleTraces(ids);
+    expect(isPromise(payload)).toBeTruthy();
   });
 
   it('@JAEGER_API/FETCH_MULTIPLE_TRACES should attach the ids as meta', () => {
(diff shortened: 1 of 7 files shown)


packages/jaeger-ui/src/actions/jaeger-api.test.js
```diff
@@ -49,11 +48,6 @@ describe('actions/jaeger-api', () => {
     expect(() => mock.verify()).not.toThrow();
   });
 
-  it('@JAEGER_API/FETCH_TRACE should return the promise', () => {
-    const { payload } = jaegerApiActions.fetchTrace(id);
-    expect(isPromise(payload)).toBeTruthy();
-  });
```
yurishkuro, line 55: Why are these tests removed? What is replacing them?

packages/jaeger-ui/src/actions/jaeger-api.test.js
```diff
@@ -174,20 +133,3 @@ describe('actions/jaeger-api', () => {
...
-      Promise.reject(2),
-      Promise.resolve(3),
-    ]);
-
-    expect(res).toEqual([
-      { status: 'fulfilled', value: 1 },
-      { status: 'rejected', reason: 2 },
-      { status: 'fulfilled', value: 3 },
-    ]);
-  });
```
EshaanAgg, line 192: I removed this test as we have now deprecated the custom implementation of `allSettled` for the builtin `Promise.allSettled`.

packages/jaeger-ui/src/actions/jaeger-api.test.js
yurishkuro: > they are essentially type-checking, which should be TypeScript's job.

how is that type-checking achieved given that the source file in question is not in TS, but JS? 

Also

> Each of these fetchXYZ functions is essentially createAction wrapper 

That's an implementation detail of those functions, but the tests were enforcing the public API. Implementations may change.

EshaanAgg: Ahh, my bad. I completely missed the `.js` extension. If we have to enforce the public API of these functions, then it would best to keep checking if we have a promise returned. I will revert the tests.

EshaanAgg:
I have reverted the tests testing the public API, and removed the `is-promise` library in favour of a simple custom implementation that checks the supplied object must be PromiseLike, that is, have a `then` and `catch` function. I have removed the analogous tests for the `readJsonFile` function as the file is in TS, and we can use explicit return type-assertions to make sure of the same.

yurishkuro:
Have you considered converting packages/jaeger-ui/src/actions/jaeger-api.js to Typescript?

EshaanAgg:
I tried to do so but later decided to avoid doing it for two reasons:
- We would have to go back to our custom implementation of `allSettled`
When changing the file to TS and using `allSettled`, we get the following TS error, asking us to increase the compiler to a more later one. Do you think this can lead to breaking changes?
```
Property 'allSettled' does not exist on type 'PromiseConstructor'. Do you need to change your target library? Try changing the 'lib' compiler option to 'es2020' or later. ts(2550)
```
- Loose types 
Even after flipping to TS, due to the nature of the `redux-actions` library, we get not very concrete types like the following:
```
const fetchAllServiceMetrics: ActionFunction<Action<any>, undefined, undefined, undefined, undefined>
const fetchDeepDependencyGraph: ActionFunctionAny<ActionMeta<any, {
    query: any;
}>>
const fetchServices: ActionFunction<Action<any>, undefined, undefined, undefined, undefined>
```
I don't think these types enforce the `PromiseLike` interface that you had earlier mentioned, thus making the exercise moot.

yurishkuro (approved):
Thanks!

Issue #2093, opened 2024-01-06
Figure out what `yarn start` in plexus is supposed to do


yurishkuro:
In `packages/plexus/package.json` there is a `start` command. It does not work currently unless the changes below are applied (committed in #2094 for easier reuse). However, even with these changes a web server is started that doesn't seem to do anything, a request to http://localhost:5000/  only shows `Cannot GET /`.

What is supposed to happen with this command? Is it perhaps just a remnant of auto-generated webpack before Plexus was a separate module? This ticket is to figure this out and post findings (e.g. look through the history of commits to see the original intent).

```diff
diff --git a/packages/plexus/webpack-factory.js b/packages/plexus/webpack-factory.js
index 89538c5f..c5120d27 100644
--- a/packages/plexus/webpack-factory.js
+++ b/packages/plexus/webpack-factory.js
@@ -72,27 +72,22 @@ function makeDevConfig() {
   const config = {
     entry,
     mode: 'development',
-    devtool: 'cheap-module-eval-source-map',
+    devtool: 'cheap-module-source-map',
     output: {
       path: join(__dirname, 'build'),
       publicPath: '/',
       filename: 'assets/[name].js',
     },
+    stats: 'normal',
     devServer: {
       port: 5000,
       hot: false,
       historyApiFallback: true,
-      overlay: true,
-      index: 'index',
-      contentBase: join(__dirname, 'build'),
-      staticOptions: {
-        extensions: ['.htm', '.html'],
+      client: {
+        overlay: true,
       },
-      stats: {
-        all: false,
-        errors: true,
-        timings: true,
-        warnings: true,
+      static: {
+        directory: join(__dirname, 'build'),
       },
     },
```

Wise-Wizard:
My understanding so far about yarn start is it performs the following in sequence:
1) Task Clean:  Removes the lib and dist directories, effectively cleaning up any previously generated files.
2) Task Bundle Worker: Uses Webpack to generate the UMD bundle for the layout.worker module. The configuration is specified in webpack.layout-worker.config.js.
3) Task Dev Server:  Uses webpack-dev-server to start a development server. The configuration is specified in webpack.dev.config.js (localhost:5000). 
The configuration of both web.layout and web.dev makes use of original webpack-factory file itself and I presume the reason for Cannot Get / being shown up is due to a missing index.html file whch would serve as entry point for React projects I guess.

yurishkuro:
[...]

PR #2094, opened 2024-01-06
Fix `start` in `packages/plexus/package.json`

yurishkuro:
- Part of #2093


yurishkuro:
Resolved by #2538

PR #2096, opened 2024-01-07
setup OSSF Scorecard workflow 

mmorel-35:
-  Improve OSSF Score
## Description of the changes
- setup OSSF Scorecard workflow 
- Pin github Actions versions
- set permissions on workflows

-  CI

## Checklist

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -1,0 +1,56 @@
+name: Scorecard supply-chain security
+
+on:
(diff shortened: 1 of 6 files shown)


yurishkuro (approved):

.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,56 @@
+name: Scorecard supply-chain security
+
+on:
```
yurishkuro, line 3: I would also suggest manual dispatch as an option

PR #2133, opened 2024-01-24
Fix temp files for depcheck

yurishkuro:
It was creating always the same files on Mac, probably because it requires XXXX to be the end of the pattern.


albertteoh (dismissed):

albertteoh (approved):

jkowall (approved):
lgtm!

PR #2135, opened 2024-01-26
Use import from /dist/ instead of symlink

zacanger:
This is an insignificant issue, so I didn't open an issue ticket beforehand. There's a symlink here to a file in node_modules. It's small problem, but the symlink is unnecessary. I came across this because I have this repo as a submodule in another project, and back up those changes (and a lot of other repos) with `rsync`, the sync checks output to see if there are errors. I would guess this also causes problems on Windows without WSL.

## Description of the changes
Change the import. This Just Works™ in Vite.

`yarn lint && yarn test && yarn start`, and `yarn build`, and grepping `react-vis` clases in the assets.

Screenshot 2024-01-26 at 11 10 46

Screenshot 2024-01-26 at 11 10 37

## Checklist


yurishkuro (approved):
not sure if importing from /dist/ is a good practice, but certainly not worse than the symlink

thanks

zacanger:
> not sure if importing from /dist/ is a good practice, but certainly not worse than the symlink
> 
> thanks

It's definitely not, but [that package](https://github.com/uber/react-vis/blob/master/packages/react-vis/package.json#L7) doesn't specify a `style` in package.json (like [this one](https://github.com/twbs/bootstrap/blob/main/package.json#L40)).

PR #2145, opened 2024-01-31
Support uploading traces in OpenTelemetry format (OTLP JSON)

NavinShrinivas:
- Part of https://github.com/jaegertracing/jaeger/issues/4949

## Description of the changes
- Adds a redux middleware to make API calls to convert OTLP traces to jaeger traces when uploaded

- Unit test have been written, but apart form that OTLP traces were fed into the frontend (when compiled using all-in-one) to get correct behabiour

## Checklist


packages/jaeger-ui/tsconfig.json
```diff
@@ -13,5 +13,5 @@
 // limitations under the License.
 {
   "extends": "../../tsconfig",
-  "include": ["src/**/*.tsx", "typings"],
+  "include": ["src/**/*.tsx", "typings"]
```
yurishkuro, line 16: run `yarn install --frozen-lockfile` and then `yarn prettier`

packages/jaeger-ui/src/middlewares/index.js
```diff
@@ -45,4 +47,29 @@ export const historyUpdateMiddleware = store => next => action => {
   return next(action);
 };
 
+export const transformOTLPMiddleware = store => next => action => {
```
yurishkuro, line 50: is it supposed to be called from somewhere?

packages/jaeger-ui/src/middlewares/index.js
```diff
@@ -45,4 +47,29 @@ export const historyUpdateMiddleware = store => next => action => {
...
+  if (action.type === String([`${loadJsonTraces}_FULFILLED`])) {
```
yurishkuro, line 51: can you explain what's happening here?

packages/jaeger-ui/src/middlewares/index.js
```diff
@@ -45,4 +47,29 @@ export const historyUpdateMiddleware = store => next => action => {
...
+    // Check if action.payload is OTLP and make API call if so
+    // We are allowed to change the action.payload here
+    //
+    if ('resourceSpans' in action.payload) {
```
yurishkuro, line 55: 1. please add a unit test
2. did you try to get an official JSON output for OTEL data? E.g. by running OTEL Collector with debugexporter (as described in the ticket). I would suggest saving a trimmed down version of that output as a test fixture file here and loading it in the unit test

packages/jaeger-ui/src/middlewares/index.js
NavinShrinivas: The middleware is applied to listen to actions in configure-store.js, hence these middelwares are triggered when the particular store is dispatched. I don't have to call them explicitly anywhere

packages/jaeger-ui/src/middlewares/index.js
NavinShrinivas: Im checking if the actions thats being invoked is loadJsonTrace_FULLFILLED (which is what is called after the file upload is finished) After which I check if its a OTLP type trace and make an API call. 

If an API call was made I make a custom action and pass it on the same reducer using the `next` variable(its a function) If not it just forwards the actioon to the default reducer

packages/jaeger-ui/src/middlewares/index.js
NavinShrinivas: > 2. did you try to get an official JSON output for OTEL data? E.g. by running OTEL Collector with debugexporter (as described in the ticket). I would suggest saving a trimmed down version of that output as a test fixture file here and loading it in the unit test

Yes I have tested the enter pipeline from uploading OTLP debug expoerted traces to having them show up in the UI....I'll look int adding them as test fixuters

packages/jaeger-ui/src/middlewares/index.js
yurishkuro: > if the actions thats being invoked is `loadJsonTrace_FULLFILLED`

but why is it `${loadJsonTraces}` instead? Is `loadJsonTraces` a constant somewhere?

packages/jaeger-ui/src/middlewares/index.js
yurishkuro: is it the fact that it's exported what makes it included/called?

packages/jaeger-ui/src/middlewares/index.js
NavinShrinivas: Yes it's a constant defined in file-upload-api.tsx (not sure about this file name)

packages/jaeger-ui/src/middlewares/index.js
NavinShrinivas: yes, and here https://github.com/jaegertracing/jaeger-ui/blob/60e5ce43746c42e4a8cf4cbb77ff2eeed2a0c46c/packages/jaeger-ui/src/utils/configure-store.js#L41-L45 

the middle wares are imported and applied

packages/jaeger-ui/src/middlewares/index.js
NavinShrinivas: This is now handled in index.test.js, a new folder called `fixtures` was created to house these trace examples

packages/jaeger-ui/src/middlewares/fixtures/JaegerTrace.json
```diff
@@ -0,0 +1,344 @@
+{
```
yurishkuro, line 1: let's trim these fixtures to a single span. There's no value in multiple spans, the important part is really just the beginning of JSON, to ensure the right structure is used.

NavinShrinivas: I have implemented this and pushed the change, ive also change the fixture file names to be consistent with frontend

packages/jaeger-ui/src/middlewares/index.js
yurishkuro: but it's not a string constant:
```
export const loadJsonTraces = createAction(
  '@FILE_READER_API/LOAD_JSON',
  fileList => readJsonFile(fileList),
  fileList => ({ fileList })
);
```
yet you're using it as if it's a string or renders into a string.

NavinShrinivas: from redux code : 

https://github.com/reduxjs/redux-toolkit/blob/261ff262745ff894c108944c5cdc7f49c6d61214/src/createAction.ts#L261-L281

And : 
https://stackoverflow.com/questions/61223096/also-the-action-creator-overrides-tostring-so-that-the-action-type-becomes-it
`createAction()` overrides the functions toString method (which is what is called during `${}`

NavinShrinivas:
Hey, are you waiting for commits to meet the test requirement? I'd like to more this pr forward with necessary changes

yurishkuro:
please make sure the CI is green.

yurishkuro (approved):

yurishkuro:
I found one issue. When I try to upload OTLP-formatted file, I am getting an exception in the UI code:
image
in this function
```
function loadJsonDone(state, { payload }) {
  const processed = payload.data.map(transformTraceData);
```
The `payload` here is still OTLP-formatted, so it has no `.data` and `.map` throws. I don't fully understand how these "actions" work, but is it perhaps possible to ensure that your new transformer action is called _before_ the execution reaches into `loadJsonDone`?

yurishkuro:
And on a related question - is it possible to ensure that we have a unit test that catches this?

packages/jaeger-ui/src/middlewares/index.js
```diff
@@ -70,7 +70,7 @@ export const loadJsonTracesMiddleware = store => next => action => {
   } else {
     return next(action);
   }
-  return next(action); // Consistent returns
+  return undefined;
```
yurishkuro, line 73: why is this the right behavior?

yurishkuro: I assume calling the next action is a pattern that needs to be maintained, it looks like a dangerous shortcut when you change it just to work around another problem. The right fix is to control the order of these middlewares, so that OTLP translator is invoked before the rest of the handlers.

packages/jaeger-ui/src/utils/readJsonFile.tsx
```diff
@@ -21,7 +23,18 @@ export default function readJsonFile(fileList: { file: File }): Promise<string>
         return;
       }
       try {
-        resolve(JSON.parse(reader.result));
+        const traceObj = JSON.parse(reader.result);
+        if ('resourceSpans' in traceObj) {
+          JaegerAPI.transformOTLP(traceObj)
```
yurishkuro, line 28: you previously wrote that transformOTLP action will be called automatically anyway. Does it mean it will be called twice now?

packages/jaeger-ui/src/utils/readJsonFile.tsx
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import JaegerAPI from '../api/jaeger';
+
 export default function readJsonFile(fileList: { file: File }): Promise<string> {
   return new Promise((resolve, reject) => {
```
yurishkuro, line 18: This looks like a generic function that just loads and parses a JSON file. It doesn't have anything to do with specifically loading traces (could be loading a config file). I think your changes need to go in upload-related components which are calling this `readJsonFile` function, and once they receive the JSON and start processing it then the new logic to intercept OTLP should be placed.

yurishkuro (approved):
I tried a few other ways, but there's a complication with JaegerAPI returning a future while all other code paths are synchronous. Will merge as is.

yurishkuro:
thanks!

NavinShrinivas:
> thanks!

It was fun working on this issues. Learnt a lot :) , any other issue that i work on that beginner friendly?

yurishkuro:
https://github.com/search?q=org%3Ajaegertracing+label%3A%22good+first+issue%22&type=issues&state=open

PR #2168, opened 2024-02-16
Use JSON for table values only for objects

bobrik:
Without this numbers are treated as json, which styles them with an offset:

image

With this change applied:

image

It's possible that there's a better solution for this.

https://github.com/jaegertracing/jaeger-ui/pull/1724/files#r1491924866

Manually tested.

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.tsx
@@ -52,11 +52,27 @@ const stringListMarkup = (value: any[]) => (
   </div>
 );
 
-const stringMarkup = (value: string) => (
-  <div className="json-markup">
-    <span className="json-markup-string">{value}</span>
-  </div>
-);
+const scalarMarkup = (value: string | Number | Boolean) => {
+  let className;
+  switch (typeof value) {
+    case 'boolean': {
+      className = 'json-markup-bool';
+      break;
+    }
+    case 'number': {
+      className = 'json-markup-number';
+      break;
+    }
+    default: {
+      className = 'json-markup-string';
+    }
+  }
+  return (
+    <div className="json-markup">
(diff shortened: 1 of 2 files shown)


yurishkuro:
I actually like that it shows numbers differently

bobrik:
I can understand the improvement from the numbers having a different color, but a random offset is not something you see in any software showing tables. It just looks broken to me.

yurishkuro:
I think the way to fix the offset would be via changing the styling, not the business logic.

bobrik:
The styling we apply to the table cell is based on the `json-markup-child` class, which is logically wrong, as the table cell is not a child of any json, because there is no json there to begin with:

image

The `margin` style is applied to the actual json children to provide the tree structure, so if you drop it, you're left with this:

image

Without my change numbers are nested into `json-markup-number`, while strings aren't, because they have their own condition branch that rightfully bypasses treating them as json.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.tsx
```diff
@@ -95,6 +93,8 @@ function formatValue(key: string, value: any) {
         }}
       />
     );
+  } else {
+    content = stringMarkup(parsed);
   }
```
yurishkuro, line 98: I think we should continue checking type for number and Boolean and use the corresponding styles.  I would rename `stringMarkup` to `scalarMarkup` and do that check internally in order to switch between `json-markup-string`, `json-markup-number`, and `json-markup-bool`. This way the styling will be consistent with  the JSON rendering, but without extra indentation.

bobrik:
With the latest changes:

image

yurishkuro (approved):
Thanks!

yurishkuro:
looks like the existing test cases are missing a case for boolean value, can you add it?

bobrik:
Sure, added.

yurishkuro (approved):
🎉

PR #2182, opened 2024-02-24
Treat exception key in log as error

bobrik:

This is how OpenTelemetry distringuishes errors:

* https://opentelemetry.io/docs/specs/otel/trace/exceptions/

I also unified computation of the condition into `isErrorSpan`.

With this change:

image

image

Manually tested. Happy to add tests if this makes sense.

## Checklist


yurishkuro (commented):
I don't think this is semantically correct. The presence of an exception does not indicate that the span itself was in error, there is span.status field for that (which in Jaeger translation becomes error=true). I think it would be better to use a different state for this in Jaeger, maybe a different icon.

And I think another nice to have is to color the log ticks on the span bar differently for exceptions. The log entries in details could also use the same icon as above.

bobrik:
I think you're right, instead of doing this:

```go
span.RecordError(err)
```

I should be doing this:

```go
span.SetStatus(codes.Error, err.Error())
```

PR #2273, opened 2024-04-20
Added testcases

NitinRamnani:
## Description of the changes
- Added testcases

- Ran command `yarn test`
ticks test js

utils test js

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/Ticks.test.js
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/Ticks.test.js
@@ -22,4 +22,9 @@ describe('<Ticks>', () => {
     const wrapper = shallow(<Ticks endTime={200} numTicks={5} showLabels startTime={100} />);
     expect(wrapper).toBeDefined();
   });
+
+  it('should match snapshot', () => {
+    const wrapper = shallow(<Ticks endTime={200} numTicks={5} showLabels startTime={100} />);
+    expect(wrapper).toMatchSnapshot()
+  });
 });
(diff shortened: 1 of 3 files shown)


packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/Ticks.test.js
```diff
@@ -22,4 +22,9 @@ describe('<Ticks>', () => {
     const wrapper = shallow(<Ticks endTime={200} numTicks={5} showLabels startTime={100} />);
     expect(wrapper).toBeDefined();
   });
+
+  it('should match snapshot', () => {
+    const wrapper = shallow(<Ticks endTime={200} numTicks={5} showLabels startTime={100} />);
+    expect(wrapper).toMatchSnapshot()
```
yurishkuro, line 28: we are trying to move away from snapshot testing. What specifically are you trying to validate?

NitinRamnani: Trying to validate that if same props are passed, component should render same html. So in future if someone accidentally changes anything in the component, it will be caught by this testcase.

yurishkuro: > component should render same html

but the component is not obligated to render the same HTML. Its implementation can change and the test will fail, even though the implementation can still be functionally correct. Tests should be validating use-visible behavior, e.g. you asked for 5 ticks - the result should have 5 ticks.

NitinRamnani: You are correct but if we are changing implementation then we should also update the testcase. Snapshot testcase make sure until we haven't made any change HTML shouldn't also change there by help us in catching any accidentally changes.

Issue #2272, opened 2024-04-20
Help with eslint v9 upgrade


yurishkuro:
#2271 is failing to bump eslint versions:
  1. [x] We need to change our build to use Node 20 in `.github/workflow/**` and `.nvmrc`. It would be nice to [enforce it via package.json](https://stackoverflow.com/a/29349773/2673284).
  2. [x] This error: `ESLintIgnoreWarning: The ".eslintignore" file is no longer supported. Switch to using the "ignores" property in "eslint.config.js": https://eslint.org/docs/latest/use/configure/migration-guide#ignoring-files`
  3. [ ] New error: `[eslint       ] Error: Could not find config file.`

Baalekshan:
@yurishkuro unit-tests workflow is failing in my commit, how can i resolve

yurishkuro:
@Baalekshan the lint step is still failing with
```
[eslint       ] Error: Could not find config file.
[eslint       ]     at locateConfigFileToUse (/home/runner/work/jaeger-ui/jaeger-ui/node_modules/eslint/lib/eslint/eslint.js:350:21)
[eslint       ]     at async calculateConfigArray (/home/runner/work/jaeger-ui/jaeger-ui/node_modules/eslint/lib/eslint/eslint.js:385:49)
[eslint       ]     at async ESLint.lintFiles (/home/runner/work/jaeger-ui/jaeger-ui/node_modules/eslint/lib/eslint/eslint.js:815:25)
[eslint       ]     at async Object.execute (/home/runner/work/jaeger-ui/jaeger-ui/node_modules/eslint/lib/cli.js:500:23)
[eslint       ]     at async main (/home/runner/work/jaeger-ui/jaeger-ui/node_modules/eslint/bin/eslint.js:153:22)
[eslint       ] error Command failed with exit code 2.
```

Baalekshan:
@yurishkuro  Is this a concern ? Since, it was safely merged without any errors in the workflow. I didn't find this error in the PR raised only faced this in my forked repo.

yurishkuro:
@Baalekshan the change we merged was good, but it did not resolve this ticket, because the upgrade is still failing, just on a different error.

Baalekshan:
Oh right, what could be done now. Should I try fixing these errors? If so what can I do.

yurishkuro:
The upgrade is trying to bump eslint from 8.x to 9.x, there could be some breaking changes causing this issue.

[...]

PR #2274, opened 2024-04-20
Require Node version 20+

Baalekshan:
#2272 

## Description of the changes
- 

- 

## Checklist


.eslintrc.js
```diff
@@ -19,6 +19,15 @@ module.exports = {
     jest: true,
     jasmine: true,
   },
+  // Add any other files or directories you want to ignore
+  ignorePatterns: [
```
yurishkuro, line 23: we need to delete `.eslintignore` to verify that this section is equivalent

Baalekshan: @yurishkuro  .eslintignore is deleted.

yurishkuro (approved):
Thanks!

PR #2275, opened 2024-04-23
Improve trace page title with data and unique emoji (fixes #2256)

### LINKED ISSUE #2256: [Feature]: Make good use of page titles
### Requirement

As a Jaeger UI user, I want page titles to include the same data as trace names so that I can disambiguate browser tabs instead of having a dozen of them named "Jaeger UI".

### Problem

Currently all Jaeger UI pages are titled "Jaeger UI", which is not descriptive and doesn't help me when I have a dozen tabs open trying to debug an issue.

### Proposal

Make use of `<Helmet title="…" />` in `TracePageHeader` to make the page title become:

```
<short trace id>: <operation name> | <service name> — Jaeger UI
```

### Open questions

What the exact format of the page title should be? I think it should go from most unique bit to less unique bit, so it should start with the short trace id and end with the service name.

nox:
- This solves #2256.

## Description of the changes
- Trace page titles are changed to `<unique emoji> <short trace id>: <operation name> (<service name>) — Jaeger UI`.

- Page titles are checked in tests.

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageHeader.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageHeader.tsx
@@ -159,7 +159,7 @@ export function TracePageHeaderFn(props: TracePageHeaderEmbedProps & { forwarded
 
   return (
     <header className="TracePageHeader">
-      <Helmet title={`${traceShortID}: ${trace.tracePageTitle} — Jaeger UI`} />
+      <Helmet title={`${trace.traceEmoji} ${traceShortID}: ${trace.tracePageTitle} — Jaeger UI`} />
       <div className="TracePageHeader--titleRow">
         {toSearch && (
           <Link className="TracePageHeader--back" to={toSearch}>
--- a/packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageHeader.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageHeader.tsx
@@ -28,7 +28,6 @@ import { TUpdateViewRangeTimeFunction, IViewRange, ViewRangeTimeUpdate, ETraceVi
 import LabeledList from '../../common/LabeledList';
 import NewWindowIcon from '../../common/NewWindowIcon';
 import TraceName from '../../common/TraceName';
-import { getTraceName } from '../../../model/trace-viewer';
 import { TNil } from '../../../types';
 import { Trace } from '../../../types/trace';
 import { formatDatetime, formatDuration } from '../../../utils/date';
@@ -151,7 +150,7 @@ export function TracePageHeaderFn(props: TracePageHeaderEmbedProps & { forwarded
 
   const title = (
     <h1 className={`TracePageHeader--title ${canCollapse ? 'is-collapsible' : ''}`}>
-      <TraceName traceName={getTraceName(trace.spans)} />{' '}
+      <TraceName traceName={trace.traceName} />{' '}
       <small className="u-tx-muted">{trace.traceID.slice(0, 7)}</small>
     </h1>
   );
--- a/packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageHeader.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageHeader.tsx
@@ -20,6 +20,7 @@ import _values from 'lodash/values';
 import { IoArrowBack, IoFileTrayFull, IoChevronForward } from 'react-icons/io5';
 import { Link } from 'react-router-dom';
 
+import { Helmet } from 'react-helmet';
 import AltViewOptions from './AltViewOptions';
 import KeyboardShortcutsHelp from './KeyboardShortcutsHelp';
 import SpanGraph from './SpanGraph';
@@ -148,15 +149,17 @@ export function TracePageHeaderFn(props: TracePageHeaderEmbedProps & { forwarded
       return { ...rest, value: renderer(trace) };
     });
 
+  const traceShortID = trace.traceID.slice(0, 7);
+
   const title = (
     <h1 className={`TracePageHeader--title ${canCollapse ? 'is-collapsible' : ''}`}>
-      <TraceName traceName={trace.traceName} />{' '}
-      <small className="u-tx-muted">{trace.traceID.slice(0, 7)}</small>
+      <TraceName traceName={trace.traceName} /> <small className="u-tx-muted">{traceShortID}</small>
     </h1>
   );
 
   return (
     <header className="TracePageHeader">
+      <Helmet title={`${traceShortID}: ${trace.tracePageTitle} — Jaeger UI`} />
       <div className="TracePageHeader--titleRow">
         {toSearch && (
           <Link className="TracePageHeader--back" to={toSearch}>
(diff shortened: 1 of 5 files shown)


packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageHeader.tsx
```diff
@@ -149,15 +150,18 @@ export function TracePageHeaderFn(props: TracePageHeaderEmbedProps & { forwarded
...
     <h1 className={`TracePageHeader--title ${canCollapse ? 'is-collapsible' : ''}`}>
-      <TraceName traceName={getTraceName(trace.spans)} />{' '}
-      <small className="u-tx-muted">{trace.traceID.slice(0, 7)}</small>
+      <TraceName traceName={trace.traceName} /> <small className="u-tx-muted">{traceShortID}</small>
     </h1>
   );
 
   return (
     <header className="TracePageHeader">
+      <Favicon url={trace.traceEmojiFaviconURL} />
```
yurishkuro, line 163: my understanding was that emoji will be the first char in the title, not replacing the whole Jaeger favicon. I don't like the latter - I think it breaks the implied contract between the website and the browser. 

Some websites alter the favicon (e.g. gmail shows actual msg count, others just add a red dot indicating unread messages).
image

 I wonder if this would work here, i.e. keep the overall favicon background as Jaeger logo but reserve a corner of it with a flat background and an emoji on top.

nox: The emoji is not legible if I do that. Could we make it a runtime setting whether we want to replace the favicon? My colleagues like the idea a lot and the favicon just wastes tab bar estate bringing no additional value for us. In any case, I made the favicon thing be its own commit at the very end of the list of commits, so we can easily apply the rest.

yurishkuro: yeah, we can add it to config. Meanwhile, I'd still suggest including emoji in the title (regardless of where it replaces favicon or not)

nox: Ack. I removed the last commit putting the emoji in a favicon. Will open another PR for configurable favicons.

packages/jaeger-ui/src/model/trace-viewer.tsx
```diff
@@ -44,10 +49,59 @@ export function _getTraceNameImpl(spans: Span[]) {
...
+
+  // prettier-ignore
+  const emojiSet = [
+    '🐶', '🐱', '🐭', '🦊', '🐨', '🐮', '🐷', '🐸', '🐵', '🐔', '🐤', '🦆',
+    '🦉', '🐝', '🦋', '🐢', '🦀', '🐳', '🐊', '🦒', '🪶', '🦩', '🐉', '🍄',
+    '🌸', '🌜', '⭐️', '🔥', '🌪️', '💧', '🍏', '🍊', '🍉', '🍒', '🥦', '🌽',
+    '🍠', '🥐', '🥖', '🥚', '🧀', '🍗', '🍟', '🍕', '🍣', '🍤', '🍙', '🍪',
+    '⚽️', '🏀', '🥎', '🎹', '🎲', '🎮', '🧩', '🚗', '🚲', '🛞', '🚂', '⛺️',
+    '💿', '📞', '⏰', '💡', '🔌', '💎', '🪚', '🧲', '💣', '🧬', '🔑', '🎀',
+    '📬', '🗑️', '📘', '📌', '🔍', '🔒', '🩷', '❗️', '🎵', '🔔', '🏴', '🚩',
```
yurishkuro, line 95: some of these are quite similar to common UX elements, I would avoid them, e.g. '⭐️', '❗️', '🗑️', '📌', '🔍', '🔒', '💡'

packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageHeader.tsx
```diff
@@ -149,15 +149,17 @@ export function TracePageHeaderFn(props: TracePageHeaderEmbedProps & { forwarded
       return { ...rest, value: renderer(trace) };
     });
 
+  const traceShortID = trace.traceID.slice(0, 7);
+
   const title = (
     <h1 className={`TracePageHeader--title ${canCollapse ? 'is-collapsible' : ''}`}>
-      <TraceName traceName={getTraceName(trace.spans)} />{' '}
-      <small className="u-tx-muted">{trace.traceID.slice(0, 7)}</small>
+      <TraceName traceName={trace.traceName} /> <small className="u-tx-muted">{traceShortID}</small>
```
yurishkuro, line 156: please keep `{' '}`

nox: Sure, I think `yarn prettier` is what removed it though.

packages/jaeger-ui/src/model/trace-viewer.tsx
nox: Sure.

packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageHeader.tsx
nox: I removed it, but `yarn prettier` put it back. Is it not only relevant when it's a trailing space at the end of line?

yurishkuro: ok do what prettier says otherwise the CI won't pass

nox: Done.

yurishkuro (approved):

packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageHeader.tsx
```diff
@@ -149,15 +149,17 @@ export function TracePageHeaderFn(props: TracePageHeaderEmbedProps & { forwarded
...
     </h1>
   );
 
   return (
     <header className="TracePageHeader">
+      <Helmet title={`${trace.traceEmoji} ${traceShortID}: ${trace.tracePageTitle} — Jaeger UI`} />
```
SOF3, line 162: Any thoughts on optionally including the timestamp? That would be more intuitive than the trace ID.

yurishkuro:
@nox how do you actually use this new feature? I have several traces open in Chrome tabs. They all have our standard favicon, on the trace page the new emojis are not showing anywhere. The only place where I see them is if I expand the tabs list dropdown (something that I almost never do as a navigation mechanism):
image

bobrik:
There's a bit of a problem: you need to open the tab for the emoji to appear (first two tabs were not opened).

Other than that this is how it looks like in Chrome for me:

image

You need to have a reasonable number of tabs open for the emojis to be visible, too many tabs and they disappear.

yurishkuro:
Yes, the tabs need to be wider to see the custom title. I had too many tabs open.

I still think the favicon should remain as Jaeger, for branding. But it would be good to include the emoji into other places showing trace, for consistency, such as search page, and the main trace view (which will be shared between different trace visualizations).
image
 

Another glitch I noticed: when I open a new tab with a trace, it first opens with the standard "Jaeger UI" title. If I wait for page to load completely, it then changes to emoji/uuid title, but if I switch tab away before page loads, it stays as "Jaeger UI" until I switch back to it (then it changes). Don't know if this behavior is avoidable given how React router works.

PR #2369, opened 2024-07-04
Color log markers by user-specified log fields

### LINKED ISSUE #2361: [Feature]: Colormap log markers by tag
### Requirement

As a Jaeger user workign with large traces with many logs on each span, I want to be able to distinguish various types of logs more easily.

### Problem

When I see a large span like this, it is difficult to figure out what is going on at a glance
image

### Proposal

Add a button "Log colormap" in the trace detail page, probably next to the "Find" menu.

The button opens a dialog in which the user can select a "key", either a dropdown based on all distinct values from `.spans[].logs[].key`or a manual input dialog.

When this dialog is closed, all log markers

### Open questions

_No response_

SOF3:
- 
- Resolves #2361

## Description of the changes
- Add a settings dialog to TracePage
- Add an option to select log marker color keys
- Change log marker colors from black to colorgenerator-based (same color generator as span bars, but with only 2/3 brightness to avoid producing identical colors)

image

- 

(This is my first time working with a reactjs project, would appreciate pointers to what tests I need to write)

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageHeader.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageHeader.tsx
@@ -23,6 +23,7 @@ import { Link } from 'react-router-dom';
 import { Helmet } from 'react-helmet';
 import AltViewOptions from './AltViewOptions';
 import KeyboardShortcutsHelp from './KeyboardShortcutsHelp';
+import TracePageSettings from './TracePageSettings';
 import SpanGraph from './SpanGraph';
 import TracePageSearchBar from './TracePageSearchBar';
 import { TUpdateViewRangeTimeFunction, IViewRange, ViewRangeTimeUpdate, ETraceViewType } from '../types';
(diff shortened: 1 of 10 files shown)


yurishkuro:
what is the meaning of multiple colors on the same tick?

image

packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageHeader.tsx
```diff
@@ -58,6 +59,9 @@ type TracePageHeaderEmbedProps = {
   showViewOptions: boolean;
   slimView: boolean;
   textFilter: string | TNil;
+  availableTagKeys: string[];
+  selectedMarkerColorKey: string[];
+  setMarkerColorKey: (tagKey: string[]) => void;
```
yurishkuro, line 64: These don't seem relevant in any way to the "page header" type. Is it possible to use the Context API to keep the knowledge of the settings only to the components that actually need this knowledge, such as the Settings popup and the SpanBar that renders the ticks?

PR #2417, opened 2024-07-19
Draft: Upgrading from raven-js to sentry/browser

### LINKED ISSUE #2412: Help upgrading from raven-js
We are using a deprecated https://www.npmjs.com/package/raven-js package in [packages/jaeger-ui/src/utils/tracking/](./packages/jaeger-ui/src/utils/tracking/). Maybe we can use `@sentry/browser` for that, or maybe we can just implement the needed functionality internally since we use it only to capture some basic error information to submit to the tracking backend.

BenzeneAlcohol:
Resolves #2412 

## Description of the changes
I've migrated a part of the things, but I want to know whether the approach is correct or it needs to be different. Hence a draft MR. The names of files, variables etc haven't been changed, this is just in progress. 

Test cases for now, the old ones with the same payload. 

## Checklist


packages/jaeger-ui/src/utils/tracking/conv-raven-to-ga.tsx
```diff
@@ -36,6 +37,16 @@ const FETCH_SYMBOLS = [
...
 ];
 
+interface SentryTransportOptions {
+  url: string;
+  data: any;
+  auth: {
+    sentry_version: string;
+    sentry_client: string;
+    sentry_key: string;
+  };
```
yurishkuro, line 47: why is this relevant to Jaeger? We're not submitting anything to Sentry.

BenzeneAlcohol: It was there in `RavenTransportOptions`, so I just had it here as well. Will remove.

package.json
```diff
@@ -67,5 +67,8 @@
     "hooks": {
       "pre-commit": "npm-run-all -ln --parallel lint test"
     }
+  },
+  "dependencies": {
+    "@sentry/browser": "^8.18.0"
```
anshgoyalevil, line 72: That `@sentry/browser` dep should be installed inside the `jaeger-ui` directory instead of the root workspace.

BenzeneAlcohol: Didn't get you?

BenzeneAlcohol: Oh nvm, got it

BenzeneAlcohol:
If you see `packages/jaeger-ui/src/utils/tracking/ga.tsx`, autoBreadcrumbs there are configured to accept `location` as well to be recorded for eveyr event. However , from sentry's docs for breadcrumbs, it seems like the location param is absent. What do we do we here?

https://docs.sentry.io/platforms/javascript/configuration/integrations/breadcrumbs/

BenzeneAlcohol:
The PR is almost over, I need to change file names and function names. For now the existing unit tests pass, but any other ways to test this properly? @yurishkuro

yurishkuro:
You can set up your own property in GA and test that Jaeger UI generates the desired events, as well as compare them with events generated by the UI before your change. You will probably need to induce some exceptions, e.g. by using Upload function in Search tab and uploading a malformed text.

BenzeneAlcohol:
> You can set up your own property in GA and test that Jaeger UI generates the desired events, as well as compare them with events generated by the UI before your change. You will probably need to induce some exceptions, e.g. by using Upload function in Search tab and uploading a malformed text.

Alright, I'll try that. Meanwhile, can you review the code and see if there are any obvious mistakes?

packages/jaeger-ui/package.json
```diff
@@ -50,6 +50,7 @@
     "@ant-design/compatible": "^5.1.3",
     "@jaegertracing/plexus": "0.2.0",
     "@pyroscope/flamegraph": "0.21.4",
+    "@sentry/browser": "^8.18.0",
```
yurishkuro, line 53: should there be a corresponding change to `yarn.lock`?

packages/jaeger-ui/src/types/tracking.tsx
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { RavenStatic } from 'raven-js';
+import * as Sentry from '@sentry/browser';
```
yurishkuro, line 15: can we import just what's needed, not `*`?

packages/jaeger-ui/src/utils/tracking/conv-raven-to-ga.tsx
```diff
@@ -169,9 +160,7 @@ function convNav(to: string) {
 //    "dp" - fetch the dependency data
 // And, "NNN" is a non-200 status code.
 type TCrumbData = {
```
yurishkuro, line 162: can you explain what "breadcrumb" means in this context? I saw a link you posted to sentry docs, but they didn't explain it either, just started talking details.

packages/jaeger-ui/src/utils/tracking/ga.tsx
```diff
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 import _get from 'lodash/get';
-import Raven, { RavenOptions, RavenTransportOptions } from 'raven-js';
+import * as Sentry from '@sentry/browser';
```
yurishkuro, line 16: can we not use `*`?

packages/jaeger-ui/src/utils/tracking/ga.tsx
```diff
@@ -164,24 +171,31 @@ const GA: IWebAnalyticsFunc = (config: Config, versionShort: string, versionLong
...
     }
     if (isErrorsEnabled) {
-      const ravenConfig: RavenOptions = {
-        autoBreadcrumbs: {
-          xhr: true,
-          console: false,
-          dom: true,
-          location: true,
+      Sentry.init({
+        dsn: 'https://fakedsn@omg.com/1',
```
yurishkuro, line 175: ?

BenzeneAlcohol: Same DSN used in the old raven config, so I kept it that way

packages/jaeger-ui/src/utils/tracking/conv-raven-to-ga.tsx
BenzeneAlcohol: Will explain this tomorrow, late night here.

BenzeneAlcohol: breadcrumb in this context is basically any event that is captured. Everything is referred to as a breadcrumb. It basically lets us capture everything leading up to that event/error.

packages/jaeger-ui/package.json
BenzeneAlcohol: Yes. I will commit that file as well.

packages/jaeger-ui/src/types/tracking.tsx
```diff
@@ -22,7 +22,7 @@ export interface IWebAnalyticsFunc {
 
 export default interface IWebAnalytics {
   init: () => void;
-  context: boolean | RavenStatic | null;
+  context: boolean | typeof BrowserClient | null;
```
yurishkuro, line 25: This field is rather confusing to me. Is it a readable property for the user of IWebAnalytics interface? Or is it an input to the implementation of the interface? If it's the latter, why does it need to be exposed in the interface?

Not specifically related to the PR, but I am trying to understand the structure of the code.

packages/jaeger-ui/src/utils/tracking/fixtures.js
```diff
@@ -212,7 +212,7 @@ tr
 ! Type! A very long message that will be truncated and re~
 cic2i2i{.LabeledList.TracePageHeader--overviewItems}`;
 
-export const RAVEN_TO_GA = deepFreeze({
+export const SENRTY_TO_GA = deepFreeze({
```
yurishkuro, line 215: typo
```suggestion
export const SENTRY_TO_GA = deepFreeze({
```

packages/jaeger-ui/src/utils/tracking/conv-sentry-to-ga.tsx
```diff
@@ -247,20 +236,14 @@ function compressCssSelector(selector: string) {
 // The chronological ordering of the breadcrumbs is older events precede newer
 // events. This ordering was kept because it's easier to see which page events
 // occurred on.
-interface ICrumb {
```
yurishkuro, line 250: it seems like we used an abstraction over Raven type (even though abstraction itself is a bit coupled, but an abstraction nonetheless, i.e. could be implemented by something other than Raven). What's the reason to remove it instead of just adapting? Adapting seems to only require optionality on the fields, which would minimize the changes throughout this file.

yurishkuro (commented):
will you be able to test against Google Analytics?

BenzeneAlcohol:
> will you be able to test against Google Analytics?

Could be better if someone who already has an idea about GA does it. Can you do it? Else I can go through the docs and setup, but will need one or two days.

packages/jaeger-ui/src/types/tracking.tsx
BenzeneAlcohol: So here, `IWebAnalytics` is an interface of the object that is returned by the function `IWebAnalyticsFunc` that takes in 3 variables as the input.

This is used in `ga.tsx` as function `GA`. There, if you notice, context is assigned to `BrowserClient`, and not an instance of it. That is why I have kept context as either a boolean, null, or a type of BrowserClient object, and not an instance of it.

Even I am a little confused as to the places it being used.

BenzeneAlcohol: It is set depending on whether error tracking is enabled on not. It needs to be there on the interface because depending on the config, this needs to be set. 

It is used in L56-57 in `packages/jaeger-ui/src/utils/tracking/index.tsx`

packages/jaeger-ui/src/utils/tracking/conv-sentry-to-ga.tsx
BenzeneAlcohol: This does make sense. But why I thought of using the `Breadcrumb` interface from Sentry was because this function `convSentryToGa` is called from `trackSentryError` in L192, `packages/jaeger-ui/src/utils/tracking/ga.tsx`. 

So there, the transport options come from the `Event` interface in Sentry, which has a breadcrumb field which uses the `Breadcrumb` interface. I just thought of keeping it uniform everywhere and since that field already uses the same interface, used it here as well.

yurishkuro:
> Could be better if someone who already has an idea about GA does it. Can you do it? Else I can go through the docs and setup, but will need one or two days.

I don't have any of that set up either. At minimum you can enable debug mode and see what the tracking logs into console and compare it with the before version.

BenzeneAlcohol:
i'll do.

BenzeneAlcohol:
Hey, I got caught up with a bit of work stuff. I'll test this PR tomorrow and day after, and hopefully by this weekend we can get this done. @yurishkuro

BenzeneAlcohol:
I tried, but couldn't really find a way to troubleshoot using just the console.log statements. I went through a bunch of tests that use this config in a lot of places, so since they have passed, I am 90% sure this works. It is just the other 10% I'm a little concerned about.

yurishkuro:
Is it possible to test with a real GA account / property? There's also a [Chrome extension][1] that claims to help with testing.

[1]: https://chromewebstore.google.com/detail/analyticsgtmpixel-debugge/canpneabbfipaelecfibpmmjbdkiaolf

BenzeneAlcohol:
> Is it possible to test with a real GA account / property? There's also a [Chrome extension](https://chromewebstore.google.com/detail/analyticsgtmpixel-debugge/canpneabbfipaelecfibpmmjbdkiaolf) that claims to help with testing.

Will checkt that out

BenzeneAlcohol:
Hi, I'm sorry I was out of station for the past week. Is there anything I have to do now, seeing you have pushed a few commits?

yurishkuro:
Thanks @BenzeneAlcohol - this was merged as part of #2509

BenzeneAlcohol:
Great to know! :)

PR #2421, opened 2024-08-29
Horizontal Scroll Bar Span Graph Timeline

### LINKED ISSUE #2419: [Feature]: Add Horizontal Scrollbar for Viewing Long Traces in Jaeger UI
### Requirement

In our business, we frequently need to analyze and debug complex distributed systems that involve dozens or even hundreds of microservices. When we query these systems in Jaeger, some traces can become very long. The current Jaeger UI does not handle the display of long traces well, requiring users to resize the browser window or zoom out to see the entire trace.

We would like to improve the user experience by adding a horizontal scrollbar to the trace view. This enhancement would significantly improve the efficiency of our developers and operations teams when analyzing long traces. It would also help us identify and resolve issues in our systems more quickly, ultimately leading to greater system stability and higher user satisfaction.

### Problem

**Description:**
When working with long traces in the Jaeger UI, the current design requires users to either extend the browser window width or zoom out to view the full trace. This is particularly challenging when the trace is very long, as users may find it cumbersome to navigate the trace timeline. As shown in the attached screenshot, the trace details are cut off and only visible by resizing the browser or zooming out significantly.

image

**Steps to Reproduce:**

Query a long trace in Jaeger that spans beyond the typical width of a browser window.
Observe that the full trace is not visible without resizing the window or zooming out.
Notice that the UI lacks a horizontal scrollbar, which would allow easy navigation through the trace.

**Environment:**

Jaeger version: 1.56
Browser: chrome
Operating System: macos

### Proposal

**Expected Behavior:**

The Jaeger UI should provide a horizontal scrollbar when the trace extends beyond the viewport width.
Users should be able to scroll horizontally to view the entire trace timeline without needing to adjust the window size or zoom level.

**Current Workaround:**

Users are currently required to either zoom out in the browser or increase the width of the browser window to view the entire trace. This workaround is not user-friendly, especially for very long traces.

### Open questions

_No response_

CoryBarney:
-  Resolves #2419

## Description of the changes
- Adding horizontal scrollbar to gain a clickable interface to scrub though the trace timeline
- Adding tests for said above addition

- Manually and with added Jest

https://github.com/user-attachments/assets/e4ca58a4-4337-45f8-830e-da5221b613af

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TracePageHeader/SpanGraph/ViewingLayer.css
+++ b/packages/jaeger-ui/src/components/TracePage/TracePageHeader/SpanGraph/ViewingLayer.css
@@ -52,6 +52,15 @@ limitations under the License.
   fill: #f44;
 }
 
+/* Styles for greying out the area outside the marker points */
+.ViewingLayer--area {
+  fill: #fff;
+}
+
+.ViewingLayer--area.isActive {
+  fill: rgba(214, 214, 214, 0.5);
+}
+
 .ViewingLayer--fullOverlay {
   bottom: 0;
   cursor: col-resize;
@@ -72,4 +81,31 @@ limitations under the License.
 
 .ViewingLayer:hover > .ViewingLayer--resetZoom {
   display: unset;
+}
+
+/* Scroll Bar Styles */
+.ViewingLayer--scrollBar {
+  position: relative;
+  overflow-x: hidden;
+  overflow-y: hidden;
+  height: 20px;
+  visibility: hidden;
+}
+
+.ViewingLayer--scrollBar:hover {
+  visibility: visible;
+  overflow-x: auto;
+}
+
+.ViewingLayer--scrollThumb {
+  position: absolute;
+  height: 100%;
+  background-color: rgba(0, 0, 0, 0.2);
+  border-radius: 4px;
+  cursor: pointer;
+  transition: background-color 0.2s;
+}
+
+.ViewingLayer--scrollThumb:hover {
+  background-color: rgba(0, 0, 0, 0.4);
 }
--- a/packages/jaeger-ui/src/components/TracePage/TracePageHeader/SpanGraph/ViewingLayer.test.js
+++ b/packages/jaeger-ui/src/components/TracePage/TracePageHeader/SpanGraph/ViewingLayer.test.js
@@ -29,7 +29,7 @@ function getViewRange(viewStart, viewEnd) {
   };
 }
 
-describe('<SpanGraph>', () => {
+describe('<ViewingLayer>', () => {
   polyfillAnimationFrame(window);
 
   let props;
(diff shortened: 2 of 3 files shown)


CoryBarney:
Got better direction on requirements making some changes

yurishkuro:
maybe a small bug - when there is a selection, the Hand icon for dragging only shows in the top part of the selection (correctly), but dragging also works if started outside of that selection in the top part (where hand icon is not shown). I think it's better to limit dragging to work as moving only within the selection.

CoryBarney:
Will do, thanks for pointing that out!

CoryBarney:
I think this is more in line with what your goal was, as well I noticed that the vertical red bar is present on the any place you can start dragging to create anchor points to I put that in. As well I synced up the cursor icon between the ViewingLayer and the TimelineViewingLayer so that they are the crosshair but I can swap that out to be whatever icon if you think the cross is no good.

Let me know and ill probably cut this into it's own PR as it kind of when on a tangent and don't want to overload it

https://github.com/user-attachments/assets/181e3c5e-ad57-4273-9312-fbd817ad1f99

yurishkuro:
still seeing this bug https://github.com/jaegertracing/jaeger-ui/pull/2421#issuecomment-2319111167

CoryBarney:
https://github.com/user-attachments/assets/dde3ea22-3e29-402b-86a9-4a4660c417f0

I tried to replicate the functionality as close as possible to the google inspector and it has it global kind of like an invisible scroll bar. I can limit its functionality to just the physical range but I feel like being able to move and use your scroll wheel is a better UX.

yurishkuro:
ok, it's fine to do like Chrome does, but I see that in Chrome the drag indeed still works both over and outside of the current selection, but it also shows the hand icon in both cases (in the top part).

CoryBarney:
Currently it is setup to:

- Show the hand icon open in the to 40%
- The crosshair for making anchors in the bottom 60%
- the red line which might be better to replace the crosshair with to be more in line
- When clicked in the top 40% of the timeline scrubbing begins:
   - the icon is changed to a grab icon regardless of where it is in the web page until you release you click
   - Scrubbing moves left and right with the mouse curser regardless of where you are in the window until you release the click
 
 Should I get a better recorder as it looks fine to me in first person without the weird none precision clicks that get shown in the recording? I'm grasping at straws what the bug is as the open hand icon is just present within the 40% and dragging closed hand is present to show you are still clicking regardless of where in the screen. I can try to cut it off for the upper section but I think that is a regression.
 
 Am I misunderstanding something or having a difference in opinion (I do see the moment the crosshair comes up and will try to get the pixel sorted out)

[08-31-2024_jaeger_pr_review.webm](https://github.com/user-attachments/assets/9513979e-a8b1-42a6-9e38-a5bb861ab32e)

yurishkuro:
so what I meant was:
  * when you have a selection and hover it, you have crosshair in the lower part, and hand in the upper part
  * but if you move the cursor horizontally outside of the selection, it turns into an arrow, even though the dragging functionality is the same as within the selection (top part: pan, bottom part: new selection)

It would be great to have the cursor to be consistently crosshair / hand when there is a selection (and just crosshair when there is no current selection).

CoryBarney:
i can swap on property on the css and swap what ever area to what needs be. I put everything back to default cursor and have the to range in the middle of the anchors a grab icon cause we need to let the users know that there is that functionality but I can remove it.


https://github.com/user-attachments/assets/7a66fa3f-25d7-4a6d-89b1-c89fa8b52b78

yurishkuro:
Maybe I am not explaining correctly. I am concerned that these two positions have different icons:

image
image

Since in the 2nd one the selection can still be panned by dragging, then for consistency it should be showing hand icon as well (when actually dragging it should change to fist in both cases).

In contrast, below I didn't expect a hand, because you are still doing selection (pointer cursor), not panning (hand/fist cursor)
image

Srishti-j18:
I would like to work on this PR.

Issue #2427, opened 2024-09-09
[Bug]: CriticalPath functions fail if root span has a `FOLLOWS_FROM` reference in another trace


swythan:
### What happened?

I'm unable to view traces originating from Airflow (coming in via OTLP to an opentelemetry-collector and the Jaeger exporter) because the root activity has a `FOLLOWS_FROM` link to a span in another trace. 

Both `getChildOfSpans` and `computeCriticalPath` have trouble with this situation. The former prevents the page from loading. The latter seems to just slow it down a lot and it loads eventually.

I believe the critical bit is:
```json
{
    "data": [
        {
            "traceID": "19845b251f80e4b7331f4d446b0a557b",
            "spans": [
                {
                    "traceID": "19845b251f80e4b7331f4d446b0a557b",
                    "spanID": "331f4d446b0a557b",
                    "operationName": "test_delta_instrumentation",
                    "references": [
                        {
                            "refType": "FOLLOWS_FROM",
                            "traceID": "a5f0b2f965afac53d7b68670c59b163a",
                            "spanID": "ff300a3a5e0b44db"
                        }
                    ],
```

N.B. I appreciate we're running an old version of the Jaeger UI (an upgrade is in progress), but these methods seem to still be essentially the same in the repo.

### Steps to reproduce

1. Use the "Upload" function on the Jaeger UI to load the the attached file: 
[simplified-follows_from-trace.json](https://github.com/user-attachments/files/16931339/simplified-follows_from-trace.json)
2. Click through to the trace page
3. Note that there is an error in the Javascript Console and the page never loads

### Expected behavior

These functions should ignore `FOLLOWS_FROM` spans in other traces and the page should load and display properly.

### Relevant log output

```shell
From the browser Javascript console:

TypeError: Cannot read properties of undefined (reading 'childSpanIds')
    at index-6410cb48.js:815:40808
    at Map.forEach (<anonymous>)
    at getChildOfSpans (index-6410cb48.js:815:40642)
    at criticalPathForTrace (index-6410cb48.js:815:42909)
    at et (index-6410cb48.js:228:19520)
    at TracePageImpl.render (index-6410cb48.js:815:48823)
    at kj (index-6410cb48.js:40:6570)
    at ij (index-6410cb48.js:40:6348)
    at Wk (index-6410cb48.js:40:47291)
    at Vk (index-6410cb48.js:40:41925)
```

### Jaeger backend version

[...]

PR #2428, opened 2024-09-09
Workaround for root spans with alien `FOLLOWS_FROM` references

swythan:
- Workaround/solution for #2427

## Description of the changes

These were the changes I had to make locally (in the Chrome debug tools) to prevent the page from crashing when the root span had a `FOLLOWS_FROM` reference from a different trace.

Only by making these changes directly in the minified sources in Chrome/Edge's "Local Overrides". It stopped things failing, but I'm not sure whether the actual critical path algorithm still works afterwards.

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/CriticalPath/index.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/CriticalPath/index.tsx
@@ -46,38 +46,40 @@ const computeCriticalPath = (
 ): criticalPathSection[] => {
   const currentSpan: Span = spanMap.get(spanId)!;
 
-  const lastFinishingChildSpan = findLastFinishingChildSpan(spanMap, currentSpan, returningChildStartTime);
-  let spanCriticalSection: criticalPathSection;
+  if (currentSpan) {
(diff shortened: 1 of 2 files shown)


packages/jaeger-ui/src/components/TracePage/CriticalPath/index.tsx
```diff
@@ -46,38 +46,40 @@ const computeCriticalPath = (
 ): criticalPathSection[] => {
   const currentSpan: Span = spanMap.get(spanId)!;
 
-  const lastFinishingChildSpan = findLastFinishingChildSpan(spanMap, currentSpan, returningChildStartTime);
-  let spanCriticalSection: criticalPathSection;
+  if (currentSpan) {
```
yurishkuro, line 49: I would like to investigate upstream of this call why this could happen, but just a nit on the code:
```suggestion
  if (!currentSpan) {
    return criticalPath;
  }
```

swythan: That's what comes of re-typing the fix from the minified code in the browser into the real source code. 

Honestly: I'd advise you do check out what is going on here properly. I wanted to provide what I'd done to work around the problem rather than just filing a bug report and walking off. I didn't have the tools available to really work through it properly, though.

PR #2433, opened 2024-09-22
refactor: update LoadingIndicator component

Ghaby-X:
This PR aims to ensure compatibility of the LoadingIndicator component with future React releases. The LoadingIndicator component was using the older defaultProps syntax, which will be removed in future React versions. 

## Description of the changes
The following changes have been made to the LoadingIndicator component:

- Removed the defaultProps object from the LoadingIndicator component.
- Updated LoadingIndicator component to use default parameter values instead.
- Updated instances where components were called, to match rendered snapshots

- Default tests provided for the LoadingIndicator component was used to ensure it renders exactly as before the changes were made

## Checklist


packages/jaeger-ui/src/components/DeepDependencies/SidePanel/DetailsPanel.tsx
```diff
@@ -159,7 +159,7 @@ export class UnconnectedDetailsPanel extends React.PureComponent<TProps, TState>
         )}
         {this.state.detailsLoading && (
           <div className="Ddg--DetailsPanel--LoadingWrapper">
-            <LoadingIndicator className="Ddg--DetailsPanel--LoadingIndicator" />
+            <LoadingIndicator centered={false} className="Ddg--DetailsPanel--LoadingIndicator" small={false} />
```
yurishkuro, line 162: false values are defaults, why do we need to set them explicitly?

Ghaby-X: False values are indeed default. However, it seems there was a mismatch between the tests snapshot and the actual rendered component. 
The snapshots explicitly included the false values even though it could be ignored, so I wrote the components to match them.

yurishkuro: I would rather update the snapshots

PR #2435, opened 2024-09-23
POC: migrate to yarn 3

pavolloffay:
See https://yarnpkg.com/migration/guide 

```
yarn set version 3.8.5
```

## Which problem is this PR solving?
- 

## Description of the changes
- 

## How was this change tested?
- 

## Checklist
- [ ] I have read https://github.com/jaegertracing/jaeger/blob/master/CONTRIBUTING_GUIDELINES.md
- [ ] I have signed all commits
- [ ] I have added unit tests for the new functionality
- [ ] I have run lint and test steps successfully
  - for `jaeger`: `make lint test`
  - for `jaeger-ui`: `yarn lint` and `yarn test`

PR #2459, opened 2024-09-26
Upgrade yarn to v3.8.5

andreasgerstmayr:
- Upgrade from yarn classic to yarn v3.8.5 (ref #2434 and #2435)

## Description of the changes
- Upgrade from yarn classic to yarn v3.8.5

<del>The `lint` step is broken at the moment, because TypeScript incorrectly resolves the typings for react-router v5 with the types from react-router v6 (installed as a dependency of `react-router-dom-v5-compat`) instead of `@types/react-router-dom`. I'm not sure why this works correctly with yarn v1 though. See https://github.com/remix-run/react-router/discussions/9892</del>

- jaeger all-in-one & `yarn start`

## Checklist


.gitignore
```diff
@@ -22,6 +22,16 @@ yarn-error.log
...
 
+# yarn
+# https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored
+.pnp.*
+.yarn/*
+!.yarn/patches
+!.yarn/plugins
+!.yarn/releases
+!.yarn/sdks
+!.yarn/versions
```
yurishkuro, line 33: Why would we not ignore these?

andreasgerstmayr: I copied this list straight from the yarn docs here: https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored
There's an explanation in that FAQ entry for each path.

andreasgerstmayr: Regarding checking in yarn sources, afaics we have two options when using yarn v2+:
* use the experimental [corepack](https://yarnpkg.com/corepack) feature
* or commit the yarn sources

see https://github.com/yarnpkg/berry/issues/5748 for more details

andreasgerstmayr:
Hm, somehow the s390x build is stuck because of the upgrade to yarn v3. I'll see with IBM if we can get this resolved.

.gitignore
yurishkuro: `corepack` seems preferable, I am strongly against vendoring yarn code in the repo (feels insane to me that they would even propose this as a default setup).

In the issue there were mentions of problems with how corepack operates depending on versions. Probably not an issue for jaeger-ui.

Q: why yarn 3.8 and not 4.x?

andreasgerstmayr: The reason for yarn 3.8 was that we want to use cachi2 (used to ensure the build is hermetic, i.e. it caches dependencies to support building Jaeger UI without network access), which at the moment [only supports yarn v3](https://github.com/containerbuildsystem/cachi2/blob/main/docs/yarn.md) (I assume v4 might also be supported, but I didn't test it).

Alternatively, we could also use `npm`. It also supports workspaces now (in case this was the reason to switch to yarn in the first place). What do you think?

yurishkuro: I don't really follow. We have a yarn.lock file that ensures repeatable build. Are we worried about supply chain attack?

andreasgerstmayr: Yes, it's about supply chain attacks. Our downstream builds need to be [hermetic according to SLSA](https://slsa.dev/spec/v0.1/requirements#hermetic) and include a Software Bill of Materials (SBOM) etc.

You're right, effectively the yarn package manager already supports these properties. We need to support multiple languages and package managers though, so this part of the build pipeline is handled by an external tool like [cachi2](https://github.com/containerbuildsystem/cachi2) for our downstream builds.

yurishkuro:
We went with npm instead #2462

PR #2488, opened 2024-11-14
Bump cross-spawn to 7.0.5

rubenvp8510:
- Override  cross-spaw transitive dependency

## Description of the changes
- Set  cross-spawn to 7.0.5

## Checklist


yurishkuro:
1. why this specific dependency
2. usually upgrades (by the bot) are done directly in the lock file, not via override (which only creates a problem in the future)

yurishkuro:
I don't know how bots are doing it exactly, but can the change be only made in the lock file?

rubenvp8510:
> I don't know how bots are doing it exactly, but can the change be only made in the lock file?

I've already include the change only on the lock file! Thanks

yurishkuro (approved):

PR #2504, opened 2024-11-30
Pad start formatting function

drewcorlin1:
Given a traceparent header like 00-00000000000000001b3ef83a5d15ed1b-61409d351b3a5d99-01, jaeger will store (and thus display + interpolate) the trace ID as 1b3ef83a5d15ed1b. When attempting to use the shortened trace ID to link to logs some tracing integrations (in my case, logs using [pino](https://www.npmjs.com/package/pino) instrumented with the [OpenTelemetry pino instrumentation](https://www.npmjs.com/package/@opentelemetry/instrumentation-pino)) will use the full length trace ID, making jaeger link patterns not work.

This PR adds a padStart formatting function, heavily inspired by javascript's [String.padStart](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart) to format the trace ID in those cases. If approved, I'll update the documentation in https://github.com/jaegertracing/documentation to mention this function and it's behavior.

In addition to this specific function, I'd like feedback on the approach of passing arguments to formatting functions with the space-delimited approach (eg #{traceID | padStart 32 0}).

### DIFF
--- a/packages/jaeger-ui/src/model/link-patterns.test.js
+++ b/packages/jaeger-ui/src/model/link-patterns.test.js
@@ -349,6 +349,11 @@ describe('computeTraceLink()', () => {
       url: 'http://example.com/?traceID=#{traceID}&traceName=#{traceName}&startTime=#{startTime}&endTime=#{endTime}&duration=#{duration}',
       text: 'third link (#{traceID}, #{traceName}, #{startTime}, #{endTime}, #{duration})',
     },
+    {
+      type: 'traces',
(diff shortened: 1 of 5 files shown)


yurishkuro (commented):
please add tests for edge cases

drewcorlin1:
Add for this and for `epoch_micros_to_date_iso`

yurishkuro (approved):

PR #2501, opened 2024-11-15
Allow formatting link parameter values as ISO date #2487

drewcorlin1:
## Which problem is this PR solving?
Resolves #2487

## Description of the changes
Add the ability to format link variables.

## How was this change tested?
Unit tests and by running it locally with the following UI config
```javascript
const DEFAULT_CONFIG = {
  "linkPatterns": [
    {
      "type": "traces",
      "url": "https://myOpenSearch.com/_dashboards/app/discover#/?_g=(time:(from:'#{startTime | epoch_micros_to_date_iso}',to:'#{endTime | epoch_micros_to_date_iso}'))&_a=(index:filebeat-all,query:(language:kuery,query:'#{traceID}'))",
      "text": "Logs"
    }
  ]
};
```

## Checklist
- [x] I have read https://github.com/jaegertracing/jaeger/blob/master/CONTRIBUTING_GUIDELINES.md
- [x] I have signed all commits
- [x] I have added unit tests for the new functionality
- [x] I have run lint and test steps successfully
  - for `jaeger`: `make lint test`
  - for `jaeger-ui`: `yarn lint` and `yarn test`

PR #2507, opened 2024-12-01
Add new formatting function "add"

drewcorlin1:
In addition to #2501 and #2504, allowing any numbers (in my immediate case a date in microseconds since epoch) to be offset by a number is useful to allow users to see logs before/after their span/trace timestamps by a configurable offset. 

## Description of the changes
This adds a `add` function and gives the ability to chain formatting functions together (eg `#{endTime | add 60000000 | epoch_micros_to_date_iso}`)

Unit tests and with the following UI config
```
const DEFAULT_CONFIG = {
  "linkPatterns": [
    {
      "type": "traces",
      "url": "https://myKibana/_dashboards/app/discover#/?_g=(time:(from:'#{startTime | shift_epoch_micros -60000000 | epoch_micros_to_date_iso}',to:'#{endTime | shift_epoch_micros 60000000 | epoch_micros_to_date_iso}'))&_a=(columns:!(_source),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:filebeat-all,key:json.traceId,negate:!f,params:(query:'42c7668cd8d887da'),type:phrase),query:(match_phrase:(json.traceId:'42c7668cd8d887da')))),index:filebeat-all)",
      "text": "Logs"
    }
  ]
};
```

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/utils/link-formatting.tsx
+++ b/packages/jaeger-ui/src/utils/link-formatting.tsx
@@ -14,64 +14,62 @@
 
 import { Trace } from '../types/trace';
 
-function getFormatFunctions<T = Trace[keyof Trace]>(): Record<
+const formatFunctions: Record<
   string,
-  (value: T, ...args: string[]) => T | string | number
-> {
-  return {
-    epoch_micros_to_date_iso: microsSinceEpoch => {
-      if (typeof microsSinceEpoch !== 'number') {
-        console.error('epoch_micros_to_date_iso can only operate on numbers, ignoring formatting', {
-          value: microsSinceEpoch,
-        });
-        return microsSinceEpoch;
-      }
+  <T extends Trace[keyof Trace]>(value: T, ...args: string[]) => T | string | number
+> = {
+  epoch_micros_to_date_iso: microsSinceEpoch => {
+    if (typeof microsSinceEpoch !== 'number') {
+      console.error('epoch_micros_to_date_iso() can only operate on numbers, ignoring formatting', {
+        value: microsSinceEpoch,
+      });
+      return microsSinceEpoch;
+    }
 
-      return new Date(microsSinceEpoch / 1000).toISOString();
-    },
-    pad_start: (value, desiredLengthString: string, padCharacter: string) => {
-      if (typeof value !== 'string') {
-        console.error('pad_start can only operate on strings, ignoring formatting', {
-          value,
-          desiredLength: desiredLengthString,
-          padCharacter,
-        });
-        return value;
-      }
-      const desiredLength = parseInt(desiredLengthString, 10);
-      if (Number.isNaN(desiredLength)) {
-        console.error('pad_start needs a desired length as second argument, ignoring formatting', {
-          value,
-          desiredLength: desiredLengthString,
-          padCharacter,
-        });
-      }
+    return new Date(microsSinceEpoch / 1000).toISOString();
+  },
+  pad_start: (value, desiredLengthString: string, padCharacter: string) => {
+    if (typeof value !== 'string') {
+      console.error('pad_start() can only operate on strings, ignoring formatting', {
+        value,
+        desiredLength: desiredLengthString,
+        padCharacter,
+      });
+      return value;
+    }
+    const desiredLength = parseInt(desiredLengthString, 10);
+    if (Number.isNaN(desiredLength)) {
+      console.error('pad_start() needs a desired length as second argument, ignoring formatting', {
+        value,
+        desiredLength: desiredLengthString,
+        padCharacter,
+      });
+    }
 
-      return value.padStart(desiredLength, padCharacter);
-    },
+    return value.padStart(desiredLength, padCharacter);
+  },
 
-    add: (value, offsetString: string) => {
-      if (typeof value !== 'number') {
-        console.error('add can only operate on numbers, ignoring formatting', {
-          value,
-          offsetString,
-        });
-        return value;
-      }
+  add: (value, offsetString: string) => {
+    if (typeof value !== 'number') {
+      console.error('add() needs a numeric offset as an argument, ignoring formatting', {
+        value,
+        offsetString,
+      });
+      return value;
+    }
 
-      const offset = parseInt(offsetString, 10);
-      if (Number.isNaN(offset)) {
-        console.error('add needs a valid offset in microseconds as second argument, ignoring formatting', {
-          value,
-          offsetString,
-        });
-        return value;
-      }
+    const offset = parseInt(offsetString, 10);
+    if (Number.isNaN(offset)) {
+      console.error('add() needs a valid offset in microseconds as second argument, ignoring formatting', {
+        value,
+        offsetString,
+      });
+      return value;
+    }
 
-      return value + offset;
-    },
-  };
-}
+    return value + offset;
+  },
+};
 
 export function getParameterAndFormatter<T = Trace[keyof Trace]>(
   parameter: string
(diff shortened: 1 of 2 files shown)


yurishkuro:
As described the use case doesn't make sense to me. How does an operator of Jaeger install know what possible time skew might exist?

I can see the other use case: you want to see the logs from +-1min of the span timestamp, and need to transform the ts into an interval. If that's what you meant please update the description. 

As far as implementation, can this be simply an "add" function? It does not care about the units (up to the user), and by allowing negative arguments it can adjust ts in either direction.

packages/jaeger-ui/src/utils/link-formatting.tsx
```diff
@@ -49,31 +49,66 @@ function getFormatFunctions<T = Trace[keyof Trace]>(): Record<
...
+        console.error('add can only operate on numbers, ignoring formatting', {
+          value,
+          offsetString,
+        });
+        return value;
+      }
+
+      const offset = parseInt(offsetString, 10);
+      if (Number.isNaN(offset)) {
+        console.error('add needs a valid offset in microseconds as second argument, ignoring formatting', {
```
yurishkuro, line 64:
```suggestion
        console.error('add() needs a numeric offset as an argument, ignoring formatting', {
```

packages/jaeger-ui/src/utils/link-formatting.tsx
```diff
@@ -49,31 +49,66 @@ function getFormatFunctions<T = Trace[keyof Trace]>(): Record<
 
       return value.padStart(desiredLength, padCharacter);
     },
+
+    add: (value, offsetString: string) => {
+      if (typeof value !== 'number') {
+        console.error('add can only operate on numbers, ignoring formatting', {
```
yurishkuro, line 55:
```suggestion
        console.error('add() can only operate on numbers, ignoring formatting', {
```

packages/jaeger-ui/src/utils/link-formatting.tsx
```diff
@@ -49,31 +49,66 @@ function getFormatFunctions<T = Trace[keyof Trace]>(): Record<
...
-    console.error('Unrecognized format function name, ignoring formatting', {
-      parameter,
-      formatFunctionName,
-      validValues: Object.keys(formatFunctions),
-    });
-  }
+  const chainedFormatFunctions = parts
+    .slice(1)
+    .map(formatPart => {
+      const [formatFunctionName, ...args] = formatPart.split(' ');
```
yurishkuro, line 91: what happens if there are multiple spaces?

packages/jaeger-ui/src/utils/link-formatting.tsx
```diff
@@ -16,7 +16,7 @@ import { Trace } from '../types/trace';
 
 function getFormatFunctions<T = Trace[keyof Trace]>(): Record<
```
yurishkuro, line 17: why is this a function and not just a const array `formatFunctions`?

packages/jaeger-ui/src/utils/link-formatting.tsx
```diff
@@ -49,31 +49,66 @@ function getFormatFunctions<T = Trace[keyof Trace]>(): Record<
...
-  const formatFunction = formatFunctions[formatFunctionName];
-  if (!formatFunction) {
-    console.error('Unrecognized format function name, ignoring formatting', {
-      parameter,
-      formatFunctionName,
-      validValues: Object.keys(formatFunctions),
-    });
-  }
+  const chainedFormatFunctions = parts
+    .slice(1)
```
yurishkuro, line 89:
```suggestion
    .slice(1) // skip the input value
```

packages/jaeger-ui/src/utils/link-formatting.tsx
```diff
@@ -49,31 +49,66 @@ function getFormatFunctions<T = Trace[keyof Trace]>(): Record<
...
 }
 
 export function getParameterAndFormatter<T = Trace[keyof Trace]>(
   parameter: string
 ): {
   parameterName: string;
-  formatFunction: ((value: T) => T | string) | null;
+  formatFunction: ((value: T) => T | string | number) | null;
 } {
   const parts = parameter.split('|').map(part => part.trim());
```
yurishkuro, line 82: could this be something like this?
```suggestion
  const (parameterName, ...formats) = parameter.split('|').map(part => part.trim());
```

packages/jaeger-ui/src/utils/link-formatting.tsx
```diff
@@ -49,31 +49,66 @@ function getFormatFunctions<T = Trace[keyof Trace]>(): Record<
...
-    });
-  }
+  const chainedFormatFunctions = parts
+    .slice(1)
+    .map(formatPart => {
+      const [formatFunctionName, ...args] = formatPart.split(' ');
+      const formatFunction = formatFunctions[formatFunctionName];
+      if (!formatFunction) {
+        console.error(
+          'Unrecognized format function name, ignoring formatting. Other formatting functions may be applied',
```
yurishkuro, line 95: include `formatFunctionName` in the log

packages/jaeger-ui/src/utils/link-formatting.tsx
```diff
@@ -49,31 +49,66 @@ function getFormatFunctions<T = Trace[keyof Trace]>(): Record<
...
+            parameter,
+            formatFunctionName,
+            validValues: Object.keys(formatFunctions),
+          }
+        );
+        return null;
+      }
+      return (val: T) => formatFunction(val, ...args);
+    })
+    .filter((fn): fn is NonNullable<typeof fn> => fn !== null);
```
yurishkuro, line 106: odd statement, NonNullable is compared against null?

I think you could use flatMap() instead and return `[]` in cases of errors such that it will be automatically skipped

packages/jaeger-ui/src/utils/link-formatting.tsx
```diff
@@ -49,31 +49,66 @@ function getFormatFunctions<T = Trace[keyof Trace]>(): Record<
...
 export function getParameterAndFormatter<T = Trace[keyof Trace]>(
   parameter: string
 ): {
   parameterName: string;
-  formatFunction: ((value: T) => T | string) | null;
+  formatFunction: ((value: T) => T | string | number) | null;
 } {
   const parts = parameter.split('|').map(part => part.trim());
   const parameterName = parts[0];
   if (parts.length === 1) return { parameterName, formatFunction: null };
```
yurishkuro, line 84: is this early exit necessary? Due to formatting errors you may still end up with an empty list of formatters, why not handle it at the end?

packages/jaeger-ui/src/utils/link-formatting.tsx
```diff
@@ -49,31 +49,66 @@ function getFormatFunctions<T = Trace[keyof Trace]>(): Record<
...
 
-  const formatFunction = formatFunctions[formatFunctionName];
-  if (!formatFunction) {
-    console.error('Unrecognized format function name, ignoring formatting', {
-      parameter,
-      formatFunctionName,
-      validValues: Object.keys(formatFunctions),
-    });
-  }
+  const chainedFormatFunctions = parts
```
yurishkuro, line 88:
```suggestion
  const formatters = parts
```

packages/jaeger-ui/src/utils/link-formatting.tsx
drewcorlin1: It's included in the object that's the second argument to this `console.error()` call

packages/jaeger-ui/src/utils/link-formatting.tsx
drewcorlin1: This is a [type predicate](https://www.typescriptlang.org/docs/handbook/advanced-types.html#using-type-predicates) which says that it returns a boolean indicating if `fn` is non-null. It may make more sense reading it with a curly brace-wrapped function body
```typescript
.filter((fn): fn is NonNullable<typeof fn> => {
  return fn !== null;
});
```

I could use flatMap, but would just add more allocations for the same behavior right?

packages/jaeger-ui/src/utils/link-formatting.tsx
drewcorlin1: Fair point. I usually prefer early returns but recognize its a stylistic difference so can remove this check

packages/jaeger-ui/src/utils/link-formatting.tsx
drewcorlin1: Good catch, handling this + adding a test case

packages/jaeger-ui/src/utils/link-formatting.tsx
drewcorlin1: Irrelevant as of https://github.com/jaegertracing/jaeger-ui/pull/2507#discussion_r1864980255

drewcorlin1:
I don't think I can add labels. @yurishkuro would you be able to add the "changelog:bugfix-or-minor-feature" label (or whatever you deem appropriate)

yurishkuro (approved):

Issue #2092, opened 2024-01-06
[Feature]: UI configuration to display the entire trace ID


brunexgeek:
### Requirement

As a system architect, who monitors the behavior of production environments using Jaeger, I would like to be able to view the entire trace ID in the trace search and trace details.

### Problem

We generate our own IDs and the trace ID is sequential (similar to [Snowflake ID](https://en.wikipedia.org/wiki/Snowflake_ID)). The problem is that the UI only shows the first characters of the ID, which means that IDs like `0003b24be26bc00a` and `0003b24be22d400a` look the same in the listings (they both appear as `0003b24`). This also happens on the trace details page. The span ID, on the other hand, is shown in its entirety.

### Proposal

Add an option in the UI configuration to show the entire trace ID.

professorabhay:
Hey, @brunexgeek ! I would like to give it a try. 
Is there anything else that can help me to resolve it?

brunexgeek:
Hello there! Considering the case of sequential IDs, I think that's it. I thought we could just add an option to display the entire ID, disabled by default. I'm new in Jaeger to be honest. Thanks!

yurishkuro:
reopening - need to mention the new option in the UI configuration docs on the website

karankoder:
Hi @yurishkuro,
Can i add the new option in the docs?

yurishkuro:
@karankoder just do it, no need to ask

abhayporwals:
I think these changes required in the `jaegertracing/documentation` if I'm not wrong 🤔

yurishkuro:
yes

PR #2536, opened 2025-01-03
Add config option to allow displaying full traceID

avinpy-255:
- Resolves: https://github.com/jaegertracing/jaeger-ui/issues/2092
## Description of the changes
- I've added a funtion to show the full trace ID in the trace details page

- here is the screenshot of the function Full ID
Screenshot from 2025-01-03 18-00-08

- after clicking the Full ID
Screenshot from 2025-01-03 17-59-45

## Checklist


MAX-786:
looks great ... but IMHO it takes extra space as it's showing the User same information as it's already available on top. what you can do is that make the shorten traceID clickable and then clicking on it would expand to show Full ID and again clicking on it would collapse it. This would also clean-up the UI and IMO would be a better UX design

avinpy-255:
> looks great ... but IMHO it takes extra space as it's showing the User same information as it's already available on top. what you can do is that make the shorten traceID clickable and then clicking on it would expand to show Full ID and again clicking on it would collapse it. This would also clean-up the UI and IMO would be a better UX design

Thanks! I'll make TraceID clickable for better UX

avinpy-255:
Hey @MAX-786 
As suggested, I've successfully updated the ui here is the example: 
[Screencast from 2025-01-13 19-53-13.webm](https://github.com/user-attachments/assets/32fcafc1-d174-4677-aa07-2be3619a806b)

MAX-786:
@avinpy-255 Hi,
yepp, wait for the review from the maintainers.

yurishkuro:
The original issue was asking for **configuration** to make the trace ID displayed in full. There is no configuration added in this PR. The configuration can be specifying the desired length from the ID to display, e.g. 32 would display full ID.

packages/jaeger-ui/src/constants/default-config.tsx
```diff
@@ -20,6 +20,8 @@ import getVersion from '../utils/version/get-version';
 import { version } from '../../package.json';
 import { Config } from '../types/config';
 
+export const TraceIDLength = 7;
```
yurishkuro, line 23: we don't use constants for anything else, just use the value directly

packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageHeader.tsx
```diff
@@ -149,7 +150,7 @@ export function TracePageHeaderFn(props: TracePageHeaderEmbedProps & { forwarded
       return { ...rest, value: renderer(trace) };
     });
 
-  const traceShortID = trace.traceID.slice(0, 7);
+  const traceShortID = trace.traceID.slice(0, TraceIDLength);
```
yurishkuro, line 153: add a test?

avinpy-255: ok, I will add a test for tracIDlength

packages/jaeger-ui/src/types/config.tsx
```diff
@@ -135,6 +135,9 @@ export type Config = {
   // TODO when is it useful?
   scripts?: readonly TScript[];
 
+  // traceIDLength controls the length of the trace ID displayed in the UI.
+  traceIDLength?: number;
```
yurishkuro, line 139:
```suggestion
  traceIdDisplayLength?: number;
```

packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageHeader.tsx
```diff
@@ -149,7 +150,8 @@ export function TracePageHeaderFn(props: TracePageHeaderEmbedProps & { forwarded
       return { ...rest, value: renderer(trace) };
     });
 
-  const traceShortID = trace.traceID.slice(0, 7);
+  const traceIDLength = getConfigValue('traceIDLength');
+  const traceShortID = trace.traceID.slice(0, traceIDLength);
```
yurishkuro, line 154: This doesn't play nice with styling when very long.

image

I think the ID needs to be in smaller font.

yurishkuro (commented):
The new format should be applied consistently across the UI. I would recommend a dedicated component for trace ID display (I would also make it have a popup tooltip showing the full length string if the default max size is smaller)

image

image

packages/jaeger-ui/src/components/common/TraceIdDisplayLength.tsx
```diff
@@ -0,0 +1,39 @@
...
+import React from 'react';
+import { getConfigValue } from '../../utils/config/get-config';
+import './TraceIdDisplayLength.css';
+
+type Props = {
+  traceId: string;
+  className?: string;
+};
+
+export function TraceIdDisplayLength({ traceId, className = '' }: Props) {
```
yurishkuro, line 24:
```suggestion
export function TraceId({ traceId, className = '' }: Props) {
```
"DisplayLength" is implementation detail irrelevant to the user.

packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageHeader.tsx
```diff
@@ -37,7 +37,7 @@ import { getTraceLinks } from '../../../model/link-patterns';
 import './TracePageHeader.css';
 import ExternalLinks from '../../common/ExternalLinks';
 import { getTargetEmptyOrBlank } from '../../../utils/config/get-target';
-import TraceIdDisplayLength from '../../common/TraceIdDisplayLength';
+import TraceId from '../../common/TraceIdDisplayLength';
```
yurishkuro, line 40:
```suggestion
import TraceId from '../../common/TraceId';
```

packages/jaeger-ui/src/components/common/TraceId.tsx
```diff
@@ -0,0 +1,34 @@
...
+  className?: string;
+};
+
+export function TraceId({ traceId, className = '' }: Props) {
+  const traceIdDisplayLength = getConfigValue('traceIdDisplayLength') || 7;
+  const traceIdDisplay = traceId ? traceId.slice(0, traceIdDisplayLength) : '';
+  const lengthClass = traceIdDisplayLength === 7 ? 'TraceIDLength--short' : 'TraceIDLength--full';
+
+  return (
+    <small className={`TraceIDLength ${lengthClass} u-tx-muted ub-pr2 ${className} `}>{traceIdDisplay}</small>
```
yurishkuro, line 30: there was no `ub-pr2` style in packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageHeader.tsx. Do you need to pass addl args?

packages/jaeger-ui/src/types/config.tsx
```diff
@@ -135,6 +135,9 @@ export type Config = {
   // TODO when is it useful?
   scripts?: readonly TScript[];
 
+  // traceIDLength controls the length of the trace ID displayed in the UI.
```
yurishkuro, line 138:
```suggestion
  // traceIdDisplayLength controls the length of the trace ID displayed in the UI.
```

packages/jaeger-ui/src/components/common/TraceId.tsx
avinpy-255: You're right about the `ub-pr2` class not being in `TracePageHeader.tsx`. I was worried about conflicts with the existing `TraceID `(TraceHeader.tsx) display in the `<small>` tag which might affect `TraceHeader.test.js`, but I understand now that as long as we maintain the same structure with the new `TraceID` component and update tests accordingly, it will work fine. I'll remove the class from `TraceId.tsx`

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceHeader.tsx
```diff
@@ -86,9 +87,7 @@ export default function TraceHeader(props: Props) {
           {traceID ? (
             <React.Fragment>
               <TraceName key="name" traceName={traceName} error={error} state={state} />{' '}
-              <small key="id" className="u-tx-muted ub-pr2">
-                {traceID.slice(0, 7)}
-              </small>
+              <TraceId traceId={traceID} />
```
yurishkuro, line 90: I didn't get the answer to my question. Previously there was `ub-pr2` class added, not it's no longer added by the component itself. Why?

avinpy-255: The `ub-pr2` class is used for padding right between the `traceID` and the `TraceTimelineLink`.

Initially, I thought the test file for `TraceHeader.tsx` would check if the `traceID` had the `ub-pr2` class, which is why I implemented the `ub-pr2` class in the `TraceID.tsx `component.

Then I realised this approach was wrong, so I removed the`ub-pr2` class because this specific type of padding is only used in the Traces Compare page.

I specifically removed the `ub-pr2` in `TraceHeader.tsx` from my component because after adding my component to the Compare page with the `ub-pr2` class, the following issues occurred:
- The UI looks odd when users try to view a `traceID` of length 32
- While it may not affect users on desktop screens, the effect is noticeable on laptop screens

For these reasons, I decided not to use the `ub-pr2` class.

yurishkuro: My point was that `ub-pr2` was used in some scenarios but not everywhere. You seem to be approaching this as if it has to be either always present or never. Why can't it be used the same way as before? Your new component can accept additional styles from the caller.

avinpy-255: Ok I'm adding `ub-pr2`, in `TraceHeader.tsx`

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ResultItemTitle.tsx
```diff
@@ -107,7 +108,7 @@ export default class ResultItemTitle extends React.PureComponent<Props> {
           {duration != null && <span className="ub-right ub-relative">{formatDuration(duration)}</span>}
           <h3 className="ResultItemTitle--title">
             <TraceName error={error} state={state} traceName={traceName} />
-            <small className="ResultItemTitle--idExcerpt">{traceID.slice(0, 7)}</small>
```
yurishkuro, line 110: `ResultItemTitle--idExcerpt`?

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceHeader.tsx
yurishkuro: don't see it

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceHeader.tsx
```diff
@@ -93,9 +94,7 @@ export default function TraceHeader({
           {traceID ? (
             <React.Fragment>
               <TraceName key="name" traceName={traceName} error={error} state={state} />{' '}
-              <small key="id" className="u-tx-muted ub-pr2">
```
yurishkuro, line 96: what was the role of `key="id"` attribute?

packages/jaeger-ui/src/components/common/TraceId.test.js
```diff
@@ -0,0 +1,100 @@
+// Copyright (c) 2023 The Jaeger Authors
```
yurishkuro, line 1: 2025

packages/jaeger-ui/src/components/common/TraceId.tsx
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) 2023 The Jaeger Authors
```
yurishkuro, line 1: 2025

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceHeader.tsx
yurishkuro: How can the new component accept additional attributes? For example, in `TraceSpanView/index.tsx` we use `common/SearchableSelect.tsx` and pass an extra attribute `data-testid=`, which is not explicitly declared in `SearchableSelect` yet still compiles, and the `SearchableSelect` includes them via `{...props}`.

packages/jaeger-ui/src/components/common/TraceId.tsx
yurishkuro, line 1:
```suggestion
// Copyright (c) 2025 The Jaeger Authors
```

packages/jaeger-ui/src/components/common/TraceId.test.js
yurishkuro, line 1:
```suggestion
// Copyright (c) 2025 The Jaeger Authors
```

packages/jaeger-ui/src/components/common/TraceId.css
```diff
@@ -0,0 +1,28 @@
+/*
+Copyright (c) 2018 Uber Technologies, Inc.
```
yurishkuro, line 2:
```suggestion
Copyright (c) 2025 The Jaeger Authors
```

yurishkuro (approved):

PR #1940, opened 2023-11-03
Remove deprecated plexus/DirectedGraph

yurishkuro:
Hard to tell what was the intention/distinction between plexus/Digraph and plexus/DirectedGraph, but there seemed to be an intention to remove it:
  * the latter is not used anywhere in `jaeger-ui` or internally in `plexus` (other than being exported)
  * the history log shows changes by @tiffon  to `DirectedGraph` stopped but changes to `Digraph` continued
  * there are comments in several places "Update import after killing `DirectedGraph`"

Open issue: plexus/demo does use DirectedGraph

PR #2548, opened 2025-01-05
[clean-up]: Remove deprecated plexus/DirectedGraph

hari45678:
- Resolves #1940

## Description of the changes
- This PR removes the DirectedGraph component from plexus lib. The component was anyway looking ugly on demos (big arrows), which might be due to the fact that it stopped receiving changes.

- npm run start in plexus package.

## Checklist


hari45678:
A new version of plexus can also be published to npmjs

Steps:
- Change the version in plexus/package.json to `0.3.0` (for a minor up).
- run `npm run build`
- `npm login`
- `npm publish`
- done

packages/plexus/demo/src/index.tsx
```diff
@@ -447,87 +444,6 @@ class Demo extends React.PureComponent<{}, TState> {
             />
           </div>
         </div>
-        <h1>Directed graph with cycles - dot edges</h1>
```
yurishkuro, line 450: Rather than deleting, could we instead swap it for the Digraph? It looks like this example is using some other data sets

hari45678: Hm, we can use that data. I will try

hari45678:
I have changed the demos to align with Digraph props and types.

yurishkuro (approved):
Thanks!

Issue #2556, opened 2025-01-05
[Deps] Remove dependency on redux-form


yurishkuro:
The library is unmaintained and its [readme](https://github.com/redux-form/redux-form?tab=readme-ov-file#%EF%B8%8F-attention-%EF%B8%8F) says form state should not be used in Redux. 

Here is some analysis of Jaeger dependencies: https://github.com/jaegertracing/jaeger-ui/pull/1261#issuecomment-1465321804

yaten2302:
Hey @yurishkuro , in this issue, we've to remove the dependency of redux-form, but what should be used instead of that to render?
I was checking out the codebase and I think that redux form is being used [redux-form-field-adapter.tsx](https://github.com/jaegertracing/jaeger-ui/blob/main/packages/jaeger-ui/src/utils/redux-form-field-adapter.tsx), could you please guide that instead of this what should be used for form state?

yurishkuro:
I don't know what we should use, this is part of this issue to figure that out.

kalio007:
Hello @yurishkuro Please i'll love to be assigned this task

yaten2302:
Hey @yurishkuro , I was searching of some redux-form alternatives, I found - [`react final form`](https://final-form.org/react). Could you please confirm that should I use this to migrate from redux-form?

yurishkuro:
Do we actually need an additional framework? We only have a handful of simple forms, I think React can support them without any other libraries.

yaten2302:
Oh, okay, got it, will do it in pure react 👍 
BTW, I think I mentioned the wrong file here, the `redux-form` lib is present in these files mainly:
1. https://github.com/jaegertracing/jaeger-ui/blob/main/packages/jaeger-ui/src/components/Monitor/ServicesView/index.tsx
2. https://github.com/jaegertracing/jaeger-ui/blob/main/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.tsx
3. https://github.com/jaegertracing/jaeger-ui/blob/main/packages/jaeger-ui/src/components/SearchTracePage/SearchForm.jsx
4. https://github.com/jaegertracing/jaeger-ui/blob/main/packages/jaeger-ui/src/middlewares/index.js
5. https://github.com/jaegertracing/jaeger-ui/blob/main/packages/jaeger-ui/src/reducers/index.js

Could you please guide that have I mentioned it correctly @yurishkuro ?

yashpandey06:
[...]

PR #2562, opened 2025-01-07
[fix]: Remove redux-form dependency from Monitor page

hari45678:
- part of #2556 

## Description of the changes
- This PR removes dependency of Monitor page on `redux-form`.
- To keep this PR small and easy to review, I haven't removed other instances of `redux-form` in this same PR. Neither have I removed `redux-form` package as of now.
- `Antd` `SearchableSelect` requires loading type to be `boolean | undefined` so I have changed the related types in other files to fix ts-lint errors.

### Why this change was needed?
- Monitor page has a few selectors that doesn't need any global state sync to other components. `redux-form` is anyway deprecating so in-built react state management is a good choice.
- React uses `state` object for state management, and `props` for injecting props to the component. `redux-form` use to inject those Dropdown Selector values `selectedService`, `selectedSpanKind`, and `selectedTimeFrame` from redux store.
- This PR moves that to in-built react state, so we won't have to inject those from store anymore. 

- `npm run test`
- `npm run start` with monitor page view. everything seems to be working fine.

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/Monitor/ServicesView/__snapshots__/index.test.js.snap
+++ b/packages/jaeger-ui/src/components/Monitor/ServicesView/__snapshots__/index.test.js.snap
@@ -14,19 +14,13 @@ exports[`<MonitorATMServicesView> ATM snapshot test 1`] = `
         >
           Service
         </h2>
-        <Field
-          component={[Function]}
-          name="service"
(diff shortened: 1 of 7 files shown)


yurishkuro (approved):
Thanks!

PR #2563, opened 2025-01-07
init: remove redux-form dependency from packages/jager-ui/src/middlewares/index.js

### LINKED ISSUE #2556: [Deps] Remove dependency on redux-form

yaten2302:
- 

## Description of the changes
- This PR attempts to remove dependency of redux-form from jager-ui. 

## Checklist


yurishkuro:
cc @hari45678

yurishkuro:
@yaten2302 this does not fix the whole issue, so please don't use `Fixes #` syntax as it may close the issue prematurely.

packages/jaeger-ui/src/middlewares/index.js
```diff
@@ -26,13 +25,19 @@ export { default as trackMiddleware } from './track';
...
   ) {
     store.dispatch(fetchServiceOperations(action.payload));
-    store.dispatch(change('searchSideBar', 'operation', 'all'));
+    store.dispatch({
+      type: 'FORM_CHANGE',
+      meta: {
+        form: 'searchSideBar',
+        field: 'operation',
+      }
+    })
```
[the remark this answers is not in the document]
yaten2302: Hey, should I create a new func similar to `change`? I was checking about it on the [redux-form repo](https://github.com/redux-form/redux-form/blob/master/src/actions.js#L237-L247). What change does is that it save the value to the field in the given form, so will creating a new function do? I was also checking out you [PR](https://github.com/jaegertracing/jaeger-ui/pull/2562/files#diff-3feeebf160ae6670b08f25b846765ccd375319f7175e77067c2ffc0a817c9400R203), in this I think you've created new functions to handle any change in the form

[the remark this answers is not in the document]
hari45678: That depends. You would have to do some manual review of the codebase and watch out for the places where that middleware is actually dispatching actions.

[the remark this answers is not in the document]
yaten2302: Got it, I'll check in the codebase all the use cases of redux-form components and actions, and then will make the changes accordingly 👍
Will ask about it here or on public slack channel if I've any doubt!
Thanks

packages/jaeger-ui/src/middlewares/index.js
```diff
@@ -26,13 +25,19 @@ export { default as trackMiddleware } from './track';
  */
 export const loadOperationsForServiceMiddleware = store => next => action => {
```
yurishkuro, line 26: where is this even used? I only see a declaration of this function
```
$ rg loadOperationsForServiceMiddleware

packages/jaeger-ui/src/middlewares/index.js
27:export const loadOperationsForServiceMiddleware = store => next => action => {
```

yurishkuro: I just tried removing this function and the test for it and the UI builds just fine.

yaten2302: Yes, @yurishkuro , I think the other function is also not being used anywhere, should I remove both of these? <br>

image

yurishkuro: If tests and build succeed then feel free to remove.

hari45678: It is used. Here's how:

https://github.com/jaegertracing/jaeger-ui/blob/1ee19c24a6b34ad7f1f75a6ec7cc4938de0eefde/packages/jaeger-ui/src/utils/configure-store.js#L24

And why `rg` didn't showed the exact usage of `loadOperationsForServiceMiddleware` is because that function object is being injected in a different way over here:

https://github.com/jaegertracing/jaeger-ui/blob/1ee19c24a6b34ad7f1f75a6ec7cc4938de0eefde/packages/jaeger-ui/src/utils/configure-store.js#L41-L46

Also, in SearchForm (where we actually select a service, select operation, start time, end time etc.), this middleware is used to load the operations related to the selected service.

The reason why tests work is because most of the middlewares have to be mocked while testing.

If you want to actually check what this middle ware is doing, remove it, then go to Search Form on `/search` page on Jaeger-UI and select a service from Service selector. Now try selecting an Operation from Operation selector. You won't see any option in that Operation selector.

hari45678: My test `middlewares/index.test.js` also failed after removing this function

yaten2302: Thanks for your explanation @hari45678 , now I understood how this `loadOperationsForServiceMiddleware` is being used 👍 

BTW, just wanted to ask that for running the tests - `npm test --path-to-test`, is this the correct command?

Also, I've one more doubt, that in here, I've changed the `service`,  `<Field>` tag to `<SearchableSelect>` and removed the `component={}` and it's working fine. The only change which I'm able to see is:

<b>Before:</b>
image

<b>After:</b>
image

But, on removing this, a blank screen is being shown:
image

image

Currently, I'm trying to make another function to manage this, to display a component and remove the dependency from `AdaptedSelect`, `AdaptedInput`, etc..., but could you please guide that why is this happening, or any link to the file or func due to which this is occurring?

hari45678: Appreciate that you are trying but what's the point of creating a PR in first place if someone else has to figure out everything?

yaten2302: Yep, sorry, @hari45678 , I'm trying to figure out why is this going wrong, once done I'll just ask is my approach correct or not and then try to debug further 👍 
Thanks and Apologies for inconvenience

yaten2302: Hi @hari45678 , wanted to share an update, I was trying to remove the `<Field />` dependency from the `SearchForm.jsx`, I was trying to get and set the values selected by the user. The values are getting selected but not getting displayed, currently I'm working on it 👍 <br>
image

When I'm selecting jaeger over here, it's getting logged to the console, but not getting selected/shown here.

yaten2302:
Hi @hari45678 @yurishkuro , apologies for the delay on this, now I've pushed all the changes that were required to remove the redux-form dependency from `SearchForm.jsx`, could you please have a look at it and LMK if any changes are required?
Thanks

Issue #2532, opened 2024-12-26
[Deps] Upgrade react-router-dom to 7.x


yurishkuro:

Abiral-2724:
hello @yurishkuro , assign the task to me

yurishkuro:
@Abiral-2724 consider it assigned

Abiral-2724:
@yurishkuro , react-router-dom updated successfully. Check the PR .

Srishti-j18:
Hi, is this issue open to work on? I see one PR related to this, but there is no recent update.

yurishkuro:
@Parship999 this got marked stale even though it has help-wanted which we wanted to exempt from stale bot
```
          exempt-issue-labels: 'do-not-expire,help-wanted'
```

ghost:
> [@Parship999](https://github.com/Parship999) this got marked stale even though it has help-wanted which we wanted to exempt from stale bot
> 
> ```
>           exempt-issue-labels: 'do-not-expire,help-wanted'
> ```

it's due to label name mismatch. Github issues use help wanted (with space) but stale workflow expecting help-wanted (with hyphen). I will fix it

PR #2553, opened 2025-01-05
[revert]: revert redux and react-redux dependency upgrades

hari45678:
Reverts jaegertracing/jaeger-ui#2537
Reason: https://github.com/jaegertracing/jaeger-ui/pull/2542#issuecomment-2571520772
Part of #2530

PR #2565, opened 2025-01-07
[Deps]: Upgraded react-router-dom

yashpandey06:
- 
- Fixes: #2532

## Description of the changes
- Upgraded react router dom to 7.x.x dependency

## Checklist


packages/jaeger-ui/src/components/App/index.jsx
```diff
@@ -88,49 +65,21 @@ export default class JaegerUIApp extends Component {
     return (
       <ConfigProvider theme={jaegerTheme}>
         <Provider store={store}>
-          <HistoryProvider history={history}>
```
yurishkuro, line 91: what is the implication of removing history here?

packages/jaeger-ui/src/components/App/index.jsx
```diff
@@ -1,49 +1,26 @@
-// Copyright (c) 2017 Uber Technologies, Inc.
```
yurishkuro, line 1: why are you changing headers?

packages/jaeger-ui/src/components/App/index.jsx
yashpandey06: Yeah , it should'nt be removed ...

packages/jaeger-ui/src/components/App/index.jsx
yashpandey06: @yurishkuro  restored

packages/jaeger-ui/src/components/App/index.jsx
```diff
@@ -89,47 +81,21 @@ export default class JaegerUIApp extends Component {
...
-                    <Redirect to={searchPath} />
-                  </Route>
-
-                  <Route>
-                    <NotFound />
-                  </Route>
-                </Switch>
-              </Page>
-            </Router>
+            <Page>
```
yurishkuro, line 84: previously there was a history attribute
```
<Router history={history}>
```
what's the implication of removing it?

Disclaimer: I am not a React developer, don't know how any of this works.

yashpandey06: ## Explanation for Significant Changes in Routing (React Router v6+)

### 1. `<Router history={history}>` → `<BrowserRouter>`
- **Reason:** 
  In React Router v6+, the external `Router` component with a custom `history` object has been replaced with `<BrowserRouter>`. React Router v6+ handles the navigation internally, which simplifies routing management by eliminating the need for manually handling `history`.

### 2. `<Switch>` → `<Routes>`
- **Reason:** 
  The `<Switch>` component is replaced with `<Routes>` in React Router v6+. The `<Routes>` component is used to wrap route definitions and ensures that only the first matching route is rendered. This change improves routing performance and reduces the need for complex route management.

### 3. `<Redirect>` → `<Navigate>`
- **Reason:** 
  React Router v6+ has removed the `<Redirect>` component. Instead, it introduces `<Navigate>`, which is used for programmatic navigation and redirection. `<Navigate>` simplifies the redirection logic by offering a cleaner and more declarative approach.

### 4. Removal of `HistoryProvider`
- **Reason:** 
  With the introduction of `useNavigate` and the internal `history` handling , the need for a custom `HistoryProvider` has been eliminated. Routing and navigation are now directly managed using the `useNavigate` hook, reducing boilerplate code and improving the structure of the application.
  
  
  cc : @yurishkuro

packages/jaeger-ui/src/components/App/index.jsx
yashpandey06: @yurishkuro the need of custom History is not there in v6+ versions of react-router-dom as they handle 
it internally well .

packages/jaeger-ui/src/components/App/index.jsx
```diff
@@ -11,39 +11,29 @@
...
 import processScripts from '../../utils/config/process-scripts';
 import prefixUrl from '../../utils/prefix-url';
+import { store } from '../../utils/configure-store';
 
 import '../common/vars.css';
 import '../common/utils.css';
 import 'antd/dist/reset.css';
 import './index.css';
-import { history, store } from '../../utils/configure-store';
-import { HistoryProvider } from '../../utils/useHistory';
```
yurishkuro, line 46: Use `../../utils/useHistory` used anywhere else? Should it be deleted?

packages/jaeger-ui/package.json
```diff
@@ -77,8 +77,7 @@
     "react-is": "^18.2.0",
     "react-json-view-lite": "2.0.1",
     "react-redux": "^8.1.2",
-    "react-router-dom": "5.3.4",
-    "react-router-dom-v5-compat": "^6.24.0",
+    "react-router-dom": "7.1.1",
```
yurishkuro, line 80: I expect this needs changed to package-lock

packages/jaeger-ui/src/components/App/index.jsx
yashpandey06: @yurishkuro  yeah i believe ....this is significant to notice that we can even replace useHistory hook and withProps component functionality with useNavigate .

package-lock.json
```diff
@@ -20115,13 +20019,12 @@
         "react-icons": "^5.0.1",
         "react-is": "^18.2.0",
         "react-json-view-lite": "2.0.1",
-        "react-redux": "^9.2.0",
```
yurishkuro, line 20118: @hari45678 this looks like a left over from #2553?

package-lock.json
```diff
@@ -20160,7 +20063,7 @@
         "jest-junit": "^16.0.0",
         "less": "4.2.0",
         "react-test-renderer": "^18.3.1",
-        "rollup-plugin-visualizer": "^5.13.1",
+        "rollup-plugin-visualizer": "^5.11.0",
```
yurishkuro, line 20066: any idea what's causing this downgrade?

packages/jaeger-ui/src/components/App/index.jsx
yurishkuro: I see you remove the screenshot of the docs, please add a link to those docs instead

package-lock.json
hari45678: yes actually. i hand picked the changes from original PR. This got left out.

anshgoyalevil:
Looking at the changes. Have you tested them locally? @yashpandey06

PR #2566, opened 2025-01-08
fix TraceDiff graph pan and zoom issue

anshgoyalevil:
- Fix pan and zoom issue on TraceDiff graph.
## Description of the changes
- This PR fixes the PAN and ZOOM functionality, which was caused by a minor CSS issue on the graph's parent that had relative position carried down the DOM to graph.

- viewed the changes in browser by running jaeger-ui in dev, and jaeger monitor docker compose example.

## Checklist


anshgoyalevil:
@yurishkuro It was a trivial fix so I think we can get it solved without waiting for a mentee to actually dig in

yurishkuro (approved):

yurishkuro:
Thanks!

PR #2569, opened 2025-01-09
[fix]: Remove redux-form dependency from sort selector

hari45678:
- part of #2556 

## Description of the changes
- This PR removes the dependency of Sort Selector on redux-form.
- We could previously sort the traces on homepage, and the value we were selecting on this Sort Selector was reflecting to the SearchTrace page that is responsible to rendering the Sorted Traces list and sorting the traces. This state exchange was previously done by redux-form but now we are directly passing it using react state and component props.
- We were also tracking the changes done by a user in this sort selector using a middleware tied to the redux-actions. Now we are not using redux for form state management so tracking logic is directly moved to the part where a sort action takes place, i.e., `handleSortChange`
- Removed @types/redux-form as there is no .ts/x file left that is using redux-form.

- npm run test
- npm run start and using the sort selector

## Checklist


packages/jaeger-ui/src/components/SearchTracePage/SearchForm.track.tsx
```diff
@@ -49,10 +46,6 @@ export function trackFormInput(
   trackEvent(CATEGORY_SERVICE, serviceName);
 }
 
-export const middlewareHooks = {
-  [constants.FORM_CHANGE_ACTION_TYPE]: (_: Store<ReduxState>, action: any) => {
```
anshgoyalevil, line 53: was is only used for tracking Sort Selector changes?

hari45678: Yes, Looking into the codebase, this `middlewareHooks` had only one if condition that was checking if the action dispatched via `redux-form` has the `.meta` property as `sortBy` which is possible for only sort selector.
I couldn't find other pages (other than Trace Results page) where a similar sorting selector is used.

yurishkuro (approved):

PR #2571, opened 2025-01-09
Fix: Deep clone trace data for consistency

### LINKED ISSUE #2483: [Bug]: Inconsistent Data Displayed in Flamegraph
### What happened?

When analyzing the trace flamegraph, some data appears to be inaccurate. Specifically, when there are two spans at the same level in the flamegraph, switching to the "trace timeline" view and then returning to the flamegraph causes the data to change. The duration of the spans seems to increase upon revisiting the flamegraph.

### Steps to reproduce

1. Open the trace flamegraph for a trace containing two spans at the same level.
2. Note the duration and positioning of the spans.
3. Switch to the "trace timeline" view.
4. Return to the flamegraph view.
5. Observe that the duration and/or positioning of the spans have changed.

### Expected behavior

Switching between views causes changes in the flamegraph display, showing inaccurate durations for the spans.

### Relevant log output

_No response_

### Screenshot

_No response_

### Additional context

_No response_

### Jaeger backend version

v1.53.0

### SDK

Jaeger v1.53.0
Commit b620f0e
Build 2024-01-08T18:05:40Z
Jaeger UI v1.37.0

### Pipeline

_No response_

### Stogage backend

_No response_

### Operating system

_No response_

### Deployment model

Kubernetes

### Deployment configs

_No response_

Zen-cronic:
Resolves #2483 

## Description of the changes
- deep cloned the trace data before passing it to @pyroscope/flamegraph

- npm run test
- visual comparison with this traces data [trace.zip](https://github.com/user-attachments/files/18364826/trace.zip):

Timeline view: 

timeline

Flamegraph view:

flamegraph

Navigating back and forth between flamegraph and timeline views no longer show inconsistent data

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceFlamegraph/index.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceFlamegraph/index.tsx
@@ -13,12 +13,14 @@
 // limitations under the License.
 import React from 'react';
 import { FlamegraphRenderer, convertJaegerTraceToProfile } from '@pyroscope/flamegraph';
+import _cloneDeep from 'lodash/cloneDeep';
 
 import '@pyroscope/flamegraph/dist/index.css';
 import './index.css';
 
 const TraceFlamegraph = ({ trace }: any) => {
-  const convertedProfile = trace && trace.data ? convertJaegerTraceToProfile(trace.data) : null;
+  // Cloned b/c convertJaegerTraceToProfile or FlamegraphRenderer can possibly mutate the trace
+  const convertedProfile = trace && trace.data ? convertJaegerTraceToProfile(_cloneDeep(trace.data)) : null;
 
   return (
     <div className="Flamegraph-wrapper">
--- a/packages/jaeger-ui/src/components/TracePage/TraceFlamegraph/index.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceFlamegraph/index.tsx
@@ -20,6 +20,7 @@ import './index.css';
 
 const TraceFlamegraph = ({ trace }: any) => {
   // Cloned b/c convertJaegerTraceToProfile or FlamegraphRenderer can possibly mutate the trace
+  // https://github.com/jaegertracing/jaeger-ui/issues/2483
   const convertedProfile = trace && trace.data ? convertJaegerTraceToProfile(_cloneDeep(trace.data)) : null;
 
   return (


packages/jaeger-ui/src/components/TracePage/TraceFlamegraph/index.tsx
```diff
@@ -13,12 +13,14 @@
 // limitations under the License.
 import React from 'react';
 import { FlamegraphRenderer, convertJaegerTraceToProfile } from '@pyroscope/flamegraph';
+import _cloneDeep from 'lodash/cloneDeep';
```
Zen-cronic, line 16: An alternative is to use the global `structuredClone` method, but it's still not supported in [jsdom](https://github.com/jsdom/jsdom/issues/3363). The workaround would be: 

```javascript
// in TracePage/TraceFlamegraph/index.test.js
beforeAll(() => {
  global.structuredClone = jest.fn((obj) => JSON.parse(JSON.stringify(obj)));

});

afterAll(() => {
  structuredCloneSpy.mockRestore();
});

```
But we already have lodash installed, so might as well use it

packages/jaeger-ui/src/components/TracePage/TraceFlamegraph/index.tsx
```diff
@@ -13,12 +13,14 @@
...
 
 import '@pyroscope/flamegraph/dist/index.css';
 import './index.css';
 
 const TraceFlamegraph = ({ trace }: any) => {
-  const convertedProfile = trace && trace.data ? convertJaegerTraceToProfile(trace.data) : null;
+  // Cloned b/c convertJaegerTraceToProfile or FlamegraphRenderer can possibly mutate the trace
```
yurishkuro, line 22:
```suggestion
  // Cloned b/c convertJaegerTraceToProfile or FlamegraphRenderer can possibly mutate the trace
  // https://github.com/jaegertracing/jaeger-ui/issues/2483
```

yurishkuro:
does it actually resolve the original issue? Seems like a partial solution (a good one, to keep the trace unchanged).

Zen-cronic:
> does it actually resolve the original issue? Seems like a partial solution (a good one, to keep the trace unchanged).

Yep, b/c now `me-java-rpgle-std::SELECT postgres.MU24020F` in the timeline view is 8.17s, which is ≈ `0.14 min` in the flamegraph. 
likewise, `me-java-rpgle-std::CHAIN Execution` is 10.57ms, displayed as `<0.01 min` in the flamegraph.

 These values don't change no matter how many times we switch views.

yurishkuro (approved):

yurishkuro:
Thanks!

PR #2580, opened 2025-01-10
Update: React router dom from v5 to v6

Zen-cronic:
- Part of #2532 & #1825 (duplicate issues?) 
- Bridge towards v7 migration

## Description of the changes
- [x] Incremental Migrate to v6: migration guide: https://github.com/remix-run/react-router/discussions/8753
- [x] remove unused modules/dependencies

- `npm run test`

## Commands Ran

- `npm install @types/history@^4.7.11`, removed in #1637 but added back b/c the pkg became fully typed only in [v5.0.0](https://github.com/remix-run/history/releases/tag/v5.0.0)
- `npm uninstall react-router-dom-v5-compat` to uninstall compat pkg
- `npm install react-router-dom@^v6.28.1` to install v6
- `npm uninstall @types/react-router-dom` b/c the rrd pkg became fully typed in [v6.0.0](https://github.com/remix-run/react-router/releases/tag/v6.0.0)

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.tsx
+++ b/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.tsx
@@ -112,7 +112,7 @@ export class UnconnectedSearchResults extends React.PureComponent<SearchResultsP
   };
 
   onDdgViewClicked = () => {
-    const { history, urlQueryParams } = this.props;
+    const { location, history, urlQueryParams } = this.props;
     const urlState = queryString.parse(location.search);
     const view = urlState.view && urlState.view === 'ddg' ? EAltViewActions.Traces : EAltViewActions.Ddg;
     trackAltView(view);


packages/jaeger-ui/src/components/App/index.test.js
```diff
@@ -18,7 +18,7 @@ import shallow from '../../utils/ReactShallowRenderer.test';
 import JaegerUIApp from './index';
 
 describe('JaegerUIApp', () => {
-  it('does not explode', () => {
+  it.skip('does not explode', () => {
```
Zen-cronic, line 21: TODO: fix snapshot after each semver major migration

Zen-cronic:
Migration to v6 done! All the unit tests pass

TODOs:

- [ ] fix error on "Compare" page (`TraceDiff`): 
Error: `HTTP Error: strconv.ParseUint: parsing "...": invalid syntax`
URL: `/api/traces/...`
Response body: `	{
  "data": null,
  "total": 0,
  "limit": 0,
  "offset": 0,
  "errors": [
    {
      "code": 400,
      "msg": "strconv.ParseUint: parsing \"...\": invalid syntax"
    }
  ]
}`

- [ ] Continue v7 migration

yurishkuro:
please do not bundle 6 & 7 migration into one PR, let's get a green CI and merge

yurishkuro:
please document in the description how the dependency files were changed (which commands did you run). For example, why are `"peer": true` entries being removed? Are you using a different version of npm?

packages/jaeger-ui/src/components/App/TraceIDSearchInput.test.js
```diff
@@ -29,9 +29,9 @@ describe('<TraceIDSearchInput />', () => {
     history = createMemoryHistory();
     render(
       <HistoryProvider history={history}>
```
yurishkuro, line 31: why in `packages/jaeger-ui/src/components/App/index.jsx` the history provider is removed but here it's not? What is the motivation for either decision?

packages/jaeger-ui/src/components/App/index.jsx
```diff
@@ -30,7 +30,7 @@ import SearchTracePage from '../SearchTracePage';
 import { ROUTE_PATH as searchPath } from '../SearchTracePage/url';
 import TraceDiff from '../TraceDiff';
 import { ROUTE_PATH as traceDiffPath } from '../TraceDiff/url';
-import TracePage from '../TracePage';
+import TracePage from '../TracePage/index';
```
Zen-cronic, line 33: TODO: remove `/index`

packages/jaeger-ui/package.json
```diff
@@ -48,7 +48,7 @@
     "@ant-design/compatible": "^5.1.3",
     "@jaegertracing/plexus": "0.2.0",
     "@pyroscope/flamegraph": "0.21.4",
-    "@sentry/browser": "^8.18.0",
+    "@sentry/browser": "^8.48.0",
```
Zen-cronic, line 51: change back to 8.18.0 to match [current](https://github.com/jaegertracing/jaeger-ui/blob/1ee19c24a6b34ad7f1f75a6ec7cc4938de0eefde/packages/jaeger-ui/package.json#L51)

Zen-cronic:
> please document in the description how the dependency files were changed (which commands did you run). For example, why are `"peer": true` entries being removed? Are you using a different version of npm?

noted. i'm on node v22.12.0 and npm 10.9.0. 
The `peer:true` entries seem to be related to the`antd` library

packages/jaeger-ui/src/components/App/TraceIDSearchInput.test.js
Zen-cronic: In `packages/jaeger-ui/src/components/App/index.jsx`,  `HistoryProvider` is removed b/c rrd v6 works without being wrapped with that component.

But in this test and other [tests](https://github.com/jaegertracing/jaeger-ui/blob/1ee19c24a6b34ad7f1f75a6ec7cc4938de0eefde/packages/jaeger-ui/src/utils/withRouteProps.test.js#L39), the `history` object from it's namesake package is used for assertions. So, removing `HistoryProvider` would mean a rewrite for these tests.

yurishkuro: but what's the story with the `history` package, aren't we going to have to remove it altogether as incompatible going forward? It's ok if that clean-up can be done incrementally, we can defer it to other PRs.

Zen-cronic: `history` is no longer a peer dep in [v6](https://github.com/remix-run/react-router/discussions/8753#top) and later. And yes, we should have a separate PR for removing it from this repo. 

And we need to consider how to replace the function of that package - currently, it's used to provide the navigation props (e.g., `history.push(/)`) in class-based components wrapped with `withRouteProps`. Because they're class based, we can't use hooks like `useNavigation`). This discussion is better continued in another PR :)

packages/jaeger-ui/src/components/App/index.jsx
```diff
@@ -88,49 +87,21 @@ export default class JaegerUIApp extends Component {
...
-                  <Route path={qualityMetricsPath}>
-                    <QualityMetrics />
-                  </Route>
-                  <Route path={monitorATMPath}>
-                    <MonitorATMPage />
-                  </Route>
+          <Routes>
+            <Route path="/" element={<Page />}>
+              <Route index element={<SearchTracePage />} />
+              <Route path={searchPath} element={<SearchTracePage />} />
```
Zen-cronic, line 93: should use relative paths in v7 migration: e.g., `search` instead of `/search`

packages/jaeger-ui/src/components/App/TraceIDSearchInput.test.js
yurishkuro: Specifically, the migration guide you linked recommends using `navigate` instead of `history`.

yurishkuro: And the docs say BrowserRouter incorporates the history in it.

packages/jaeger-ui/src/components/App/index.jsx
```diff
@@ -42,8 +42,7 @@ import '../common/vars.css';
 import '../common/utils.css';
 import 'antd/dist/reset.css';
 import './index.css';
-import { history, store } from '../../utils/configure-store';
-import { HistoryProvider } from '../../utils/useHistory';
```
yurishkuro, line 46: can this module be deleted?

packages/jaeger-ui/src/components/SearchTracePage/index.test.js
```diff
@@ -43,11 +43,13 @@ import transformTraceData from '../../model/transform-trace-data';
 import { store as globalStore } from '../../utils/configure-store';
 
 const AllProvider = ({ children }) => (
-  <BrowserRouter>
+  <MemoryRouter>
```
yurishkuro, line 46: please elaborate on this swap

packages/jaeger-ui/src/types/index.tsx
```diff
@@ -56,7 +56,7 @@ export type ReduxState = {
     error: ApiError | TNil;
   };
   embedded: EmbeddedState;
-  router: Router & {
+  router: typeof Router & {
```
yurishkuro, line 59: just to confirm, are we sure we actually need this variable now in the redux state?

package-lock.json
```diff
@@ -20103,7 +19951,7 @@
         "@ant-design/compatible": "^5.1.3",
         "@jaegertracing/plexus": "0.2.0",
         "@pyroscope/flamegraph": "0.21.4",
-        "@sentry/browser": "^8.48.0",
+        "@sentry/browser": "^8.18.0",
```
yurishkuro, line 19954: What causes this downgrade? Before you're running install/uninstall commands, are you on the latest set of dependencies (`npm ci`)?

yurishkuro (commented):
overall lgtm, but please address questions / comments.

packages/jaeger-ui/src/components/App/index.jsx
Zen-cronic: not in this PR, other unrelated [tests](https://github.com/jaegertracing/jaeger-ui/blob/1ee19c24a6b34ad7f1f75a6ec7cc4938de0eefde/packages/jaeger-ui/src/utils/withRouteProps.test.js#L39) depend on it. We should remove history on its own PR

packages/jaeger-ui/src/components/SearchTracePage/index.test.js
Zen-cronic: apparently has no effect on the tests (either works). Will revert

packages/jaeger-ui/src/components/App/Page.tsx
```diff
@@ -62,7 +63,9 @@ export class PageImpl extends React.Component<TProps> {
               <TopNav />
             </Header>
           )}
-          <Content className={contentCls}>{this.props.children}</Content>
+          <Content className={contentCls}>
+            <Outlet />
```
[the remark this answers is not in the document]
Zen-cronic: this addresses one of the action items in #1825 `Use <Outlet /> instead of props.children to render nested routes inside <Page />`. It's the recommended approach for nested routes, [docs.](https://reactrouter.com/tutorials/address-book#nested-routes-and-outlets)

or we can start using `<Outlet/>` only in the v7 migration PR.

package-lock.json
Zen-cronic: it seems that the current dep version is `8.18.0`? https://github.com/jaegertracing/jaeger-ui/blob/1ee19c24a6b34ad7f1f75a6ec7cc4938de0eefde/packages/jaeger-ui/package.json#L51

yurishkuro: oh. I hate it when GH glitches like that. It's showing fake diffs, how to even trust its diff...

packages/jaeger-ui/src/components/App/TraceIDSearchInput.test.js
Zen-cronic: > Specifically, the migration guide you linked recommends using `navigate` instead of `history`.

yep, we should replace `history` with `navigate` from the `useNavigate` hook. But that change can't be made unless the class-based components (like [this](https://github.com/jaegertracing/jaeger-ui/blob/1ee19c24a6b34ad7f1f75a6ec7cc4938de0eefde/packages/jaeger-ui/src/components/SearchTracePage/index.jsx#L87)) are rewritten into functional components.

packages/jaeger-ui/src/types/index.tsx
Zen-cronic: found a usage here: https://github.com/jaegertracing/jaeger-ui/blob/1ee19c24a6b34ad7f1f75a6ec7cc4938de0eefde/packages/jaeger-ui/src/model/path-agnostic-decorations/index.tsx#L36

packages/jaeger-ui/src/components/TraceDiff/url.tsx
```diff
@@ -24,16 +24,15 @@ export type TDiffRouteParams = {
   b?: string | undefined;
 };
 
-export const ROUTE_PATH = prefixUrl('/trace/:a?\\.\\.\\.:b?');
-
-const ROUTE_MATCHER = { path: ROUTE_PATH, strict: true, exact: true };
+export const ROUTE_PATH = prefixUrl('/trace-diff/:a?/:b?');
```
Zen-cronic, line 27: since regex route matching is dropped in rrd [v6](https://reactrouter.com/en/6.14.0/start/faq#what-happened-to-regexp-routes-paths), a different approach is implemented.

`a` & `b` traces are optional parameter. They're prefixed with `/trace-diff` instead of `/trace` b/c `/trace/:id` is the pattern for [TracePage](https://github.com/jaegertracing/jaeger-ui/blob/1ee19c24a6b34ad7f1f75a6ec7cc4938de0eefde/packages/jaeger-ui/src/components/TracePage/url/index.tsx#L21)

packages/jaeger-ui/src/components/App/Page.tsx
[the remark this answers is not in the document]
Zen-cronic: what do you think? @yurishkuro  
Even with `Outlet` implemented here, both the ui and the unit tests work like before - so no breaking changes. But we can revert it as well.

package-lock.json
```diff
@@ -2208,7 +2204,6 @@
       "resolved": "https://registry.npmjs.org/@ctrl/tinycolor/-/tinycolor-3.6.1.tgz",
       "integrity": "sha512-SITSV6aIXsuVNV3f3O0f2n/cgyEDWoSqtZMYiAmcsYHydcKrOz3gUxB/iXd/Qf08+IZX4KpgNbvUdMBmWz+kcA==",
       "license": "MIT",
-      "peer": true,
```
Zen-cronic, line 2211: need be reverted?

package-lock.json
yurishkuro: It is 8.48 in the lock file though:
```
$ grep sentry/browser package-lock.json
    "node_modules/@sentry/browser": {
      "resolved": "https://registry.npmjs.org/@sentry/browser/-/browser-8.48.0.tgz",
        "@sentry/browser": "^8.48.0",
```

yurishkuro (commented):
the Deep Dependency Graph on the Search page stopped working

package-lock.json
Zen-cronic: in the original repo, the current versions in the lock file is `8.48.0` as well. 
https://github.com/jaegertracing/jaeger-ui/blob/1ee19c24a6b34ad7f1f75a6ec7cc4938de0eefde/package-lock.json#L4121

so i updated the version in `package.json` to match that

packages/jaeger-ui/src/components/App/Page.tsx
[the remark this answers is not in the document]
yurishkuro: let's keep it

Zen-cronic:
> the Deep Dependency Graph on the Search page stopped working

it might involve converting a few class-based components into functional ones; i'll keep exploring solutions

yurishkuro:
If we have to, can we do it in a separate PR and merge it before this one?

Zen-cronic:
the previous [commit](https://github.com/jaegertracing/jaeger-ui/pull/2580/commits/b0ac58a9eccd962e73f67a5fa92e10eeb1203193) fixes the error of DDG not being displayed (due to a missing `history` object for navigation in the class components). 

No components were rewritten into function components in this PR, but in [this](https://github.com/jaegertracing/jaeger-ui/pull/2588). The rewriting process would block the migration process b/c to use the rrdv6/7 hooks (i.e.g, `useLocation`, `useNavigation`) the components must be functional. So, a different workaround is used, explained below:

- before, the class components (e.g., `SearchTracePage`, `SearchResults`, `TracesDdgImpl`) uses the redux state to keep track of the url params and the `location` object to check the current url.  
- However, after upgrading to rrdv6, the `location` object is not in sync with the url in the redux state. 
- So, i passed existing `urlQueryParams` props and/or a custom prop for the search url to the children components. This resulted in the `SearchResults` component correctly rendering the DDG:

error-extra-url-query-params
 
But there are a few errors (so a few tests are failing):

1. Clicking on a node in the scatter plot should navigate to `/trace/:id`. The url does change but the respective component is not. That's because the `TraceTimelineViewer` (& related components) are reading from the redux state which isn't updated. 

error-scatter-plot-node

2. Using the Find search bar while on the DDG page erases the existing url params and removes the graph. This is because the [location object](https://github.com/jaegertracing/jaeger-ui/blob/1ee19c24a6b34ad7f1f75a6ec7cc4938de0eefde/packages/jaeger-ui/src/components/common/UiFindInput.tsx#L113) that the search bar component reads from is not aware of the other url params:
https://github.com/jaegertracing/jaeger-ui/blob/1ee19c24a6b34ad7f1f75a6ec7cc4938de0eefde/packages/jaeger-ui/src/components/common/UiFindInput.tsx#L113


error-uifind

3. Clicking on the `Find Traces` form by default should set the `view` params to `traces`, not `ddg`:

error-find-traces-should-traces

packages/jaeger-ui/src/components/App/index.jsx
```diff
@@ -89,47 +89,21 @@ export default class JaegerUIApp extends Component {
       <ConfigProvider theme={jaegerTheme}>
         <Provider store={store}>
           <HistoryProvider history={history}>
```
Zen-cronic, line 91: update: `HistoryProvider` is still needed by the router b/c `history` is used for navigation:
https://github.com/jaegertracing/jaeger-ui/blob/1ee19c24a6b34ad7f1f75a6ec7cc4938de0eefde/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.tsx#L116

`useNavigate` can't be used inside a class-based component.

packages/jaeger-ui/src/components/DeepDependencies/traces.tsx
```diff
@@ -36,7 +36,9 @@ const svcOp = memoizeOne((service, operation) => ({ service, operation }));
 
 // export for tests
 export function mapStateToProps(state: ReduxState, ownProps: TOwnProps): TReduxProps {
-  const urlState = getUrlState(ownProps.location.search);
+  const locationUrlState = getUrlState(ownProps.location.search);
+  const urlState = { ...locationUrlState, ...ownProps.urlQueryParams };
```
Zen-cronic, line 40: merging with the passed `urlQueryParams` b/c `location` and the redux state router are out of sync

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/__snapshots__/DiffSelection.test.js.snap
```diff
@@ -93,7 +93,7 @@ exports[`DiffSelection renders multiple traces as expected 1`] = `
     className="DiffSelection--message"
   >
     <Link
-      to="/trace/trace-id-0...trace-id-1?cohort=trace-id-0&cohort=trace-id-1"
+      to="/trace-diff/trace-id-0/trace-id-1?cohort=trace-id-0&cohort=trace-id-1"
```
Zen-cronic, line 96: TODO: the second `cohort` param should be separated by an `&`

packages/jaeger-ui/src/components/DeepDependencies/traces.tsx
```diff
@@ -63,15 +65,19 @@ export function mapStateToProps(state: ReduxState, ownProps: TOwnProps): TReduxP
...
 };
 
 // export for tests
 export class TracesDdgImpl extends React.PureComponent<TOwnProps & TReduxProps> {
   render(): React.ReactNode {
-    const { location } = this.props;
-    const urlArgs = queryString.parse(location.search);
+    const { navigateTo } = this.props;
+
+    const urlArgs = queryString.parse(navigateTo.split('/search?')[1]);
```
Zen-cronic, line 77: TODO fix:
parsing the manually passed url prop (that contains the correct params) like this is very hacky.

Zen-cronic:
Blocked by #2588

Zen-cronic:
more [issue](https://github.com/salvoravida/redux-first-history/issues/117) of `useLocation` and `location` not in sync

PR #2586, opened 2025-01-11
[ci]: Add workflow to guard against growing bundle size

### LINKED ISSUE #2581: Add CI workflow to guard against increases in the bundle size
Using GH actions like `action/cache` we can store an artifact for each commit on `main` that will record the current size of the UI assets bundle. We want to have a check in the CI that will fail if the bundle size goes up by more than a certain percentage (say 2% to start with, we can adjust later).

hari45678:
- fixes #2581 

## Description of the changes
This GH Workflow would go through the following steps:

1. Calculate the actual bundle size by using `build` directory generated when we run `npm run build`
2. Save that size to a `new_bundle_size.txt` file
3. Now fetch the `bundle_size.txt` file from cache, if present.
4. If cache hits, compare the `bundle_size.txt` (contains old size) with the `new_bundle_size.txt` (contains new size).
5. If its more than 2%, fail the CI
6. Else pass it.
7. If cache miss, save the `bundle_size.txt` to cache
8. Finally, if everything's alright, save the current bundle size file to cache. (but only on pushes to main branch)

- Example run: https://github.com/hari45678/jaeger-ui/actions/runs/12724047648/job/35469797958#step:8:27

## Checklist

### DIFF
--- a/.github/workflows/check_bundle.yml
+++ b/.github/workflows/check_bundle.yml
@@ -5,6 +5,11 @@ on:
     
   push:
     branches: [main]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ (github.event.pull_request && github.event.pull_request.number) || github.ref || github.run_id }}
+  cancel-in-progress: true
+
 jobs:
   build-and-check:
     runs-on: ubuntu-latest
@@ -61,7 +66,7 @@ jobs:
           fi
           
       - name: Update bundle size file
-        run: rm -f bundle_size.txt && mv new_bundle_size.txt bundle_size.txt
+        run: mv new_bundle_size.txt bundle_size.txt
         
       - name: Save new bundle size
         if: github.event_name == 'push' && github.ref == 'refs/heads/main'
--- a/.github/workflows/check_bundle.yml
+++ b/.github/workflows/check_bundle.yml
@@ -35,10 +35,10 @@ jobs:
           
       - name: Restore previous bundle size
         id: cache-bundle-size
-        uses: actions/cache@v4
+        uses: actions/cache/restore@v4
         with:
           path: bundle_size.txt
-          key: bundle-size-jaeger-ui
+          key: jaeger-ui-bundle-size
           
       - name: Compare bundle sizes
         if: steps.cache-bundle-size.outputs.cache-hit == 'true'
@@ -66,4 +66,4 @@ jobs:
         uses: actions/cache/save@v4
         with:
           path: bundle_size.txt
-          key: bundle-size-jaeger-ui
+          key: jaeger-ui-bundle-size
--- a/.github/workflows/check_bundle.yml
+++ b/.github/workflows/check_bundle.yml
@@ -15,17 +15,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871 # v4.2.1
-      
-      - name: Read .npmrc
-        id: npmrc
-        run: |
-          NODE_VERSION=$(grep 'node=' .npmrc | cut -d'=' -f2)
-          echo "node_version=$NODE_VERSION" >> $GITHUB_OUTPUT
-          
       - name: Setup Node.js
         uses: actions/setup-node@39370e3970a6d050c480ffad4ff0ed4d3fdee5af # v4.1.0
         with:
-          node-version: ${{ steps.npmrc.outputs.node_version }}
+          node-version-file: '.nvmrc'
           cache: 'npm'
           
       - name: Install dependencies
--- a/.github/workflows/check_bundle.yml
+++ b/.github/workflows/check_bundle.yml
@@ -1,74 +1,71 @@
 name: Bundle Size Check
-
 on:
   pull_request:
     branches: [main]
     
   push:
     branches: [main]
-
 jobs:
   build-and-check:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


hari45678:
Just noticed, we already do a build in Lint and Build workflow. Should I include the bundle_check part in that workflow instead of creating a separate one?

yurishkuro:
I think it's good to have this one separate, because if it fails we can occasionally override it and still merge (when we deem the size change legitimate), we probably won't make this check required.

.github/workflows/check_bundle.yml
```diff
@@ -0,0 +1,74 @@
...
+    
+  push:
+    branches: [main]
+
+jobs:
+  build-and-check:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: packages/jaeger-ui
```
yurishkuro, line 15: Why do we need this? We always build from the root.

.github/workflows/check_bundle.yml
```diff
@@ -0,0 +1,74 @@
...
+
+    steps:
+      - uses: actions/checkout@v4 # v4.2.1
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4 # v4.1.0
+        with:
+          node-version: '20'
```
yurishkuro, line 23: We should really be reading the version from .npmrc, in all workflows.

.github/workflows/check_bundle.yml
```diff
@@ -0,0 +1,74 @@
...
+      - name: Calculate bundle size
+        run: |
+          TOTAL_SIZE=$(du -sb build | cut -f1)
+          echo "$TOTAL_SIZE" > new_bundle_size.txt
+          echo "Total bundle size: $TOTAL_SIZE bytes"
+      - name: Restore previous bundle size
+        id: cache-bundle-size
+        uses: actions/cache@v4
+        with:
+          path: /home/runner/work/jaeger-ui/jaeger-ui/packages/jaeger-ui/bundle_size.txt
```
yurishkuro, line 41: Why do we need to depend on absolute path? Can we read it from some gh var?

.github/workflows/check_bundle.yml
```diff
@@ -0,0 +1,74 @@
...
+          key: bundle-size-jaeger-ui
+
+      - name: Compare bundle sizes
+        run: |
+          if [ -f bundle_size.txt ]; then
```
yurishkuro, line 46: Use if: check on success of the previous step instead

yurishkuro:
It doesn't look like your new workflow ran, perhaps a syntax error

hari45678:
> It doesn't look like your new workflow ran, perhaps a syntax error

It did run https://github.com/jaegertracing/jaeger-ui/actions/runs/12724148042/job/35470007592?pr=2586

I would try fixing as per the comments though

.github/workflows/check_bundle.yml
hari45678: That's because the build folder is located at the `packages/jaeger-ui` location. Steps get some what easy on working with that dir. Though I can make it work with root dir as well

.github/workflows/check_bundle.yml
hari45678: I will add that fix in a separate PR for other workflows (for this, would add in this PR)

.github/workflows/check_bundle.yml
```diff
@@ -0,0 +1,71 @@
...
+          echo "Size change: $PERCENTAGE_CHANGE%"
+          if [ $PERCENTAGE_CHANGE -gt 2 ]; then
+            echo "❌ Bundle size increased by more than 2% ($PERCENTAGE_CHANGE%)"
+            exit 1
+          else
+            echo "✅ Bundle size change is within acceptable range ($PERCENTAGE_CHANGE%)"
+          fi
+          
+      - name: Update bundle size file
+        run: rm -f bundle_size.txt && mv new_bundle_size.txt bundle_size.txt
```
yurishkuro, line 64: just `mv` not enough? It should override the file (unless permissions are bad)

hari45678: uhm yes actually

hari45678:
Looks like some issue. How is it getting a cache hit?

.github/workflows/check_bundle.yml
```diff
@@ -0,0 +1,76 @@
...
+jobs:
+  build-and-check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871 # v4.2.1
+      
+      - name: Read .npmrc
+        id: npmrc
+        run: |
+          NODE_VERSION=$(grep 'node=' .npmrc | cut -d'=' -f2)
```
yurishkuro, line 22: Did you mean to read `.nvmrc`? It contains `20`, but `.npmrc` only has `engine-strict=true`

hari45678: my bad, a typo

hari45678: got a better approach on googling

yurishkuro (approved):
looks great!

hari45678:
There was an issue which I just fixed in the latest commit. Actually `[actions/cache](https://github.com/actions/cache/blob/main/restore/README.md#restore-action)` restores the cache but also saves it as a post action. `action/cache/restore` doesn't do that step.

yurishkuro (approved):

PR #2588, opened 2025-01-16
Refactor: Rewrite search page related components into functional comp…

Zen-cronic:
- Part of #2532 
- Supporting PR/PoC for #2580 

## Description of the changes
- Converted several class-based components into functional ones, to enable the use of useHooks from rrd v6

- `npm run test`
- Currently, many failing tests - will address if this concept works

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/DeepDependencies/traces.tsx
+++ b/packages/jaeger-ui/src/components/DeepDependencies/traces.tsx
@@ -30,13 +30,19 @@ import transformTracesToPaths from '../../model/ddg/transformTracesToPaths';
 
 import { TDdgStateEntry } from '../../types/TDdgState';
 import { ReduxState } from '../../types';
+import { useLocation } from 'react-router-dom-v5-compat';
 
 // Required for proper memoization of subsequent function calls
 const svcOp = memoizeOne((service, operation) => ({ service, operation }));
 
 // export for tests
 export function mapStateToProps(state: ReduxState, ownProps: TOwnProps): TReduxProps {
-  const urlState = getUrlState(ownProps.location.search);
+  // ownProps.location.search !== state.router.location.search
+  // Graph displayed only when the state.router.location.search is used, which contains all the params
+  // ownProps.location.search contains only the `?view=dgg` param
+  // const urlState = getUrlState(ownProps.location.search);
+  const urlState = getUrlState(state.router.location.search);
+
   const { density, operation, service, showOp: urlStateShowOp } = urlState;
   const showOp = urlStateShowOp !== undefined ? urlStateShowOp : operation !== undefined;
   let graphState: TDdgStateEntry | undefined;
@@ -65,10 +71,11 @@ type TOwnProps = {
   location: Location;
 };
 
+type TProps = TOwnProps & TReduxProps;
+
 // export for tests
-export class TracesDdgImpl extends React.PureComponent<TOwnProps & TReduxProps> {
-  render(): React.ReactNode {
-    const { location } = this.props;
+export const TracesDdgImpl: React.FC<TProps> = (props) => {
+    const location = useLocation();
     const urlArgs = queryString.parse(location.search);
     const { end, start, limit, lookback, maxDuration, minDuration, view } = urlArgs;
     const extraArgs = { end, start, limit, lookback, maxDuration, minDuration, view };
@@ -77,10 +84,10 @@ export class TracesDdgImpl extends React.PureComponent<TOwnProps & TReduxProps> 
         baseUrl={ROUTE_PATH}
         extraUrlArgs={extraArgs}
         showSvcOpsHeader={false}
-        {...this.props}
+        {...props}
       />
     );
   }
-}
+// }
 
 export default connect(mapStateToProps)(TracesDdgImpl);
--- a/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.tsx
+++ b/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.tsx
@@ -16,8 +16,8 @@
 
 import * as React from 'react';
 import { Select } from 'antd';
-import { History as RouterHistory, Location } from 'history';
-import { Link } from 'react-router-dom';
+import { History as RouterHistory } from 'history';
+import { useNavigate, useLocation, Link, useSearchParams } from 'react-router-dom-v5-compat';
 import queryString from 'query-string';
 
 import AltViewOptions from './AltViewOptions';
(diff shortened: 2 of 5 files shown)


Zen-cronic:
Update: 

This is the default page after clicking "Find Traces" form:
Notice the params in the url.

before-missing-params

The DDG page functions as expected (after clicking on "Deep Dependency Graph").
**But the page url lost its many of its params**:
 
missing-params

This is the expected result, with full params:

working-ddg


I'll explain further in the review section.

packages/jaeger-ui/src/components/DeepDependencies/traces.tsx
```diff
@@ -30,13 +30,19 @@ import transformTracesToPaths from '../../model/ddg/transformTracesToPaths';
...
 const svcOp = memoizeOne((service, operation) => ({ service, operation }));
 
 // export for tests
 export function mapStateToProps(state: ReduxState, ownProps: TOwnProps): TReduxProps {
-  const urlState = getUrlState(ownProps.location.search);
+  // ownProps.location.search !== state.router.location.search
+  // Graph displayed only when the state.router.location.search is used, which contains all the params
+  // ownProps.location.search contains only the `?view=dgg` param
+  // const urlState = getUrlState(ownProps.location.search);
+  const urlState = getUrlState(state.router.location.search);
```
Zen-cronic, line 44: as said in the inline comments, the ddg is displayed only with the url params stored in the redux state.

packages/jaeger-ui/src/components/DeepDependencies/traces.tsx
```diff
@@ -65,10 +71,11 @@ type TOwnProps = {
...
 };
 
+type TProps = TOwnProps & TReduxProps;
+
 // export for tests
-export class TracesDdgImpl extends React.PureComponent<TOwnProps & TReduxProps> {
-  render(): React.ReactNode {
-    const { location } = this.props;
+export const TracesDdgImpl: React.FC<TProps> = (props) => {
+    const location = useLocation();
```
Zen-cronic, line 78: i expected `useLocation` to provide us with the same value as `state.router.location` due to the use of `redux-first-history` in the store:
https://github.com/jaegertracing/jaeger-ui/blob/1ee19c24a6b34ad7f1f75a6ec7cc4938de0eefde/packages/jaeger-ui/src/utils/configure-store.js#L55

that's what the library is for - to sync the `history` object shared by `react-redux` and rrd. From the [docs](https://github.com/salvoravida/redux-first-history?tab=readme-ov-file#main-goal), in rrdv6, we should get `useLocation() === state.router.location`, but that's not the case here.

Zen-cronic: our current implementation stores the url (the params) in the redux store, and related components subscribe to the store and update it via `mapStateToProps`. E.g.,
https://github.com/jaegertracing/jaeger-ui/blob/1ee19c24a6b34ad7f1f75a6ec7cc4938de0eefde/packages/jaeger-ui/src/components/DeepDependencies/traces.tsx#L38

But as said above, the `useLocation` hook doesn't provide us with the synchronized `location` object. That's probably why the url doesn't get updated correctly:
https://github.com/jaegertracing/jaeger-ui/blob/1ee19c24a6b34ad7f1f75a6ec7cc4938de0eefde/packages/jaeger-ui/src/components/DeepDependencies/traces.tsx#L39

Zen-cronic: folks have been facing similar [issues](https://github.com/salvoravida/redux-first-history/issues/95).

we shouldn't rely anymore on the redux state anymore for navigating url routes and updating query params, and instead use rrdv5-compat and rrdv6's useHooks.

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.tsx
```diff
@@ -96,153 +94,156 @@ export function createBlob(rawTraces: TraceData[]) {
...
-    const { location, history } = this.props;
+  const onDdgViewClicked = () => {
     const urlState = queryString.parse(location.search);
     const view = urlState.view && urlState.view === 'ddg' ? EAltViewActions.Traces : EAltViewActions.Ddg;
     trackAltView(view);
-    history.push(getUrl({ ...urlState, view }));
-  };
+    // Navigated to `/search?view=ddg`
+    // should be the full `/search?end=1736992218835000&limit=20&lookback=1h&maxDuration&minDuration&service=frontend&start=1736988618835000&view=ddg` or &view=traces
+    navigate(getUrl({ ...urlState, view }));
```
Zen-cronic, line 133: the discrepancy between `location` provided by `useLocation` and what's in the redux state. 
This is what's making the ddg displayed correctly under the wrong url (missing params).

hari45678:
@Zen-cronic I am working on removing `redux-form` dependency from past few days. We would eventually have to remove `history` package as rrd v7 handles that internally. And as a result, `redux-first-history` would need to be removed. We can rewrite the class components to functional components gradually, one by one, otherwise some errors might stay uncaught by the automated tests. I would volunteer by your side in rewriting these components.

For this PR, my recommendation would be to go for one chain of related changes only and testing that change manually.

Zen-cronic:
thanks for your suggestion, hari. 

my goal is to unblock #2580 by following this official v5-v6 migration [guide](https://github.com/remix-run/react-router/discussions/8753).

from now on, here's my process:

1. Rewrite a component if it meets this criteria:
- if it reads `location` from the redux state. e.g. [SearchTracePage](https://github.com/jaegertracing/jaeger-ui/blob/1ee19c24a6b34ad7f1f75a6ec7cc4938de0eefde/packages/jaeger-ui/src/components/SearchTracePage/index.jsx#L267) and/or
- if it updates the query params to the redux state e.g. [UiFindInput](https://github.com/jaegertracing/jaeger-ui/blob/1ee19c24a6b34ad7f1f75a6ec7cc4938de0eefde/packages/jaeger-ui/src/components/common/UiFindInput.tsx#L115)

2. Replace `history`, `location`, and query params in the redux state with useHooks b/c routing should be decoupled from the store

3. repeat this process for another set of related components

packages/jaeger-ui/src/components/DeepDependencies/traces.tsx
yurishkuro: sounds like a regression. Jaeger UI links are supposed to be shareable and capture the complete state.

yurishkuro:
not sure what the state of this, it shows 0 file changes

PR #2591, opened 2025-01-16
[CI] Fix cache resolution and syntax in check binary size workflow 

### LINKED ISSUE #2590: [CI]: Improper cache resolution and conditional statement handling in check bundle size workflow .
### What happened?

Looking at CI runs for multiple for check bundle size workflow . There are cache save fails on due to cache being already saved with the same cache key . This is not a acceptable behaviour as the cache should stay up to date with the current bundle size . 

### Steps to reproduce

Logs of CI runs . 

### Expected behavior

-> Ensure proper cache save mechanism without overwriting . 
-> Ensure proper cache retrieval during workflow runs .
-> Ensure proper syntax for conditional statements .

### Relevant log output

```shell

```

### Screenshot

_No response_

### Additional context

_No response_

### Jaeger backend version

_No response_

### SDK

_No response_

### Pipeline

_No response_

### Stogage backend

_No response_

### Operating system

_No response_

### Deployment model

_No response_

### Deployment configs

```shell

```

chahatsagarmain:
- Resolves #2590

## Description of the changes
- 

- 

## Checklist

### DIFF
--- a/.github/workflows/check_bundle.yml
+++ b/.github/workflows/check_bundle.yml
@@ -39,9 +39,11 @@ jobs:
         with:
           path: bundle_size.txt
           key: jaeger-ui-bundle-size
+          restore-keys: |
+            jaeger-ui-bundle-size
           
       - name: Compare bundle sizes
-        if: steps.cache-bundle-size.outputs.cache-hit == 'true'
+        if: steps.cache-bundle-size.outputs.cache-matched-key != ''
         run: |
           OLD_BUNDLE_SIZE=$(cat bundle_size.txt)
           NEW_BUNDLE_SIZE=$(cat new_bundle_size.txt)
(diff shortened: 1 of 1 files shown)


yurishkuro (approved):

hari45678:
@yurishkuro This PR is creating new cache key on each commit to the main branch. That would result in cache miss on subsequent PRs.

yurishkuro:
it does create new key, but it uses a pattern-match to find them which picks the latest. Using a fixed key for some reason means cache will not be updated after the first value.

hari45678:
Yes, but it isn't picking up the latest cache key in this (jaeger-ui) repository, maybe because here exists an exact key match, so it is picking up that stale cache.
For example, in bundle check workflow in this PR https://github.com/jaegertracing/jaeger-ui/pull/2593, the [action](https://github.com/jaegertracing/jaeger-ui/actions/runs/12826517038/job/35766581059?pr=2593#step:7:14) is picking up the key `jaeger-ui-bundle-size`, and not the latest key `jaeger-ui-bundle-size-12822565240` which was created with bundle check [action](https://github.com/jaegertracing/jaeger-ui/actions/runs/12822565240/job/35755694309#step:10:10) trigger with latest push to main branch after this PR was merged.

yurishkuro:
Then we still don't have a solution

chahatsagarmain:
@yurishkuro we need to delete the old exact key match , from action -> caches .

hari45678:
> @yurishkuro we need to delete the old exact key match , from action -> caches .

yes that should solve this issue

yurishkuro:
deleted

Issue #2576, opened 2025-01-09
Upgrade antd and icons


yurishkuro:
Automatic upgrades are failing
  * https://github.com/jaegertracing/jaeger-ui/pull/2516
  * https://github.com/jaegertracing/jaeger-ui/pull/2518

There are probably some minor API changes. But since this affects the visualization, a fix should include before/after screenshots of the changed components, to verify continuity.

yurishkuro:
#2579 reverted in #2587

coderboy-yash:
i want to work on this issue, I have some clarifying questions:-
1. I have to update the antd dependency to v 5.23.1 ?
2. I have to update the react-icons dependency to v 5.4.0 ?
3. I am right or there is something I am required to do ?

PR #2595, opened 2025-01-18
Updated react-icons and antd version

coderboy-yash:
- #2576 
## Description of the changes
- updated dependency of react-icons to v5.4.0
- updated dependency of antd to v5.23.1

- npm run test
- npm run start 

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,5 +65,10 @@
     "hooks": {
       "pre-commit": "npm-run-all -ln --parallel lint test"
     }
+  },
+  "dependencies": {
+    "@ctrl/tinycolor": "^4.1.0",
+    "antd": "^5.23.1",
+    "react-icons": "^5.4.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "eslint-plugin-jsx-a11y": "^6.8.0",
     "eslint-plugin-react": "7.37.1",
     "husky": "9.1.0",
-    "jsdom": "25.0.1",
+    "jsdom": "^26.0.0",
     "npm-run-all2": "7.0.1",
     "prettier": "3.4.2",
     "rxjs-compat": "6.6.7",
--- a/packages/jaeger-ui/src/components/common/LoadingIndicator.tsx
+++ b/packages/jaeger-ui/src/components/common/LoadingIndicator.tsx
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 import React from 'react';
-import { LuLoader2 } from 'react-icons/lu';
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -22,7 +22,7 @@
     "eslint-plugin-jsx-a11y": "^6.8.0",
     "eslint-plugin-react": "7.37.1",
     "husky": "9.1.0",
-    "jsdom": "25.0.1",
+    "jsdom": "^26.0.0",
```
yurishkuro, line 25: how is this related to ant/icons upgrade?

package.json
```diff
@@ -65,5 +65,10 @@
     "hooks": {
       "pre-commit": "npm-run-all -ln --parallel lint test"
     }
+  },
+  "dependencies": {
```
yurishkuro, line 69: what is the reason for adding these as top level dependencies?

packages/jaeger-ui/src/components/common/LoadingIndicator.tsx
```diff
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 import React from 'react';
-import { LuLoader2 } from 'react-icons/lu';
+import { LuLoader } from 'react-icons/lu';
```
yurishkuro, line 16: this is icons change, why is it bundled with antd upgrade?

yurishkuro (changes requested):
for any icons changes include before & after screenshots

package.json
coderboy-yash: > how is this related to ant/icons upgrade?

no this is not related to ant/icons update , actually I made a pr to the issue #2584 previous to this issue and after that I made this pr so those changes also get included to this pr .
My mistake:- I made changes in the same code for every pr.
I am removing this mistake and will correct this pr soon.

package.json
coderboy-yash: these are added by miskake at the top level , these had to be added in the jaeger UI of packages folder.

packages/jaeger-ui/src/components/common/LoadingIndicator.tsx
coderboy-yash: Previously I merged the changes of antd and react-icons together but after reading your comment I thought you want to do it separately now I have made another pr for this, which handles only react-icons version update.
#2603 you can check this I have resolved all the errors that you mentioned here.
Please let me know if you want to make any other changes to this.

yurishkuro:
both upgrades were done already. Thanks for the help anyway.

Issue #2596, opened 2025-01-18
[Bug]: Support for defaultProps in Function Components to be Removed


its-me-abhishek:
### What happened?

The following warning appears in the console while running the application:
`Support for defaultProps will be removed from function components in a future major release. Use JavaScript default parameters instead.`

This warning points to the `LoadingIndicator` component here (and exists in other components as well), which currently relies on `defaultProps` for setting default values. As this feature will be deprecated in upcoming React versions, the codebase should transition to using default parameter values in function component definitions.

### Steps to reproduce

1. Run the application using the default settings.
2. Navigate to the link to the UI in browser
3. Observe the console for the warning.

### Expected behavior

The defaultProps should be removed from the codebase and migration should be done similar to [this](https://sophiabits.com/blog/stop-using-defaultprops) as discussed previously on Slack

### Relevant log output

```shell
Support for defaultProps will be removed from function components in a future major release. Use JavaScript default parameters instead.
```

### Components to be updated

* [x] LoadingIndicator
* [x] BreakableText
* [x] ErrorMessage
* [x] NewWindowIcon
* [ ] SearchForm
* [x] ScatterPlot
* [x] TraceHeader
* [x] Ticks
* [x] TimelineRow
* [x] AccordianKeyValues
* [x] AccordianLogs
* [x] AccordianText
* [x] KeyValuesTable
* [x] MiniMap

ADI-ROXX:
Working on it

yurishkuro:
@its-me-abhishek could you enumerate the scope of this issue? How many components need to be updated?

its-me-abhishek:
hey, 13 components are yet to be updated:

![Image](https://github.com/user-attachments/assets/10701e94-5fa2-4999-ad56-2921e9e98959)

yurishkuro:
@its-me-abhishek I suggest putting them in the description as a checkbox list (using `* [ ] name` syntax), so that we can track progress.

its-me-abhishek:
Done! will try to fix them in a similar sequence asap

its-me-abhishek:
@yurishkuro, `SearchForm.jsx` contains defaultProps, but destructuring with an interface (with Typescript) is what needs to be done to remove defaultProps and propTypes. Shall we convert it to a .tsx? [Reference](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#removed-proptypes-and-defaultprops)

[...]

PR #2601, opened 2025-01-20
[LoadingIndicator]: Replace defaultProps with destructuring

its-me-abhishek:
- Part of #2596

## Description of the changes
- Updated the LoadingIndicator component to remove defaultProps so as to avoid the deprecation warnings.
- Updated the snapshots so as to match the new LoadingIndicator.

- Running `npm ci` `npm run update-snapshots` and `npm test` passes all the tests
- Tested manually by rendering LoadingIndicator using various props

## Checklist


packages/jaeger-ui/src/components/common/LoadingIndicator.tsx
```diff
@@ -25,8 +25,13 @@ type LoadingIndicatorProps = {
   style?: React.CSSProperties;
 };
 
-export default function LoadingIndicator(props: LoadingIndicatorProps) {
-  const { centered, vcentered, className, small, ...rest } = props;
+export default function LoadingIndicator({
+  centered = false,
+  vcentered,
```
yurishkuro, line 30: is this supposed to be `= false`? Why declare it here otherwise?

packages/jaeger-ui/src/components/common/LoadingIndicator.tsx
```diff
@@ -25,8 +25,13 @@ type LoadingIndicatorProps = {
...
+  className = undefined,
+  small = false,
+  ...rest
+}: LoadingIndicatorProps) {
```
yurishkuro, line 34: is `LoadingIndicatorProps` type needed at all anymore? If you add `style` attribute there's no more `...rest` that needs the type declaration.

packages/jaeger-ui/src/components/common/LoadingIndicator.tsx
its-me-abhishek: it is the destructurised replacement default props, thus, it would be false until not assigned any value, just like the defaultprops

packages/jaeger-ui/src/components/common/LoadingIndicator.tsx
its-me-abhishek: this can be removed, I had kept this just in case sometime in the future we decide to add more values on LoadingIndicatorProps. Can remove if you agree to do so.

yurishkuro: but what's the point of duplicating information? Having everything in one place in the signature of the function is self-documenting, without needing to loop up a type definition elsewhere which does not have any other use. Let's remove it.

its-me-abhishek: Just removed the props, could not remove `...rest` as it is being used by the `data-testid` being passed on by TraceName, or shall we introduce a new field for the same?
this is my implementation so far: 
```

export default function LoadingIndicator({
  centered = false,
  vcentered,
  className = undefined,
  small = false,
  style,
  ...rest
}: {
  centered?: boolean;
  vcentered?: boolean;
  className?: string;
  small?: boolean;
  style?: React.CSSProperties;
}) {
  const cls = `
    LoadingIndicator
    ${centered ? 'is-centered' : ''}
    ${vcentered ? 'is-vcentered' : ''}
    ${small ? 'is-small' : ''}
    ${className || ''}
  `;

  return <LuLoader2 className={cls}{...rest} style={style} />;
}
```

packages/jaeger-ui/src/components/common/LoadingIndicator.tsx
yurishkuro: but why `centered = false` then? I like consistency. Plus, since you remove Props type, `vcentered` now has no type.

packages/jaeger-ui/src/components/common/LoadingIndicator.tsx
```diff
@@ -17,28 +17,27 @@ import { LuLoader2 } from 'react-icons/lu';
...
-type LoadingIndicatorProps = {
+export default function LoadingIndicator({
+  centered = false,
+  vcentered,
+  className = undefined,
+  small = false,
+  style,
+  ...rest
+}: {
   centered?: boolean;
```
yurishkuro, line 28: is it really optional if it has a default value?

its-me-abhishek: I tried not to change the previously implemented props, though, even I had though that these should have been non-optional, changed these in my latest commit
- removed optional feat from centered, classname, and style
- since classname is not an empty string, and not undefined, removed the check `    ${className || ''}` in class-formation for loading indicator

yurishkuro (dismissed):

yurishkuro:
make sure `npm run lint` is successful on your machine

its-me-abhishek:
this linting issue 
```
Type '{ className: string; }' is missing the following properties from type '{ centered: boolean; vcentered?: boolean | undefined; className: string; small: boolean; style?: CSSProperties | undefined; }': centered, smallts(2739)
```
is there in all the components, either i will have to explicitly mention all the non optional props that are not passed yet, from all the components that use loading indicator, or make these optional in the loading indicator itself.
what shall i proceed with?

yurishkuro:
maybe make `centered, small` optional again?

its-me-abhishek:
Done, the lint tests pass now

yurishkuro (approved):

yurishkuro (approved):

yurishkuro:
Thanks!

PR #2603, opened 2025-01-21
updated react-icons

coderboy-yash:
#2518 of #2576 

## Description of the changes
- I have installed react-icons version "5.4.0" 
- after installation this error was shown in the console
image
after going through the react-icons website I found that there is no LuLoader2 component in react-icons version"5.4.0" and some more people also mentioned the issue with LuLoader2 component.
so I have replace the LuLoader2 component with LuLoader component of react-icons.
### loader icon before changes 
image
### loader icon after changes
image
- npm run test
- npm run start

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/common/LoadingIndicator.tsx
+++ b/packages/jaeger-ui/src/components/common/LoadingIndicator.tsx
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 import React from 'react';
-import { LuLoader2 } from 'react-icons/lu';
+import { LuLoader } from 'react-icons/lu';
 
 import './LoadingIndicator.css';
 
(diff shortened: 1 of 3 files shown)


yurishkuro (commented):
lock file has many unrelated changes

packages/jaeger-ui/src/components/common/LoadingIndicator.tsx
```diff
@@ -34,7 +34,7 @@ export default function LoadingIndicator(props: LoadingIndicatorProps) {
     ${small ? 'is-small' : ''}
     ${className || ''}
   `;
-  return <LuLoader2 className={cls} {...rest} />;
+  return <LuLoader className={cls} {...rest} />;
```
yurishkuro, line 37: why? There is LuLoaderCircle which looks more similar to the previous one.

yurishkuro:
addressed in #2518. Thanks for investigation that made the fix easy.

PR #1627, opened 2023-08-01
Add tracing to jaeger-ui package

psk001:
## Which problem is this PR solving?
Resolves [issue 2307](https://github.com/jaegertracing/jaeger-ui/issues/1307) 

## Short description of the changes
Adds tracing module, which handles trace management and connection to jaeger all-in-one

PR #2608, opened 2025-01-22
Add UI tracing

### LINKED ISSUE #1307: [Fun task]: Add React tracing
### Requirement

Understand and visualize the behavior of Jaeger UI via traces

### Problem

Currently we only start traces at the query service.

### Proposal

Perhaps follow suggestions in the blog?

https://developers.redhat.com/articles/2023/03/22/how-enable-opentelemetry-traces-react-applications

### Open questions

How can we send trace data from the browser to Jaeger backend? Should the query-service implement a dedicated endpoint where frontend can report traces?

MAX-786:
Resolves #1307 

## Description of the changes
- Lets you enable tracing the 'jaeger-ui' itself.

### Additional step to fix CORS error (for now):

You have to run docker image of jaeger-all-in-one (if that's what you are using) with following options:
```bash
 docker run -d --rm --name jaeger   -p 16686:16686   -p 4318:4318   -e COLLECTOR_OTLP_ENABLED=true   -e COLLECTOR_OTLP_HTTP_CORS_ALLOWED_ORIGINS="http://localhost:5173"   -e COLLECTOR_OTLP_HTTP_CORS_ALLOWED_HEADERS="Content-Type"   jaegertracing/all-in-one:latest
 ```

## Demo

1. Run your collector. 
For example, we can use `jaeger-all-in-one` docker image which includes `jaeger-collector` and I am setting explicitly, allowed CORS origins/headers (see the above bash command).
2. Run Jaeger UI. 
You can send traces from jaeger UI to your collector via config. You can checkout more here. Like for sending UI traces to jaeger-collector you can use following config:
```json
"tracing": {
      "endpoint": "http://localhost:4138/v1/traces",
      "serviceName": "jaeger-ui"
  }
```

By default it is disabled and it's default config are as follows:

```js
// ... rest config  
tracing: {
      endpoint: '', // Disabled by default
      serviceName: 'jaeger-ui',
  },
```
Demo below, click here and there and see the traces show up in jaeger-ui.

https://github.com/user-attachments/assets/8403cfd4-653a-4bd8-9663-da9f3222642d

- still testing

## Checklist


MAX-786:
Hey @yurishkuro , 
sorry :) for coming back to you so late (got stuck in academic stuff) but here i am and back to the point, I opened a draft PR and   as you guessed we require additional CORS settings for now. 
And I am also looking into a way to turn on/off tracing jaeger-ui.

packages/jaeger-ui/tsconfig.json
```diff
@@ -13,5 +13,5 @@
 // limitations under the License.
 {
   "extends": "../../tsconfig",
-  "include": ["src/**/*.tsx", "typings"]
+  "include": ["src/**/*.tsx", "typings", "src/opentelemetry.ts"]
```
yurishkuro, line 16: why not?
```suggestion
  "include": ["src/**/*.tsx", "typings", "src/**/*.ts"]
```

packages/jaeger-ui/src/opentelemetry.ts
```diff
@@ -0,0 +1,56 @@
...
+import { FetchInstrumentation } from '@opentelemetry/instrumentation-fetch';
+import { UserInteractionInstrumentation } from '@opentelemetry/instrumentation-user-interaction';
+import { registerInstrumentations } from '@opentelemetry/instrumentation';
+import { ZoneContextManager } from '@opentelemetry/context-zone';
+import { Resource } from '@opentelemetry/resources';
+import { SemanticResourceAttributes } from '@opentelemetry/semantic-conventions';
+
+try {
+  const exporter = new OTLPTraceExporter({
+    url: 'http://localhost:4318/v1/traces', // why relative URL doesn't work? '/v1/traces' ???
```
yurishkuro, line 28: this doesn't address my main concerns in #1627. Why `localhost`? If I deploy Jaeger in production and access as `jaeger.mycompany.com`, what does `localhost` have to do with the destination for traces?

MAX-786: I am still getting into OTLP, and as you see i tried with multiple ways to make it work with `relative` path and then proxying it where ever jaeger exits, i was configuring proxy in ```vite.config.mts``` something like :
```js
server: {
  proxy: {

    '/v1/traces': {
      target: 'http://localhost:4318',  // OTLP endpoint
      secure: false,
      changeOrigin: true, 
    },
  },
},
```

But then in `opentelemetry.ts` it gets stuck at exporting which results in blocking react to render anything... I might be doing something wrong perhaps. 
I am still thinking what you mean here ([comment](https://github.com/jaegertracing/jaeger-ui/pull/1627#discussion_r1306574068)). When you said we need UI config option, means like there should something in UI itself or in the codebase like above we got `vite.config.tmts` where operator can configure .
hmmm let me look deeper ig or anything you want to add ?

yurishkuro: Jaeger UI is historically served by `jaeger-query` binary that is mostly a read-only service, but it does have an internal tracing that could be configured via OTEL env vars to send the data to another service (by default it sends to localhost which happens to work when running in all-in-one mode). When the UI part involves its own tracing, it needs to export the data. The question is where.

Option 1: to an external URL, e.g. a `jaeger-collector` running somewhere. This was the approach in the previous PR where the endpoint was configurable and passed to the UI, except that it wasn't passed using the right mechanism https://github.com/jaegertracing/jaeger-ui/pull/1627#discussion_r1280711610. This will work as long as the collector is set up with a correct CORS policy.

Option 2: UI can export data back into `query-service` via a dedicated endpoint, e.g. we could mount an OTLP receiver in the `query-service` as well. This avoids any cross-site issues but requires a lot more implementation to support this endpoint and then somehow export the data elsewhere.

I would recommend going with option 1. It just means we need to have a configuration option in the UI config where the user can specify the OTLP endpoint to export the data.

I am adding it to the issue description.

MAX-786:
>  This was the approach in the previous PR where the endpoint was configurable and passed to the UI, except that it wasn't passed using the right mechanism

Looked for standard way of using configs and also implemented type for tracing config.
Now, what operators need to do is just give endpoint of where they want to send traces via configs (can set serviceName optionally ) otherwise it's disable by default. 
Is this the way you were looking for? as option 1

packages/jaeger-ui/package.json
```diff
@@ -88,7 +88,16 @@
...
+    "u-basscss": "2.0.1",
+    "@opentelemetry/context-zone": "^1.30.1",
+    "@opentelemetry/exporter-trace-otlp-http": "^0.57.1",
+    "@opentelemetry/instrumentation-document-load": "^0.44.0",
+    "@opentelemetry/instrumentation-user-interaction": "^0.44.0",
+    "@opentelemetry/sdk-trace-web": "^1.30.1",
+    "@opentelemetry/instrumentation": "^0.57.1",
+    "@opentelemetry/resources": "^1.30.1",
+    "@opentelemetry/sdk-trace-base": "^1.30.1",
+    "@opentelemetry/semantic-conventions": "^1.28.0"
```
yurishkuro, line 100: list should be sorted and in the right position

packages/jaeger-ui/src/types/config.tsx
```diff
@@ -87,6 +87,14 @@ export type StorageCapabilities = {
   archiveStorage?: boolean;
 };
 
+export type TracingConfig = {
+  // tracing jaeger-ui configuration
+  otlp?: {
```
yurishkuro, line 92: overkill, can skip this nesting

yurishkuro (commented):
lgtm. Two open questions:
 * did you test it? please document the steps in the description and include screenshots of captured traces
 * how we can test this automatically? for example, we could run `jaeger` in a container and the UI separately via `npm run start`, load some UI pages and validate that traces are stored in  `jaeger`. But I am not sure how to "load some UI pages", do we need to run some headless browser with JS support?

MAX-786:
I'll try to answer your questions with best of my knowledge so..
> - did you test it? please document the steps in the description and include screenshots of captured traces

Yes, I should've done it earlier my mistake.

> - how we can test this automatically? for example, we could run jaeger in a container and the UI separately via npm run start, load some UI pages and validate that traces are stored in jaeger

hmm... yes that's exactly what i shown in above demo but..

>  I am not sure how to "load some UI pages", do we need to run some headless browser with JS support?

one thing we can do is click on certain links and buttons and see if it's exporting traces to collector or not? and for loading some UI pages part, wouldn't rendering components would trigger documentLoad traces? 

Btw in the last commit i set a check, so the file is not loaded if tracing is disabled.

yurishkuro:
> one thing we can do is click on certain links and buttons and see if it's exporting traces to collector or not? and for loading some UI pages part, wouldn't rendering components would trigger documentLoad traces?

we can test manually, my question was about automated testing during CI. For example, what if something changes in OTEL about how instrumentation is supposed to be initialize, like an extra step that we're not doing, how would our CI tests catch that?

Regarding the screencast you uploaded, is it correct that all UI-produced traces only have UI service in them? I was expecting something that actually spans UI and server, given that the server also captures traces.

MAX-786:
> we can test manually, my question was about automated testing during CI. For example, what if something changes in OTEL about how instrumentation is supposed to be initialize, like an extra step that we're not doing, how would our CI tests catch that?

Hmm, I know that unit tests can check if the setup code is correctly structured. Maybe we can write tests that verify if all necessary components are initialized. Like checking if the WebTracerProvider is created, if the exporter is set up with the right URL, and if the span processors and instrumentations are added. We could mock the OpenTelemetry modules and see if the right functions are called. Also, maybe integration tests that start a test collector and see if traces are actually sent when the UI interacts. That way, if OTEL changes something, the tests would fail if our code doesn't adapt.

> Regarding the screencast you uploaded, is it correct that all UI-produced traces only have UI service in them? I was expecting something that actually spans UI and server, given that the server also captures traces.

yes, right now UI service only only track client-side browser activities.

yurishkuro:
> yes, right now UI service only only track client-side browser activities.

can you elaborate why that is? If activities like Click initiate a trace and result in a backend request, the UI should be able to pass the trace context to that request such that the server side spans will be in the same trace. This was my main motivation for this ticket in the first place, otherwise I don't think the UI tracing is particularly useful.

PR #2622, opened 2025-01-24
[TimelineRow]: Replace defaultProps with destructuring

its-me-abhishek:
- Part of #2596

## Description of the changes
- Remove defaultProps from the TimelineRow component, replace with destructuring

- Run `npm run test` `npm run update-snapshots`

## Checklist


packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineRow.tsx
```diff
@@ -16,42 +16,42 @@ import * as React from 'react';
...
-
-function TimelineRowCell(props: ITimelineRowCellProps) {
-  const { children, className = '', width, style, ...rest } = props;
+function TimelineRowCell({
+  children,
+  className = '',
+  width,
+  style = {},
+  onClick,
+  ...rest
```
yurishkuro, line 40: I don't follow this. What can be possible provided via `...rest` if all the fields from the type are already mentioned as arguments? E.g. `<TimelineRowCell data-testid="x">` should be an error because `data-testid` is not part of the type declaration.

its-me-abhishek: `data-testid` does not give error as of now, but does give an error when `...rest` is removed. shall I replace `...rest` with `data-testid`?

yurishkuro: https://github.com/jaegertracing/jaeger-ui/pull/2536#discussion_r1935751824

its-me-abhishek:
I have removed the ...rest from both components now, the issue has been fixed too

yurishkuro (approved):

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineRow.tsx
```diff
@@ -16,31 +16,33 @@ import * as React from 'react';
...
   className?: string;
-};
+}) {
+  return <div className={`flex-row ${className}`}>{children}</div>;
+}
 
-interface ITimelineRowCellProps extends React.HTMLAttributes<HTMLDivElement> {
+function TimelineRowCell({
+  children,
+  className = '',
```
yurishkuro, line 31: I see code coverage complaining about this line. Maybe since it's already declared optional it doesn't need to have a default, and in the code use `${className || ''}`?

its-me-abhishek: makes sense. I have updated it now

yurishkuro (approved):

Issue #2530, opened 2024-12-26
[Deps]: Upgrade react-redux to 9.x


yurishkuro:
This is automated attempt that failed due to breaking changes https://github.com/jaegertracing/jaeger-ui/pull/2527

But overall not that many components use react-redux
```
$ grep -rn react-redux packages/
packages//jaeger-ui/package.json:79:    "react-redux": "^8.1.2",
packages//jaeger-ui/src/components/Monitor/ServicesView/index.tsx:24:import { connect } from 'react-redux';
packages//jaeger-ui/src/components/App/index.jsx:16:import { Provider } from 'react-redux';
packages//jaeger-ui/src/components/App/Page.tsx:19:import { connect } from 'react-redux';
packages//jaeger-ui/src/components/App/TopNav.tsx:19:import { connect } from 'react-redux';
packages//jaeger-ui/src/components/QualityMetrics/index.tsx:17:import { connect } from 'react-redux';
packages//jaeger-ui/src/components/common/UiFindInput.tsx:21:import { connect } from 'react-redux';
packages//jaeger-ui/src/components/DeepDependencies/index.tsx:19:import { connect } from 'react-redux';
packages//jaeger-ui/src/components/DeepDependencies/traces.tsx:20:import { connect } from 'react-redux';
packages//jaeger-ui/src/components/DeepDependencies/SidePanel/DetailsPanel.tsx:18:import { connect } from 'react-redux';
packages//jaeger-ui/src/components/DeepDependencies/Graph/DdgNodeContent/index.tsx:20:import { connect } from 'react-redux';
packages//jaeger-ui/src/components/TraceDiff/TraceDiffGraph/TraceDiffGraph.tsx:18:import { connect } from 'react-redux';
packages//jaeger-ui/src/components/TraceDiff/TraceDiff.tsx:17:import { connect } from 'react-redux';
packages//jaeger-ui/src/components/DependencyGraph/index.jsx:19:import { connect } from 'react-redux';
packages//jaeger-ui/src/components/TracePage/index.tsx:22:import { connect } from 'react-redux';
packages//jaeger-ui/src/components/TracePage/TraceTimelineViewer/index.tsx:16:import { connect } from 'react-redux';
packages//jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx:17:import { connect } from 'react-redux';
[...]

PR #2623, opened 2025-01-24
[deps]: Upgrade redux and react-redux to latest version

hari45678:
- resolves https://github.com/jaegertracing/jaeger-ui/issues/2530

## Description of the changes
- With redux 5, type of an action is now a `string`. According to them, this change is to make the action obj serializable. Change log: https://redux.js.org/usage/migrations/migrating-rtk-2#action-types-must-be-strings
- In the test, due to some import error, redux no longer has configurable properties, and jest requires them to be configurable for spying and changing their behavior. [More context](https://stackoverflow.com/a/72885576/17456074) on this.
- Since we are using `requireActual` in `jest.mock`, it would anyway be the actual implementation of functions for other test cases.

- `npm run start` is working fine

## How was this upgrade carried out?
1. `npm uninstall react-redux -w packages/jaeger-ui` to remove `react-redux`
2. `npm uninstall redux -w packages/jaeger-ui` to remove `redux`
3. `npm install react-redux -w packages/jaeger-ui` to install  react-redux latest version. This installed the latest version but the package was nested in jaeger-ui node modules so had to do step 4.
4. Even with 2nd step, there was an entry of redux v4.2.1 in lockfile and package was installed in root node_modules.
5. I manually removed that entry from lockfile, and again ran `npm install redux -w packages/jaeger-ui`. This worked.
6. Again ran `npm uninstall react-redux -w packages/jaeger-ui` to remove `react-redux` nested entry in lockfile
7. `npm install react-redux -w packages/jaeger-ui` to install  react-redux latest version. This finally created an entry in root node_modules in lockfile.

## Checklist


yurishkuro (approved):
Thanks!

Issue #2610, opened 2025-01-23
[Feature]: Refactor class components to functional components


MAX-786:
### Requirement

Jaeger UI still have a number of core components that are React class-based components. This prevents the use of hooks in them, and newest DX on them.

Refactor all the possible components in Jaeger UI that still are class-based components. Convert them to functional components, whenever possible.

### Problem & Motivation

The idea is also while moving to them, use hooks to abstract the functionality present in Redux and in data fetching so in a near future we can easily switch using the hooks to use other props provider.

### Proposal & Implementation
Refactor should be done carefully and providing tests where they do not exist. The lifecycle of the components should be preserved while watching out for edge cases and UX problems.

## Deliverables
IMO It should be delivered all at once, as a major breaking change. Ideally, we won't change the contract of any of the components, so no new props passed down, no return anything to the tree that we were not returning before, unless it's justified.

If the contract has to be broken, then it must be properly documented and exposed in the upgrade guide.

Here there is a preliminary list as of Jun 7 2025:

```
$ grep -rl 'extends React..*Component' packages/jaeger-ui/src
packages/jaeger-ui/src/utils/DraggableManager/demo/DraggableManagerDemo.tsx
packages/jaeger-ui/src/utils/DraggableManager/demo/DividerDemo.tsx
packages/jaeger-ui/src/utils/DraggableManager/demo/RegionDemo.tsx
packages/jaeger-ui/src/components/Monitor/ServicesView/index.tsx
packages/jaeger-ui/src/components/Monitor/ServicesView/serviceGraph.tsx
packages/jaeger-ui/src/components/Monitor/ServicesView/operationDetailsTable/index.tsx
packages/jaeger-ui/src/components/Monitor/ServicesView/operationDetailsTable/opsGraph.tsx
packages/jaeger-ui/src/components/App/TraceIDSearchInput.tsx
packages/jaeger-ui/src/components/App/Page.tsx
packages/jaeger-ui/src/components/QualityMetrics/index.tsx
packages/jaeger-ui/src/components/QualityMetrics/ScoreCard.tsx
packages/jaeger-ui/src/components/QualityMetrics/CountCard.tsx
packages/jaeger-ui/src/components/QualityMetrics/MetricCard.tsx
packages/jaeger-ui/src/components/QualityMetrics/Header.tsx
packages/jaeger-ui/src/components/QualityMetrics/BannerText.tsx
[...]

PR #2639, opened 2025-01-27
Migrate NameSelector from class based to function based component

AdiIsHappy:
Converted NameSelector in #2610 

## Description of the changes
- Refactored class components to functional components for NameSelector.tsx.
- Used React hooks in the component.
- Updated the tests to work with the new component.

- Tests are modified.

## Checklist


AdiIsHappy:
@yurishkuro I have migrated the component from class to function. But in the test, there are two tests. particularly 
- attempts to focus the filter input when the component updates
- throws Error when attempting to clear when required
which I am not able to convert; currently both are commented, and I am trying to fix them. other tests that required migration have been migrated.

yurishkuro:
Do we have a way of testing this in the UI?

yurishkuro:
stale PR

PR #2643, opened 2025-02-03
Prepare release v1.66.0

jkowall:
Prepare v1.66.0


packages/jaeger-ui/package.json
```diff
@@ -1,132 +1,69 @@
 {
```
yurishkuro, line 1: the only change the package.json should've been a version bump in one line.

PR #2667, opened 2025-02-14
[AccordianKeyValues]: replace defaultProps with destructuring

its-me-abhishek:
- Part of #2596

## Description of the changes
- removed defaultProps from AccordianKeyValues

- npm ci
- npm test

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/AccordianKeyValues.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/AccordianKeyValues.tsx
@@ -23,20 +23,9 @@ import { KeyValuePair, Link } from '../../../../types/trace';
 
 import './AccordianKeyValues.css';
 
-type AccordianKeyValuesProps = {
-  className?: string | TNil;
-  data: KeyValuePair[];
-  highContrast?: boolean;
-  interactive?: boolean;
-  isOpen: boolean;
-  label: string;
-  linksGetter: ((pairs: KeyValuePair[], index: number) => Link[]) | TNil;
-  onToggle?: null | (() => void);
-};
-
 // export for tests
 export function KeyValuesSummary(props: { data?: KeyValuePair[] }) {
-  const { data } = props;
+  const { data } = props || null;
   if (!Array.isArray(data) || !data.length) {
     return null;
   }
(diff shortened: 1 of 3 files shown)


packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/AccordianKeyValues.tsx
```diff
@@ -23,20 +23,9 @@ import { KeyValuePair, Link } from '../../../../types/trace';
...
-  isOpen: boolean;
-  label: string;
-  linksGetter: ((pairs: KeyValuePair[], index: number) => Link[]) | TNil;
-  onToggle?: null | (() => void);
-};
-
 // export for tests
 export function KeyValuesSummary(props: { data?: KeyValuePair[] }) {
-  const { data } = props;
+  const { data } = props || null;
```
yurishkuro, line 28: `props || null` means props or null - props are not nullable.

yurishkuro, line 28: don't follow this. If props are empty you replace them with null? How then can you deconstruct null into `{ data }` object?

yurishkuro: > `props || {}`

why do you need even that? `props` are not nullable in the signature, so they are already required to be at least `{}`

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/AccordianKeyValues.tsx
```diff
@@ -23,20 +23,9 @@ import { KeyValuePair, Link } from '../../../../types/trace';
...
-  highContrast?: boolean;
-  interactive?: boolean;
-  isOpen: boolean;
-  label: string;
-  linksGetter: ((pairs: KeyValuePair[], index: number) => Link[]) | TNil;
-  onToggle?: null | (() => void);
-};
-
 // export for tests
 export function KeyValuesSummary(props: { data?: KeyValuePair[] }) {
```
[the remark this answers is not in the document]
yurishkuro: doesn't answer my question. Previous bug is not the reason to repeat it.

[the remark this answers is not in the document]
its-me-abhishek: was unaware of that, in that case I guess

```
export function KeyValuesSummary({ data }: { data: KeyValuePair[] }) {
```
works well here, and also passes the CI, as only data was being passed through props so I have replaced it, and removed the nullable assignment

yurishkuro (approved):

yurishkuro:
can you add tests to in crease coverage?

yurishkuro (approved):

its-me-abhishek:
still shows as uncovered, will try to find a fix for this

Issue #2572, opened 2025-01-09
[Feature]: TraceGraph should indicate which mode is selected


Zen-cronic:
### Requirement

To provide better visual aid when users view the TraceGraph page.

### Problem

Currently, the default view for the `TraceGraph` is `Service` mode. 

![Image](https://github.com/user-attachments/assets/8e21bd5d-1aa1-4513-834d-b2f73232b61e)

The user then select a different mode (e.g., Time). On hover, it shows the mode.

![Image](https://github.com/user-attachments/assets/fdd6f3f7-6fcc-4b96-a9e5-7eea9744f27f)

After hover and press, all 3 buttons for the modes remain the same visually, although the graph view has changed to Time:

![Image](https://github.com/user-attachments/assets/58edaa24-43c0-4328-9650-ecd802cba73d)

### Proposal

For mode selection use [radio](https://ant.design/components/radio) buttons instead of buttons.

eyeaadil:
I would like to add this feature.

Saquib45:
Hi @Zen-cronic , I’ve set up the Jaeger UI locally and have started working on this feature.
I’ve already updated the sidebar buttons to use Ant Design radio buttons with proper styling, and I’m now checking if the highlighting behavior depends on the CSS changes or if it’s related to the trace data.
I’m also planning to load sample traces to test and confirm the behavior.
Will share updates here as I progress.

Please let me know if there are any specific traces or scenarios you’d like me to test against.

Saquib45:
I've opened PR #3034 to fix this:
https://github.com/jaegertracing/jaeger-ui/pull/3034

let me know any further modifications needed @yurishkuro

PR #2668, opened 2025-02-15
Feature: Changed to Radio buttons, also shows which mode is selected (#2572)

vaishnavi192:
Solves #2572 issue
[Feature]: TraceGraph should indicate which mode is selected

To provide better visual aid when users view the TraceGraph page. I was able to change the appearance of buttons to radio buttons and also changed the selection mode view.

It was tested by running JaegerUI locally and Docker all in one image on localhost as said in instructions. 
Before Image:
image

After Image:
Screenshot 2025-02-15 232557

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceGraph/TraceGraph.css
+++ b/packages/jaeger-ui/src/components/TracePage/TraceGraph/TraceGraph.css
@@ -57,6 +57,21 @@ limitations under the License.
   padding-bottom: 0.3rem;
 }
 
+.TraceGraph--menu label {
+  display: flex;
+  align-items: center;
+  cursor: pointer;
+}
+
+.TraceGraph--menu input[type="radio"] {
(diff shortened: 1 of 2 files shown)


yurishkuro:
I don't see why it need radio buttons, it should be enough to style the existing buttons such that the current one stands out, e.g. by giving it a more solid fill color.

vaishnavi192:
> I don't see why it need radio buttons, it should be enough to style the existing buttons such that the current one stands out, e.g. by giving it a more solid fill color.

The person mentioned to use radio buttons so I did that should I change it to solid colour or I think radio buttons also works right?
https://github.com/jaegertracing/jaeger-ui/issues/2572

yurishkuro:
radio buttons look uglier than what we had before and take more space.

vaishnavi192:
okay I'll work on it

PR #2673, opened 2025-02-18
Feat/migrate  page ,index and TraceIdSearch to functional component

Mubashirshariq:
Migrating Class components to functional components
/attempt #2610

## Description of the changes
migrate page,index and traceidsearch

- 

## Checklist


packages/jaeger-ui/src/components/App/Page.tsx
```diff
@@ -36,38 +37,27 @@ type TProps = {
 const { Header, Content } = Layout;
 
 // export for tests
-export class PageImpl extends React.Component<TProps> {
-  componentDidMount() {
-    const { pathname, search } = this.props;
+export const PageImpl: React.FC<TProps> = ({ children, embedded, pathname, search }) => {
```
yurishkuro, line 40: the argument should have a type (TProps)

Mubashirshariq: Since React.FC<TProps> already applies the type to the function parameters.Do i need to explicitly give type TProps to arguments

packages/jaeger-ui/src/components/App/Page.tsx
```diff
@@ -36,38 +37,27 @@ type TProps = {
...
+  useEffect(() => {
     trackPageView(pathname, search);
```
yurishkuro, line 42: previously this was called conditionally. What happens now?

> `if (pathname !== nextPathname || search !== nextSearch) {`

Mubashirshariq: @yurishkuro  The conditional check ` if (pathname !== nextPathname || search !== nextSearch)` was removed because useEffect's dependency array [pathname, search] handles this optimization automatically. The effect will only run when either dependency changes, making the manual comparison redundant. This achieves the same optimization with cleaner code.

packages/jaeger-ui/src/components/App/index.jsx
```diff
@@ -77,62 +77,60 @@ const jaegerTheme = {
...
 
-export default class JaegerUIApp extends Component {
-  constructor(props) {
-    super(props);
-    JaegerAPI.apiRoot = DEFAULT_API_ROOT;
-    processScripts();
-  }
+// Initialize API and process scripts
+JaegerAPI.apiRoot = DEFAULT_API_ROOT;
+processScripts();
```
yurishkuro, line 82: why is this ok to call this function unconditionally on importing index.jsx? Isn't there some lifecycle thing in functional components when it can be called? Is there an official recommendation for functions like this?

Mubashirshariq: i thought this is just a configuration step,which we can call at module level,but the `processScripts` interacts with DOM so it need to be in useEffect.updated it with useEffect so that it will run whenver the component mounts

packages/jaeger-ui/src/components/App/index.jsx
```diff
@@ -77,11 +77,11 @@ const jaegerTheme = {
...
 
-// Initialize API and process scripts
-JaegerAPI.apiRoot = DEFAULT_API_ROOT;
-processScripts();
-
 const JaegerUIApp = () => {
+  useEffect(() => {
+    JaegerAPI.apiRoot = DEFAULT_API_ROOT;
+    processScripts();
+  }, []);
```
hari45678, line 84: `useEffect` runs after the component mounts initially, while constructors are executed before the component mounts.
A better alternative would be to use a custom hook that uses a ref/state to keep track of the functional. Read a nice [article here](https://dev.to/bytebodger/constructors-in-functional-components-with-hooks-280m#:~:text=const%20useConstructor%20%3D%20(callBack%20%3D%20()%20%3D%3E%20%7B%7D)%20%3D%3E%20%7B%0A%20%20const%20hasBeenCalled%20%3D%20useRef(false)%3B%0A%20%20if%20(hasBeenCalled.current)%20return%3B%0A%20%20callBack()%3B%0A%20%20hasBeenCalled.current%20%3D%20true%3B%0A%7D) that explains why functional components lack this `constructor` alternative and what should be the workaround.

Mubashirshariq: but i think here it needs to be inside the `useEffect` since processScripts() directly interacts with DOM,so it should be rendered every time the component is mounted

yurishkuro:
@Mubashirshariq are you still planning to complete this?

PR #2676, opened 2025-03-02
Replace react-vis with recharts in ScatterPlot

hari45678:
- part of https://github.com/jaegertracing/jaeger-ui/issues/1597

## Description of the changes
- This PR replaces react-vis in ScatterPlot component with Recharts

- manually and testing lib

## Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4554,21 +4554,46 @@
         "@types/node": "*"
       }
     },
+    "node_modules/@types/d3-array": {
+      "version": "3.2.1",
+      "resolved": "https://registry.npmjs.org/@types/d3-array/-/d3-array-3.2.1.tgz",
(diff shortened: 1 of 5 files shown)


hari45678:
Some screenshots representing previous vs new implementation:

| Previous | New |
|----------|-------|
|image|image|
|image|image|

yurishkuro:
hmm, 8% growth in bundle size. What would be the change once we remove react-viz (which this PR is not removing)? Would it make sense to fix the SPM views too, so that we can actually remove the dependency for better comparison? I know it will make the PR larger, but we could split it later and merge as two PRs, I just want to understand the impact on the bundle size.

hari45678:
Yes, makes sense. Since react bundles uses Tree Shaking, this bundle size might even grow bigger with SPM views having imports from recharts, so that would be a more accurate comparison. Will create another PR on top of this branch with SPM views too

Issue #2534, opened 2025-01-03
[LFX] Upgrade charts and graphs


yurishkuro:
This is a project proposed as part of LFX Mentorship term https://github.com/jaegertracing/jaeger/issues/6470 ⬅ **read this first**.

[Jaeger](https://www.jaegertracing.io/) is an open-source, distributed tracing platform designed to monitor and troubleshoot microservices-based systems. Jaeger UI pioneered many new visualizations for analyzing distributed traces. However, over time, it accumulated views that utilize different and sometimes deprecated viz libraries. The objective is to standardize charting / graphing libraries used in Jaeger UI, upgrade their dependencies, and add new visualization features. Specifically:

  * #1597
  * Consolidate DAG and Force Directed Graph (FDG) views under System Architecture into one (using the internal graph library Plexus)
    * See also #1466
  * Upgrade Plexus dependencies (e.g. #1081 and #339) to actively supported libraries
  * Normalize look and feel of different graphs
    * DAG and FDG
    * Deep dependency graph (accessible from Search results)

Desired outcomes:
  * Jaeger UI is not using deprecated dependencies
  * Consistent look and feel of graphs
  * Bonus: side panel for details for a given node
  * Bonus: overlaying metrics on the graph (as edge annotations and node coloring to reflect health / error rates)
  * Bonus: varying node displays depending on the type of node and implementation language

Risks / open questions
  * The FDG view was originally designed to handle large graphs (it was able to visualize complete topology of Uber, around 3000 micro-services at the time). It is not clear how Plexus would behave given such high number of nodes and whether the layout will be legible. It will need to be benchmarked during the project.

yashpandey06:
@yurishkuro Great !!.  I would love to be a part of it !!

jayg2309:
@yurishkuro I'd love to be a part of this. Please let me know the further process.

yurishkuro:
@jayg2309 it is described in the parent issue linked at the top

coderboy-yash:
i have started working on it, will make a pr soon

yurishkuro:
@coderboy-yash this is proposed as a 3-month. We are not asking to make PRs for this specific issue until the project starts. We have plenty of other open issues in the UI that you can solve to demonstrate your skillset.

PR #2691, opened 2025-03-14
Add Depth and Layout controls and sfdp layout to DAG view

hari45678:
- part of #2534 

## Description of the changes
- This PR removes `react-vis-force`
- Removed FDG view completely.
- Added some selector options to DAG view
   - The Layout can be changed between Hierarchical (dot), and FDP
   - Both layouts can have an option to focus on a service by selecting the service from Dropdown.
   - The selected service can have an option to change the depth of nearest neighbors shown on the screen.

- manually

## Checklist


hari45678:
Some tests would fail. Need to work on that

hari45678:
I used this function to generate a large list of dependencies
https://gist.github.com/hari45678/e4711eb8f0c8944be24c2fb21f930e78

For example:
Copy paste the following file's content
https://gist.github.com/hari45678/33b538fcee8549bfcdc88a2afd9ec50a

to
jaeger-ui/packages/jaeger-ui/src/components/DependencyGraph/index.jsx

hari45678:
I used a different function previously (to test 3K nodes) but that functional directly generates a list of nodes and edges
https://gist.github.com/hari45678/3c9d9893a7b09f02e5e19db2002ae19c

packages/jaeger-ui/src/constants/index.tsx
```diff
@@ -14,7 +14,7 @@
 
 export const TOP_NAV_HEIGHT = 46;
 
-export const FALLBACK_DAG_MAX_NUM_SERVICES = 100;
+export const FALLBACK_DAG_MAX_NUM_SERVICES = 500;
```
yurishkuro, line 17: please add a comment describing what this parameter affects

yurishkuro (commented):
a couple minor nits, that can be addressed in follow-up PRs:
  1. when I switched to FDG (of HotROD diagram) the bottom left node was partially covered by the zoom controls. But the chart only allows to zoom in, not zoom out, so no way to make the graph smaller. We should either allocate less space for it or (even better) support zoom out to at least 75%
  2. The "reload" icon to unset focal service is not very intuitive. Maybe add a Reset button instead. Also, I was expecting simply erasing the value to undo depth filtering, but it appears to be sticky.
  4. In the DDG view we have the focal service highlighted with different border

hari45678:
> a couple minor nits, that can be addressed in follow-up PRs:
> 
> 1. when I switched to FDG (of HotROD diagram) the bottom left node was partially covered by the zoom controls. But the chart only allows to zoom in, not zoom out, so no way to make the graph smaller. We should either allocate less space for it or (even better) support zoom out to at least 75%
> 2. The "reload" icon to unset focal service is not very intuitive. Maybe add a Reset button instead. Also, I was expecting simply erasing the value to undo depth filtering, but it appears to be sticky.
> 3. In the DDG view we have the focal service highlighted with different border

The 2nd and 3rd should be straightforward. Would make changes for them in a follow-up PR. For the 1st one, I will try if that is something we can control

packages/jaeger-ui/src/constants/index.tsx
hari45678: done

yurishkuro (approved):

PR #2696, opened 2025-03-19
Feature/nl search

Oloruntobi1:


yurishkuro:
Curious PR.

Oloruntobi1:
Haha should i leave it? I dont know im having doubts.

Here is how it would look btw.

Screenshot 2025-03-19 at 18 55 33

And i think its a great idea. If you and others like that is

yurishkuro:
I think it could be a fun feature, assuming that (a) it's accurate enough, and (b) that's not very complicated to setup. I am not too familiar with the world of LLM APIs, ideally there would be a gateway where requests could be sent and the user can choose which model is used to answer the question. I wouldn't want to build the ability to talk to different models from Jaeger UI, just one API.

Oloruntobi1:
> I think it could be a fun feature

Thanks @yurishkuro.

> assuming that (a) it's accurate enough
The accuracy depends on the model tbh and the prompt. And its doing well for now (gets it right 6/10). But again if the prompt is refined enough with more examples im sure we can achieve 9/10 even with not so strong (local) models.

> and (b) it's not very complicated to setup.
No its not at all. Our good friend docker is there to help and Ollama is pretty easy to install and use. And i went with Ollame because it gives that `local` vibe. But i plan to make it flexible so you can call external ai models if they choose to.

> ideally there would be a gateway where requests could be sent and the user can choose which model is used to answer the question. I wouldn't want to build the ability to talk to different models from Jaeger UI, just one API.

Hmnnn this could work. Thanks. Just tinkering for now tbh but i guess i was quite excited with my first draft and made a PR lol. Got this working in just a few hours! So lets see a week?

Maybe if it get your concept working, we merge it?

Issue #2700, opened 2025-03-22
[Feature]: Align visuals of DDG and SDG


yurishkuro:
### Requirement

The graphs shown in the UI should use similar visualizations.

### Problem

DDG viz

<img width="508" alt="Image" src="https://github.com/user-attachments/assets/41ef9736-20f0-49f3-8876-38c222f15193" />

SDG viz

<img width="440" alt="Image" src="https://github.com/user-attachments/assets/bb24b18c-1f64-4b93-9e44-b2414c7e210e" />

And there are different capabilities in each. There is no reason for them to be different.

Aditya-132:
can i work on this issue

yurishkuro:
@hari45678 is there anything left in this? If not please close.

hari45678:
@yurishkuro Yes this can be closed now

Issue #2699, opened 2025-03-22
DAG View Improvements


hari45678:
Tracking the DAG view improvement tasks in this ticket.

- [ ] when switched to FDG (of HotROD diagram) the bottom left node was partially covered by the zoom controls. But the chart only allows to zoom in, not zoom out, so no way to make the graph smaller. We should either allocate less space for it or (even better) support zoom out to at least 75%
- [x] The "reload" icon to unset focal service is not very intuitive. Maybe add a Reset button instead. Also allow erasing the depth value
- [x] In the DDG view we have the focal service highlighted with different border. Do something similar for DAG view
- [x] Add an upper limit to the number of nodes that can be viewed without selecting a focal service. Beyond which, force user to select focal service.

yurishkuro:
Note - I dded it as a sub-issue of the main ticket #2534

PR #2710, opened 2025-03-26
Feat: change DAG styling and add Search functionality

hari45678:
- part of https://github.com/jaegertracing/jaeger-ui/issues/2700, https://github.com/jaegertracing/jaeger-ui/issues/2699

## Description of the changes
- This PR adds a new Search Bar functionality in the DAG view that would allow users to search the node. The matching node would be highlighted in yellowish color just like in DDG
- This also adds node highlight for focal service

- Locally

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/DependencyGraph/out.txt
+++ b/packages/jaeger-ui/src/components/DependencyGraph/out.txt
@@ -1,0 +1,291 @@
+
+> jaeger-ui-monorepo@0.0.1 test
+> npm run --workspaces test -- DAG.test
+
+
+> @jaegertracing/plexus@0.2.0 test
+> echo 'NO TESTS YET' DAG.test
+
+NO TESTS YET DAG.test
+
+> jaeger-ui@1.67.0 test
+> jest --maxWorkers=50% DAG.test
+
+  console.error
+    Error: Uncaught [Error: Inline worker is not supported]
+        at reportException (/home/hariomgupta/jaeger/jaeger-ui/node_modules/jest-environment-jsdom/node_modules/jsdom/lib/jsdom/living/helpers/runtime-script-errors.js:66:24)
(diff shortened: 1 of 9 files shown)


hari45678:
Normal

image

Focal Service

image

Node Search

image

Node Search with Focal Service as Value

image

packages/jaeger-ui/src/components/DependencyGraph/out.txt
```diff
@@ -0,0 +1,291 @@
+
+> jaeger-ui-monorepo@0.0.1 test
```
yurishkuro, line 2: checked in by mistake?

yurishkuro (approved):
Awesome!

But a question though, if you bring SDG and DDG functionality roughly in line, does it mean we'll have largely duplicated code? In other words, is there an option for converging to a single implementation in the first place? Or are they sufficiently different because of underlying one-hop / multi-hop data structures? Even in that case it still feels uncomfortable that the look&feel will be duplicated and annoyingly slightly different.

packages/jaeger-ui/src/components/DependencyGraph/out.txt
hari45678: yes, my bad. removed

hari45678:
> Awesome!
> 
> But a question though, if you bring SDG and DDG functionality roughly in line, does it mean we'll have largely duplicated code? In other words, is there an option for converging to a single implementation in the first place? Or are they sufficiently different because of underlying one-hop / multi-hop data structures? Even in that case it still feels uncomfortable that the look&feel will be duplicated and annoyingly slightly different.

I will try to eventually bring both of them inline. For example, currently, UIFindInput is reused from common components.
Similarly, that Hop thing in DDG, I will try to incorporate that into SDG by removing the current depth input, so can use the same components from DDG if they are re-usable

yurishkuro (approved):

Issue #669, opened 2020-12-21
Trace Statistics miscounts by-tag groups


ab-pm:
I've used Trace Statistics and when selecting tags in the "Group By" and "Sub-Group" dropdowns, I noticed that the subgroup counts don't add up to the group counts. Looking closer, some spans seemed to get counted multiple times - they appear only once in my trace but the "count" column shows the value "2" in the table. After ensuring the trace JSON data is correct, I stepped through the rendering in the debugger and found the culprit to be

https://github.com/jaegertracing/jaeger-ui/blob/6827f700f9e5bc85c2db23493f7e37730f39e865/packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx#L445-L449

`diffNames` contains the values for the sub-grouped tag, and the nested loops around these lines aggregate all matching spans into the `resultValue`. However, when iterating through the tags of the span, only the tag value is accounted for - it doesn't check whether the tag name is actually the one that is (sub-)grouped by! Indeed, I had a few spans where multiple tags had the same content value, which were getting counted multiple times in this loop.

I think the fix should be as simple as prefixing the `if` condition by `tempArray[l].tags[a].key === selectedTagKeySecond &&`. However similar loops can be found elsewhere in the code, I'm not certain which of them need fixing as well (e.g. [line 326](https://github.com/jaegertracing/jaeger-ui/blob/6827f700f9e5bc85c2db23493f7e37730f39e865/packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx#L326), and [this](https://github.com/jaegertracing/jaeger-ui/blob/6827f700f9e5bc85c2db23493f7e37730f39e865/packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx#L586-L593) looks especially sketchy with the nesting).

jpkrohling:
@ab-pm sorry for not picking this up sooner. Do you have a trace JSON that can be used to reproduce this problem?

ab-pm:
Here's a stripped-down trace (I cannot post the full traces we collect):
```json
{
    "data": [
        {
            "traceID": "03009dceaecb9370",
            "spans": [
                {
                    "traceID": "03009dceaecb9370",
                    "spanID": "03009dceaecb9370",
                    "operationName": "query",
                    "references": [],
                    "startTime": 1628086238194148,
                    "duration": 6675,
                    "tags": [],
                    "logs": [],
[...]

PR #2717, opened 2025-04-12
Fix grouping on trace statistics page for tags

DamianMaslanka5:
Closes https://github.com/jaegertracing/jaeger-ui/issues/669

I noticed this issue (and found existing issue on github) when I started working on improving performance of calculating the groups.  
I wanted to fix this bug before I start improving performance of trace statistics page.

Previously spans with different tags,
but the same value were displayed as a single group:

## Data
* span1 - tag1: value1
* span2 - tag2: value1

## Expected result (grouped by tag1)

```
Group  | Count
---------------
value1 | 1
Other (without tag) | 1
```

## Previous result (grouped by tag1)

```
Group  | Count
---------------
value1 | 2
```

## Before
before

## After

after

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.test.js
+++ b/packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.test.js
@@ -26,6 +26,7 @@ import traceWithOverlappingChildren from './tableValuesTestTrace/traceWithOverla
 import traceWithSingleChildLongerThanParentAndStartsAfterParent from './tableValuesTestTrace/traceWithSingleChildLongerThanParentAndStartsAfterParent.json';
 import traceWithThreeShortChildren from './tableValuesTestTrace/traceWithThreeShortChildren.json';
 import traceWithTwoChildrenStartedAtTraceStart from './tableValuesTestTrace/traceWithTwoChildrenStartedAtTraceStart.json';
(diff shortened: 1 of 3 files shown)


yurishkuro:
@DamianMaslanka5 I am not sure I follow the issue. In the Data section your table shows a single span `span1` with two tags. If you group by one of them the expected output is a single row. Did you mean the 2nd row in the `Data` section to be `span2 - tag2: value1`?

DamianMaslanka5:
@yurishkuro Yes, it was supposed to be `span1` and `span2` instead of two `span1`.

I updated the description, and force pushed to fix the commit message

yurishkuro (approved):
I wonder if we used some JS-equivalent of Python's pandas lib (like Danfo.js) whether this whole file could've been way shorter. All it's really doing is some stats computation on dataframe and grouping.

DamianMaslanka5:
> I wonder if we used some JS-equivalent of Python's pandas lib (like Danfo.js) whether this whole file could've been way shorter. All it's really doing is some stats computation on dataframe and grouping.

I feel like introducing another library is not a good idea :)

I think current code is not great, and needs some cleaning up, but I hope it will be better after I make some changes related to performance. Since there are some performance issue, introducing another library might make it worse.

Btw current code related to drange, on the trace statistics page is slow (300ms for 10k spans), I will rewrite this code at some point, to not use `drange` - [`drange in computeSelfTime`](https://github.com/jaegertracing/jaeger-ui/blob/main/packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx#L60)

PR #2716, opened 2025-04-06
Improve performance when expanding/collapsing span details

DamianMaslanka5:
Improve performance when expanding/collapsing span details (tags/process)

Related to https://github.com/jaegertracing/jaeger-ui/issues/645

Memoize "calculation" of criticalPathSections per span, when span is not collapsed

## Prepare data

Run `go run cmd/tracegen/main.go -service abcd -traces 1 -spans 300` from the jaeger repo

## Perf difference

Total render time from the profiler available in react devtools, after collapsing Tags section

### With these changes

14ms

image

### Changes from master

548ms

image

https://github.com/user-attachments/assets/0bfe8f3e-eac4-4d2e-a119-62644c453103

Issue #645, opened 2020-10-01
UI is extremely slow for traces with > 2000 spans


shooshx:
## Requirement - what kind of business use case are you trying to solve?
Tracing and debugging a complex microservices system

## Problem - what in Jaeger blocks you from solving the requirement?

One of the key processes has traces from 7 different services. A single trace can get up to 5000 spans
When such a span is displayed in the UI the UI get extremely slow.
- Clicking a single span to see the tags takes about 5 seconds to open
- Loading the page and switching to the tab also takes a long time
With such a span, the top canvas that shows a summary of all the spans is basically useless since everything is so tiny in it

## Proposal - what do you suggest to solve the problem or improve the existing situation?
- Fix the issue that makes opening the tags of a span take so long
- Create DOM elements only for the spans that are in view so that the page DOM tree would not be so huge
- Allow resizing the top canvas so that it could be useful

joe-farro:
The optimizations you've suggested were implemented years ago.im wondering
if your running a very old version or if there's been a serious
regression...

What version of jaeger-query are you running?

shooshx:
How can know what version it is?
Clicking About->Documentation brings me to this page:  https://www.jaegertracing.io/docs/1.17/
Looking at the code I'm seeing a link to ./static/js/main.2763c360.chunk.js
Running this:
document.getElementsByTagName("*").length
In a trace page with 2819 spans returns 8619, so yea, it's definitely creating a few DOM elements for every span

shooshx:
Can I please get any reply...? is this project being actively maintained?

rubenvp8510:
Hi, How are you running the query service? Are you using a docker images? all-in one?  if this is the case, which tag are you using?

Other way to find the version is executing the following:

- for all-in-one : `docker exec -it <docker-container-id> /go/bin/all-in-one-linux version`
- not all-in-one : `docker exec -it <query-docker-container-id> /go/bin/query-linux version`
- or if you have access to the binaries in case you are not executing on containers `<path-of-your-binary>/query-linux version.


As Joe mentioned, the optimizations you suggested should be already implemented but this could be a regression. I'll do some profiling to see if that is the case.

shooshx:
Running jaeger query 1.17.1
Is this something that got better since 1.17.1 ?

rubenvp8510:
[...]

PR #2718, opened 2025-04-12
Reduce load time of trace page by deferring critical path tooltip

DamianMaslanka5:
~I am opening this PR without an issue, I understand that is might not be an acceptable change, feel free to close this PR if you don't agree with this change.~

Improve performance (~300ms for 500 spans) by loading tooltip only on hover

* Related issue in antd repo https://github.com/ant-design/ant-design/issues/48468
* My other changes which improve performance related to critical path https://github.com/jaegertracing/jaeger-ui/pull/2716
* Towards https://github.com/jaegertracing/jaeger-ui/issues/645, to improve general performance with large amount of spans

## Before

criticalPathTooltip

## After

criticalPathNoTooltip

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBar.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBar.tsx
@@ -57,6 +57,19 @@ function toPercentInDecimal(value: number) {
 function SpanBarCriticalPath(props: { criticalPathViewStart: number; criticalPathViewEnd: number }) {
   const [shouldLoadTooltip, setShouldLoadTooltip] = useState(false);
 
+  const criticalPath = (
+    <div
+      data-testid="SpanBar--criticalPath"
+      className="SpanBar--criticalPath"
+      onMouseEnter={() => setShouldLoadTooltip(true)}
+      style={{
+        background: 'black',
+        left: toPercentInDecimal(props.criticalPathViewStart),
+        width: toPercentInDecimal(props.criticalPathViewEnd - props.criticalPathViewStart),
+      }}
+    />
+  );
+
   // Load tooltip only when hovering over critical path segment
   // to reduce initial load time of trace page by ~300ms for 500 spans
   if (shouldLoadTooltip) {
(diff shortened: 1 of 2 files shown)


yurishkuro (commented):
I understand what you're trying to do (effectively reduce the size of the DOM), but the resulting behavior is unintuitive, from a user perspective the tooltip would work or not "randomly" (unless they know about 500 spans threshold).

Isn't there another mechanism in React where the tooltip doesn't need to be physically added to the DOM but still can be shown dynamically?

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBar.tsx
```diff
@@ -54,6 +54,55 @@ function toPercentInDecimal(value: number) {
...
+      className="SpanBar--criticalPath"
+      onMouseEnter={() => {
+        setShouldLoadTooltip(true);
+        setIsHovered(true);
+      }}
+      onMouseLeave={() => setIsHovered(true)}
+      style={{
+        background: 'black',
+        left: toPercentInDecimal(props.criticalPathViewStart),
+        width: toPercentInDecimal(props.criticalPathViewEnd - props.criticalPathViewStart),
```
DamianMaslanka5, line 100: **note:** I am ok with a little bit of duplicated code here and in L81, because it is contained in a single component.

But let me know if this a problem, I can add another component (`SpanBarCriticalPathWithoutTooltip`?) to remove the duplication

DamianMaslanka5:
`I understand what you're trying to do (effectively reduce the size of the DOM), but the resulting behavior is unintuitive, from a user perspective the tooltip would work or not "randomly" (unless they know about 500 spans threshold).`

I agree this is not perfect.

`Isn't there another mechanism in React where the tooltip doesn't need to be physically added to the DOM but still can be shown dynamically?`

Thanks for the question, it made me think a bit more. Tbh I am not very familiar with react, so I am open to all suggestions.

---

I made changes to create the tooltip only on hover 41ef333798b14aceaa29f10cf4d746f34bfe297f, I just added another commit, and removed previous changes in the same commit, just remember to squash the commits when merging (or I can squash this locally and force push), I am not sure if you squash all commits by default, on merge

## Render time in profiler looks good

image

## Tooltips work correctly


firefox_f27lZW4NM5

DamianMaslanka5:
@yurishkuro, I updated this branch

1. I added test to avoid reducing code coverage
2. I removed the initial commit, which disabled the tooltip when 500 spans existed. Now tooltip is loaded only when hovering mouse over critical path.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBar.tsx
```diff
@@ -54,6 +54,52 @@ function toPercentInDecimal(value: number) {
...
+        placement="top"
+        // defaultOpen is needed to show the tooltip when shouldLoadTooltip changes to true
+        defaultOpen
+        title={
+          <div>
+            A segment on the <em>critical path</em> of the overall trace/request/workflow.
+          </div>
+        }
+      >
+        <div
```
yurishkuro, line 74: can we avoid duplicating this code segment?

DamianMaslanka5: Yes, see https://github.com/jaegertracing/jaeger-ui/pull/2718#discussion_r2041031917

I will change this tomorrow

yurishkuro (approved):

Issue #2179, opened 2024-02-20
[Bug]: Critical path is taxing on CPU when expanding span tags


bobrik:
### What happened?

We upgraded from v1.46 to v1.54 and some users noticed slowness when expanding tags. With some CPU tracing I was able to pinpoint #1871 as the culprit.

### Steps to reproduce

1. Run Jaeger v1.54
2. Load a large trace
3. Try expanding/collapsing spans
4. Observe slowness

### Expected behavior

Everything is snappy as it was.

### Screenshot

<img width="1399" alt="image" src="https://github.com/jaegertracing/jaeger-ui/assets/89186/5f737a33-e09f-43c4-92be-f9803ddd5569">

<img width="1399" alt="image" src="https://github.com/jaegertracing/jaeger-ui/assets/89186/6614ef3d-3b5a-4631-a9eb-54531358993b">

### Jaeger backend version

v1.54

MiirzaBaig:
Is it still on?
if yes would like to work on this!

yurishkuro:
yes, there is an unfinished PR

MiirzaBaig:
noted!

BenzeneAlcohol:
This is happening because `mergeChildrenCritialPath` has a recursive function `findAllDescendants` that keeps calling descendents (children) of the span. In the wort case, each span could be checked once and hence the time complexity here is already O(n^2). And then filtering and merging adds some extra time.

We could simplify this by using a map, that stores the spans. This will take O(n) to create, but once created search operations would be processed in constant time. Now, using a stack we can push all the spans into the stack. If they have a child, they are pushed as well. In O(n) time we can evaluate the stack, and an O(m) time to filter and merge. Makes the time complexity better than the previous approach.

I can give a detailed explanation in the PR, if this idea sounds good and I can proceed. 

cc: @yurishkuro

yurishkuro:
@BenzeneAlcohol pre-building a map makes total sense. However, you may notice that we already build a tree when loading a trace (packages/jaeger-ui/src/selectors/trace.tsx). I would really like to have a way where that tree is not thrown away but reused for critical path and any other visitors. I think we keep rebuilding this tree multiple times for every trace.

BenzeneAlcohol:
Yeah, I looked into it. I'll come up with an algo later today or tomorrow and post it here.

Issue #2720, opened 2025-04-15
[Bug]: error while computing critical path for a trace InternalError: too much recursion


DamianMaslanka5:
### What happened?

Page with trace details is unresponsive for few seconds. Critical path is not rendered.

### Steps to reproduce

1. Run `go run cmd/tracegen/main.go -service abcd -traces 1 -spans 10000`
2. Go to trace details

### Expected behavior

1. Page should be more responsive with large amount of spans. 
1. No error in console.

### Relevant log output

```shell

```

### Screenshot

![Image](https://github.com/user-attachments/assets/6c4c00ad-0eb5-4881-a5af-f9ebeaea3c64)

### Deployment configs

```shell

```

PR #2722, opened 2025-04-17
Improve performance of expanding and collapsing spans

DamianMaslanka5:
closes https://github.com/jaegertracing/jaeger-ui/issues/2179

## Description of the changes
Use Set and Map to make lookups faster.

This changes makes collapsing and expanding spans with large (10k) amount of children 200-500ms faster.

## Before

image

## After

image

## How to test

1. Generate data - `go run cmd/tracegen/main.go -service abcd -traces 1 -spans 10000`
1. Collapse root span

firefox_4ekX2LDxUX

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
@@ -159,6 +159,8 @@ function getCssClasses(currentViewRange: [number, number]) {
   });
 }
 
+const memoizedSpanByID = memoizeOne((spans: Span[]) => new Map(spans.map(x => [x.spanID, x])));
+
 function mergeChildrenCriticalPath(
   trace: Trace,
   spanID: string,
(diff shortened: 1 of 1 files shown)


packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
```diff
@@ -159,6 +159,8 @@ function getCssClasses(currentViewRange: [number, number]) {
   });
 }
 
+const memoizedSpanByID = memoizeOne((spans: Span[]) => new Map(spans.map(x => [x.spanID, x])));
```
yurishkuro, line 162: fyi https://github.com/jaegertracing/jaeger-ui/issues/2179#issuecomment-2228691694

yurishkuro (approved):

yurishkuro:
@DamianMaslanka5 how is this PR different from #1975?

DamianMaslanka5:
> @DamianMaslanka5 how is this PR different from #1975?

@yurishkuro, #1975 tries to optimize `computeCriticalPath`, but `computeCriticalPath` is executed only once ([this function is memoized](https://github.com/jaegertracing/jaeger-ui/blob/main/packages/jaeger-ui/src/components/TracePage/CriticalPath/index.tsx#L108)), when the trace details page is loaded

So #1975 will not improve performance issue described in https://github.com/jaegertracing/jaeger-ui/issues/2179

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
DamianMaslanka5: I know, but I am not sure if there is a good place for this data.  

Creating a Map or a Set should not have measurable impact on performance, because it is usually done once per trace.  
Also, if any refactoring is needed, then it would probably be best to do this in another PR.

yurishkuro:
Is that PR still needed then? In your experiments with large traces how much initial calculation takes?

DamianMaslanka5:
> Is that PR still needed then? In your experiments with large traces how much initial calculation takes?

@yurishkuro, yes https://github.com/jaegertracing/jaeger-ui/pull/1975 is still needed, without changes from https://github.com/jaegertracing/jaeger-ui/pull/1975, `computeCriticalPath` takes 200ms for 10k spans, with changes from https://github.com/jaegertracing/jaeger-ui/pull/1975 `computeCriticalPath` is not visible in the profiler, also https://github.com/jaegertracing/jaeger-ui/issues/2720 is no longer an issue

PR #2723, opened 2025-04-24
[Fix]: make reset icon in SDG more intuitive

hari45678:
- part of https://github.com/jaegertracing/jaeger-ui/issues/2699

## Description of the changes
- Make the Reset icon more intuitive

- Manually, Unit tests

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/DependencyGraph/DAG.test.js
+++ b/packages/jaeger-ui/src/components/DependencyGraph/DAG.test.js
@@ -213,6 +213,26 @@ describe('<DAG>', () => {
     expect(digraph.props.edges).toHaveLength(0);
   });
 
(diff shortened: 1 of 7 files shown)


yurishkuro:
> Make the Reset icon more intuitive

what does "more intuitive" mean specifically, can you elaborate? Screenshots?

hari45678:
The reset button was previously an icon but it now looks like

image

Also, we were not allowed to erase/delete the Max Depth input. That is now allowed, 
For example, a user wants to focus on a service but doesn't want a max depth (basically wants to view the whole graph). Now this behavior is supported. Previously, they were required to have some number in max depth field

yurishkuro (approved):

PR #2724, opened 2025-04-24
Improve performance of Trace Statistics page when grouping by tag

DamianMaslanka5:
Related to https://github.com/jaegertracing/jaeger-ui/issues/645

## Description of the changes
Tested with 10k spans - `go run cmd/tracegen/main.go -service abcd -traces 1 -spans 10000`

Previously getValueTagIsPicked took 2s, now this function takes less than 1ms because it is not visible in a profiler

## Before

image

## After

Now only `computeSelfTime` is slow, when only first dropdown is used for grouping results

image


yurishkuro:
Thanks!

PR #2731, opened 2025-04-28
Add ability to use typescript in tests

DamianMaslanka5:
Better DX, Find References in VS Code doesn't work correctly when functions are used in test.js files

## Description of the changes
Allow using TS in tests

Docs which describe how to setup jest with TS https://jestjs.io/docs/getting-started#type-definitions


packages/jaeger-ui/tsconfig.json
```diff
@@ -13,5 +13,5 @@
 // limitations under the License.
 {
   "extends": "../../tsconfig",
-  "include": ["src/**/*.tsx", "typings"]
+  "include": ["src/**/*.tsx", "src/**/*.ts", "src/**/*.json", "typings"]
```
yurishkuro, line 16: why `.json`

yurishkuro, line 16:
```suggestion
  "include": ["src/**/*.tsx", "src/**/*.ts", "typings"]
```

yurishkuro (dismissed):

yurishkuro (approved):

packages/jaeger-ui/tsconfig.json
DamianMaslanka5: Thanks for merging this, I didn't have time to look into this previously.

TS was complaining about importing the json file in the test (but you know this now)

`Error: [tsc-lint     ] packages/jaeger-ui/src/components/TracePage/TraceStatistics/generateDropdownValue.test.ts(18,23): error TS6307: File '/home/runner/work/jaeger-ui/jaeger-ui/packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValuesTestTrace/testTrace.json' is not listed within the file list of project '/home/runner/work/jaeger-ui/jaeger-ui/packages/jaeger-ui/tsconfig.lint.json'. Projects must list all files or use an 'include' pattern.`

TS is configured to also provide autocomplete for imported json files https://stackoverflow.com/a/50674344 (I didn't know that TS can do this :)), so I guess that's why processing json files must be enabled in tsconfig.json.

image

PR #2739, opened 2025-04-29
Migrate PopupSql test

nojaf:
- Part of https://github.com/jaegertracing/jaeger-ui/issues/1668

## Description of the changes
- Migrates `PopupSql` test

- Test itself.

## Checklist


packages/jaeger-ui/src/components/TracePage/TraceStatistics/PopupSql.test.js
```diff
@@ -13,27 +13,41 @@
...
+    render(<PopupSql {...defaultProps} />);
+  });
+
+  it('renders the header', () => {
+    expect(screen.getByRole('heading', { name: /Tag: "SQL"/i })).toBeInTheDocument();
+  });
+
+  it('renders the SQL content in the textarea', () => {
+    // RTL textarea role implicitly looks for `value` attribute, we need getByDisplayValue
+    // Also checking the value is wrapped in quotes as per the component logic
```
yurishkuro, line 40: is there actually wrapping logic in the component? I don't see it. Wouldn't wrapping be achieved via CSS anyway?

nojaf: This is about it being wrapped in double quotes. I'll update the comment.

PR #2742, opened 2025-04-29
Migrate ResultItem tests

nojaf:
- Part of https://github.com/jaegertracing/jaeger-ui/issues/1668

## Description of the changes
- Migrates `ResultItem` test

- Test itself.

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ResultItem.test.js
+++ b/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ResultItem.test.js
@@ -30,7 +30,10 @@ const renderWithRouter = ui => {
 let trace; // Use let to allow modification in tests
 
 beforeEach(() => {
-  // Reset trace data before each test
+  // Reset trace data before each test.
+  // Some tests modify the trace object (e.g., adding tags).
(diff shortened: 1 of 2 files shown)


packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ResultItem.test.js
```diff
@@ -13,33 +13,97 @@
...
-const trace = transformTraceData(traceGenerator.trace({}));
+// Helper function to wrap component with Router
+const renderWithRouter = ui => {
+  return render(ui, { wrapper: MemoryRouter });
+};
+
+let trace; // Use let to allow modification in tests
+
+beforeEach(() => {
+  // Reset trace data before each test
```
yurishkuro, line 33: Why do we need to reset it for each test? Is something modifying the data?

nojaf: The test case that modifies the trace data is:
`it('<ResultItem /> should render error icon on ServiceTags that have an error tag'`
This test directly adds an error tag to one of the spans within the trace object.

yurishkuro (approved):

PR #2767, opened 2025-05-14
Rewrite computeSelfTime to improve performance on Trace Statistics page

DamianMaslanka5:
Stop using drange

For 10k spans, 1s => 1ms, computeSelfTime is no longer visible in a profiler

To test:
1. Generate data: go run cmd/tracegen/main.go -service abcd -traces 1 -spans 10000
1. Go to Trace Statistics
1. Change Group By to Operation Name
1. Change Group By to Service Name

Related to https://github.com/jaegertracing/jaeger-ui/issues/645

- Existing tests

## Before

image

## After

image

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
@@ -42,61 +42,61 @@ function getChildOfSpans(parentID: string, allSpans: Span[]): Span[] {
   return memoizedParentChildOfMap(allSpans)[parentID] || [];
 }
 
-function computeSelfTime(span: Span, allSpans: Span[]): number {
-  if (!span.hasChildren) return span.duration;
+function computeSelfTime(parentSpan: Span, allSpans: Span[]): number {
+  if (!parentSpan.hasChildren) return parentSpan.duration;
 
-  let spanSelfTime = span.duration;
-  let previousSpanEndTime = span.startTime;
+  let parentSpanSelfTime = parentSpan.duration;
+  let previousChildEndTime = parentSpan.startTime;
 
-  const children = getChildOfSpans(span.spanID, allSpans).sort((a, b) => a.startTime - b.startTime);
+  const children = getChildOfSpans(parentSpan.spanID, allSpans).sort((a, b) => a.startTime - b.startTime);
 
-  const parentSpanEndTime = span.startTime + span.duration;
+  const parentSpanEndTime = parentSpan.startTime + parentSpan.duration;
 
   for (let index = 0; index < children.length; index++) {
     const child = children[index];
 
-    const spanEndTime = child.startTime + child.duration;
-    const spanStartsAfterParentEnded = child.startTime > parentSpanEndTime;
-    const spanEndsBeforePreviousSpan = spanEndTime < previousSpanEndTime;
+    const childEndTime = child.startTime + child.duration;
+    const childStartsAfterParentEnded = child.startTime > parentSpanEndTime;
+    const childEndsBeforePreviousChild = childEndTime < previousChildEndTime;
 
     // parent |..................|
     // child    |.......|
-    // child     |.....|                      - spanEndsBeforePreviousSpan is true, skipped
-    // child                         |......| - spanStartsAfterParentEnded is true, skipped
-    if (spanStartsAfterParentEnded || spanEndsBeforePreviousSpan) {
+    // child     |.....|                      - childEndsBeforePreviousChild is true, skipped
+    // child                         |......| - childStartsAfterParentEnded is true, skipped
+    if (childStartsAfterParentEnded || childEndsBeforePreviousChild) {
       continue;
     }
 
     let nonOverlappingStartTime = child.startTime;
     let nonOverlappingDuration = child.duration;
-    const spanStartsBeforePreviousSpanEnds = child.startTime < previousSpanEndTime;
+    const childStartsBeforePreviousChildEnds = child.startTime < previousChildEndTime;
 
     // parent |.....................|
     // child    |.......|
-    // child        |.....|                  - spanStartsBeforePreviousSpanEnds is true
-    // child                |.....|          - spanStartsBeforePreviousSpanEnds is false
-    if (spanStartsBeforePreviousSpanEnds) {
-      const diff = previousSpanEndTime - child.startTime;
+    // child        |.....|                  - childStartsBeforePreviousChildEnds is true
+    // child                |.....|          - childStartsBeforePreviousChildEnds is false
+    if (childStartsBeforePreviousChildEnds) {
+      const diff = previousChildEndTime - child.startTime;
       nonOverlappingDuration = child.duration - diff;
-      nonOverlappingStartTime = previousSpanEndTime;
+      nonOverlappingStartTime = previousChildEndTime;
     }
     // last span which can be included in self time calculation, because it ends after parent span ends
     // parent |......................|
     // child                      |.....|        - last span included in self time calculation
     // child                       |.........|   - skipped
-    else if (spanEndTime > parentSpanEndTime) {
-      const diff = spanEndTime - parentSpanEndTime;
+    else if (childEndTime > parentSpanEndTime) {
+      const diff = childEndTime - parentSpanEndTime;
 
       nonOverlappingDuration = child.duration - diff;
-      spanSelfTime -= nonOverlappingDuration;
+      parentSpanSelfTime -= nonOverlappingDuration;
       break;
     }
 
-    spanSelfTime -= nonOverlappingDuration;
-    previousSpanEndTime = nonOverlappingStartTime + nonOverlappingDuration;
+    parentSpanSelfTime -= nonOverlappingDuration;
+    previousChildEndTime = nonOverlappingStartTime + nonOverlappingDuration;
   }
 
-  return spanSelfTime;
+  return parentSpanSelfTime;
 }
 
 function computeColumnValues(trace: Trace, span: Span, allSpans: Span[], resultValue: StatsPerTag) {
--- a/packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
@@ -60,40 +60,32 @@ function computeSelfTime(parentSpan: Span, allSpans: Span[]): number {
     const childEndsBeforePreviousChild = childEndTime < previousChildEndTime;
 
     // parent |..................|
-    // child    |.......|
+    // child    |.......|                     - previousChild
     // child     |.....|                      - childEndsBeforePreviousChild is true, skipped
     // child                         |......| - childStartsAfterParentEnded is true, skipped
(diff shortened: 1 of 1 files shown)


packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
```diff
@@ -45,21 +44,47 @@ function getChildOfSpans(parentID: string, allSpans: Span[]): Span[] {
...
+  const children = getChildOfSpans(span.spanID, allSpans).sort((a, b) => a.startTime - b.startTime);
+
+  const parentSpanEndTime = span.startTime + span.duration;
+
+  for (let index = 0; index < children.length; index++) {
+    const child = children[index];
+
+    const spanEndTime = child.startTime + child.duration;
+    const spanStartsAfterParentEnded = child.startTime > parentSpanEndTime;
+    const spanEndsBeforePreviousSpan = spanEndTime < previousSpanEndTime;
```
yurishkuro, line 60: the naming is confusing. Can we stick with `parentXyz` and `childXyz` to make it easier to follow?

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
```diff
@@ -45,21 +44,47 @@ function getChildOfSpans(parentID: string, allSpans: Span[]): Span[] {
...
+    if (spanStartsAfterParentEnded || spanEndsBeforePreviousSpan) {
+      continue;
+    }
+
+    let nonOverlappingStartTime = child.startTime;
+    let nonOverlappingDuration = child.duration;
+
+    const spanStartsBeforePreviousSpanEnds = child.startTime < previousSpanEndTime;
+    if (spanStartsBeforePreviousSpanEnds) {
```
yurishkuro, line 69: can we add add diagrams to illustrate the scenarios? something like
```
// child(n-1):  |.......|
// child(n):         |......|
```

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
DamianMaslanka5: IMO these names are accurate.

```
parent |..................|
child    |.......|
child     |.....|                      - spanEndsBeforePreviousSpan is true, this span is skipped
child                         |......| - spanStartsAfterParentEnded is true, this span is skipped
```

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
DamianMaslanka5: WDYT about using format like this? Also see my previous comments, looks like example for L59,L60 will be also useful.

```
parent |.....................|
child    |.......|
child        |.....|                  - spanStartsBeforePreviousSpanEnds is true
child                |.....|          - spanStartsBeforePreviousSpanEnds is false
```

And example for L75

```
parent |......................|
child                      |.....|        - last span included in self time calculation
child                       |.........|   - skipped
```

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
yurishkuro: current naming is just `spanXyz`, which is confusing because there are two spans involved in every step and it's not clear which one is referenced by `spanXyz` name

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
yurishkuro: yes, in every if() there is an implied configuration which could be represented by a diagram like that to make the logic easier to follow.

DamianMaslanka5: I added the comments in 7b7286d2eaf8bdda84b97170cb575010169efd88

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
DamianMaslanka5: I think I misunderstood your initial comment, this is a general issue for the whole function, not just the lines you commented on? I see now what that I used `span` in the parameter of the function and `span` for children.

Can you confirm that these names are less confusing? Or if you have other suggestion it would be helpful if you could provide examples.
* `previousSpanEndTime` => `previousChildEndTime`
* `spanEndTime` => `childEndTime`
* `spanEndsBeforePreviousSpan` => `childEndsBeforePreviousChild` 
* `spanStartsAfterParentEnded` => `childStartsAfterParentEnded` 
* `spanStartsBeforePreviousSpanEnds` => `childStartBeforePreviousChildEnds`
* `spanSelfTime` => `parentSelfTime`?

yurishkuro: Yes that's better

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
```diff
@@ -43,23 +42,61 @@ function getChildOfSpans(parentID: string, allSpans: Span[]): Span[] {
...
+    }
+
+    let nonOverlappingStartTime = child.startTime;
+    let nonOverlappingDuration = child.duration;
+    const childStartsBeforePreviousChildEnds = child.startTime < previousChildEndTime;
+
+    // parent |.....................|
+    // child    |.......|
+    // child        |.....|                  - childStartsBeforePreviousChildEnds is true
+    // child                |.....|          - childStartsBeforePreviousChildEnds is false
```
yurishkuro, line 77: why isn't this case handled (full subtraction of span's time from parent)? The `else` part below only covers when child extends past the parent.

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
```diff
@@ -43,23 +42,61 @@ function getChildOfSpans(parentID: string, allSpans: Span[]): Span[] {
...
-    spanRange.subtract(10 * child.startTime, 10 * (child.startTime + child.duration) - 1);
-  });
-  return Math.round(spanRange.length / 10);
+function computeSelfTime(parentSpan: Span, allSpans: Span[]): number {
+  if (!parentSpan.hasChildren) return parentSpan.duration;
+
+  let parentSpanSelfTime = parentSpan.duration;
+  let previousChildEndTime = parentSpan.startTime;
+
+  const children = getChildOfSpans(parentSpan.spanID, allSpans).sort((a, b) => a.startTime - b.startTime);
```
yurishkuro, line 51: I think it might be more efficient if the full `allSpans` array was sorted once, and then direct children arrays picked from it.

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
```diff
@@ -43,23 +42,61 @@ function getChildOfSpans(parentID: string, allSpans: Span[]): Span[] {
...
+
+  const parentSpanEndTime = parentSpan.startTime + parentSpan.duration;
+
+  for (let index = 0; index < children.length; index++) {
```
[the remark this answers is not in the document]
DamianMaslanka5: IMO this is exactly the same algorithm, just implemented a bit differently.

Initially wanted to keep the code dumb, so I wanted to avoid using Math.max/min. But using Math.max/Math.min should be fine, because comments with examples are very helpful. Now I don't have any preference which version is easier to understand. 

Please let me know which version you prefer. I made the changes in 03532d7707a051d612e38a55fd329c93a06ad3f5

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
DamianMaslanka5: I am not sure if this actually will be faster. But I am almost sure that it is not worth spending any time on optimizing this.

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
DamianMaslanka5: This if is for special cases, when child doesn't have any overlap with previous span, then `child.duration` is subtracted from `parentSpanSelfTime`, see `let nonOverlappingDuration = child.duration;`

Code in the newest commit is a bit different.

These links are to the previous commit, without Math.min/max

1. [let nonOverlappingDuration = child.duration;](https://github.com/jaegertracing/jaeger-ui/blob/5935bd1528ed6596cf73abd99fc070ee3265224d/packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx#L71)
1. [parentSpanSelfTime -= nonOverlappingDuration;](https://github.com/jaegertracing/jaeger-ui/blob/5935bd1528ed6596cf73abd99fc070ee3265224d/packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx#L95)

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
```diff
@@ -43,23 +42,61 @@ function getChildOfSpans(parentID: string, allSpans: Span[]): Span[] {
...
+    const child = children[index];
+
+    const childEndTime = child.startTime + child.duration;
+    const childStartsAfterParentEnded = child.startTime > parentSpanEndTime;
+    const childEndsBeforePreviousChild = childEndTime < previousChildEndTime;
+
+    // parent |..................|
+    // child    |.......|
+    // child     |.....|                      - childEndsBeforePreviousChild is true, skipped
```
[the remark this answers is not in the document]
DamianMaslanka5: I included this change in 03532d7707a051d612e38a55fd329c93a06ad3f5

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.tsx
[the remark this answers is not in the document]
yurishkuro: I would expect the biggest cost to come from sorting the children repeatedly, making the whole thing an O(N^3) complexity, while I think it's possible to do it with O(N^2) if we sort first, the construct the tree from that sorted order.

As for the algorithm, min/max makes it a lot easier (for me at lest) to reason as to what's happening, since it makes the logic linear, vs. reasoning about which edge case in the diagrams represent and whether all edge cases are actually covered. In my algo I only need to think about `lastEnd` boundary moving up and each span measuring against it, not against any "previous" spans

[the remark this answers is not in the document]
DamianMaslanka5: > I would expect the biggest cost to come from sorting the children repeatedly, making the whole thing an O(N^3) complexity, while I think it's possible to do it with O(N^2) if we sort first, the construct the tree from that sorted order.

`computeSelfTime` is no longer visible in a profiler, so even if it can be faster, I just don't think it is worth spending time on this, because any change to this function will not be visible. Also see https://github.com/jaegertracing/jaeger-ui/pull/2767#discussion_r2107716747

yurishkuro (approved):

Issue #2764, opened 2025-05-12
Upgrade Jaeger-UI to React v19


yurishkuro:
Part of LFX Mentorship Summer 2025 term: https://github.com/jaegertracing/jaeger/issues/7117

Applications must be submitted via LFX program website.

- Description: The current version of React used by Jaeger-UI is v18. This project aims to upgrade Jaeger-UI to React v19, in order to stay up to date with the dependencies.
- Expected Outcome:
  - A working version of Jaeger-UI upgraded to React v19, with all the features and functionality of the current version. 
  - This may include a need to upgrade or replace [some other dependencies](https://github.com/jaegertracing/jaeger-ui/issues/2395) that may not be compatible with the latest React.
- Recommended Skills: Javascript, React, HTML, CSS, Web development, UI/UX design, Testing
- Mentor(s):
  - Yuri Shkuro (@yurishkuro)
  - Jonah Kowall (@jkowall)

Previous similar projects:
  - https://github.com/jaegertracing/jaeger-ui/issues/998

How to start / apply:
  - See https://github.com/jaegertracing/jaeger/issues/7117

eyeaadil:
Hey  @yurishkuro ,Myself Md Adil, a 3rd-year undergraduate student with a strong background in React, JavaScript, and modern web development. I’m highly interested in contributing to this upgrade of Jaeger-UI to React v19 as part of the LFX Mentorship Summer 2025 term.

I've reviewed the project goals and understand the importance of maintaining compatibility while upgrading to the latest version of React. I’m confident in my ability to handle dependency updates, identify and resolve breaking changes, and ensure the UI remains fully functional post-upgrade. I'm also familiar with testing strategies and UI/UX best practices.

I'd love the opportunity to start contributing and learn from this experience. Please let me know how I can get started, or if there are any prerequisites or contribution guidelines I should follow first.

Looking forward to your guidance!

Anuj054:
**Subject: Interest in Contributing to Jaeger-UI React v19 Upgrade – LFX Mentorship Summer 2025**

Hello @yurishkuro  and @jkowall ,

I hope you're doing well. I recently came across the LFX Mentorship Summer 2025 opportunity for the project “Upgrade Jaeger-UI to React v19” (issue [#7115](https://github.com/jaegertracing/jaeger/issues/7115)), and I’m very interested in contributing to it.

[...]

PR #2771, opened 2025-05-16
chore: upgrade root dependencies and fix eslint peer conflict

abhayporwals:
## Chore: Upgrade Root-Level Dev Dependencies #2764 

> _Pre-Selection Work for LFX TERM-2 Project_

This PR upgrades several development dependencies in the root `package.json` to their latest stable versions to ensure improved stability and compatibility with modern tooling.

### Updated Packages

| Package                             | From       | To        |
|-------------------------------------|------------|-----------|
| `@babel/eslint-parser`              | ^7.24.6    | ^7.27.1   |
| `@typescript-eslint/eslint-plugin` | ^8.32.0    | ^8.32.1   |
| `@typescript-eslint/parser`         | ^8.32.0    | ^8.32.1   |
| `eslint`                            | ^8.57.0    | ^8.57.1   |
| `eslint-config-prettier`            | 10.1.1     | 10.1.5    |
| `eslint-plugin-jsx-a11y`            | ^6.8.0     | ^6.10.2   |
| `eslint-plugin-react`               | 7.37.1     | 7.37.5    |
| `husky`                             | 9.1.0      | 9.1.7     |
| `jsdom`                             | ^26.0.0    | ^26.1.0   |
| `prettier`                          | 3.5.0      | 3.5.3     |
| `typescript`                        | 5.8.2      | 5.8.3     |

### ESLint Version Note

- `eslint` was updated only to `^8.57.1` (not upgraded to v9) due to compatibility issues with `eslint-config-airbnb@19.0.4`, which requires `eslint@^7.32.0 || ^8.2.0`.
- Attempting to install `eslint@9.x` causes peer dependency resolution errors.
- We'll revisit ESLint v9 migration once dependent packages are compatible.

### Flat Config Migration

ESLint 9 uses the new flat config format via `eslint.config.ts`, replacing older `.eslintrc` styles.  
Currently, we are still using the legacy format. Migration to flat config will be handled in a separate PR.

### Notes

- Linting, formatting, and pre-commit hooks continue to work as expected.

### Follow-up Tasks

- Migrate to flat ESLint config (`eslint.config.ts`)
- Investigate updating or replacing `eslint-config-airbnb` for ESLint v9 compatibility

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,31 +13,31 @@
         "packages/jaeger-ui"
       ],
       "devDependencies": {
-        "@babel/eslint-parser": "^7.24.6",
-        "@typescript-eslint/eslint-plugin": "^8.32.0",
-        "@typescript-eslint/parser": "^8.32.0",
+        "@babel/eslint-parser": "^7.27.1",
+        "@typescript-eslint/eslint-plugin": "^8.32.1",
+        "@typescript-eslint/parser": "^8.32.1",
         "depcheck": "1.4.7",
-        "eslint": "8.57.0",
+        "eslint": "^8.57.1",
         "eslint-config-airbnb": "19.0.4",
-        "eslint-config-prettier": "10.1.1",
+        "eslint-config-prettier": "10.1.5",
         "eslint-plugin-import": "2.31.0",
-        "eslint-plugin-jsx-a11y": "^6.8.0",
-        "eslint-plugin-react": "7.37.1",
-        "husky": "9.1.0",
-        "jsdom": "^26.0.0",
+        "eslint-plugin-jsx-a11y": "^6.10.2",
+        "eslint-plugin-react": "7.37.5",
+        "husky": "9.1.7",
+        "jsdom": "^26.1.0",
         "npm-run-all2": "8.0.1",
-        "prettier": "3.5.0",
+        "prettier": "3.5.3",
         "rxjs-compat": "6.6.7",
-        "typescript": "5.8.2"
+        "typescript": "5.8.3"
       },
       "engines": {
         "node": ">=20"
       }
     },
     "node_modules/@adobe/css-tools": {
-      "version": "4.4.0",
-      "resolved": "https://registry.npmjs.org/@adobe/css-tools/-/css-tools-4.4.0.tgz",
-      "integrity": "sha512-Ff9+ksdQQB3rMncgqDK78uLznstjyfIf2Arnh22pW8kBpLs6rpKDwgnZT46hin5Hl1WzazzK64DOrhSwYpS7bQ==",
+      "version": "4.4.2",
+      "resolved": "https://registry.npmjs.org/@adobe/css-tools/-/css-tools-4.4.2.tgz",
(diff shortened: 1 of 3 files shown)


anshgoyalevil:
> npm install was tested with --legacy-peer-deps to ensure dependency tree resolution

We should avoid the `--legacy-peer-deps` flag because of the way it works. [Official npm docs](https://docs.npmjs.com/cli/v11/using-npm/config#:~:text=Use%20of%20legacy%2Dpeer%2Ddeps%20is%20not%20recommended%2C%20as%20it%20will%20not%20enforce%20the%20peerDependencies%20contract%20that%20meta%2Ddependencies%20may%20rely%20on) discourage the use of this flag as a solution for upgrading the dependencies.

abhayporwals:
Tested again and working fine.

yurishkuro:
what was the process for this upgrade?

abhayporwals:
> what was the process for this upgrade?

I used npx npm-check-updates -u at the root level to upgrade outdated dependencies listed in the main package.json. This helped clean up deprecated or vulnerable packages, improve compatibility, and align with modern tooling standards. After updating, we removed node_modules and package-lock.json, then ran a fresh npm install to ensure consistency. All builds and dev scripts were tested to confirm nothing broke post-upgrade.

yurishkuro:
We don't do upgrades in this manner. It introduces too many changes at once.

yurishkuro:
Most of the changes are to a patch number, not very important. We have renovate bit that performs these upgrades individually.

abhayporwals:
> We don't do upgrades in this manner. It introduces too many changes at once. 

Got it, thanks for the feedback!

Just to clarify, the upgrades here were mostly minor and didn’t require any actual code changes, these were mainly outdated dependencies that had simple version bumps. That said, I understand your point about not introducing too many changes at once.

We can definitely keep this PR open for now and revisit it later if needed. In the meantime, I’d like to shift focus to more structured upgrades starting with React itself. I think upgrading React first will help us uncover any actual compatibility issues, and once that’s in place, we can more confidently address the surrounding packages in smaller batches. Let me know if that sounds like a good approach!

yurishkuro:
The upgrade here broke some unit tests. A useful change is to find out why (patch versions are unlikely to break tests).

abhayporwals:
> The upgrade here broke some unit tests. A useful change is to find out why (patch versions are unlikely to break tests).

image

now, it will work fine!

yurishkuro:
We do not chase after patch upgrades of all dependencies, if we did I'd be merging renovatebot PRs every day. If you are trying to upgrade a major or minor version, then limit the PR to just that one dependency.

abhayporwals:
Yeah, that completely makes sense, we can close this PR for now.

Meanwhile, I noticed that ESLint has a major update (v9), and there's already an open issue for upgrading it. I'd be happy to work on that in a new PR. I've looked into it a bit already, and there are a few breaking changes -> for example, eslint-config-airbnb doesn't support v9, and we'd need to find an alternative or adjust the config. Also, v9 uses the new flat config format instead of .eslintrc, so that'll require some migration.

Let me know if you'd like me to take that on!

abhayporwals:
> > Let me know if you'd like me to take that on!
> 
> go for it, but aim for small, targeted PRs

Sure things, review noted! 🫡

PR #2775, opened 2025-05-17
[Fix]: Reduce default minimum allowed zoom

hari45678:
- fixes https://github.com/jaegertracing/jaeger-ui/issues/2699

## Description of the changes
- This PR allows reducing the default minimum zoon value so less number of nodes doesn't get too scaled

- Manually

## Checklist


hari45678:
Current:
image

Previous:
image

yurishkuro (approved):

PR #2792, opened 2025-05-20
fix: disable archive button for in-memory storage

### LINKED ISSUE #2201: [Bug]: Archive button is enabled by default
### What happened?

Not sure if this a bug in the UI or in the Jaeger itself, but we have upgraded our Jaeger instances controlled with Jaeger Operator from 1.51 to 1.54. This updated UI from 1.35 to 1.38. Now on at least all-in-one configuration Archive button is enabled by default.

### Steps to reproduce

1. Install Jaeger Operator 1.54
2. Use this simple YAML to deploy the instance:
```
apiVersion: jaegertracing.io/v1
kind: Jaeger
metadata:
  name: jaeger
  namespace: default
spec:
  strategy: allInOne
  allInOne:
    options:
      query:
        base-path: /
  ingress:
    ingressClassName: nginx
    hosts:
      - yourhost.example.com
    pathType: Prefix
  storage:
    options:
      memory:
        max-traces: 100000
```

### Expected behavior

Archive button should be disabled by default, as per documentation.

### Relevant log output

_No response_

### Screenshot

_No response_

### Additional context

It is probably due changes in https://github.com/jaegertracing/jaeger-ui/pull/1944.

### Jaeger backend version

1.54

### SDK

_No response_

### Pipeline

_No response_

### Stogage backend

memory

### Operating system

_No response_

### Deployment model

Jaeger Operator for Kubernetes

### Deployment configs

```shell
apiVersion: jaegertracing.io/v1
kind: Jaeger
metadata:
  name: jaeger
  namespace: default
spec:
  strategy: allInOne
  allInOne:
    options:
      query:
        base-path: /
  ingress:
    ingressClassName: nginx
    hosts:
      - yourhost.example.com
    pathType: Prefix
  storage:
    options:
      memory:
        max-traces: 100000
```

VijetaPriya47:
- Resolves #2201

## Description of the changes
- Disables the archive button for in-memory storage in the TracePage component.
- Displays a warning message if a user attempts to archive a trace while using in-memory storage.
- Prevents users from mistakenly archiving traces that will not persist after a restart.

- Verified that the archive button is hidden when Jaeger is running with in-memory storage.
- Confirmed that a warning message is shown if archive is attempted with in-memory storage.
- Tested the functionality to ensure that the warning message appears correctly.

## Documentation

**Note:**  
The main documentation in the [jaegertracing/jaeger](https://github.com/jaegertracing/jaeger) repository should be updated to reflect this new behavior and clarify that persistent storage is required for archiving traces.  

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/index.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/index.tsx
@@ -21,6 +21,7 @@ import _mapValues from 'lodash/mapValues';
 import _memoize from 'lodash/memoize';
 import { connect } from 'react-redux';
 import { bindActionCreators, Dispatch } from 'redux';
+import { message } from 'antd';
 
 import ArchiveNotifier from './ArchiveNotifier';
 import { actions as archiveActions } from './ArchiveNotifier/duck';
@@ -285,7 +286,10 @@ export class TracePageImpl extends React.PureComponent<TProps, TState> {
   };
 
   archiveTrace = () => {
-    const { id, archiveTrace } = this.props;
+    const { id, archiveTrace, storageCapabilities } = this.props;
+    if (storageCapabilities?.storageType === 'memory') {
+      message.warning('Warning: Archiving with in-memory storage is not persistent. Traces will be lost on restart.');
+    }
     archiveTrace(id);
   };
 
@@ -362,6 +366,8 @@ export class TracePageImpl extends React.PureComponent<TProps, TState> {
 
     const isEmbedded = Boolean(embedded);
     const hasArchiveStorage = Boolean(storageCapabilities?.archiveStorage);
+    const isMemoryStorage = storageCapabilities?.storageType === 'memory';
+    const showArchiveButton = !isEmbedded && archiveEnabled && hasArchiveStorage && !isMemoryStorage;
     const headerProps = {
       focusUiFindMatches: this.focusUiFindMatches,
       slimView,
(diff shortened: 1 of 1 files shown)


packages/jaeger-ui/src/components/TracePage/index.tsx
```diff
@@ -285,7 +286,10 @@ export class TracePageImpl extends React.PureComponent<TProps, TState> {
   };
 
   archiveTrace = () => {
-    const { id, archiveTrace } = this.props;
+    const { id, archiveTrace, storageCapabilities } = this.props;
+    if (storageCapabilities?.storageType === 'memory') {
+      message.warning('Warning: Archiving with in-memory storage is not persistent. Traces will be lost on restart.');
+    }
```
hari45678, line 292: The warning message is shown after the user clicks the archive button, but the button itself is still visible. This could be confusing for users since they can still click it

VijetaPriya47: I feel like instead of removing the button altoghther, let's make just disable the archiving and show the warning on hovering, but the button would still be clickable, cause why remove functionality.

packages/jaeger-ui/src/components/TracePage/index.tsx
```diff
@@ -362,6 +366,8 @@ export class TracePageImpl extends React.PureComponent<TProps, TState> {
 
     const isEmbedded = Boolean(embedded);
     const hasArchiveStorage = Boolean(storageCapabilities?.archiveStorage);
+    const isMemoryStorage = storageCapabilities?.storageType === 'memory';
```
yurishkuro, line 369: the UI was never made aware of the actual storage type implemented by the backend. Where is this coming from? There is no such property in storageCapabilities. But there is a flag that tells if archive storage is enabled.

packages/jaeger-ui/src/components/TracePage/index.tsx
yurishkuro: I agree, the button can be disabled and the tooltip say that archive storage is disabled in the backend (it should not mention anything about in-memory, it's completely irrelevant).

yurishkuro:
not the right implementation

PR #2811, opened 2025-05-24
test(calculateTraceDagEV): improve coverage and add mapFollowsFrom unit tests

vishvamsinh28:
This PR enhances the test coverage and robustness of the calculateTraceDagEV

- Retains existing test logic for calculating nodes in the trace DAG using real-world testTrace.json.
- Adds new unit tests for mapFollowsFrom, validating conditional logic based on span references (CHILD_OF, FOLLOWS_FROM).
- Maintains modular and clear assertions using assertData() for per-node validation.

This update ensures better coverage of internal decision logic and guards against regression in span reference mapping behavior.

Coverage before
Screenshot 2025-05-24 at 2 59 24 PM

Coverage after
Screenshot 2025-05-24 at 3 01 58 PM


packages/jaeger-ui/src/components/TracePage/TraceGraph/calculateTraceDagEV.test.js
```diff
@@ -49,3 +48,41 @@ describe('calculateTraceDagEV', () => {
...
+          },
+        ],
+      },
+    ];
+
+    const result = mapFollowsFrom(mockEdges, mockNodes);
+    expect(result[0].followsFrom).toBe(false);
+  });
+
+  it('sets followsFrom true if node does NOT have CHILD_OF reference and e.to is number', () => {
```
yurishkuro, line 71: what happens if the references array has both? it is dependent on the order?

vishvamsinh28: If a span has both CHILD_OF and FOLLOWS_FROM references, the code treats it as CHILD_OF, regardless of order. This is because it checks for the presence of any CHILD_OF using some(), and sets followsFrom = !hasChildOf. So followsFrom is only true when there’s no CHILD_OF reference at all.

yurishkuro (approved):

PR #2812, opened 2025-05-24
Fix array return pattern in `LabeledList` component

ghost:
The `LabeledList` component currently returns an array from its map function without proper React Fragment keys.

**Issues:**
- The divider element doesn't have a guaranteed unique key when it exists
- Arrays should be wrapped in Fragments with proper keys

## Description of the changes
- Wraped array elements in React.Fragment with proper key attribution
- Move the key prop from the `<li>` element to the React.Fragment

## Checklist


yurishkuro:
I don't understand what problem this solves.

yurishkuro (approved):

PR #2823, opened 2025-05-25
Fix React Fragment key issues in multiple components

ghost:
While working on some other issues in the codebase, I found additional components with similar React Fragment and key issues. These components had the same React Fragment and key patterns (like in #2812 ) that needed to be solved for consistency and react best practices

## Description of the changes
1. `KeyValuesTable.tsx`
- Changed `<>` to `<React.Fragment key={i}>` for proper key attribution
2. `highlightMatches.tsx`
- Ensured consistent return types by wrapping plain strings in `<span>` elements
- Added proper keys to all elements returned from map function
3. `SearchResults/index.tsx`
- Added keys to `React.Fragment` elements in conditional rendering

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/common/FilteredList/__snapshots__/highlightMatches.test.js.snap
+++ b/packages/jaeger-ui/src/components/common/FilteredList/__snapshots__/highlightMatches.test.js.snap
@@ -3,13 +3,21 @@
 exports[`highlightMatches(query, text) matches contains: been-kat-br[eak]able 1`] = `
 <span>
   <span>
-    been-kat-br
+    <span
+      key="text-1"
+    >
+      been-kat-br
+    </span>
     <mark
       key="eak2"
     >
       eak
     </mark>
-    able
+    <span
+      key="text-3"
+    >
+      able
+    </span>
   </span>
 </span>
 `;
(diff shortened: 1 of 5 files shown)


packages/jaeger-ui/src/components/common/FilteredList/__snapshots__/highlightMatches.test.js.snap
```diff
@@ -3,45 +3,73 @@
 exports[`highlightMatches(query, text) matches contains: been-kat-br[eak]able 1`] = `
 <span>
   <span>
-    been-kat-br
+    <span
```
yurishkuro, line 6: I thought adding Fragment was generally a no-op, how come there's a new `<span>` element added to the DOM?

packages/jaeger-ui/src/components/common/FilteredList/highlightMatches.tsx
```diff
@@ -29,7 +29,7 @@ function toHighlights(isHighlightGroup: TIsHighlightGroupFn, match: RegExpMatchA
       if (isHighlightGroup(i)) {
         return <mark key={`${tx + i}`}>{tx}</mark>;
       }
-      return tx;
+      return <span key={`text-${i}`}>{tx}</span>;
```
yurishkuro, line 32: why use span instead of Fragment?

Parship999:
Thanks for clarifying, I understand your feedbacks and will make the changes as suggested.

packages/jaeger-ui/src/components/common/FilteredList/highlightMatches.tsx
yurishkuro: ?

yurishkuro (approved):

Issue #2531, opened 2024-12-26
[Deps] Upgrade history package to 5.x


yurishkuro:

jayesh9747:
@yurishkuro  i want to work on this issue can you assign me ?

jayesh9747:
@yurishkuro  can you describe in details like what should excalty  waht outcome you want ?

yurishkuro:
The desired outcome is that we don't have a stale dependency - we either upgrade to the latest version or if the dependency is deprecated altogether we find something else that is maintained.

jayesh9747:
@yurishkuro, after updating the history version to 5.x, how can I check if it is compatible with this repository and passes all the tests?

yurishkuro:
You run the tests, npm run test, npm run lint.

We already know that it's not compatible, the automated upgrade is failing CI.

jayesh9747:
@yurishkuro 

As I searched on Google to find an alternative to History.js, I discovered that no direct alternatives are available. Therefore, I prefer to resolve the test cases that were filed instead.

![Image](https://github.com/user-attachments/assets/1d673b3e-8aaa-4927-bdb3-c7398c3bf24a)

yurishkuro:
If it is unmaintained it's not enough to fix the tests, we need to actively get rid of it.

yurishkuro:
But fixing the test can be a first step.

jayesh9747:
@yurishkuro  sure , will PR very soon.

manzil-infinity180:
@yurishkuro 

I upgraded to 5.X and tested it, two test cases failing one for the `history.length` and other one is for` toMatchSnapshot()`

```bash
 console.log
      {
        "index": 1,
        "action": "PUSH",
        "location": {
          "pathname": "/trace/MOCK-TRACE-ID",
          "search": "",
          "hash": "",
          "state": null,
          "key": "8kaa6msa"
        }
      }

      at Object.log (src/components/App/TraceIDSearchInput.test.js:50:13)

    console.log
      {
        "index": 0,
        "action": "POP",
        "location": {
          "pathname": "/",
          "search": "",
          "hash": "",
          "state": null,
          "key": "jxeeyogw"
        }
      }

      at Object.log (src/components/App/TraceIDSearchInput.test.js:57:13)

  ● <TraceIDSearchInput /> › does not push to history on falsy input value

    expect(received).toEqual(expected) // deep equality

    Expected: 1
    Received: undefined

[...]

PR #2834, opened 2025-05-27
[Deps] Upgrade history package to 5.x #2531

7908837174:
History Package Upgrade

This PR upgrades the history package from v4.6.3 to v5.3.0 in the Jaeger UI project.

 Changes Made

1. Updated `history` dependency from v4.6.3 to v5.3.0 in package.json
2. Created a polyfill (`history-polyfill.ts`) to bridge API differences between history v4 and v5
3. Updated `configure-store.js` to use the polyfill
4. Fixed test cases that were failing due to API changes in history v5

## Implementation Details

The main issue was that history v5 has API changes that are incompatible with react-router-dom v5, which the project currently uses. The key difference is that history v5 doesn't have a `length` property, which was used in tests and possibly in the application code.

The polyfill adds the missing `length` property to the history object, making it compatible with code that expects the v4 API while using the v5 package.

## Future Considerations

This is a temporary solution until the project can be migrated to use react-router-dom v6+ with functional components and the useNavigate hook. The polyfill should be removed during that migration.

## Testing

All tests are now passing with the history v5 package.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,144 +1,24 @@
-[![Build Status][ci-img]][ci] [![Coverage Status][cov-img]][cov] [![FOSSA Status][fossa-img]][fossa] [![OpenSSF Scorecard](https://api.securityscorecards.dev/projects/github.com/jaegertracing/jaeger-ui/badge)](https://securityscorecards.dev/viewer/?uri=github.com/jaegertracing/jaeger-ui)
+# History Package Upgrade
 
-# Jaeger UI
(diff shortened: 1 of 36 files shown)


vishvamsinh28:
Read https://github.com/jaegertracing/jaeger-ui/pull/2836#issuecomment-2913790773

kallal79:
It works!

PR #2836, opened 2025-05-27
Eslint 9 upgrade - Help with eslint upgrade #2272

7908837174:
Since we're using the FlatCompat utility from @eslint/eslintrc, our solution should maintain compatibility with the existing ESLint configuration while allowing us to use ESLint 9.

The key components of our solution are: 

1. eslint.config.js: Uses FlatCompat to load the existing configuration

2. package.json: Updated with new dependencies and set to use ESM

3. Ignore patterns: Moved from .eslintrc.js to the flat config        

  
## Checklist


kallal79:
Thanks for the update! Tell me what is my mistake in the same...?

Issue #633, opened 2020-09-08
Search on attributes is broken with some characters


MirkoRuggiero:
Hello,

I encountered some issues on the search function using _Tags_ as filter criteria using specific characters as starting or/and ending character of the value string.

Following some examples of tags that have been set on single span, entry for the search and outcome:

### **Tag set    --    Search input    --    Outcome**

hash=#value#  **--**  hash=#value#  **--**  not found
ampersand=&value&  **--**  ampersand=&value&  **--**  Internal error 500
plus=+value+  **--**  plus=+value+  **--**  not found
question=?value?  **--**  question=?value?  **--**  not found

hash=#value  **--**  hash=#value  **--**  not found
ampersand=&value&  **--**  ampersand=&value  **--** found
plus=+value+  **--**  plus=+value  **--**  found
question=?value?  **--**  question=?value  **--**  found

hash=#value  **--**  hash=#value#  **--**  not found
ampersand=&value  **--**  ampersand=&value&  **--**  Internal error 500
plus=+value  **--**  plus=+value+  **--**  found
question=?value  **--**  question=?value?  **--**  found

It seems that some characters inhibit the search on tags (# completely, + and ? only when are both at the beginning and at the end of the value string) while the ampersand breaks the URL.

Are these known behaviours? I tried to search but have found nothing referring to this.
Or is there any documentation I'm missing? 

Cheers!

MirkoRuggiero:
As update:

it seems that "escaping" the hash does the trick

hash=#value# -- hash=\\#value\\# -- found

also it appears that some regular expressions interpretation is going on during the search, given that also the search input string

hash=.+value.+ 
hash=.*value.* 
hash=.value. 

return the correct trace.

jpkrohling:
@oalbacha, would you be interested in taking a look at this one?

oalbacha:
Yes please

jpkrohling:
It's yours then. Let me know if you need help.

fatmcgav:
Just to add, the "double escaping" trick worked for searching a tag value that contains `&` or `:` aswell :+1: 

Would be great to not _have_ to escape them though...

7908837174:
Hey @maintainers,

It seems that escaping the hash (#) with \\ allows searches, while certain regex-like inputs (hash=.+value.+) also return expected results. However, ampersand (&) still breaks the URL, and + and ? block searches when both prefixed and suffixed.

[...]

PR #2838, opened 2025-05-27
Fix search special characters-Search on attributes is broken with some characters #633 

7908837174:
Hash character (#) test cases:

hash=#value# - Now works correctly because the # characters are encoded as %23

hash=#value - Now works correctly for the same reason.

Ampersand character (&) test cases:

ampersand=&value& - Now works correctly because & characters are encoded as %26

ampersand=&value - Now works correctly for the same reason.

Plus character (+) test cases:

plus=+value+ - Now works correctly because + characters are encoded as %2B

plus=+value - Now works correctly for the same reason.

Question mark character (?) test cases:

question=?value? - Now works correctly because ? characters are encoded as %3F

question=?value - Now works correctly for the same reason...

## Checklist

### DIFF
--- a/README-search-fix.md
+++ b/README-search-fix.md
@@ -1,0 +1,61 @@
+# Fix for Search on Attributes with Special Characters
+
+This PR fixes issue #633 where searching on attributes with special characters like `#`, `?`, `+`, and `&` was broken.
+
+## Problem
+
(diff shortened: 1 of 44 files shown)


kallal79:
thanks for the update!

PR #2841, opened 2025-05-28
Migrate `TraceDiffHeader.test.js` from Enzyme to RTL

Parship999:
- Resolves part of #1668 

## Description of the changes
Migrated this test files from Enzyme to RTL:
- `TraceDiffHeader.test.js`

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceDiffHeader.test.js
+++ b/packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceDiffHeader.test.js
@@ -131,7 +131,7 @@ describe('TraceDiffHeader', () => {
     };
   });
 
-  it('renders as expected', () => {
+  it('renders UI elements and trace names correctly', () => {
     render(<TraceDiffHeader {...props} />);
 
     expect(screen.getByText('A', { selector: 'h1' })).toBeInTheDocument();


yurishkuro:
>Current coverage is acceptable?

no, we don't go down in coverage

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/CohortTable.test.js
```diff
@@ -42,190 +36,143 @@ describe('CohortTable', () => {
...
-    beforeEach(() => {
-      updateRowSelection();
-    });
-
     it('defaults selectedRowKeys to empty array', () => {
-      updateWrapper({ current: undefined });
-      updateRowSelection();
-      expect(rowSelection.selectedRowKeys).toEqual([]);
+      render(<CohortTable {...props} current={undefined} />);
+      expect(screen.getByText('Service & Operation')).toBeInTheDocument();
```
yurishkuro, line 73: how does this test the same condition as before (empty array)?

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/CohortTable.test.js
```diff
@@ -42,190 +36,143 @@ describe('CohortTable', () => {
...
     });
 
-    it('calls props.selectTrace on row selection', () => {
-      rowSelection.onChange([cohort[1].id, cohort[2].id], [cohort[1], cohort[2]]);
+    it('calls props.selectTrace on row selection', async () => {
+      render(<CohortTable {...props} />);
+      props.selectTrace(cohort[1].id);
       expect(selectTrace).toHaveBeenCalledWith(cohort[1].id);
```
yurishkuro, line 79: this doesn't make sense - you are calling a function (which is a mock) and then testing that you call it. How does it test that the function would be called on the row selection?

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/CohortTable.test.js
Parship999: yes you're right, it doesn't properly verify the condition...Ok I am working on this

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceDiffHeader.test.js
```diff
@@ -87,100 +87,96 @@ describe('TraceDiffHeader', () => {
...
   });
 
   it('renders as expected', () => {
-    expect(wrapper).toMatchSnapshot();
+    render(<TraceDiffHeader {...props} />);
+    expect(screen.getByText('A')).toBeInTheDocument();
+    expect(screen.getByText('B')).toBeInTheDocument();
+    expect(screen.getByText('VS')).toBeInTheDocument();
+    expect(
+      screen.getAllByText((content, node) => {
```
yurishkuro, line 112: add a comment explaining what this is searching for and why

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceDiffHeader.test.js
```diff
@@ -87,100 +87,96 @@ describe('TraceDiffHeader', () => {
...
+        const hasText = n =>
+          n.textContent && n.textContent.replace(/\s+/g, '').includes('cohort-trace-name-1');
```
yurishkuro, line 114: why does `.includes('cohort-trace-name-1')` need removing spaces first?

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceDiffHeader.test.js
```diff
@@ -87,100 +87,96 @@ describe('TraceDiffHeader', () => {
...
+    props = {
+      a: cohort[1],
+      b: cohort[2],
+      cohort,
+      diffSetA,
+      diffSetB,
+    };
   });
 
   it('renders as expected', () => {
```
yurishkuro, line 106: this is doing more than that, describe the actual test / condition

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceDiffHeader.test.js
```diff
@@ -87,100 +87,96 @@ describe('TraceDiffHeader', () => {
...
+        const childrenDontHaveText = Array.from(node?.children || []).every(childNode => !hasText(childNode));
+        return nodeHasText && childrenDontHaveText;
+      }).length
+    ).toBeGreaterThan(0);
   });
 
   it('handles trace without spans', () => {
-    wrapper.setProps({ a: cohort[0] });
+    render(<TraceDiffHeader {...props} a={cohort[0]} />);
+    expect(screen.getByText('A')).toBeInTheDocument();
```
yurishkuro, line 124: how is this testing "without spans"?

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceDiffHeader.test.js
```diff
@@ -87,100 +87,96 @@ describe('TraceDiffHeader', () => {
...
   });
 
   it('handles absent a', () => {
-    wrapper.setProps({ a: null });
-    expect(wrapper).toMatchSnapshot();
+    render(<TraceDiffHeader {...props} a={null} />);
+    expect(screen.getByText('B')).toBeInTheDocument();
```
yurishkuro, line 129: couldn't condition like getByText('B') be easily matched by other unrelated things?

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceDiffHeader.test.js
```diff
@@ -87,100 +87,96 @@ describe('TraceDiffHeader', () => {
...
-    popovers.forEach(popover => expect(popover.prop('open')).toBe(false));
-
-    getPopoverProp(0, 'onOpenChange')(true);
-    expect(getPopoverProp(0, 'open')).toBe(true);
-    expect(getPopoverProp(1, 'open')).toBe(false);
-
-    getPopoverProp(1, 'onOpenChange')(true);
-    expect(getPopoverProp(0, 'open')).toBe(false);
-    expect(getPopoverProp(1, 'open')).toBe(true);
+  it('manages visibility correctly', async () => {
```
yurishkuro, line 143: non-descriptive

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceDiffHeader.test.js
```diff
@@ -87,100 +87,96 @@ describe('TraceDiffHeader', () => {
...
+    expect((await screen.findAllByText('Service & Operation')).length).toBeGreaterThan(0);
+    await userEvent.click(chevrons[1]);
+  });
 
-    // repeat onOpenChange call to test that visibility remains correct
-    getPopoverProp(1, 'onOpenChange')(true);
-    expect(getPopoverProp(0, 'open')).toBe(false);
-    expect(getPopoverProp(1, 'open')).toBe(true);
+  describe('_diffSetTrace method', () => {
+    it('calls diffSetA when which is "a"', () => {
```
yurishkuro, line 155: "when which"?

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceDiffHeader.test.js
Parship999: if the rendered text has extra spaces due to formatting, the test will still pass as long as the non space characters match

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceDiffHeader.test.js
Parship999: Yeah...Ok I will add CSS class `.TraceDiffHeader--label`

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/CohortTable.test.js
```diff
@@ -42,190 +37,145 @@ describe('CohortTable', () => {
...
+      render(<CohortTable {...props} />);
+      const radioButtons = screen.getAllByRole('radio');
+      await userEvent.click(radioButtons[1]);
       expect(selectTrace).toHaveBeenCalledWith(cohort[1].id);
     });
 
     it('calculates checkbox props for selected and current record with error', () => {
-      expect(rowSelection.getCheckboxProps(cohort[0])).toEqual({ disabled: true });
+      render(<CohortTable {...props} />);
+      expect(screen.getByText('selected index 0')).toBeInTheDocument();
```
yurishkuro: the test title and expectation are confusing

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/CohortTable.test.js
```diff
@@ -42,190 +37,145 @@ describe('CohortTable', () => {
...
     });
 
     it('calculates checkbox props for selected and current record without error', () => {
-      expect(rowSelection.getCheckboxProps({ ...cohort[0], state: fetchedState.DONE })).toEqual({});
+      render(<CohortTable {...props} cohort={[{ ...cohort[0], state: fetchedState.DONE }]} />);
+      expect(screen.getByText('Service & Operation')).toBeInTheDocument();
```
yurishkuro: how does that validate the condition?

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/CohortTable.test.js
```diff
@@ -42,190 +37,145 @@ describe('CohortTable', () => {
...
+          }}
+        />
+      );
+      expect(screen.getByText('Service & Operation')).toBeInTheDocument();
     });
 
     it('calculates checkbox props for not selected record', () => {
-      expect(rowSelection.getCheckboxProps(cohort[1])).toEqual({});
+      render(<CohortTable {...props} />);
+      expect(screen.getByText('Service & Operation')).toBeInTheDocument();
```
yurishkuro: how is this different from the first test?

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/CohortTable.test.js
```diff
@@ -42,190 +37,145 @@ describe('CohortTable', () => {
...
     });
   });
 
   it('renders shortened id', () => {
-    const idRenderer = getRowRenderer('id');
-    const traceID = 'trace-id-longer-than-eight-characters';
-    const renderedId = shallow(idRenderer(traceID));
-    expect(renderedId.text()).toBe(traceID.slice(0, 7));
+    render(<CohortTable {...props} />);
+    expect(screen.getAllByText(props.current.slice(0, 7)).length).toBeGreaterThan(0);
```
yurishkuro: not following. The original test was much more explicit, with a fake trace id

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/CohortTable.test.js
```diff
@@ -42,190 +37,145 @@ describe('CohortTable', () => {
...
+    render(<CohortTable {...props} />);
+    expect(screen.getByText('selected index 0')).toBeInTheDocument();
+    expect(
+      screen.getAllByText((content, node) => {
+        const hasText = n => n.textContent && n.textContent.replace(/\s+/g, '').includes('apierror');
+        const nodeHasText = hasText(node);
+        const childrenDontHaveText = Array.from(node?.children || []).every(childNode => !hasText(childNode));
+        return nodeHasText && childrenDontHaveText;
+      }).length
+    ).toBeGreaterThan(0);
```
yurishkuro: don't follow this complex logic. Title says "renders TraceName fragment", can we not just look for the trace name in the output (and probably state & error as in the original test)?

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/CohortTable.test.js
```diff
@@ -42,190 +37,145 @@ describe('CohortTable', () => {
...
-    const testTrace = cohort[1];
-    const renderedTraceNameColumn = shallow(
-      // traceNameRenderer returns a React Fragment, wrapper div helps enzyme
-      <div>{traceNameColumnRenderer('unused argument', testTrace)}</div>
-    );
-
-    expect(renderedTraceNameColumn.find(Tag).length).toBe(0);
-    expect(renderedTraceNameColumn.find(TraceName).length).toBe(1);
+    render(<CohortTable {...props} cohort={[cohort[1]]} current={cohort[1].id} selection={{}} />);
+    expect(screen.getByText('Service & Operation')).toBeInTheDocument();
```
yurishkuro: how is this checking TraceName?

yurishkuro (commented):
general feedback - you are not sticking to the original spirit of the tests, the validations need to be checking the specific conditions for which the test was written.

Parship999:
@yurishkuro I have implemented the suggested changes. Let me know if there is anything else needed

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/CohortTable.test.js
```diff
@@ -42,190 +37,151 @@ describe('CohortTable', () => {
...
-    beforeEach(() => {
-      updateRowSelection();
-    });
-
     it('defaults selectedRowKeys to empty array', () => {
-      updateWrapper({ current: undefined });
-      updateRowSelection();
-      expect(rowSelection.selectedRowKeys).toEqual([]);
+      render(<CohortTable {...props} current={undefined} />);
+      expect(screen.queryByRole('radio', { checked: true })).not.toBeInTheDocument();
```
yurishkuro: why does this need `checked: true`? Since you didn't simulate any clicks this condition is always true.

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/CohortTable.test.js
```diff
@@ -42,190 +37,151 @@ describe('CohortTable', () => {
...
+    it('calls props.selectTrace on row selection', async () => {
+      render(<CohortTable {...props} />);
+      const radioButtons = screen.getAllByRole('radio');
+      await userEvent.click(radioButtons[1]);
       expect(selectTrace).toHaveBeenCalledWith(cohort[1].id);
     });
 
-    it('calculates checkbox props for selected and current record with error', () => {
-      expect(rowSelection.getCheckboxProps(cohort[0])).toEqual({ disabled: true });
+    it('disables radio button for selected and current record with error state', () => {
```
yurishkuro: what is the difference between "selected" and "current" record?

yurishkuro: and how does the "error state" factor into this test?

Parship999:
I will make the changes in the next commit. Let me know if you spot anything else that needs fixing

vishvamsinh28:
1. Missing State Visibility Assertions for Popovers
The Enzyme suite validated toggle behavior of the popovers, including mutually exclusive visibility and internal state (tableVisible). This logic is not currently tested in the RTL version. Consider adding assertions to verify that when one popover opens, the other is closed.

2. Popover Toggle Test Needs More Specificity
The current popover interaction test only checks for the presence of 'Service & Operation', without confirming which trace is active or ensuring previous ones closed. Enhance this with visibility assertions tied to the correct trace content.

3. Test Name Mismatch (Direct Prop Call Test)
The test "calls diffSetA and diffSetB correctly" manually calls the functions instead of triggering them via user interaction. Either revise the test name to reflect this or replace it with a UI-driven test that validates those callbacks through user behavior.

4. Redundant/Complex DOM Traversal
The test that checks trace name hierarchy uses a getAllByText callback with child traversal logic. This can be fragile. Consider simplifying using getByTestId('traceName') or asserting directly on known spans/text.

5. Missing Case: traceID === null
The component displays “Select a Trace...” when traceID is missing. This edge case wasn’t covered in either the original or migrated suite. It would be good to include it now.

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceDiffHeader.test.js
```diff
@@ -87,100 +96,141 @@ describe('TraceDiffHeader', () => {
...
+    expect(screen.getByText('VS')).toBeInTheDocument();
+
+    const traceHeaders = screen.getAllByTestId('TraceDiffHeader--traceHeader');
+    expect(within(traceHeaders[0]).getByText('cohort-trace-name-1')).toBeInTheDocument();
+    expect(within(traceHeaders[1]).getByText('cohort-trace-name-2')).toBeInTheDocument();
   });
 
-  it('handles trace without spans', () => {
-    wrapper.setProps({ a: cohort[0] });
+  it('handles a trace without spans array', () => {
```
yurishkuro: please do not rewrite test titles

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceDiffHeader.test.js
```diff
@@ -87,100 +96,141 @@ describe('TraceDiffHeader', () => {
...
-    const shouldCall = {
-      a: diffSetA,
-      b: diffSetB,
-    };
-    const shouldNotCall = {
-      a: diffSetB,
-      b: diffSetA,
-    };
+    // Wait for popover transition to complete
+    await waitFor(() => {
```
yurishkuro: wait for how long? Is there a version of this function with a timeout?

ghost: yes `waitFor` accepts an options object with a timeout parameter. I will add `{ timeout: 5000 }` (5 seconds) to all the `waitFor` calls in the updated code to address this. The default timeout for `waitFor` is usually 1000ms, but adding an explicit timeout allows for longer waits if needed for UI transitions.

yurishkuro:
@vishvamsinh28 please review

yurishkuro:
@vishvamsinh28 please review

vishvamsinh28:
1. Lost Popover Visibility State Validation
-The original test (manages visibility correctly) explicitly verified each Popover’s open prop value using .prop('open').
-The migrated test (toggles popovers with mutual exclusion) only checks for DOM presence (getByRole('tooltip')) and lacks precise mutual exclusivity validation.
-Add assertions to confirm that when one popover opens, the other is definitively closed — not just by checking DOM presence, but by explicitly validating intended open/closed states.

2. Incomplete Test for Bound Function Props
-The original test validated trace selection via both:
title prop (which contains TraceIdInput)
content prop (which contains CohortTable)
-The migrated test only validates interaction through the content (table rows), completely omitting the TraceIdInput interaction.
-Add test cases that simulate trace selection via the popover title (TraceIdInput) to ensure both selection paths still work.

3. Weak Assertion for Span Count Handling
-Test handles trace without spans only checks that the span field renders as empty (toBeEmptyDOMElement()).
-This is a loose validation compared to asserting that span count is undefined or 0.
-Strengthen the assertion to explicitly confirm the correct fallback behavior when spans are missing.

4. Missing Internal State Validation
-Original tests validated internal component state (tableVisible) after toggling popovers.
-Migrated tests assume state works correctly based on tooltip visibility, but don’t validate the actual internal state transitions.
-Use component-level testing utilities (if feasible) or mock callbacks more explicitly to assert state transitions.

5. Test Name Doesn’t Match What’s Being Tested
-Test handles trace without spans implies it's checking span handling.
-In reality, the test mostly verifies trace name rendering; span count is loosely implied and not explicitly validated.
-Either rename the test to reflect its actual focus, or improve the assertions to match the current test name.

6. Unnecessary Async Complexity
-toggles popovers with mutual exclusion uses await and waitFor for popover visibility checks.
-These async steps complicate the test without improving clarity or specificity.
-Simplify the async flow where possible, or move focus to validating core open/close logic with better precision.

@Parship999 Could you clarify why changes were made to the TraceHeader component, given that it's already mocked in the test file?

cc : @yurishkuro

ghost:
Thanks for the review.
I'm closing this PR as it contains several issues and it's been over a month since it was opened, I honestly no longer recall the original reasoning behind some of the changes. I’ll create a new, clean PR and follwing all the feedbacks from the start

PR #2859, opened 2025-06-03
Migrate `QualityMetrics/Header.test.js` from Enzyme to RTL

ghost:
- Resolves part of #1668 

## Description of the changes
- Replaced Enzyme tests with RTL in `Header.test.js`

## Checklist


packages/jaeger-ui/src/components/QualityMetrics/Header.test.js
```diff
@@ -50,46 +49,59 @@ describe('Header', () => {
...
+      expect(screen.getByRole('spinbutton')).toBeInTheDocument();
+      expect(screen.getByText(/in hours/i)).toBeInTheDocument();
+      expect(screen.getByText(/Service:/i)).toBeInTheDocument();
     });
 
     it('renders props.lookback when state.ownInputValue is `undefined`', () => {
-      expect(wrapper.find(InputNumber).prop('value')).toBe(lookback);
+      render(<Header {...props} />);
+      const input = screen.getByRole('spinbutton');
+      expect(input.value).toBe(String(lookback));
```
yurishkuro, line 73: how is this testing "state.ownInputValue is undefined" condition?

yurishkuro (commented):
I don't recommend mixing -> functional component changes with test->RTL changes. They should be done independently.

packages/jaeger-ui/src/components/QualityMetrics/Header.test.js
```diff
@@ -50,46 +49,65 @@ describe('Header', () => {
...
-      expect(wrapper.find(InputNumber).prop('value')).toBe(lookback);
+      const { rerender } = render(<Header {...props} />);
+      const input = screen.getByRole('spinbutton');
+      expect(input.value).toBe(String(lookback));
+
+      rerender(<Header {...props} lookback={10} />);
+      expect(input.value).toBe('10');
     });
 
     it('renders state.ownInputValue when it is not `undefined` regardless of props.lookback', () => {
```
vishvamsinh28, line 76: The original test was validating that when state.ownInputValue is set, it takes precedence over props.lookback. The migrated test is just checking DOM manipulation, not the component's internal state management logic

packages/jaeger-ui/src/components/QualityMetrics/Header.test.js
```diff
@@ -50,46 +49,65 @@ describe('Header', () => {
...
     });
 
     it('updates state with numeric value, then clears state and calls props.setLookback after debounce', () => {
-      const value = 42;
-      wrapper.find(InputNumber).prop('onChange')(value);
+      render(<Header {...props} />);
+      const input = screen.getByRole('spinbutton');
+      const inputNumber = input.closest('.ant-input-number');
+      const inputNumberInstance =
+        inputNumber && inputNumber[Object.keys(inputNumber).find(k => k.startsWith('__reactFiber$'))];
```
vishvamsinh28: This approach is fragile, version-dependent, and doesn't actually test the component properly.

packages/jaeger-ui/src/components/QualityMetrics/Header.test.js
```diff
@@ -50,46 +49,65 @@ describe('Header', () => {
...
+      fireEvent.change(input, { target: { value: '27' } });
+      expect(input.value).toBe('27');
+
+      rerender(<Header {...props} lookback={10} />);
+      expect(input.value).toBe('27');
     });
   });
 
   describe('setting lookback', () => {
     it('no-ops for string values', () => {
```
vishvamsinh28, line 120: Original test verified the component's handleInputChange logic that filters out string values. The migrated version just tests DOM behavior, not the component's protective logic

vishvamsinh28:
@Parship999 you haven't updated the tests, they're same.

Parship999:
> @Parship999 you haven't updated the tests, they're same.

I will update the tests in next commit. Can you please review the comments

vishvamsinh28:
1. Lost string input validation
The original Enzyme test explicitly verified that string inputs are ignored by the component (onChange('foo') → no state update or callback).
This validation is missing in the RTL version. The current tests implicitly rely on Ant Design's InputNumber to reject invalid input, but we should still explicitly validate that setLookback is not triggered when a string is entered — to ensure the component's logic remains intact.

2. No direct state validation (ownInputValue)
The original test accessed wrapper.state('ownInputValue') to confirm how state was updated and cleared. RTL intentionally avoids internal state inspection, however, the effects of ownInputValue should still be validated indirectly, such as:

- Showing the user input immediately after change
- Ignoring prop changes while ownInputValue is active
- Falling back to props.lookback once debounce clears state

This ensures we maintain confidence that the component behaves as expected even without testing state directly.

yurishkuro (commented):
please do not rename tests, it makes it hard to compare

packages/jaeger-ui/src/components/QualityMetrics/Header.test.js
```diff
@@ -50,47 +49,105 @@ describe('Header', () => {
...
+      expect(input.value).toBe(userInput); // Proves state.ownInputValue is being used
+
+      // Additional verification: changing back to original props still shows state value
+      rerender(<Header {...props} />);
+      expect(input.value).toBe(userInput);
     });
   });
 
   describe('setting lookback', () => {
-    it('no-ops for string values', () => {
```
yurishkuro, line 78: what about this test?

packages/jaeger-ui/src/components/QualityMetrics/Header.test.js
```diff
@@ -50,47 +49,105 @@ describe('Header', () => {
...
     });
 
     it('renders props.lookback when state.ownInputValue is `undefined`', () => {
-      expect(wrapper.find(InputNumber).prop('value')).toBe(lookback);
+      render(<Header {...props} />);
+      const input = screen.getByRole('spinbutton');
+      expect(input.value).toBe(String(lookback));
     });
 
-    it('renders state.ownInputValue when it is not `undefined` regardless of props.lookback', () => {
```
yurishkuro, line 70: what about this test?

vishvamsinh28:
LGTM!

@yurishkuro can you take another look at the tests too?

packages/jaeger-ui/src/components/QualityMetrics/Header.test.js
```diff
@@ -50,47 +49,114 @@ describe('Header', () => {
...
     });
 
     it('renders state.ownInputValue when it is not `undefined` regardless of props.lookback', () => {
-      const ownInputValue = 27;
-      wrapper.setState({ ownInputValue });
-      expect(wrapper.find(InputNumber).prop('value')).toBe(ownInputValue);
+      const { rerender } = render(<Header {...props} />);
+      const input = screen.getByRole('spinbutton');
+
+      // Set the internal state by triggering user input (simulates setState({ ownInputValue: 27 }))
```
yurishkuro: > simulates setState({ ownInputValue: 27 })

how does it do that? `ownInputValue` in the test is only a local variable name (you could easily rename it to `x`), not the same as `state.ownInputValue`

ghost: I've updated it to be clearer about simulating a user event. Thanks!

yurishkuro (approved):

PR #2868, opened 2025-06-07
test: Migrate highlightMatches.test.js and ListItem.test.js from Enzyme to RTL

vishvamsinh28:
Rewrote highlightMatches.test.js and ListItem.test.js using React Testing Library, removing Enzyme and snapshot usage. Preserved full test coverage with DOM-based assertions for all existing test cases.


packages/jaeger-ui/src/components/common/FilteredList/ListItem.test.js
```diff
@@ -30,87 +29,115 @@ describe('<ListItem>', () => {
...
-    props.data.setValue.mockReset();
-    wrapper = shallow(<ListItem {...props} />);
+    defaultProps.data.setValue.mockReset();
   });
 
-  it('renders without exploding', () => {
-    expect(wrapper).toMatchSnapshot();
+  it('renders and highlights correct option', () => {
+    renderItem();
+    expect(screen.getByText('a')).toBeInTheDocument();
```
yurishkuro: we should use full words like apple/banana. Who knows which part of UI is being picked up just by letter "a", maybe it's part of some label.

yurishkuro (approved):

PR #2877, opened 2025-06-11
Migrate `UiFindInput.test.js` from Enzyme to RTL

ghost:
- Resolves part of #1668 

## Description of the changes
- Replaced Enzyme tests with RTL in `UiFindInput.test.js`

## Checklist


ghost:
@yurishkuro @vishvamsinh28 
I've made changes following your previous feedbacks on my PRs. Could you please review it and let me know if everything is fine?

packages/jaeger-ui/src/components/common/UiFindInput.test.js
```diff
@@ -112,58 +119,63 @@ describe('UiFind', () => {
       });
     });
 
-    it('no-ops if value is unchanged', () => {
-      wrapper.find(Input).simulate('change', { target: { value: '' } });
+    it('no-ops if value is unchanged', async () => {
```
vishvamsinh28, line 145: The current implementation of no-ops if value is unchanged doesn't actually test the same scenario as the original. The test unmounts and remounts the component, which resets context and breaks continuity. We need to validate that typing the same value as props.uiFind doesn’t trigger updateUiFind

packages/jaeger-ui/src/components/common/UiFindInput.test.js
```diff
@@ -56,42 +56,50 @@ describe('UiFind', () => {
...
-      expect(wrapper.find(Input).prop('value')).toBe(uiFind);
+      render(<UnconnectedUiFindInput {...props} uiFind={uiFind} />);
+      expect(screen.getByPlaceholderText('Find...')).toHaveValue(uiFind);
     });
 
-    it('renders state.ownInputValue when it is not `undefined` regardless of props.uiFind', () => {
-      wrapper.setProps({ uiFind });
-      wrapper.setState({ ownInputValue });
-      expect(wrapper.find(Input).prop('value')).toBe(ownInputValue);
+    it('renders state.ownInputValue when it is not `undefined` regardless of props.uiFind', async () => {
```
vishvamsinh28, line 80: Multiple tests check input values but don't verify the underlying state behavior that drives the component logic. The test name suggests we're validating state precedence logic, but we're only checking the final rendered value, not the decision-making process.

packages/jaeger-ui/src/components/common/UiFindInput.test.js
```diff
@@ -112,58 +119,63 @@ describe('UiFind', () => {
...
-    it('clears state.ownInputValue', () => {
-      wrapper.setState({ ownInputValue });
-      expect(wrapper.state('ownInputValue')).toBe(ownInputValue);
-      wrapper.find(Input).simulate('blur');
-      expect(wrapper.state('ownInputValue')).toBe(undefined);
-    });
-
-    it('triggers pending queryParameter updates', () => {
-      wrapper.find(Input).simulate('blur');
+    it('clears state.ownInputValue', async () => {
```
vishvamsinh28, line 165: The test should verify that after blur, if we type again, the component uses props.uiFind as the starting point, proving internal state was actually cleared.

packages/jaeger-ui/src/components/common/UiFindInput.test.js
```diff
@@ -112,58 +119,63 @@ describe('UiFind', () => {
...
+      expect(screen.queryByTestId('clear-icon')).not.toBeInTheDocument();
     });
 
     it('clears value immediately when clicked', () => {
-      wrapper.setProps({ uiFind });
-      findIcon().find(IoClose).simulate('click');
-
+      render(<UnconnectedUiFindInput {...props} allowClear uiFind={uiFind} />);
+      const clearIcon = screen.getByTestId('clear-icon');
+      clearIcon.dispatchEvent(new MouseEvent('click', { bubbles: true }));
```
vishvamsinh28: Should use await userEvent.click(clearIcon) for more realistic user interaction testing.

yurishkuro (approved):

PR #2895, opened 2025-06-19
test: migrated TimelineViewingLayer tests from Enzyme to React Testing Library

vishvamsinh28:
Rewrote TimelineViewingLayer tests using React Testing Library. Removed Enzyme-specific logic and improved test clarity and coverage.


yurishkuro (commented):
it is hard to compare the updated tests with the originals. Why can't we keep the same test structure / headers and only change their implementation? That makes it much easier to compare.

vishvamsinh28:
@yurishkuro sure, i was trying to improvise them but lets keep it same

yurishkuro (approved):

PR #2898, opened 2025-06-19
test: Migrate KeyValuesTable Tests from Enzyme to React Testing Library 

vishvamsinh28:
Migrated tests from Enzyme to React Testing Library for improved reliability and alignment with modern best practices.


packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.test.js
```diff
@@ -68,96 +90,122 @@ describe('<KeyValuesTable>', () => {
...
+
+    const valueElements = document.querySelectorAll('.ub-inline-block');
+    expect(valueElements).toHaveLength(data.length);
+
+    valueElements.forEach((valueDiv, i) => {
       if (data[i].expected) {
-        expect(valueDiv.render().text()).toBe(data[i].expected);
+        expect(valueDiv).toHaveTextContent(data[i].expected);
       } else if (data[i].snapshot) {
-        expect(valueDiv).toMatchSnapshot();
```
yurishkuro, line 104: I don't think this is equivalent. If we're not using snapshot then the `expected` parameter of the test should contain the desired rendered value.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.test.js
```diff
@@ -62,102 +84,137 @@ describe('<KeyValuesTable>', () => {
...
-    // so instead rely on jest snapshot comparison.
-    // Key observations:
-    // - "<xss>" key is encoded as &lt;xss&gt;
-    // - link value is rendered as <a>
-    // - xss_link value is rendered as plain string
-    { key: 'jsonkey', value: JSON.stringify(jsonValue), snapshot: true },
+    {
+      key: 'jsonkey',
+      value: JSON.stringify(jsonValue),
+      expected: ['world', 'safe', 'https://example.com', 'true', '42'],
```
yurishkuro: jsonValue has 9 entries. Let's make sure we validate all of them. This is one example where a snapshot would actually be useful, since it explicitly shows what the context should be. Could we do this manually, i.e. manually define how the DOM would look given these values and compare against that?

vishvamsinh28: done

vishvamsinh28: done

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.test.js
```diff
@@ -62,102 +84,149 @@ describe('<KeyValuesTable>', () => {
...
+        expect(valueDiv).toHaveTextContent('safe');
+        expect(valueDiv).toHaveTextContent('https://example.com');
+        expect(valueDiv).toHaveTextContent('https://example.com with "quotes"');
+        expect(valueDiv).toHaveTextContent('true');
+        expect(valueDiv).toHaveTextContent('42');
+        expect(valueDiv).toHaveTextContent('null');
+        expect(valueDiv).toHaveTextContent('x');
+        expect(valueDiv).toHaveTextContent('b');
+        expect(valueDiv).toHaveTextContent('y');
+        expect(valueDiv).toHaveTextContent('a');
```
yurishkuro: * this is easy to match on unrelated letters. We can use longer words here to avoid spurious matches.
* you can include these in the `expected` field of the test rather than explicitly listing here
* still missing some checks mentioned in "Key observations" in L73, e.g. checking for a link, and for escaped XML tag

vishvamsinh28: done

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.test.js
```diff
@@ -57,107 +79,151 @@ describe('<KeyValuesTable>', () => {
     object: { a: 'b', x: 'y' },
```
yurishkuro: please replace with longer words to not accidentally match on letters from other words

yurishkuro:
```suggestion
    object: { apple: 'banana', melon: 'mango' },
```

vishvamsinh28: done

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/KeyValuesTable.test.js
```diff
@@ -54,110 +76,154 @@ describe('<KeyValuesTable>', () => {
...
+        'safe',
+        'https://example.com',
+        'https://example.com with "quotes"',
+        'true',
+        '42',
+        'null',
+        'x',
+        'y',
+        'b',
+        'a',
```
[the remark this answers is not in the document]
yurishkuro: +1 - and this is a perfect illustration of why I wanted to change to longer words - the test still passes because it just finds these letters in other words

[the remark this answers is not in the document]
vishvamsinh28: made changes

yurishkuro (approved):

PR #2902, opened 2025-06-24
Convert `getNodeRenderers.test.js` from Enzyme to RTL

ghost:
- Resolves part of #1668 

## Description of the changes
- Convert `getNodeRenderers.test.js` from Enzyme to RTL

## Checklist


yurishkuro (commented):
@vishvamsinh28  please review

vishvamsinh28:
1. Test Intent Preservation Issue – vectorBorder
- The original Enzyme test used a snapshot to validate the full rendered output.
- The migrated RTL version uses individual assertions, which are more specific—but it misses the base className validation.
- Missing: The assertion for the always-present DdgNode--VectorBorder class.
- Add expect(circle).toHaveClass('DdgNode--VectorBorder') to restore complete validation.

2. Missing Validation – htmlEmphasis
- The test suite verifies some classNames but omits validation for the base DdgNode--HtmlEmphasis class, which is always present when a <div> is rendered.
- Missing: Base class check in tests that expect a <div> return.
- Add expect(div).toHaveClass('DdgNode--HtmlEmphasis') in applicable test cases.

cc : @yurishkuro

vishvamsinh28:
- Correct the `vectorFindColorBand` test setup
```
it('returns null if provided vertex is not in set', () => {
  // Keep the original logic - test with a vertex key that's not in the set
  const findMatches = new Set([`not-${key}`]);
  expect(getNodeRenderers(findMatches, new Map()).vectorFindColorBand(lv)).toBe(null);
});
```

- Add missing attribute validation to `vectorBorder`
```
// Add this assertion to verify the vector effect
expect(circle).toHaveAttribute('vectorEffect', 'non-scaling-stroke');
```

- Ensure base class validation in `htmlEmphasis`
The current tests are checking for the base class, but it would be clearer to make this explicit in all test cases where a div is rendered.

ghost:
@vishvamsinh28 Thanks for the feedback!
1. the test setup below already correctly validates that null is returned when the provided vertex's key is not present in the findMatches set. Acc to me this implementation is better because it uses the correct data type (Set of strings) that matches how the function actually works, while the previous one creates a set of objects which causes a type mismatch. What do you think about this?
```
it('returns null if provided vertex is not in set', () => {
   const findMatches = new Set([`not-${key}`]);
   expect(getNodeRenderers(findMatches, new Map()).vectorFindColorBand(lv)).toBe(null);
});
```

2. I have already added the base class `DdgNode--HtmlEmphasis` in all tests where a div is rendered, in my previous commit

3. `expect(circle).toHaveAttribute('vector-effect', 'non-scaling-stroke');` added this to verify the vector effect

vishvamsinh28:
1. Test Intent Preservation - vectorBorder Test
The original Enzyme test used comprehensive snapshot testing to validate the complete rendered output with all class combinations. The migrated RTL test attempts to manually verify each class condition, but there's a fundamental logic error

```
// ISSUE: This test setup is incorrect
const findMatches = new Set(findMatch ? [testLv.vertex.key] : undefined);
```

When findMatch is false, you're passing undefined to the Set constructor, which creates a Set with one element: undefined. This means findMatches.has(testLv.vertex.key) will always be false when testing the non-findMatch case, but the Set isn't actually empty.

try this 

`const findMatches = new Set(findMatch ? [testLv.vertex.key] : []);`

vishvamsinh28:
LGTM!

@yurishkuro could you also take a look ?

vishvamsinh28 (approved):

packages/jaeger-ui/src/components/DeepDependencies/Graph/getNodeRenderers.test.js
```diff
@@ -41,12 +42,45 @@ describe('getNodeRenderers', () => {
...
               const vms = new Map([[key, vm]]);
-              expect(shallow(getNodeRenderers(findMatches, vms).vectorBorder(testLv))).toMatchSnapshot();
+              const { container } = render(getNodeRenderers(findMatches, vms).vectorBorder(testLv));
+              const circle = container.querySelector('circle');
+
+              expect(circle).toBeInTheDocument();
+              expect(circle).toHaveClass('DdgNode--VectorBorder');
+              expect(circle).toHaveAttribute('vector-effect', 'non-scaling-stroke');
+
+              if (findMatch) {
```
yurishkuro: instead of adding all this boilerplate, define an inline helper function
```
function expectClass(className, shouldHave) {
  if (shouldHave) {
    expect(circle).toHaveClass(className);
  } else {
    expect(circle).not.toHaveClass(className);
  }
}

expectClass('is-findMatch', findMatch);
...
```

ghost:
```
Summary of all failing tests
 FAIL  src/components/TracePage/TraceStatistics/index.test.js (26.217 s)
  ● <TraceTagOverview> › check handler

    thrown: "Exceeded timeout of 5000 ms for a test.
    Add a timeout value to this test to increase the timeout, if this is a long-running test. See https://jestjs.io/docs/api#testname-fn-timeout."

      74 |   });
      75 |
    > 76 |   it('check handler', async () => {
         |   ^
      77 |     let componentRef;
      78 |     const TestWrapper = () => {
      79 |       const ref = React.useRef();

      at it (src/components/TracePage/TraceStatistics/index.test.js:76:3)
      at Object.describe (src/components/TracePage/TraceStatistics/index.test.js:26:1)
```

This is an exceeded timeout error...not related to my changes

yurishkuro (approved):

yurishkuro (dismissed):

yurishkuro (dismissed):

yurishkuro (dismissed):

yurishkuro (approved):

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.test.js
```diff
@@ -26,6 +26,9 @@ const transformedTrace = transformTraceData(testTrace);
 describe('<TraceTagOverview>', () => {
   let defaultProps;
 
+  // FIXME Increase timeout for this test suite due to slow Enzyme/React 18 mounting
+  jest.setTimeout(10000);
```
yurishkuro, line 30: @vishvamsinh28 I added a temporary patch to allow CI to succeed

PR #2905, opened 2025-06-26
test: Migrate <OperationTableDetails> Tests from Enzyme to React Testing Library

vishvamsinh28:
Refactored the test suite for the OperationTableDetails component to use React Testing Library instead of Enzyme.


packages/jaeger-ui/src/components/Monitor/ServicesView/operationDetailsTable/index.test.js
```diff
@@ -49,130 +61,131 @@ describe('<OperationTableDetails>', () => {
...
+    expect(screen.getByText('/PlaceOrder')).toBeInTheDocument();
   });
 
   it('render latency in seconds in the table', () => {
-    const cloneServiceOpsMetrics = {};
-    Object.assign(cloneServiceOpsMetrics, serviceOpsMetrics[0]);
+    const cloneServiceOpsMetrics = { ...serviceOpsMetrics[0] };
     cloneServiceOpsMetrics.latency = 8000;
+    render(<OperationTableDetails {...props} data={[cloneServiceOpsMetrics]} loading={false} />);
+    // this test in original just creates the data but doesn't verify it, so we just render it
```
yurishkuro: can we not do at least a single lookup for some latency values in the output?

packages/jaeger-ui/src/components/Monitor/ServicesView/operationDetailsTable/index.test.js
```diff
@@ -49,130 +61,131 @@ describe('<OperationTableDetails>', () => {
...
-    const cloneServiceOpsMetrics = {};
-    Object.assign(cloneServiceOpsMetrics, serviceOpsMetrics[0]);
+    const cloneServiceOpsMetrics = { ...serviceOpsMetrics[0] };
     cloneServiceOpsMetrics.requests = 0.02;
-    wrapper.setProps({ ...props, data: [cloneServiceOpsMetrics], loading: false });
-    expect(wrapper).toMatchSnapshot();
+    const { container } = render(
+      <OperationTableDetails {...props} data={[cloneServiceOpsMetrics]} loading={false} />
+    );
+    expect(container.querySelector('table')).toBeInTheDocument();
```
yurishkuro: doesn't seem equivalent. Why is "lower than 0.1 request rate value" critical? How is it rendered? Can we not test for that specific rendering?

packages/jaeger-ui/src/components/Monitor/ServicesView/operationDetailsTable/index.test.js
```diff
@@ -49,130 +61,131 @@ describe('<OperationTableDetails>', () => {
...
-    const cloneServiceOpsMetrics = {};
-    Object.assign(cloneServiceOpsMetrics, serviceOpsMetrics[0]);
+    const cloneServiceOpsMetrics = { ...serviceOpsMetrics[0] };
     cloneServiceOpsMetrics.requests = 0.2888;
-    wrapper.setProps({ ...props, data: [cloneServiceOpsMetrics], loading: false });
-    expect(wrapper).toMatchSnapshot();
+    const { container } = render(
+      <OperationTableDetails {...props} data={[cloneServiceOpsMetrics]} loading={false} />
+    );
+    expect(container.querySelector('table')).toBeInTheDocument();
```
yurishkuro: same as above - the original test was validating the rendering via snapshot, how are we doing it here? How does `requests = 0.2888` manifest differently?

I realize that it's hardly possible to validate how the actual graphical trend lines are rendered, but this is a "table" component, so it should also be rendering values in the table directly, and we should be able to validate those.

vishvamsinh28: done

packages/jaeger-ui/src/components/Monitor/ServicesView/operationDetailsTable/index.test.js
vishvamsinh28: done

packages/jaeger-ui/src/components/Monitor/ServicesView/operationDetailsTable/index.test.js
vishvamsinh28: done

packages/jaeger-ui/src/components/Monitor/ServicesView/operationDetailsTable/index.test.js
```diff
@@ -49,130 +61,127 @@ describe('<OperationTableDetails>', () => {
...
-    expect(wrapper.state('hoveredRowKey')).toBe(-1);
+    expect(screen.getByText('View traces')).toBeInTheDocument();
 
-    wrapper.find('.table-row').at(0).simulate('mouseenter');
-    expect(wrapper.state('hoveredRowKey')).toBe(0);
+    fireEvent.mouseLeave(tableRow);
 
-    wrapper.find('.table-row').at(0).simulate('mouseleave');
-    expect(wrapper.state('hoveredRowKey')).toBe(-1);
+    expect(screen.queryByText('View traces')).not.toBeInTheDocument();
```
yurishkuro, line 180: how does this test "highlights the row"? Should we be detecting some style change?

What's the equivalent of previous
```
expect(wrapper.state('hoveredRowKey')).toBe(0);
expect(wrapper.state('hoveredRowKey')).toBe(-1);
```

vishvamsinh28: RTL can't check component state, so added a 'table-row--hovered' class to hovered rows. This allows us to test row highlighting based on actual DOM changes.

yurishkuro (approved):

PR #2912, opened 2025-06-28
test: Migrate DetailsPanel tests from Enzyme to React Testing Library

vishvamsinh28:
Migrated all DetailsPanel component tests from Enzyme to React Testing Library, maintaining the original test cases and coverage while improving test clarity and reliability.


yurishkuro:
@codecov-ai-reviewer

packages/jaeger-ui/src/components/DeepDependencies/SidePanel/DetailsPanel.test.js
```diff
@@ -57,247 +75,195 @@ describe('<SidePanel>', () => {
...
   });
 
   describe('fetchDetails', () => {
-    it('fetches correct details url, perferring op-scoped details, or does not fetch at all', async () => {
-      const details = 'test details';
-      const columnDefs = ['test', 'column', 'defs'];
-
-      const tests = [];
-
-      ['detailUrl#{service}', undefined].forEach(detailUrl => {
```
yurishkuro, line 69: omg what was going on here?

yurishkuro (dismissed):

yurishkuro:
codecov complains about this gap
image

packages/jaeger-ui/src/components/DeepDependencies/SidePanel/DetailsPanel.test.js
vishvamsinh28: This removed code was generating 1,152 test combinations using 10 nested forEach loops that tested every possible combination of parameters (detailUrl, detailPath, operation, error scenarios, etc.). Each loop multiplied the test count: 2^7 × 3^2 = 1,152 individual test cases all running in a single test function. This approach was extremely slow to execute, nearly impossible to debug when failures occurred (imagine trying to identify which specific combination out of 1,152 failed), and tested many unrealistic edge cases that would never happen in practice. The new RTL tests replaced this with ~6 focused, readable test cases that cover the actual use cases while being much faster and easier to maintain.

vishvamsinh28: This removed code was generating 1,152 test combinations using 10 nested forEach loops that tested every possible combination of parameters (detailUrl, detailPath, operation, error scenarios, etc.). Each loop multiplied the test count: 2^7 × 3^2 = 1,152 individual test cases all running in a single test function. This approach was extremely slow to execute, nearly impossible to debug when failures occurred (imagine trying to identify which specific combination out of 1,152 failed), and tested many unrealistic edge cases that would never happen in practice. The new RTL tests replaced this with readable test cases that cover the actual use cases while being much faster and easier to maintain.

vishvamsinh28:
> codecov complains about this gap image

added an extra test to fix it

yurishkuro (approved):

PR #2913, opened 2025-06-29
Convert `serviceGraph.tsx` from class component to functional component

Parship999:
- Resolves part of  #2610 

## Description of the changes
- Converted the following components to functional components:
  - `serviceGraph.tsx`
- Related changes:
  - `serviceGraph.test.js`

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/Monitor/ServicesView/serviceGraph.test.js
+++ b/packages/jaeger-ui/src/components/Monitor/ServicesView/serviceGraph.test.js
@@ -15,9 +15,28 @@
 import React from 'react';
 import { render, screen, cleanup } from '@testing-library/react';
 import '@testing-library/jest-dom';
-import { Legend, Tooltip } from 'recharts';
-import ServiceGraph, { ServiceGraphImpl, tickFormat, Placeholder } from './serviceGraph';
+import ServiceGraph, {
+  tickFormat,
+  Placeholder,
+  getData,
+  getMetricsData,
+  calculateYDomain,
+  calculateYAxisTicks,
+  formatYAxisTick,
+  calculateNumericTicks,
+} from './serviceGraph';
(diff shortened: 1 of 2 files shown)


packages/jaeger-ui/src/components/Monitor/ServicesView/serviceGraph.test.js
```diff
@@ -15,16 +15,41 @@
 import React from 'react';
 import { render, screen, cleanup } from '@testing-library/react';
 import '@testing-library/jest-dom';
-import { Legend, Tooltip } from 'recharts';
```
yurishkuro, line 18: why is the test being changed so much? Our common practice is to refactor to functional components __without__ changing the tests, to guarantee that the behavior is the same. When you change both there's no way to tell.

yurishkuro (changes requested):
there are too many changes that are unrelated to converting to a functional component.

Parship999:
@yurishkuro I found two approaches:
1. Retain the class component as a named export for backward compatibility while introducing a functional component as the default export, ensuring all tests pass without modification.

2. completely refactor to a functional component, removing class-based logic and updating some tests.

Which approach would you recommend?

vishvamsinh28:
@Parship999 Too many changes, it's hard to review.

Parship999:
> @Parship999 Too many changes, it's hard to review.

I understand there are too many changes. Once I confirm which approach is better in this case, I’ll work on minimizing them accordingly. could you please let me know which of the two approaches you’d recomend?

yurishkuro:
@Parship999 please describe why the tests need to change so significantly. Your many other PRs only touched the components themselves, not the tests, I would like that approach better. If the tests somehow require changes please elaborate and think if it's possible to do a small preliminary change first. For example, in another PR I just approved there was a static function of the class which you changed to exported function and that require some changes at the caller callsites, we could make those as a separate PR.

PR #2915, opened 2025-07-01
Convert `QualityMetrics/Header.tsx` from class component to functional component

ghost:
- Resolves part of #2610

## Description of the changes
- Converted the following component to functional component: `Header.tsx`

## Checklist


vishvamsinh28:
The current implementation has a fundamental flaw with how the debounced function is being created. You're using useCallback to wrap the debounce call, but this actually creates a new debounced function every time the setLookback prop changes. This completely breaks the debouncing functionality because each new function call gets its own internal timer, rather than sharing one timer across all calls.

- First, you need to change useCallback to useMemo for the debounced function creation. The key difference is that useMemo will create the debounced function once and reuse it, while useCallback was creating a new one each time.
- Second, you need to add proper cleanup logic using useEffect. When the component unmounts or the debounced function changes, you should call the cancel method on the debounced function to prevent memory leaks and avoid executing stale callbacks.
- Third, you'll need to update your imports to include useMemo and useEffect from React.
- Fourth, there's an inconsistency in the parameter type for handleInputChange. You've added string to the union type, but this doesn't match the original class component and may not be necessary based on how InputNumber works.

ghost:
Thanks for the feedback! I've updated the code.

This actually broke the tests in `Header.test.js` because the component's new cleanup logic called a `.cancel()` method that didn't exist on the lodash/debounce mock. I fixed the test mock by adding a `.cancel()` function, now it makes the mock accurately reflect the library's behavior and matched the component's code.

Parship999:
@vishvamsinh28 @yurishkuro please review

ghost:
@vishvamsinh28 @yurishkuro please review.
for the codecov error just a small refactor has done, no changes to logic or tests, everything works exactly as before.

yurishkuro (approved):

PR #2921, opened 2025-07-03
Fix eslint warnings in path-agnostic-decorations component

ghost:
Screenshot 2025-07-04 031032

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/actions/path-agnostic-decorations.tsx
+++ b/packages/jaeger-ui/src/actions/path-agnostic-decorations.tsx
@@ -73,7 +73,7 @@ export function getDecoration(
 
   pendingCount = pendingCount ? pendingCount + 1 : 1;
   const { summaryUrl, opSummaryUrl, summaryPath, opSummaryPath } = schema;
-  let promise: Promise<Record<string, any>>;
+  let promise: Promise<Record<string, unknown>>;
   let getPath: string;
   let setPath: string;
   if (opSummaryPath && opSummaryUrl && operation) {
@@ -116,7 +116,10 @@ const fullActions = createActions<Promise<TNewData> | undefined>({
   [actionTypes.GET_DECORATION]: getDecoration,
 });
 
-export default (fullActions as any).jaegerUi.pathAgnosticDecorations as Record<
-  string,
-  ActionFunctionAny<Action<Promise<TNewData> | undefined>>
->;
+export default (
+  fullActions as unknown as {
+    jaegerUi: {
+      pathAgnosticDecorations: Record<string, ActionFunctionAny<Action<Promise<TNewData> | undefined>>>;
+    };
+  }
+).jaegerUi.pathAgnosticDecorations;


yurishkuro:
why is `unknown` better than `any`? I understand that you're trying to fix the lint error, but I don't think the spirit of that error is to suggest this kind of change.

ghost:
> why is `unknown` better than `any`? I understand that you're trying to fix the lint error, but I don't think the spirit of that error is to suggest this kind of change.

Thanks for the feedback!
`unknown` is type-safe, it tells the compiler "I don't know what this is," forcing to perform checks before using the value. This prevents runtime errors. 

With `Record<string, any>`, we are telling TypeScript that the promise will resolve to an object, and the values in that object can be used as anything without checks. wherease using `Record<string, unknown>`, is like we don't know the types of the values. TypeScript will then force us to check the type before you can perform any operations.

and for the export, using `as unknown as` type is a more safer way to cast. It forces us to assert the exact shape of the `fullActions` object, unlike `as any`, which would have hidden potential bugs from typos

https://stackoverflow.com/questions/51439843/unknown-vs-any/51439876#:~:text=16%20bronze%20badges-,8,without%20checking%20the%20type%20first.

yurishkuro:
Ok makes sense. But why do we have either "any" or "unknown" in this code instead of concrete type? It looks like someone might have used any as a quick hack instead of figuring out the right type.

ghost:
> Ok makes sense. But why do we have either "any" or "unknown" in this code instead of concrete type? It looks like someone might have used any as a quick hack instead of figuring out the right type.

`getDecoration` can be configured to fetch data from completely different endpoints

Its behavior is determined by the schema it finds based on the id provided. This schema contains two information:
A URL: `summaryUrl` or `opSummaryUrl`
A path t data: `summaryPath` or `opSummaryPath`.
One schema might extract a number, another might extract a string
Because `getDecoration` must handle both of these completely different response shapes, it cannot have one specific, concrete type.

```
const { summaryUrl, opSummaryUrl, summaryPath, opSummaryPath } = schema;
let promise: Promise<Record<string, unknown>>;
let getPath: string;
if (opSummaryPath && opSummaryUrl && operation) {
  promise = fetchDecoration(stringSupplant(opSummaryUrl, { service, operation }));
  getPath = stringSupplant(opSummaryPath, { service, operation });
} else {
  promise = fetchDecoration(stringSupplant(summaryUrl, { service }));
  getPath = stringSupplant(summaryPath, { service });
}
promise
  .then(res => {
    return _get(res, getPath, `\`${getPath}\` not found in response`);
  })
```

yurishkuro:
> Because getDecoration must handle both of these completely different response shapes, it cannot have one specific, concrete type.

why can it not have a generic type argument? This is what generics are for.

ghost:
A generic type argument doesn't work because the specific shape of the data is determined at runtime inside the function, not by the person calling it. The caller has no way of knowing what type to provide.

The function's behavior is which URL to hit and what data to extract depends on the id. The caller doesn't know these details.
```
// The caller provides an ID, not a type.
export function getDecoration(id: string, ...): Promise<TNewData> | undefined {
........
  // The function uses the ID to find the correct configuration.
  const schema = getDecorationSchema(id);
.......
}
```
Please let me know if I’ve misunderstood anything

yurishkuro (approved):

yurishkuro:
> The caller doesn't know these details.

why wouldn't the caller know? They are (a) formulating the query to retrieve specific type of data, and (b) then interpreting the results as that data type. It is possible that the caller itself doesn't care about the type but that just tells us there is a higher level caller who does care.

ghost:
yes the component displaying the data does know the type, but this function runs as a redux action, it only receives the id and has no context from the origignal caller.
Thanks for the discussion on this, it was a good learning experience

PR #2930, opened 2025-07-05
Fix eslint warnings in jaeger.js, NotFound and TraceIDSearchInput component

ghost:
Screenshot 2025-07-06 030056

### NotFound.tsx

* **Warning**: The code used `error: any`, which disables TypeScript's type-checking and is considered unsafe practice.
* **Change**: replaced `any` with a more specific type: `error: Error | string;`. This change makes the component type-safe.

### TraceIDSearchInput.tsx

* **Warning**: The code unsafely cast the form event to `any` to access input field value.
* **Change**: Used `event.currentTarget`, accessed the element by name with `.namedItem('idInput')`, and used optional chaining (`?.value`). It correctly identifies the form element, finds the input, and safely retrieves its value, also preventing runtime error even if the input is not found.

### jaeger.js
* The warnings `'_' is defined but never used` occur because a variable `_` is declared in a catch block but never used.

## Checklist


yurishkuro:
> Change: replaced any with a more specific type: error?: Error | string | null;. This change makes the component type-safe. It clearly define that the error prop can be a standard Error object, a simple string, or null.

seems to work just fine without null.

ghost:
@yurishkuro I have resolved the merge conflict in `TraceIDSearchInput.tsx`

ghost:
@yurishkuro the changes are ready and just need approval now

yurishkuro (approved):

PR #2955, opened 2025-07-09
test: Migrate SearchResults Component Tests from Enzyme to React Testing Library

vishvamsinh28:
Refactored the SearchResults test suite to use React Testing Library instead of Enzyme.


packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.test.js
```diff
@@ -13,139 +13,178 @@
...
-    expect(wrapper.find(LoadingIndicator).length).toBe(1);
+    render(<SearchResults {...baseProps} loading />);
+    expect(screen.getByTestId('loading')).toBeInTheDocument();
   });
 
   it('hide scatter plot if queryparam hideGraph', () => {
-    wrapper.setProps({ hideGraph: true, embed: true, getSearchURL: () => 'SEARCH_URL' });
-    expect(wrapper.find(ScatterPlot).length).toBe(0);
+    render(<SearchResults {...baseProps} hideGraph />);
+    expect(screen.queryByTestId('scatterplot')).not.toBeInTheDocument();
```
yurishkuro, line 122: nit: I think this is a bad practice for test. Asserting that something is not in the document can succeed for many reasons, maybe it's never in the document. A robust pattern for a test is to render with hideGraph=false, asset that the scallerplot IS in the document, then re-render with hideGraph=true and asset that it's no longer there.

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.test.js
```diff
@@ -13,139 +13,178 @@
...
   });
 
   it('hide DiffSelection when disableComparisons = true', () => {
-    wrapper.setProps({ disableComparisons: true });
-    expect(wrapper.find(DiffSelection).length).toBe(0);
+    render(<SearchResults {...baseProps} disableComparisons />);
+    expect(screen.queryByTestId('diffselection')).not.toBeInTheDocument();
```
yurishkuro, line 132: same comment as above

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.test.js
```diff
@@ -13,139 +13,178 @@
...
+        traceName: 'T',
+        startTime: 0,
+        duration: 1,
+        processes: {},
+        spans: [{ tags: [{ key: 'error', value: true }] }],
+      },
+    ];
+    render(<SearchResults {...baseProps} traces={errorTrace} />);
+    const scatterProps = ScatterPlot.mock.calls[0][0];
+    expect(scatterProps.data[0].color).toBe('red');
```
yurishkuro, line 168: this test for red color inside the ScatterPlot component, right? Why isn't it a test on ScatterPlot component, rather than SearchResults?

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.test.js
```diff
@@ -13,139 +13,178 @@
...
     it('deep links traces', () => {
       const uiFind = 'ui-find';
-      const spanLinks = {
-        [traces[0].traceID]: uiFind,
-      };
-      wrapper.setProps({ spanLinks });
-      const results = wrapper.find(ResultItem);
-      expect(results.at(0).prop('linkTo').search).toBe(`uiFind=${uiFind}`);
-      expect(results.at(1).prop('linkTo').search).toBeUndefined();
+      const spanLinks = { [baseTraces[0].traceID]: uiFind };
```
yurishkuro: confusing naming
```suggestion
      const spanLinks = { [baseTraces[0].traceID]: 'foobar' };
```
then check for 'foobar' below

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.test.js
```diff
@@ -13,139 +13,178 @@
...
+      render(<SearchResults {...baseProps} spanLinks={spanLinks} />);
+      const [first, second] = ResultItem.mock.calls;
+      expect(first[0].linkTo.search).toBe(`uiFind=${uiFind}`);
```
yurishkuro:
```suggestion
      expect(first[0].linkTo.search).toBe('uiFind=foobar');
```

vishvamsinh28:
@yurishkuro working on this

vishvamsinh28:
@yurishkuro updated the test

yurishkuro (approved):

PR #2957, opened 2025-07-10
Fix UI bug of convertServiceErrorRateToPercentages

pipiland2612:
- Resolves bug where y is math.NaN or null returning 0. Specifically doing `return { ...metricPoint, y: metricPoint.y! * 100 };` will return 0 for y = null or math.NaN

## Description of the changes
- Add explicit check to prevent this bugs

- npm run test

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/Monitor/ServicesView/index.tsx
+++ b/packages/jaeger-ui/src/components/Monitor/ServicesView/index.tsx
@@ -134,7 +134,13 @@ const convertServiceErrorRateToPercentages = (serviceErrorRate: null | ServiceMe
   if (!serviceErrorRate) return null;
 
   const convertedMetricsPoints = serviceErrorRate.metricPoints.map((metricPoint: Points) => {
-    return { ...metricPoint, y: metricPoint.y! * 100 };
+    const y = metricPoint.y;
+
+    if (y === null || y === undefined || Number.isNaN(y)) {
+      return { ...metricPoint, y: null };
+    }
+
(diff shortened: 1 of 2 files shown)


pipiland2612:
This bug is found while I develop spm getErrorRates method: https://github.com/jaegertracing/jaeger/pull/7298#issue-3208571811

packages/jaeger-ui/src/components/Monitor/ServicesView/index.tsx
```diff
@@ -134,7 +134,13 @@ const convertServiceErrorRateToPercentages = (serviceErrorRate: null | ServiceMe
...
 
   const convertedMetricsPoints = serviceErrorRate.metricPoints.map((metricPoint: Points) => {
-    return { ...metricPoint, y: metricPoint.y! * 100 };
+    const y = metricPoint.y;
+
+    if (y === null || y === undefined || Number.isNaN(y)) {
+      return { ...metricPoint, y: null };
+    }
+
+    return { ...metricPoint, y: y * 100 };
```
yurishkuro, line 143: what is the meaning of the constant 100 here? If this was some time conversion (like ms to s) I could see it being 1000, but 100?

yurishkuro (approved):

packages/jaeger-ui/src/components/Monitor/ServicesView/index.tsx
pipiland2612: Hi @yurishkuro it's converting from 0.xx to xx %

Issue #2649, opened 2025-02-05
Investigate why CodeQL workflow is failing


yurishkuro:
Example: https://github.com/jaegertracing/jaeger-ui/pull/2518/checks?check_run_id=36715056170

<img width="803" alt="Image" src="https://github.com/user-attachments/assets/cd62f23a-1d7d-447c-b967-0894a85d6d4c" />

ksachin7:
@yurishkuro due to this change here - https://github.com/jaegertracing/jaeger-ui/commit/dd757565b23c00f2dabe35c613a8b4de5b9cf489

yurishkuro:
why? matrix is used to parameterize the jobs. Our job was not using the variable from the matrix in any way.

ksachin7:
sorry I meant this PR - https://github.com/jaegertracing/jaeger-ui/pull/2633

ksachin7:
It started failing after changes in this PR , so I am investigating this

yurishkuro:
DCO check #2633 has nothing to do with CodeQL

abhayporwals:
Hi, @yurishkuro! I was looking into the file and it almost seems fine to me. 
So, the cause of the error maybe in the languages. I'm thinking to add `typescript` with  JS just beacuse the test fails for file with extension `.tsx`

![Image](https://github.com/user-attachments/assets/4251ee34-6146-4249-ba15-ff336b28ecd6)

NischalPaliwal:
Hi, @yurishkuro! I think we should write as:
- name: Initialize CodeQL
        uses: github/codeql-action/init@48ab28a6f5dbc2a99bf1e0131198dd8f1df78169 # v3.28.0
        with:
          languages: [ javascript ]

I think this will solve the issue if I am not wrong!

NischalPaliwal:
Hi @yurishkuro, I think the initialize CodeQL step should be like:
`- name: Initialize CodeQL
        uses: github/codeql-action/init@48ab28a6f5dbc2a99bf1e0131198dd8f1df78169 # v3.28.0
        with:
          languages: javascript-typescript`

NischalPaliwal:
Hi @yurishkuro, is this meaningful to have -
```yml
strategy:
      fail-fast: false

as we do not have any matrix element.

VijetaPriya47:
[...]

PR #2958, opened 2025-07-10
fix: enable CodeQL scan on main to resolve configuration missing error

gkbishnoi07:
- Part of #2649 

## Description of the changes
- This PR updates the `.github/workflows/codeql.yml` to ensure the CodeQL analysis job runs on both:
- `push` to `main` (baseline scan required by GitHub)
- `pull_request` to `main`

Previously, the job was restricted to only pull requests, which caused the error:
> "1 configuration not found on refs/heads/main"

With this update, GitHub Advanced Security will be able to run CodeQL on `main`, allowing proper diff analysis for PRs.

- Raised this PR to verify that the CodeQL job triggers correctly on pull request
- After merging, will push a trivial commit to `main` to confirm baseline scan runs as expected

## Checklist

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,14 +1,10 @@
 name: "CodeQL"
 
-
 on:
   workflow_dispatch:
   merge_group:
-  push:
-    branches: [main]
   pull_request:
     branches: [main]
-
   push:
     branches: [main]
 


gkbishnoi07:
@yurishkuro @jkowall Could you please approve the CodeQL workflow to run for this PR? so i can check it working or not!

gkbishnoi07:
Issue #2649  is resolved — the CodeQL workflow now runs successfully. please review and merge it! thanks

.github/workflows/codeql.yml
```diff
@@ -16,8 +16,6 @@ permissions:
 
 jobs:
   codeql-analyze:
-    # Run only on pull requests, see https://github.com/github/codeql-action/issues/1537
-    if: ${{ github.event_name == 'pull_request' }}
```
yurishkuro, line 20: this is here for a reason (described in the linked issue), you are not fixing the reason this override was added

gkbishnoi07:
@yurishkuro Can you please confirm whether we are using the default setup or the advanced setup for Code Scanning in this repository?
Reference: [GitHub Community Discussion #121836](https://github.com/orgs/community/discussions/121836)

.github/workflows/codeql.yml
gkbishnoi07: The reason we were getting the “1 configuration not found on refs/heads/main” warning is because CodeQL isn't running on the main branch at all, even though it works fine on pull requests.

That’s actually happening because of this line in the workflow:

`if: ${{ github.event_name == 'pull_request' }}`

This line tells the job to run only during pull requests, so when changes are merged to main, CodeQL doesn’t run — meaning GitHub has nothing to compare future PRs against, which causes the warning.

I completely understand that this condition was added to avoid issues with GitHub Merge Queue, where CodeQL checks get stuck if they’re marked as required. So we definitely shouldn’t remove that line, as you pointed out.

gkbishnoi07:
@yurishkuro according to me There are 2 solutions to fix the “1 configuration not found” warning.

Main goal: Run CodeQL once on main to generate a baseline (without removing the if: github.event_name == 'pull_request' line).

Solution 1: Dummy Commit
Make a harmless commit (like adding a dummy file) and merge it into main. This triggers the workflow and lets GitHub register the config on main.

Solution 2: Temporary Workflow
Create a one-time workflow that forces CodeQL to run on main. Trigger it manually or after merge, then delete the workflow.

Both methods work after the merge into main and future PRs won’t show the warning anymore. (We can safely delete the files after CodeQL runs once on the main branch)

yurishkuro:
Can we add a manual dispatch trigger to this workflow and force run it from main? We can also alter the condition to allow running from main or from PR but not from merge queue.

gkbishnoi07:
@yurishkuro Yes, we can definitely add a workflow_dispatch trigger to manually run the workflow from main.

Also, I’ll update the if: condition to allow runs from both pull_request and direct main pushes, while still avoiding merge queue by excluding synthetic refs.

`if: github.event_name == 'pull_request' || github.ref == 'refs/heads/main'`

gkbishnoi07:
@yurishkuro code updated please review!

yurishkuro (approved):

PR #2960, opened 2025-07-11
Convert `SidePanel/index.tsx` from class component to functional component

ghost:
- Resolves part of #2610

## Description of the changes
- Converted the following component to functional component: `SidePanel/index.tsx`

## Checklist


ghost:
four tests were failing because of the behavioral differences between class component lifecycle methods and the `useEffect` hook, which runs after the initial render by default. another was the `clears decoration and tracks clear` test failed because it was checking if the `selectDecoration` function was called with 0 arguments `.toHaveBeenCalledWith()`, but actually the handler calls it with `undefined`.

yurishkuro (approved):

PR #2970, opened 2025-07-14
Added scroll up down buttons in lookback dropdown

### LINKED ISSUE #469: Custom Time Range option is hard to spot
In order to see Custom Time Range option in the Lookback dropdown one has to scroll all the way down, while there's little indication that the list is even scrollable. Is it possible to make the scrollbar always visible? Or any other way to improve the usability?

image

cc @everett980 @tiffon

Parship999:
- Resolves #469

## Description of the changes
- I tried several ways to make the scrollbar always visible by modifying the css file, but it didn't work. so, I added two buttons in the lookback dropdown as a workaround to solve this issue.
- The `consistent-return` rule expects functions to either always return a value or never return a value. In this case the `useEffect` returns a cleanup function (which is the standard React pattern), so i have disabled this rule.
```
    // eslint-disable-next-line consistent-return
    return () => {
      observer.disconnect();
    };
```
- The istanbul ignore comments were added to exclude browser-specific DOM manipulation code (`document.querySelector` and `scrollBy`) that cannot be reliably tested in Jest's JSDOM environment. these lines only execute during real user interactions with scroll buttons and require a full browser environment to test properly. I have added comments for this.

Screenshot 2025-07-13 204120

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/SearchTracePage/SearchForm.jsx
+++ b/packages/jaeger-ui/src/components/SearchTracePage/SearchForm.jsx
@@ -444,6 +444,7 @@ export class SearchFormImpl extends React.PureComponent {
             disabled={submitting}
             defaultValue={DEFAULT_LOOKBACK}
             onChange={value => this.handleChange({ lookback: value })}
+            showScrollButtons
           >
             {optionsWithinMaxLookback(searchMaxLookback)}
             <Option value="custom">Custom Time Range</Option>
--- a/packages/jaeger-ui/src/components/common/SearchableSelect.css
+++ b/packages/jaeger-ui/src/components/common/SearchableSelect.css
@@ -1,0 +1,71 @@
+/*
+Copyright (c) 2025 The Jaeger Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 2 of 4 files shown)


yurishkuro:
@vishvamsinh28 please review

ghost:
@vishvamsinh28 can you please review this PR?

yurishkuro:
@Parship999 I appreciate the effort but having buttons in a scrollable dropdown is an unusual UX.

ghost:
> * I tried several ways to make the scrollbar always visible by modifying the css file, but it didn't work. so, I added two buttons in the lookback dropdown as a workaround to solve this issue.

I initially tried multiple approaches to make the scrollbar always visible by adjusting the CSS, but unfortunately, those didn’t work as expected. As a workaround, I added the two buttons.

yurishkuro:
maybe you can document which approaches you tried. There should be ways to make the scrollbar visible via CSS settings. Gemini suggests things like
  * `overflow-y: scroll`
  * `-webkit-scrollbar`

ghost:
I mainly tried using overflow-y: scroll, but it didn’t work. I found this approach through stack overflow and AI suggestions, but i don't know why it wasn’t working with the css here. I don’t remember the other approaches clearly since I tested them about a month ago. I can try again later if I find a proper solution.

ghost:
I also tried to use a custom scrollbar component but couldn't hide or disable the antd scrollbar. I think these issues are ocuuring because its an antd component

Issue #2963, opened 2025-07-11
[Feature]: Add pre-commit hooks for automated ESLint and Prettier checks


gkbishnoi07:
### Requirement

As a Jaeger UI developer and contributor I want automated pre-commit hooks that run ESLint, TypeScript, and Prettier checks so that I can ensure code quality and consistency before commits reach the repository, reducing CI/CD failures and review overhead.

### Problem

Currently, there are no automated checks preventing developers from committing code with:

- ESLint linting errors
- TypeScript compilation errors
- Inconsistent code formatting
- Missing license headers
- This leads to:

Increased review time as maintainers catch basic formatting/linting issues
Inconsistent code quality across different contributors
Developer frustration when builds fail due to easily preventable issues

### Proposal

Implement Husky pre-commit hooks that automatically run essential checks before each commit:

- TypeScript compilation (npm run tsc-lint) - Ensure no TS errors
- ESLint checks (npm run eslint) - Enforce linting standards
- Cross-platform compatibility - Work on Windows, macOS, and Linux
- Clear error messaging - Help developers understand and fix issues
- Bypass option - Allow git commit --no-verify for emergency commits

example:-

<img width="1171" height="313" alt="Image" src="https://github.com/user-attachments/assets/6a67533c-f5db-4b6a-88a4-98bdf9253ab0" />

```
echo "🔍 Running pre-commit checks..."echo "🔍 Running pre-commit checks..."
npm run tsc-lint && npm run eslintnpm run tsc-lint && npm run eslint

```

gkbishnoi07:
@yurishkuro @jkowall Could you please take a look at this feature? I'm curious why a pre-hook wasn’t implemented.

yurishkuro:
we used to have them with `husky`

gkbishnoi07:
Why were Husky hooks removed? Should I add them again?

gkbishnoi07:
@yurishkuro should I add again or not?

yurishkuro:
feel free to fix husky setup

gkbishnoi07:
Thanks for the response! I will start working on it

isagar678:
@gkbishnoi07 Are you working on the issue? If not I want to work on that

gkbishnoi07:
Hey @isagar678 Yes, I'm currently working on it. Thanks for checking!

gkbishnoi07:
@yurishkuro can we close this issue?

PR #2983, opened 2025-07-17
Setup Husky pre-commit hook with lint-staged

gkbishnoi07:
- #2963 

## Description of the changes

- Added Husky as a Git hook manager
- Set up pre-commit hook to run lint-staged
- Integrated ESLint, Prettier, and Stylelint with lint-staged
- Updated package.json with prepare script for Husky install
- Created .husky/pre-commit script

- Verified that committing with unformatted or invalid files triggers the pre-commit checks
- Confirmed staged files are linted and fixed automatically before commit
- Manually tested on PowerShell and WSL (Ubuntu) for compatibility
- Checked fallback behavior and commit abortion on errors

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "eslint-plugin-testing-library": "^7.5.4",
     "husky": "9.1.0",
     "jsdom": "^26.0.0",
-    "lint-staged": "^15.2.10",
+    "lint-staged": "^15.5.2",
     "npm-run-all2": "8.0.1",
     "prettier": "^3.6.2",
     "rxjs-compat": "6.6.7",
--- a/package.json
+++ b/package.json
@@ -26,9 +26,11 @@
     "eslint-plugin-testing-library": "^7.5.4",
     "husky": "9.1.0",
     "jsdom": "^26.0.0",
+    "lint-staged": "^15.2.10",
     "npm-run-all2": "8.0.1",
     "prettier": "^3.6.2",
     "rxjs-compat": "6.6.7",
+    "stylelint": "^16.10.0",
     "typescript": "5.8.2"
   },
   "overrides": {
@@ -48,7 +50,7 @@
     "eslint": "eslint --cache 'scripts/*.{js,jsx,ts,tsx}' 'packages/*/src/**/*.{js,jsx,ts,tsx}' 'packages/*/*.{js,jsx,ts,tsx,mts}'",
     "fmt": "npm run prettier",
     "lint": "npm-run-all -ln --parallel prettier-lint tsc-lint eslint check-license",
-    "prepare": "npm run --workspace @jaegertracing/plexus prepublishOnly",
+    "prepare": "husky install && npm run --workspace @jaegertracing/plexus prepublishOnly",
     "prettier": "prettier --write '{.,scripts}/*.{js,jsx,json,md,ts,tsx,mts}' 'packages/*/{src,demo/src}/**/!(layout.worker.bundled|react-vis).{css,js,jsx,json,md,ts,tsx}' 'packages/*/*.{css,js,jsx,json,md,ts,tsx,mts}'",
     "prettier-lint": "prettier --list-different '{.,scripts}/*.{js,jsx,json,md,ts,tsx,mts}' 'packages/*/{src,demo/src}/**/!(layout.worker.bundled|react-vis).{css,js,jsx,json,md,ts,tsx}' 'packages/*/*.{css,js,jsx,json,md,ts,tsx,mts}' || (echo '*** PLEASE RUN npm run prettier AND RESUBMIT ***' && false)",
     "test": "npm run --workspaces test --",
@@ -64,9 +66,13 @@
     "singleQuote": true,
     "trailingComma": "es5"
   },
-  "husky": {
-    "hooks": {
-      "pre-commit": "npm-run-all -ln --parallel lint test"
-    }
+  "lint-staged": {
+    "*.{js,jsx,ts,tsx}": [
+      "eslint --cache --fix",
+      "prettier --write"
+    ],
+    "*.css": [
+      "stylelint --fix"
+    ]
   }
 }
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
       "eslint --cache --fix",
       "prettier --write"
     ],
-    "*.{json,md,css}": [
+    "*.{json,css}": [
       "prettier --write"
     ]
   }
--- a/package.json
+++ b/package.json
@@ -30,7 +30,6 @@
     "npm-run-all2": "8.0.1",
     "prettier": "^3.6.2",
     "rxjs-compat": "6.6.7",
-    "stylelint": "^16.10.0",
     "typescript": "5.8.2"
   },
   "overrides": {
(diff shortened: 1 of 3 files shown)


gkbishnoi07:
@yurishkuro @jkowall please review

jkowall:
I'd like Yuri to review this one. I don't really know the setup for this part.

jkowall:
I'd like Yuri to review this one. I don't really know the setup for this part.

gkbishnoi07:
Sure, thanks! I'll wait for Yuri's review.

yurishkuro (commented):
did you run `husky init`? It creates a whole separate dir `.husky/_/` which as I understand is supposed to be checked in.

gkbishnoi07:
Yes! I ran husky init and it created the .husky/_/ directory locally. However, during commit it was being ignored by the .gitignore  file inside that directory. let me push that directory

gkbishnoi07:
@yurishkuro please check now! i pushed `.husky/_/` directory

image

yurishkuro:
if husky automatically creates .gitignore then great, but I didn't see it when I ran init. I certainly don't want to be committing files that are just generated.

gkbishnoi07:
@yurishkuro now i noticed this automatically create `.husky/_/ `directory by running **npm install** because of this command
`    "prepare": "husky install && npm run --workspace @jaegertracing/plexus prepublishOnly",
`

gkbishnoi07:
can you please download this PRs locally and try to run  ` npm install` it will automatically generate `.husky/_/` directory

gkbishnoi07:
@yurishkuro please review

package.json
```diff
@@ -48,7 +50,7 @@
     "eslint": "eslint --cache 'scripts/*.{js,jsx,ts,tsx}' 'packages/*/src/**/*.{js,jsx,ts,tsx}' 'packages/*/*.{js,jsx,ts,tsx,mts}'",
     "fmt": "npm run prettier",
     "lint": "npm-run-all -ln --parallel prettier-lint tsc-lint eslint check-license",
-    "prepare": "npm run --workspace @jaegertracing/plexus prepublishOnly",
+    "prepare": "husky install && npm run --workspace @jaegertracing/plexus prepublishOnly",
```
yurishkuro: `husky install` logs that install command is deprecated. Just need `husky`

package.json
```diff
@@ -64,9 +66,13 @@
...
-    "hooks": {
-      "pre-commit": "npm-run-all -ln --parallel lint test"
-    }
+  "lint-staged": {
+    "*.{js,jsx,ts,tsx}": [
+      "eslint --cache --fix",
+      "prettier --write"
+    ],
+    "*.css": [
+      "stylelint --fix"
```
yurishkuro: why is this here? I don't see us using stylelint anywhere

package.json
```diff
@@ -64,9 +66,13 @@
     "singleQuote": true,
     "trailingComma": "es5"
   },
-  "husky": {
-    "hooks": {
-      "pre-commit": "npm-run-all -ln --parallel lint test"
-    }
+  "lint-staged": {
+    "*.{js,jsx,ts,tsx}": [
```
yurishkuro: we run prettier on these extensions `js,jsx,json,ts,tsx,mts`

package.json
```diff
@@ -64,9 +66,13 @@
...
+      "eslint --cache --fix",
```
yurishkuro, line 70: what about tsc-lint ?

yurishkuro: and in general, you are deconstructing our main lint/fmt commands and creating the possibility of inconsistent behavior if more linters are added. I would prefer a solution where lint-stage just calls out lint command, which can be parameterized to accept a list of files (npm-run-all can pass same args to subcommands: `npm-run-all script1 script2 -- args`)

package.json
```diff
@@ -64,9 +66,13 @@
...
+      "prettier --write"
```
yurishkuro, line 71: if this modifies the files they need to be `git add`-ed before finishing pre-commit hook

gkbishnoi07: right, let me update

package.json
gkbishnoi07: Okay, my bad — just noticed it now. Will fix it.

gkbishnoi07: We’re not running the full `lint or fmt` scripts in lint-staged since they run on the entire project and would slow down every commit. Instead, we’re running `eslint` and `prettier `only on staged files via Husky for faster feedback. The full checks including tsc, license, and full-format (fmt) — will run during the PR in GitHub Actions. 

I’ve seen a lot of projects do it this way — they only run eslint and prettier on the files that were changed, just to keep the commit fast. Then the full checks like lint, tsc, and formatting run later in CI during the PR. 
what do you think??

yurishkuro: I am not suggesting we run the full lint on completely repository, I am suggesting that we run full lint (as in "all lint steps) but only on staged files. The lint command can be parameterized with file names

gkbishnoi07: Okay, so I’m creating a new script:  `"lint:file": "npm-run-all -s prettier-lint eslint"` And updating lint-staged like this:

```
"lint-staged": {
  "*.{js,jsx,ts,tsx,mts}": [
    "npm run lint:file --",
    "git add"
  ],
  "*.{json,md,css}": [
    "prettier --write",
    "git add"
  ]
}
```
I’m keeping tsc-lint and check-license out since because i believe they run on the whole codebase and not individual files. Also kept .json, .md, and .css separate because only Prettier handles those — not ESLint. Let me know if this setup works for you or if you'd prefer any changes!

gkbishnoi07:
enovo@LAPTOP-NAUH09A2:/mnt/d/jaegery-ui/jaeger-ui$ npx lint-staged
⚠ Some of your tasks use `git add` command. Please remove it from the config since all modifications made by tasks will be automatically added to the git commit index.

gkbishnoi07:
hey @yurishkuro I was trying to improve the commit workflow by running npm run lint:file during the pre-commit stage via Husky. But I’ve been running into issues where commits fail, and the error messages are either unclear or misleading.

Initially, I had this lint-staged setup:

```
"lint-staged": {
  "*.{js,jsx,ts,tsx,mts}": [
    "eslint --cache --fix",
    "prettier --write"
  ],
  "*.{json,md,css}": [
    "prettier --write"
  ]
}

```

It worked well — each command ran independently on staged files, and any failure was easy to identify and fix.

But after switching to this:  

```
"lint-staged": {
  "*.{js,jsx,ts,tsx,mts}": [
    "npm run lint:file --"
  ],
  "*.{json,md,css}": [
    "prettier --write"
  ]
}
```
And using this in package.json:

```
"scripts": {
  "lint:file": "npm-run-all -s prettier-lint eslint || exit 1"
}
```

During commit:
lint:file chains Prettier and ESLint using npm-run-all, which doesn’t work well with lint-staged. If even one file fails, the commit is blocked, and debugging is hard due to missing file references.

npm install also failed:
I’m also seeing errors like:
“Could not find a declaration file for module `'@viz-js/viz'...”`
This seems related to module resolution and may be triggered by Husky running scripts too early during setup.

Can you please advise what you’d prefer I do?

i think  this approach is best!
```
 "lint-staged": {
  "*.{js,jsx,ts,tsx,mts}": [
    "eslint --cache --fix",
    "prettier --write"
  ],
  "*.{json,md,css}": [
    "prettier --write"
  ]
}
```

package.json
yurishkuro: ?

package.json
yurishkuro: Ok let's keep it as you have it

package.json
yurishkuro:
```suggestion
    "prepare": "husky && npm run --workspace @jaegertracing/plexus prepublishOnly",
```

package.json
yurishkuro: why are you resolving conversations without any follow-up?

gkbishnoi07: I solve this on my local setup

gkbishnoi07: Now after some testing I will commit those changes which suggested by you

package.json
gkbishnoi07: thanks for the response

package.json
gkbishnoi07: You're right we're not using Stylelint yet. I initially included it as a future improvement since it's useful for CSS linting (like ESLint for JS). But this PRs is for husky so for now Im removing it and will propose its proper integration in a separate issue.

gkbishnoi07:
image

@yurishkuro its working! please review

package.json
gkbishnoi07: done! thank you

package.json
```diff
@@ -64,9 +65,13 @@
...
-  "husky": {
-    "hooks": {
-      "pre-commit": "npm-run-all -ln --parallel lint test"
-    }
+  "lint-staged": {
+    "*.{js,jsx,ts,tsx,mts}": [
+      "eslint --cache --fix",
+      "prettier --write"
+    ],
+    "*.{json,md,css}": [
```
yurishkuro:
```suggestion
    "*.{json,css}": [
```

gkbishnoi07:
in script prettier run on `.md` too  `"prettier": "prettier --write '{.,scripts}/*.{js,jsx,json,md,ts,tsx,mts}`

yurishkuro (approved):

PR #2995, opened 2025-07-21
Fix the issue of operation reset to all when reloading the Search screen

### LINKED ISSUE #2599: [Bug]: Fix regressions on SearchTracePage form
### What happened?
There are some bugs in SearchTracePage SearchForm, that are probably introduced with PR https://github.com/jaegertracing/jaeger-ui/pull/2593

- [ ] When reloading the Search screen the service remains, but the operation is reset to all (not a regression from the PR listed above but is surely an area to look at)
- [x] Submit button works even when the `maxDuration` or `minDuration` is invalid.

ghost:
- Resolves #2599
- when a user performs a search, both service and operation are saved to localStorage. when the page restarts, both service and operation are restored from localStorage, even if operations have not been loaded yet. If the operation is not valid (e.g., it was removed from the service), it gets reset to 'all'.

## Description of the changes
- Modified the `stateServicesXformer` function to detect when a selected service from localStorage has no operations loaded yet.
- keeping the services in loading state until operations are fetched, preventing the form from rendering with incomplete data
- The test I added now provides proper coverage for this logic

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/SearchTracePage/index.jsx
+++ b/packages/jaeger-ui/src/components/SearchTracePage/index.jsx
@@ -250,6 +250,15 @@ const stateServicesXformer = memoizeOne(stateServices => {
     operationsForService: opsBySvc,
     error: serviceError,
   } = stateServices;
+  const selectedService = store.get?.('lastSearch')?.service;
+  if (
+    selectedService &&
+    serviceList &&
+    serviceList.includes(selectedService) &&
+    (!opsBySvc || !opsBySvc[selectedService] || opsBySvc[selectedService].length === 0)
+  ) {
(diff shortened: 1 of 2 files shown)


ghost:
@yurishkuro now when a user performs a search, both service and operation are saved to localStorage. when the page restarts, both service and operation are restored from localStorage, even if operations have not been loaded yet. If the operation is not valid (e.g., it was removed from the service), it gets reset to 'all'.

I have found an approach to solve this issue but that approach needs a bit more changes. Also when i checked the solution by het4399 that also works. so i used that approach.
Thanks @het4399

yurishkuro (approved):

PR #3009, opened 2025-08-07
test: Migrate JaegerUIApp Component Tests to React Testing Library

vishvamsinh28:
Rewrote the JaegerUIApp component test suite using React Testing Library


packages/jaeger-ui/src/components/App/index.test.js
```diff
@@ -13,13 +13,146 @@
...
+
+  it('should render NotFound for unknown routes', () => {
+    const { getByTestId } = renderWithPath('/unknown');
+    expect(getByTestId('not-found')).toBeInTheDocument();
+  });
+
+  it('should handle root path redirect', () => {
+    const { container } = renderWithPath('/');
+    expect(container).toBeInTheDocument();
+  });
```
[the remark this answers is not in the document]
yurishkuro: +1

packages/jaeger-ui/src/components/App/index.test.js
```diff
@@ -13,13 +13,146 @@
...
+
+  it('should handle prefix path redirects', () => {
+    const { container } = renderWithPath('/prefix');
+    expect(container).toBeInTheDocument();
```
yurishkuro: what does this actually test?

packages/jaeger-ui/src/components/App/index.test.js
```diff
@@ -13,13 +13,146 @@
 // limitations under the License.
 
 import React from 'react';
-import shallow from '../../utils/ReactShallowRenderer.test';
```
yurishkuro, line 16: can we also remove ../../utils/ReactShallowRenderer.test ?

vishvamsinh28: Yes once this PR is merged, I'll open another PR to delete that file and modify the code to remove the Enzyme dependency

packages/jaeger-ui/src/components/App/index.test.js
vishvamsinh28: Right now this test is basically a smoke test. It renders the app at /prefix and only checks that something is in the DOM. Updated it now.

PR #3042, opened 2025-08-21
Fix TypeScript ref typing errors across React components

vishvamsinh28:
Resolved TypeScript errors

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow/TimelineCollapser.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow/TimelineCollapser.tsx
@@ -12,7 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// @ts-nocheck
 import React from 'react';
 import { Tooltip } from 'antd';
 import { LuChevronsRight } from 'react-icons/lu';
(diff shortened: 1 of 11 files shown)


packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow/TimelineCollapser.tsx
```diff
@@ -32,12 +31,7 @@ function getTitle(value: string) {
 }
 
 export default class TimelineCollapser extends React.PureComponent<CollapserProps> {
-  containerRef: React.RefObject<HTMLDivElement>;
-
-  constructor(props: CollapserProps) {
-    super(props);
```
yurishkuro, line 38: We don't need this?

vishvamsinh28: With the modern class property syntax containerRef = React.createRef<HTMLDivElement>(), the separate type declaration containerRef: React.RefObject<HTMLDivElement>; is redundant and can be removed. The class property declaration handles both typing and initialization.

yurishkuro: I'm asking about super(props)

vishvamsinh28: it is only required if we have a constructor

PR #3054, opened 2025-09-01
Expand the logs automatically

### LINKED ISSUE #144: Expand the logs automatically up to N?
I was thinking, would it not make sense to expand the logs automatically? E.g. always show the first 3 logs (configurable) and a `[show more...]` link.

image

ghost:
- Resolves #144 
- Logs section auto-opens and show first 3 logs

https://github.com/user-attachments/assets/3e516e2d-f4af-4552-8be3-7ef7ca062838

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/AccordianLogs.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/AccordianLogs.tsx
@@ -35,6 +35,8 @@ type AccordianLogsProps = {
   timestamp: number;
   currentViewRangeTime: [number, number];
   traceDuration: number;
+  initialVisibleCount?: number;
+  spanID?: string;
 };
 
 export default function AccordianLogs({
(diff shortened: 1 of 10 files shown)


ghost:
```
// This would fail depending on environment:
expect(DetailsPanel).toHaveBeenCalledWith(
  expect.objectContaining({
    decorationSchema: expect.objectContaining({ id: testID }),
    operation: testVertex.operation,
    service: testVertex.service,
  }),
  undefined  //sometimes this was {} instead of undefined
);
```

so the solution is to ignore the second argument.
Instead of trying to predict what the second argument will be, we extract only the first argument:
```
//only check props
expect(DetailsPanel).toHaveBeenCalledTimes(1);
const [detailsPanelProps] = DetailsPanel.mock.calls[0];
expect(detailsPanelProps).toEqual(
  expect.objectContaining({
    decorationSchema: expect.objectContaining({ id: testID }),
    operation: testVertex.operation,
    service: testVertex.service,
  })
);
```

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/DetailState.tsx
```diff
@@ -37,7 +37,7 @@ export default class DetailState {
     this.isReferencesOpen = Boolean(isReferencesOpen);
     this.isWarningsOpen = Boolean(isWarningsOpen);
     this.logs = {
-      isOpen: Boolean(logs && logs.isOpen),
+      isOpen: logs ? Boolean(logs.isOpen) : true,
```
yurishkuro, line 40: what was wrong with the original code? Why is `true` fallback a good one?

Parship999: it ensures logs are visible without an extra click, when detail is first expanded

ghost: > what was wrong with the original code? Why is `true` fallback a good one?

it ensures first 3 logs are visible without an extra click, when detail is first expanded

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/AccordianLogs.tsx
```diff
@@ -48,11 +50,71 @@ export default function AccordianLogs({
...
+    };
+
+    setTimeout(emit);
+    if (typeof window !== 'undefined' && 'requestAnimationFrame' in window) {
+      window.requestAnimationFrame(emit);
+    }
+    setTimeout(emit, 50);
+  }, []);
+
+  // Observe height changes in the content area to notify virtualized list to reflow
```
yurishkuro, line 83: I don't follow this part. If you change the state React should rerender the list. Why do you need these additional manipulations?

ghost: The state change only re-renders the AccordianLogs component itself. However, the parent `VirtualizedTraceView` uses a virtualized list (ListView) that caches row heights for performance. when the logs expand, the virtualized list does nor automatically know that this specific row's height changed, so it continues using the old cached height that causing the overlap issue (here overlapping issue occur when expanding to show all the logs).

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/AccordianLogs.tsx
```diff
@@ -48,11 +50,71 @@ export default function AccordianLogs({
...
+  React.useEffect(() => {
+    if (!interactive || !isOpen) return;
```
yurishkuro, line 85: how can the widget be not interactive?

yurishkuro (dismissed):

yurishkuro:
> codecov/project — 97.78% (-0.36%) compared to 85eb0c7

can you improve code coverage?

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/AccordianLogs.tsx
ghost: It prevents unnecessary DOM observers when AccordianLogs is used in read-only popovers (like span bar tooltips). In non-interactive mode, there are no user actions that change content height.

yurishkuro (approved):

PR #3002, opened 2025-07-30
Make `/deep-dependencies` configurable

sujalshah-bit:
## 🔴 Breaking Change

Previously Deep Dependencies view was always using `/analytics/v1/deep-dependencies` API endpoint to retrieve the data. Now this endpoint is configurable via UI config, but the default is changed to `/api/deep-dependencies` to match the change in the backend https://github.com/jaegertracing/jaeger/issues/6606 that returns dummy data for said endpoint.

Part of [#7399](https://github.com/jaegertracing/jaeger/issues/7399).

## Checklist
- [x] I have read https://github.com/jaegertracing/jaeger/blob/master/CONTRIBUTING_GUIDELINES.md
- [x] I have signed all commits
- [ ] I have added unit tests for the new functionality
- [x] I have run lint and test steps successfully
  - for `jaeger`: `make lint test`
  - for `jaeger-ui`: `npm run lint` and `npm run test`

PR #3058, opened 2025-09-05
Pr 3002

iypetrov:
Extends https://github.com/jaegertracing/jaeger-ui/pull/3002
Previously Deep Dependencies view was always using /analytics/v1/deep-dependencies API endpoint to retrieve the data. Now this endpoint is configurable via UI config, but the default is changed to /api/deep-dependencies to match the change in the backend https://github.com/jaegertracing/jaeger/issues/6606 that returns dummy data for said endpoint.

Part of https://github.com/jaegertracing/jaeger/issues/7399.

## Description of the changes
- 

- Unit tests

## Checklist


packages/jaeger-ui/src/api/jaeger.js
```diff
@@ -81,7 +81,7 @@ function getJSON(url, options = {}) {
 }
 
 export const DEFAULT_API_ROOT = prefixUrl('/api/');
-export const ANALYTICS_ROOT = prefixUrl('/analytics/');
+export const DEEP_DEPENDENCIES_ROOT = prefixUrl(getConfig().deepDependencies?.apiEndpoint);
```
yurishkuro, line 84: Use of ? indicates the value might be null, why should it be allowed?

packages/jaeger-ui/src/types/config.tsx
```diff
@@ -195,6 +195,8 @@ export type Config = {
 
   deepDependencies?: {
```
yurishkuro, line 196: I understand you're just copying preexisting pattern, but I wonder why these need to be optional if our default config will always overwrite them if not defined? It seems not using optional will make the rest of the code simpler.

PR #3067, opened 2025-09-07
Fix top bar tab order

mdwyer6:
Fixes accessibility issues with topbar's tab navigation experience.

- The tab order was starting from the search box on the right half of the topbar and subsequently jumping left to the nav items.
- There was no focus ring on the topbar nav menu items

## Description of the changes
- Tab navigation ordering in top bar now flows left to right as expected
- Tabbing into the topbar nav now shows focus ring

Screenshot of focus ring:
image

Large screen:
image

Small screen:
image

- Tested various screen sizes for visual regressions in responsive behavior
- Tested with keyboard navigation and programmatically triggering focus ring

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/App/TopNav.tsx
+++ b/packages/jaeger-ui/src/components/App/TopNav.tsx
@@ -144,7 +144,14 @@ export function TopNavImpl(props: Props) {
           NAV_LINKS.map(({ matches, to, text }) => {
             const url = typeof to === 'string' ? to : to(props);
             const key = matches(pathname) ? pathname : url;
-            return { key, label: <Link to={url}>{text}</Link> };
+            return {
+              key,
+              label: (
+                <Link style={{ outline: 'revert' }} to={url}>
+                  {text}
+                </Link>
+              ),
+            };
           })
         )}
         className="Menu--item"


packages/jaeger-ui/src/components/App/TopNav.tsx
```diff
@@ -137,29 +137,38 @@ export function TopNavImpl(props: Props) {
...
         className="Menu--item"
         mode="horizontal"
         selectable={false}
         selectedKeys={[pathname]}
+        style={{ flex: '1 1 0', minWidth: 0 }}
+      />
+      <Menu
+        theme="dark"
+        items={itemsGlobalRight}
+        className="Menu--item"
```
yurishkuro, line 166: what about `ub-right`?

mdwyer6: ub-right class just applies `float: right` as far as i can tell. this is no longer needed since i swapped the ordering of the markup to match the visual order and flexbox layout now takes care of spacing

mdwyer6: it's not super clear from the diff but i've done 3 things in this PR:

1. Swapped the order of the two menus so tab order would behave
2. Replace the float layout with equivalent flexbox one because after swapping menus the float layout doesn't work anymore (it'll wrap the second menu to a newline)
3. Unrelated with above two changes: added `outline: 'revert'` to bring back focus ring that ant design overrides

yurishkuro: > this is no longer needed since i swapped the ordering of the markup to match the visual order and flexbox layout now takes care of spacing

but we still want the 2nd menu to be flushed right, as well as the search bar.

mdwyer6: yes flexbox is taking care of that now. here is a screenshot on fairly wide monitor:
image

mdwyer6: I would've kept the float layout if I could to keep this diff tightly focused but it's a quirk of how floats work that you cannot easily get the same behavior when the order is reversed.

PR #3070, opened 2025-09-08
Replace dependency react-window

### LINKED ISSUE #3028: [deps]: Update or replace dependency react-window
### Requirement

We want to upgrade to React 19

### Problem

"react-window": "^1.8.10" doesn’t support React 19

### Proposal

It is used in a single component `FilteredList`
```
$ rg react-window packages/jaeger-ui/src
packages/jaeger-ui/src/components/common/FilteredList/ListItem.tsx
18:import { ListChildComponentProps } from 'react-window';

packages/jaeger-ui/src/components/common/FilteredList/index.tsx
20:import { FixedSizeList as VList, ListOnItemsRenderedProps, ListOnScrollProps } from 'react-window';
```

Options:
  * see if AntD supports this functionality already, it sounds pretty basic
  * possible replacement https://tanstack.com/virtual/latest

ghost:
- Resolves #3028 

## Description of the changes
- Removed: react-window
- Added: @tanstack/react-virtual
- Migration: complete functional replacement

- unit tests

## Checklist


packages/jaeger-ui/package.json
```diff
@@ -74,7 +73,7 @@
     "react-redux": "^9.2.0",
     "react-router-dom": "5.3.4",
     "react-router-dom-v5-compat": "^6.24.0",
-    "react-window": "^1.8.10",
```
yurishkuro, line 77: why can we not upgrade to react-window v2?

ghost: based on my investigation:
Option 1: upgrade to react-window v2
1. breaking changes: FixedSizeList -> SimpleList, children -> rowComponent
2. our class component would need complete rewrite anyway
3. I also found that it removes features we might need (horizontal lists, rtl support)

Option 2: migrate to tanstack virtual 
1. direct react 19 support - no intermediate upgrades needed
2. api aligns with our functional component refactor
3. single migration step instead of multiple breaking changes

Since we already had to refactor our class component to functional, tanstack virtual is the cleaner path forward. For react window we have to additionally deal with deprecated apis also

ghost:
https://github.com/bvaughn/react-window/issues/302
https://github.com/bvaughn/react-window/issues/802.
So acc to my investigation: the main issue is that react-window does not officially support react 19 yet.

packages/jaeger-ui/src/components/common/FilteredList/index.tsx
```diff
@@ -37,148 +38,161 @@ type TProps = {
...
-  filterText: string;
-  visibleStartIndex: number;
-  visibleStopIndex: number;
-  focusedIndex: number | null;
-};
+export interface IFilteredListRef {
+  focusInput: () => void;
+}
 
-export default class FilteredList extends React.PureComponent<TProps, TState> {
```
yurishkuro: please separate migration from class to functional components in a standalone PR

yurishkuro:
https://github.com/tanstack looks like a reasonable choice

packages/jaeger-ui/src/components/common/FilteredList/index.tsx
ghost: > please separate migration from class to functional components in a standalone PR

In this pr the conversion to functional component is already done with TanStack virtual. If we do a standalone pr now, we'd have to do the conversion with react-window. This will creates conflict because both PRs modify the same component but with different virtualization libraries.

yurishkuro: I understand that, but conversion to functional is a significant change and it needs to be done as a standalone change (preferably without changing tests).

yurishkuro (approved):

PR #3078, opened 2025-09-09
Convert SearchTracePage/SearchForm to functional component

JeevaRamanathan:
- #2610

## Description of the changes
- convert `packages/jaeger-ui/src/components/SearchTracePage/SearchForm.jsx` from class to functional and improved test cases.

- Ensured there is no break in the functionality in UI
- Ran the test suite using npm run test to ensure all test cases pass. 
Screenshot 2025-09-10 044525

## Checklist


packages/jaeger-ui/src/components/SearchTracePage/SearchForm.jsx
```diff
@@ -257,338 +257,338 @@ export function submitForm(fields, searchTraces) {
...
+    services,
+    submitting,
+    initialValues,
+    changeServiceHandler,
+    submitFormHandler,
+    handleChange: propHandleChange,
+  } = props;
+
+  const [formData, setFormData] = React.useState({
+    service: initialValues?.service,
```
yurishkuro: what is the value of expanding `initialValues` like this instead of just storing as an object? It's not like `setFormData` provides a way to set individual fields in this struct, it's always all or nothing.

packages/jaeger-ui/src/components/SearchTracePage/SearchForm.jsx
```diff
@@ -257,338 +257,338 @@ export function submitForm(fields, searchTraces) {
...
+      changeServiceHandler(fieldData.service);
+      setFormData(prevState => ({
+        ...prevState,
+        operation: DEFAULT_OPERATION,
       }));
     }
   };
 
-  handleSubmit = e => {
+  const handleChange = propHandleChange || internalHandleChange;
```
yurishkuro: the search form exists in only one place in the UI. What is the use case for this either/or behavior? Can we simplify to only one behavior?

packages/jaeger-ui/src/components/SearchTracePage/SearchForm.test.js
```diff
@@ -552,67 +568,75 @@ describe('SearchForm onChange handlers', () => {
...
     expect(serviceOnChange).toBeDefined();
     serviceOnChange('testService');
-    expect(handleChangeMock).toHaveBeenCalledWith({ service: 'testService' });
+    expect(defaultProps.changeServiceHandler).toHaveBeenCalledWith('testService');
 
     const operationOnChange = SearchableSelect.onChangeFns.operation;
     expect(operationOnChange).toBeDefined();
     operationOnChange('testOperation');
-    expect(handleChangeMock).toHaveBeenCalledWith({ operation: 'testOperation' });
+    expect(operationOnChange).toBeDefined();
```
yurishkuro: how is this equivalent to checking that the handler was called?

yurishkuro (commented):
it is unusual to change unit tests when performing migration to functional component. It may indicate that you're breaking the functionality.

packages/jaeger-ui/src/components/SearchTracePage/SearchForm.jsx
JeevaRamanathan: It's updated now

packages/jaeger-ui/src/components/SearchTracePage/SearchForm.jsx
JeevaRamanathan: It's updated now in one behaviour

JeevaRamanathan:
Hi @yurishkuro, https://github.com/jaegertracing/jaeger-ui/blob/30a402df6598086380d5ed963c977efc21373a05/packages/jaeger-ui/src/components/SearchTracePage/SearchForm.test.js#L432-L435

https://github.com/jaegertracing/jaeger-ui/blob/30a402df6598086380d5ed963c977efc21373a05/packages/jaeger-ui/src/components/SearchTracePage/SearchForm.test.js#L520-L534

Yeah, but in few cases like above test changes were necessary because those tests were directly extending the SearchForm class and accessing internal methods; when converting to a functional component, these class-based patterns don't work anymore since there's no class to extend or this context to access and the tests were failing.

packages/jaeger-ui/src/components/SearchTracePage/SearchForm.test.js
JeevaRamanathan: Yeah I agree, after converting to a functional component, the handleChangeMock is not working properly with the current implementation. I'll check again. Your guidance on the section would be appreciated.

JeevaRamanathan: Hi @yurishkuro, I have updated this retaining the same check. Can you please check and let know if this looks good or any further changes is required

packages/jaeger-ui/src/components/SearchTracePage/SearchForm.test.js
```diff
@@ -518,18 +520,32 @@ describe('<SearchForm>', () => {
...
-        expect(this.props.submitFormHandler).toHaveBeenCalledWith(this.state.formData);
-      }
-    }
+      return <SearchForm {...props} />;
+    };
 
     render(<TestSearchForm {...defaultProps} />);
+
+    expect(mockEvent.preventDefault).toHaveBeenCalled();
+    expect(defaultProps.submitFormHandler).toHaveBeenCalled();
```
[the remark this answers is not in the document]
JeevaRamanathan: The suggestion to use `screen.getByRole('form')` is not viable in this as form role is not exposed here.

yurishkuro:
> Ensured there is no break in the functionality in UI

how?

JeevaRamanathan:
> how?

I manually tested the SearchForm with various inputs, verified correct functionality, results, and layout visually, and also ran the test suite. Please let me know if anything was missed

packages/jaeger-ui/src/components/SearchTracePage/SearchForm.jsx
```diff
@@ -246,338 +246,329 @@ export function submitForm(fields, searchTraces) {
...
+        disabled={submitting || noSelectedService || invalid || invalidDuration}
+        data-test={markers.SUBMIT_BTN}
+      >
+        Find Traces
+      </Button>
+    </Form>
+  );
 }
 
 SearchFormImpl.propTypes = {
```
yurishkuro, line 574: why is this needed? normally functional components are just pure functions

yurishkuro:
merge conflicts

yurishkuro:
I think search form is already a functional component
```
export const SearchFormImpl: React.FC<ISearchFormImplProps> = ({
```

If there's anything extra in this PR please reopen.

PR #3099, opened 2025-09-25
Convert TraceDiff component from class to functional component

ghost:
- Resolves part of #2610

## Description of the changes
- Convert `TraceDiff.tsx` from class to functional component.

- unit tests

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/TraceDiff/TraceDiff.test.js
+++ b/packages/jaeger-ui/src/components/TraceDiff/TraceDiff.test.js
@@ -296,6 +296,28 @@ describe('TraceDiff', () => {
       expect(graphWrapper).toBeInTheDocument();
       expect(graphWrapper).toHaveStyle(`top: ${TOP_NAV_HEIGHT}px`);
     });
+
+    it('setGraphTopOffset with different height', () => {
+      const mockHeight = 50;
+      const expectedTop = TOP_NAV_HEIGHT + mockHeight;
+
+      // Mock clientHeight to trigger the state update
+      Object.defineProperty(HTMLDivElement.prototype, 'clientHeight', {
+        configurable: true,
+        value: mockHeight,
+      });
+
+      const { container, rerender } = render(<TraceDiffImpl {...defaultProps} />);
+      rerender(<TraceDiffImpl {...defaultProps} />);
+
+      const graphWrapper = container.querySelector('.TraceDiff--graphWrapper');
+      expect(graphWrapper).toHaveStyle(`top: ${expectedTop}px`);
+
+      Object.defineProperty(HTMLDivElement.prototype, 'clientHeight', {
+        configurable: true,
+        value: 0,
+      });
+    });
   });
 
   describe('mapStateToProps', () => {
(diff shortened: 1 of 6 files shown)


packages/jaeger-ui/src/components/TraceDiff/TraceDiff.test.js
```diff
@@ -88,6 +90,10 @@ describe('TraceDiff', () => {
   const getUrlSpyMockReturnValue = 'getUrlSpyMockReturnValue';
   let getUrlSpy;
 
+  const renderWithRouter = component => {
+    return render(<MemoryRouter>{component}</MemoryRouter>);
+  };
```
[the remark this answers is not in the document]
ghost: I have tried but `MemoryRouter` creates its own router context that provides the real `useNavigate` hook, which takes precedence over our mock. So even though we mocked `useNavigate`, the component was actually using the real implementation from `MemoryRouter`

ghost:
> ❌ Patch coverage is `93.75000%` with `2 lines` in your changes missing coverage. Please review.

These 2 lines cannot be covered through testing because the condition `graphTopOffset !== newGraphTopOffset` is impossible to trigger.
The callback depends on `graphTopOffset` in its dependency array, so when `graphTopOffset` changes, a new callback is created. This means `graphTopOffset` and `newGraphTopOffset` are always equal when the callback runs.
Root cause is: we can't change `graphTopOffset` without triggering a new callback, but the new callback will always have the updated `graphTopOffset` value.

packages/jaeger-ui/src/components/TraceDiff/TraceDiff.tsx
```diff
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 import * as React from 'react';
-import { History as RouterHistory } from 'history';
```
yurishkuro, line 16: why can we not separate removal of history from functional component upgrade?

yurishkuro: and what is replacing `history` ?

ghost: > why can we not separate removal of history from functional component upgrade?

yes, it can be done separately

ghost: > and what is replacing `history` ?

Are you referring to the wording here?

yurishkuro (dismissed):

yurishkuro:
is there a way not to regress the overall coverage?

packages/jaeger-ui/src/components/TraceDiff/TraceDiff.tsx
```diff
@@ -76,90 +72,96 @@ function syncStates(
...
+  traceDiffState,
+  fetchMultipleTraces,
+  forceState,
+  history,
+}: TStateProps & TDispatchProps & TOwnProps) {
+  const [graphTopOffset, setGraphTopOffset] = React.useState(TOP_NAV_HEIGHT);
+  const headerWrapperElmRef = React.useRef<HTMLDivElement | null>(null);
+
+  const setGraphTopOffsetCallback = React.useCallback(() => {
+    if (headerWrapperElmRef.current && headerWrapperElmRef.current.clientHeight !== undefined) {
```
ghost, line 89: line 98 (`setGraphTopOffset(TOP_NAV_HEIGHT);`) was unreachable in the original code. The condition `if (headerWrapperElmRef.current)` was always true because the ref gets set during render. This means the else branch (line 98) was never executed, so it couldn't be tested. 
Now after the addition of this extra check (`headerWrapperElmRef.current.clientHeight !== undefined`), we can test the else part (`setGraphTopOffset(TOP_NAV_HEIGHT);`).

yurishkuro (approved):

PR #3100, opened 2025-09-27
Remove the history instance from the App component

ghost:
- Work towards #2531

## Description of the changes
- This pr is work towards removing of history package because neither `react-router-dom v6` nor `v7` recommends it.

- It's safe to remove the `history` instance from this component because React router's `Redirect` component handles navigation internally using the router's built-in history mechanism, not our custom `history` instance. When we remove the `history` prop and switch to `BrowserRouter`, React Router automatically falls back to using the browser's `window.history` api. The `Redirect` component works with any router that provides a history context, and our test (`should handle root path redirect`) prove this by verifying that when users visit the root path `/`, they correctly see the `SearchTracePage` component rendered.

- unit tests

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/App/index.jsx
+++ b/packages/jaeger-ui/src/components/App/index.jsx
@@ -14,7 +14,7 @@
 
 import React, { Component } from 'react';
 import { Provider } from 'react-redux';
-import { Route, Redirect, Switch, Router } from 'react-router-dom';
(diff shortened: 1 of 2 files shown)


yurishkuro:
please explain in the description why you're removing it, why is it safe to remove.

ghost:
> please explain in the description why you're removing it, why is it safe to remove.

Done, explanation added in the description

yurishkuro (approved):
Thanks for good description!

PR #3108, opened 2025-09-30
Convert TracePage {TimelineCollapser} to functional component

JeevaRamanathan:
- #2610  

## Description of the changes
- convert `packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow/TimelineCollapser.tsx` to functional component

- Ensured there is no break in the functionality in UI
- Ran the test suite using npm run test to ensure all test cases pass.

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow/TimelineCollapser.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow/TimelineCollapser.tsx
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import React from 'react';
+import React, { useRef } from 'react';
 import { Tooltip } from 'antd';
 import { LuChevronsRight } from 'react-icons/lu';
 import { IoChevronForward } from 'react-icons/io5';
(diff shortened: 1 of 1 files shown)


packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow/TimelineCollapser.tsx
```diff
@@ -30,41 +30,43 @@ function getTitle(value: string) {
...
+  onExpandAll,
+  onExpandOne,
+  onCollapseAll,
+  onCollapseOne,
+}: CollapserProps) {
+  const containerRef = useRef<HTMLDivElement>(null);
 
   // TODO: Something less hacky than createElement to help TypeScript / AntD
-  getContainer = () => this.containerRef.current || document.createElement('div');
+  const getContainer = () => containerRef.current || document.createElement('div');
```
yurishkuro: do we need this? In the old code it was doing a dubious `document.createElement('div')`. I think `containerRef` above already guarantees that it's defined and you can use the constant directly

JeevaRamanathan: @yurishkuro using only containerRef.current can return null, which causes a TypeScript error since Tooltip expects an HTMLElement. Defaulting to document.body with `const getContainer = () => containerRef.current || document.body`; ensures a valid HTMLElement is always returned and resolves the type issue.

Kindly let me know if there is any concerns

JeevaRamanathan: Hi @yurishkuro, just a gentle remainder to review the PR's and provide suggestions if any

yurishkuro (approved):

PR #3110, opened 2025-10-02
Remove history instances from traces.tsx

ghost:
- Work towards #2531 

## Description of the changes
- The `traces.tsx` component can safely use `useNavigate()` because it's already wrapped in router context through the `withRouteProps`, and since it's a functional component, it can use react hooks. The external interface remains unchanged (other components still pass the same props), so there are no breaking changes. All the actual navigation logic remain in the parent component (`index.tsx`), which we updated to accept the navigate function instead of history.

- unit tests

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/DeepDependencies/traces.tsx
+++ b/packages/jaeger-ui/src/components/DeepDependencies/traces.tsx
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 import * as React from 'react';
-import { History as RouterHistory, Location } from 'history';
+import { Location, useNavigate } from 'react-router-dom-v5-compat';
 import _get from 'lodash/get';
 import memoizeOne from 'memoize-one';
 import queryString from 'query-string';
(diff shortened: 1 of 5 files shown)


packages/jaeger-ui/src/components/DeepDependencies/traces.tsx
```diff
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 import * as React from 'react';
-import { History as RouterHistory, Location } from 'history';
+import { Location, useNavigate } from 'react-router-dom-v5-compat';
```
yurishkuro, line 16: Why is using `react-router-dom-v5-compat` better? IIUC we also have an oustanding react-router 5->6 migration. Is `***v5-compat` making it compatible with v6?

ghost: Yes. `react-router-dom-v5-compat` runs React Router v6 under the hood while exposing a v5-compatible surface (via `CompatRouter`), so we can adopt v6 patterns like `useNavigate` without rewriting all v5 code (`switch`, `withRouter` etc.)

yurishkuro (approved):

PR #3149, opened 2025-11-12
Fix: Clicking dots and DDG button

ghost:
## Which problem is this PR solving?
- Resolves #3111 and https://github.com/jaegertracing/jaeger-ui/issues/3123#issuecomment-3431088792

## Description of the changes
- Replaced `history` with `useNavigate()` hook and converted `goToTrace` to `useCallback`.

https://github.com/user-attachments/assets/39df0d58-c3aa-4e71-8cbd-3cabe717db5e

## How was this change tested?
- Unit tests

## Checklist
- [x] I have read https://github.com/jaegertracing/jaeger/blob/master/CONTRIBUTING_GUIDELINES.md
- [x] I have signed all commits
- [x] I have added unit tests for the new functionality
- [x] I have run lint and test steps successfully
  - for `jaeger`: `make lint test`
  - for `jaeger-ui`: `npm run lint` and `npm run test`

PR #3150, opened 2025-11-13
Fix: Upgrade to node 24

### LINKED ISSUE #3147: [Feature]: Upgrade to Node 24
When upgrading to Node 24, there are some new conflicts in the dependencies
```
nvm use 24
rm -rf node_modules package-lock.json
npm install && npm test
```
All the above commands succeed, but the following (which is part of our CI) fails:
```
npm ls --all
...
npm error code ELSPROBLEMS
npm error invalid: ansi-styles@4.3.0 /Users/ysh/dev/jaegertracing/jaeger-ui/node_modules/ansi-styles
npm error invalid: supports-color@7.2.0 /Users/ysh/dev/jaegertracing/jaeger-ui/node_modules/supports-color
npm error invalid: react@19.2.0 /Users/ysh/dev/jaegertracing/jaeger-ui/node_modules/react
```

ghost:
- Resolves #3147 

## Description of the changes
The React 19 validation error was caused by `react-side-effect` (a transitive dependency via react-helmet). Its `package.json`:
```
json"peerDependencies": {
  "react": "^16.3.0 || ^17.0.0 || ^18.0.0"
}
```
The package does not support React 19. Our override successfully installed React 19. But `npm ls --all` do strict validation against the declared peer dependencies and make `React 19.2.0` as invalid.
Proof: https://github.com/nfl/react-helmet/issues/716

Solution: Added `react-side-effect` as a direct dependency in `package.json`. Now `npm` validates leniently, checking if React exists and is compatible -> passes with React 19.

## Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,8 @@
         "packages/jaeger-ui"
       ],
       "dependencies": {
-        "@tanstack/react-virtual": "^3.13.12"
+        "@tanstack/react-virtual": "^3.13.12",
+        "react-side-effect": "^2.1.2"
       },
       "devDependencies": {
         "@babel/cli": "7.28.0",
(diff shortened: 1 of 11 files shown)


ghost:
These failing tests are fixed in the PR: #3149

yurishkuro (approved):

yurishkuro:
> The React 19 validation error was caused by react-side-effect (a transitive dependency via react-helmet).

Are these libs not planning to to support R19? Should we be looking to remove them?

yurishkuro:
Thanks!

ghost:
> Are these libs not planning to to support R19? Should we be looking to remove them?

`react-helmet` and `react-side-effect` are unmaintained since around 2020. We should be looking to remove them. What i found on internet `react-helmet` is no longer needed, React 19 introduced built-in support for metadata tags.
We can now render <title>, <meta> and <link> tags directly inside any of our components.

yurishkuro:
@Parship12 how about this #3151?

ghost:
> @Parship12 how about this #3151?

Nice! much better solution. What about the clean re install?

yurishkuro:
> What about the clean re install?

Let's discuss on #3147

PR #3161, opened 2025-11-21
fix(bug): Handle errors in response body with status code is 200

### LINKED ISSUE #3079: [Bug]: UI Shows Infinite Loading for Non-Existent Trace ID
### What happened?

### Steps to reproduce

Infitine Loader
Image

Image

### Expected behavior

Melonps:
- Resolves #3079

## Description of the changes

In `packages/jaeger-ui/src/api/jaeger.js`
- Validate response body for `errors` array even when status is 200 
- Throw error if errors are present to trigger error handling flow

After this change, I checked to show valid error message in UI, also prevents infinity loading. 
Please see the screenshot below:

image

- Add 1 test to verify this changing in `packages/jaeger-ui/src/api/jaeger.test.js`

## Checklist


packages/jaeger-ui/src/api/jaeger.test.js
```diff
@@ -250,3 +250,35 @@ describe('fetchMetrics', () => {
...
+      })
+    );
+
+    JaegerAPI.searchTraces({ traceID }).catch(err => {
+      expect(err.message).toMatch(msg);
+      expect(err.message).toMatch(traceID);
+      expect(err.httpStatus).toBe(status);
+      expect(err.httpStatusText).toBe(statusText);
+      done();
+    });
```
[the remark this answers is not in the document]
Melonps: Indeed, I fixed it!

packages/jaeger-ui/src/api/jaeger.js
```diff
@@ -48,7 +48,19 @@ function getJSON(url, options = {}) {
 
   return fetch(`${url}${queryStr}`, init).then(response => {
     if (response.status < 400) {
-      return response.json();
+      return response.json().then(body => {
+        if (body && Array.isArray(body.errors) && body.errors.length > 0) {
```
yurishkuro: A function called `getJSON` cannot assume to know the structure of the response. Why do we need to throw exception in this lower level function instead of handling the error where needed?

Melonps: Thanks for your feedback🙏
I've refactored the code to keep `getJSON` generic, and moved the Jaeger specific error handling (where 200 responses can contain an errors array) into the searchTraces method itself.

Without this check, these errors would be silently ignored and treated as successful 200 responses.

packages/jaeger-ui/src/api/jaeger.js
```diff
@@ -123,7 +123,16 @@ const JaegerAPI = {
...
-    return getJSON(`${this.apiRoot}traces`, { query });
+    return getJSON(`${this.apiRoot}traces`, { query }).then(body => {
+      if (body && Array.isArray(body.errors) && body.errors.length > 0) {
+        const errorMessage = body.errors.map(err => getMessageFromError(err, 200)).join('; ');
+        const error = new Error(`API Error: ${errorMessage}`);
+        error.httpStatus = 200;
+        error.httpBody = JSON.stringify(body, null, 2);
+        throw error;
+      }
+      return body;
```
yurishkuro: A successful result is not mutually exclusive with some errors being present. This logic is incorrect. The errors need to be checked by the caller.

yurishkuro (changes requested):

packages/jaeger-ui/src/api/jaeger.js
Melonps: @yurishkuro 
Thanks for the review ! I understand your point now🙇

I see that `fetchMultipleTraces` already handles this pattern correctly in `packages/jaeger-ui/src/reducers/trace.js`, where it processes both payload.data and payload.errors separately. I should follow the same approach for searchTraces, right?

I'll need a bit more time to fix it. Will update the PR soon.

Melonps:
@yurishkuro 

I misidentified the root cause.  The actual problem is a trace ID normalization mismatch:
1.  Frontend sends requests using short trace IDs (e.g., "1", "xyz").
2. Backend zero-pads them to 32-char: "0000000000000001", "0000000000000xyz", as defined in [jaeger-idl](https://github.com/jaegertracing/jaeger-idl/blob/ac6612b8bea16bf20c30a34f26d23060c455b1e0/model/v1/ids.go#L40)
3. Frontend stores responses under the normalized IDs
4. Original short IDs remain stuck in FETCH_LOADING state

I think we should normalize trace IDs to 32-char hex strings on the frontend before issuing API requests by 0, since the Jaeger backend zero-pads them and we need the request and response IDs to align.

image

yurishkuro:
Where does the frontend get the short id from in the first place?

packages/jaeger-ui/src/components/TraceDiff/TraceDiffHeader/TraceIdInput.tsx
```diff
@@ -23,5 +25,11 @@ const { Search } = Input;
 
 export default function TraceIdInput(props: Props) {
   const { selectTrace } = props;
-  return <Search addonBefore="Select by Trace ID" enterButton onSearch={selectTrace} />;
+  const handleSearch = React.useCallback(
+    (value: string) => {
+      selectTrace(normalizeTraceId(value));
```
yurishkuro: Are you normalizing raw user input?

Melonps: @yurishkuro 
yes, 

> Where does the frontend get the short id from in the first place?

Agree that, but I am specifically targeting the scenario where a user manually inputs a Trace ID into the search bar

Sometimes users might input a short ID (e.g., due to a copy paste error or typing only a part of it). Currently, sending this short ID to the backend causes a mismatch with the IDL, leading to an unhandled error that results in the infinite loading state. (UI should show `Trace not found.`

I agree that we shouldn't handle short IDs internally, but to fix this UX issue (the infinite spinner), I propose applying zero-padding specifically for the search bar input. This ensures the ID format matches what the backend expects (16 or 32 characters) and prevents the UI from hanging.

Melonps: Actually, It is likely that user input 31-char trace ID in input, because added 0 to tracd Id's head from backend response (ref: # https://github.com/jaegertracing/jaeger-ui/issues/3079)

image

yurishkuro: if a user inputs invalid ID we should display an error. That the UI keeps spinning in this case is a bug that we need to fix. But we do not need to fix invalid IDs. How do you know their copy&paste omitted the leading zero and not a trailing character?

Melonps: That makes sense.
I will rework the change to implement client-side validation on the input form. If the user enters a Trace ID with an invalid length, an error will be shown immediately, which will effectively resolve the spinning UI bug while maintaining data integrity.

I'm still learning the ropes, so I really appreciate your feedback. I'll push the fixes shortly.

Melonps:
@yurishkuro 

I’m planning to validate the user input as follows.
When the user enters an invalid trace ID, the TraceHeader component will display a validation error in red:
image


Also considering another approach using `packages/jaeger-ui/src/utils/ValidatedFormField.tsx`, which would show the validation error in a popover(but, two popovers appearing at the same time)

image

image

packages/jaeger-ui/src/components/TraceDiff/TraceDiff.tsx
```diff
@@ -29,7 +29,10 @@ import pluckTruthy from '../../utils/ts/pluckTruthy';
 
 import './TraceDiff.css';
 import parseQuery from '../../utils/parseQuery';
+import { validateTraceId } from '../../utils/trace-id-validation';
 import withRouteProps from '../../utils/withRouteProps';
+import './TraceDiff.css';
```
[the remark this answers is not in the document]
Melonps: Done.

Melonps:
@yurishkuro 
Hi!, I update this PR, introducing validation function for checking traceId's digit into redux action and pass only valid IDs to fetch.
Along with it, I add ValidateError as type.

packages/jaeger-ui/src/utils/trace-id-validation.ts
```diff
@@ -0,0 +1,36 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { ValidateError } from '../types/validate-error';
+
+export function validateTraceId(traceId: string): ValidateError | null {
```
yurishkuro, line 17: The UI has no direct dependency on the trace ID being in any specific format. It's just an opaque string for the UI, it should not be adding any additional semantics to that string.

packages/jaeger-ui/src/actions/jaeger-api.js
```diff
@@ -33,6 +33,12 @@ export const fetchMultipleTraces = createAction(
   ids => ({ ids })
 );
 
+export const setTraceValidationError = createAction(
```
yurishkuro: why does this need to be at the API level instead of just handled in TraceDiff?

packages/jaeger-ui/src/reducers/trace.ts
```diff
@@ -89,6 +89,14 @@ function fetchMultipleTracesErred(state, { meta, payload }) {
   return { ...state, traces };
 }
```
yurishkuro, line 90: there is no packages/jaeger-ui/src/reducers/trace.js file in `main` branch. You fork might be stale.

yurishkuro:
if still working on this please rebase & resolve merge conflicts

Melonps:
> if still working on this please rebase & resolve merge conflicts

Okay, I will

packages/jaeger-ui/src/types/validate-error.tsx
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
yurishkuro, line 1: invalid header, see CONTRIBUTING

Melonps:
@yurishkuro I'm going to close this because I've lost track of how to proceed with the fix. I apologize for leaving the pull request open for so long.

Issue #1911, opened 2023-10-24
[Feature]: Add dark theme support to Jaeger-UI


anshgoyalevil:
### Requirement

Since Jaeger is a distributed tracing platform, we sometimes need to look at the traces the whole day. Working with a high-contrast UI becomes a pain for the eye in such cases.

### Problem

Jaeger-UI currently supports only the light theme. A dark theme would help users focus on observability rather than eye strain.

### Proposal

Since react v18, and ant-design v5, the support for dynamic theming and state management has improved.

ant-design provides three theme algorithms: `defaultAlgorithm`, `darkAlgorithm`, and `compactAlgorithm`. Our concern is with the first two.

We can change the `jaegerTheme`'s default algorithm based on the current state of the theme. For example:

```diff
const jaegerTheme = {
+ algorithm: theme.darkAlgorithm,
  token: {
    ...defaultTheme.token,
    colorPrimary: '#199',
  },
  components: {
    ...defaultTheme.components,
    Layout: {
      ...defaultTheme.components.Layout,
      bodyBg: '#fff',
      headerBg: '#404040',
      footerBg: '#fff',
      headerHeight: 48,
      headerPadding: '0 50',
....rest of code
```
will change the whole Jaeger-UI to dark mode, but currently, since we haven't designed the whole UI that way, there are many areas where the theme still remains the same, i.e., light.

State management features of react (ex, `useState` in react 18) can be used to manage the current theme's value and can be used to set the theme dynamically based on the click of a button, like in most applications.

Actionable items:
- Remove the hard-coded themes from the codebase, and make use of colors based on the current theme state's algorithm
- Go step-by-step, with small code changes/PRs at each step.
- Use statement management to change/store the current theme.
- Add a Switch kind of component to the NavBar to toggle between dark and light modes.

Relevant documentation:
- ant-design theme customization guide: https://ant.design/docs/react/customize-theme
- https://ant.design/docs/spec/dark

yurishkuro:
This is not a small issue in itself, but can be broken into many small changes that are actually `good-first-issue`s

yurishkuro:
@anshgoyalevil [this section](https://ant.design/docs/react/customize-theme#seedtoken) describes a lot of pre-defined names/tokens. Is the intention that our higher-level components should be using those tokens for their styles as well? We have a bunch of custom styles, I am wondering if they all need to be migrated, or if the approach that we can keep those custom tokens but migrate them into the themes framework to allow switching.

[...]

PR #3178, opened 2025-11-28
Dark theme ADR

yurishkuro:
Part of #1911 

Main files:
  * docs/adr/0001-design-token-based-theming.md
  * docs/adr/0001/Design-Token-Migration-Instructions.md

The rest can be deleted once the migration is finished.


ghost:
Is the expected workflow like this:
1. pick a file from `phase-1-2-checklist.md`
2. use `phase-0-5-migration-mapping.md` to replace colors with tokens
and
3. follow the rules in `Design-Token-Migration-Instructions.md` ?

yurishkuro:
Yes. As usual, small PRs. But start with the components of the search screen, then trace view.

PR #3160, opened 2025-11-21
Dark mode for Jaeger UI

jkowall:
Part of #1911

## Description of the changes
A full-featured dark mode:

- There’s a global theme provider plus a toggle button in the top nav. The app remembers your choice, updates body/top-nav gradients, link colors, and keeps typography consistent in both themes.
- The Search page was redesigned so the left panel, form controls, and FileLoader share a gradient card aesthetic with theme-aware inputs, tabs, and placeholders—everything stays readable in dark mode.
- The Monitor tab got matching treatment: the empty-state card, warning alert, dropdowns, and search field now switch colors properly, and “No Data” placeholders remain visible.
- Supporting components like the Trace ID search input and TopNav tests were updated to match the new styles.

In short, the UI now offers a polished Material-inspired dark mode with consistent gradients, legible text, and styled controls across Search and Monitor views.

Key changes:

- Introduced a full theme system: new `ThemeProvider` and `ThemeToggleButton`, plus index.css, Page.css, `TopNav.css/TopNav.tsx/test.js`, `TraceIDSearchInput` updates, and global vars.css/`utils.css` tweaks. Body/nav gradients, link colors, and persisted light/dark toggling are now in place.
- Refreshed Search experience: index.css, SearchForm.css, and SearchForm.jsx now use gradient cards, theme-aware inputs, improved label/readability, and customized Ant tabs. `FileLoader.tsx/css` picks up the same aesthetic.
- Monitor tab polish: index.css + serviceGraph.css + index.css adjust cards, alerts, “No data” placeholders, dropdowns, and search inputs for dark mode readability; index.tsx tags the warning alert for styling.
- Supporting cleanup: index.jsx wraps the tree in the new provider, Trace ID search received pill styling, and overall gradients/shadows were aligned for both themes.

In total, 21 files touched (≈777 insertions / 61 deletions) covering theming infrastructure, Search page visuals, FileLoader styling, and Monitor tab readability. 

## How was this change tested?
npm run test (full suite)

## Checklist
- [X] I have read https://github.com/jaegertracing/jaeger/blob/master/CONTRIBUTING_GUIDELINES.md
- [X] I have signed all commits
- [X] I have added unit tests for the new functionality
- [X] I have run lint and test steps successfully
  - for `jaeger`: `make lint test`
  - for `jaeger-ui`: `npm run lint` and `npm run test`

PR #3192, opened 2025-11-29
Add dark theme selector

yurishkuro:
Part of #1911 

Derived from #3160 by @jkowall, but most styling changes are excluded to keep backwards compatibility.

Must be enabled via configuration, e.g.
```
$ cat packages/jaeger-ui/jaeger-ui.config.json
{
  "themes": {
    "enabled": true
  }
}
```


kago-dk:
@yurishkuro I upgraded from v1.75.0 / v2.12.0 to v1.76.0 / v2.13.0 (Docker) without making any configuration changes, now the UI looks like this:
Screenshot 2025-12-05 at 11 22 54

Was that the intention?

My config-ui.json is unchanged.
`
{
    "archiveEnabled": false,
    "monitor.menuEnabled":true
}

`

yurishkuro:
Thanks for reporting. It's likely due to browser-level or OS level preference set to dark and it's still overriding our default to lightScheme (in the absence of config).

As a workaround you can enable the flag in the config and manually switch to light theme, which will be a sticky setting.

yurishkuro:
@kago-dk this sounds like the same issue as #3203

PR #3193, opened 2025-11-29
Remove duplicate Antd dependencies

yurishkuro:
After we upgraded to AntD v6 we should've removed the use of `"@ant-design/compatible"` which was still pulling in v5 and creating some issues.

Also in this change:
  * remove unnecessary vite plugin that was extending antd imports
  * add `allowClear` attribute to input text boxes
  * remove `clear search` button from in-page search since `allowClear` works the same way
  * AI also fixed scripts/get-tracking-version.js that was causing warnings

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19374,25 +19374,6 @@
         }
       }
     },
-    "node_modules/vite-plugin-imp": {
-      "version": "2.4.0",
-      "resolved": "https://registry.npmjs.org/vite-plugin-imp/-/vite-plugin-imp-2.4.0.tgz",
-      "integrity": "sha512-L/6/nvOw+MyNh4UxAlCZHsmKd5MitmHamqqAWB15sbUgVIEz/OQ8jpKr6kkQU0eA/AIe8fkCVbQBlP81ajrqWg==",
-      "dev": true,
-      "license": "MIT",
-      "dependencies": {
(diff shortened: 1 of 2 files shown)


bobrik:
@yurishkuro I upgrade v2.12.0 to v2.13.0 and service tags are really low contrast now. I think this is the PR responsible.

* Before:

image

* After:

image

If you look at ant examples, tags are supposed to contrast with the background:

* https://ant.design/components/tag

image

This is pretty low contrast, but it's better than nothing. In Jaeger search results our background seems to match that of the whole surrounding trace block, making it disappear.

v5.x was a lot better at it in my opinion with the border being explicit:

image

yurishkuro:
@bobrik yeah, it looked better before. We cannot undo the upgrade, have to move forward, but it's more likely the style changes are due to are internal overrides of the styles, e.g. here https://github.com/jaegertracing/jaeger-ui/blob/1ceadb6f6fb29774bfaa77a6148499ff7a04902b/packages/jaeger-ui/src/components/App/ThemeProvider.tsx#L51

When I was making theming changes in #1911 I was trying to match the before style but the before was already after AntD v6 upgrade.

bobrik:
I opened https://github.com/jaegertracing/jaeger-ui/pull/3202 to bring back the outline.

PR #3202, opened 2025-12-05
Use outlined tags for contrast

bobrik:
See: https://github.com/jaegertracing/jaeger-ui/pull/3193#issuecomment-3610048456

Before:

image

After:

image


yurishkuro (approved):

yurishkuro:
Thanks

PR #3206, opened 2025-12-10
feat: Add incomplete trace detection and adjustable search time offset

xenonnn4w:
 - Resolves jaegertracing/jaeger#2231

## Description of the changes
- Added orphan span detection to identify traces with spans referencing
non-existent parent spans (indicating potentially incomplete traces)
- Added `hasOrphanSpans` and `orphanSpanCount` fields to the Trace type
- Added orange "Incomplete" warning tag in search results for traces with
orphan spans
- Added warning banner in trace view header for incomplete traces
- Added new `search.adjustTime` config option that shifts the search end
time back by a specified duration (e.g., "1m") to exclude very recent
traces that may still be receiving spans
- UI shows the adjustment suffix in lookback dropdown options (e.g., "Last
Hour (-1m)")
- Added `traceWithOrphans` helper in demo trace generators for testing

- Added unit tests for orphan span detection in
`transform-trace-data.test.js` (5 new tests)
- Added unit tests for `applyAdjustTime` function in `SearchForm.test.js`
(7 new tests)
- Manually verified orphan detection logic via browser console
- Manually verified adjustTime config shows suffix in lookback dropdown
- All existing tests pass 

Screenshot From 2025-12-22 22-33-52

Screenshot From 2025-12-23 10-20-24

### Search Form - Lookback with Time Adjustment Toggle

>    Search form showing the Lookback field with a time adjustment toggle beside the Lookback(on the right). When `adjustTime` is              configured (e.g., "1m"), a toggle switch appears with the label "Adjusted -1m" and an info icon (?). Clicking the info icon   
reveals a popover explaining that when enabled, the search end time is adjusted back by the configured duration to exclude        very recent traces that may still be receiving spans. The toggle state is persisted in local storage, allowing users to      enable/disable the adjustment as needed.

Screenshot From 2025-12-24 11-37-42

## Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -26,7 +26,7 @@
         "eslint": "9.39.1",
         "eslint-config-prettier": "10.1.8",
         "eslint-plugin-import": "2.32.0",
-        "eslint-plugin-jest": "29.4.0",
+        "eslint-plugin-jest": "29.9.0",
         "eslint-plugin-jsx-a11y": "6.10.2",
         "eslint-plugin-prettier": "5.5.4",
         "eslint-plugin-react": "7.37.5",
@@ -10031,9 +10031,9 @@
       }
     },
     "node_modules/eslint-plugin-jest": {
-      "version": "29.4.0",
-      "resolved": "https://registry.npmjs.org/eslint-plugin-jest/-/eslint-plugin-jest-29.4.0.tgz",
-      "integrity": "sha512-mT51Vflps9lKrifFutmcN5sQqovy7ALriWOrufpXHtuAS6CsqOGpLtG8tvJFK0IKM/2pRxbcCoBPG6tRQbvkvQ==",
+      "version": "29.9.0",
+      "resolved": "https://registry.npmjs.org/eslint-plugin-jest/-/eslint-plugin-jest-29.9.0.tgz",
+      "integrity": "sha512-1H2VNoEkJEO3CzfoSvAIBRaiywXmLU7FesYct/yI4YPnzkM91bNFBMVKqNjT2C7mPGvKAkCQV14oi5rgFCqPdQ==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "eslint": "9.39.1",
     "eslint-config-prettier": "10.1.8",
     "eslint-plugin-import": "2.32.0",
-    "eslint-plugin-jest": "29.4.0",
+    "eslint-plugin-jest": "29.9.0",
     "eslint-plugin-jsx-a11y": "6.10.2",
     "eslint-plugin-prettier": "5.5.4",
     "eslint-plugin-react": "7.37.5",


xenonnn4w:
@yurishkuro, may I get your review on this?

yurishkuro:
* Please provide screenshots indicating the changes
* Please resolve conflicts

xenonnn4w:
@yurishkuro rebased and updated the pr description with the required screenshots, Thanks!

yurishkuro:
thanks for the screenshots. I would like two changes:

1. the whole new banner in the trace timeline view is overkill. You can just add an icon with the word "incomplete" in the next row next to the span count. The full text of the banner can go into a tooltip - it's low signal to always show it once the user knows what "incomplete" means.

2. the (-1m) in the time picker is distracting and unnecessary. You can put an info (?) icon next to the word Lookback and explain the -1m offset there.

xenonnn4w:
@yurishkuro updated the code and screenshots in the description with you requested changes. Thanks.

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ResultItem.css
```diff
@@ -28,3 +28,10 @@ SPDX-License-Identifier: Apache-2.0
   margin-right: 0.25rem;
   padding: 1px;
 }
+
+.ResultItem--warningIcon {
+  color: #f2711c;
```
yurishkuro: please do not introduce hardcoded colors, add a theme variable. We already have
```
  --feedback-error: #db2828;
  --feedback-warning: #f2711c;
  --feedback-success: #21ba45;
  --feedback-info: #2185d0;
```

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ResultItem.tsx
```diff
@@ -45,7 +45,16 @@ export default function ResultItem({
...
 }: Props) {
-  const { duration, services = [], startTime, traceName, traceID, spans = [] } = trace;
+  const {
+    duration,
+    services = [],
+    startTime,
+    traceName,
+    traceID,
+    spans = [],
+    hasOrphanSpans,
```
yurishkuro: this seems redundant, isn't it `== (orphanSpanCount>0)` ?

packages/jaeger-ui/src/components/SearchTracePage/SearchForm.jsx
```diff
@@ -708,19 +741,30 @@ export function mapStateToProps(state) {
       traceIDs: traceIDs || null,
     },
     searchMaxLookback: _get(state, 'config.search.maxLookback'),
+    searchAdjustTime: _get(state, 'config.search.adjustTime'),
```
yurishkuro: I don't see this new setting defined in the config schema.

I was also thinking whether this would make more sense to be UI-controlled. The actual value can come from the config, but whether to apply it or not could be a UI toggle (stored in local storage), e.g.
```
| Lookback                  adjusted -1m: [ ] |
```

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ResultItem.tsx
xenonnn4w: ah it's redundant, overlooked that.

packages/jaeger-ui/src/components/SearchTracePage/SearchForm.jsx
xenonnn4w: https://github.com/jaegertracing/jaeger-ui/pull/3206/files#diff-0722eec6ee34d1ce3215e6c7d51fb11776dd3002419dc3ac73518d61369cf27bR127

Sounds good to me!

xenonnn4w: updated the screenshot!

yurishkuro:
Can we make "Adjusted -1m" to appear on the same line as Lookback label, only flushed right? The form already has the Go button below the fold on my screen, we don't want to make it longer.

packages/jaeger-ui/src/types/trace.tsx
```diff
@@ -73,6 +73,10 @@ export type Trace = TraceData & {
   tracePageTitle: string;
   traceEmoji: string;
   services: { name: string; numberOfSpans: number }[];
+  // Indicates the trace has spans referencing parent spans that don't exist in the trace
+  hasOrphanSpans?: boolean;
```
yurishkuro: redundant

packages/jaeger-ui/src/types/config.tsx
```diff
@@ -118,6 +118,13 @@ export type Config = {
     // maxLimit configures the "search depth" parameter.
     // The interpretation of search depth varies between different backends.
     maxLimit: number;
+
+    // adjustTime shifts the search end time back by the specified duration.
```
yurishkuro:
```suggestion
    // adjustTime shifts the search start/end time back by the specified duration.
```

packages/jaeger-ui/src/model/transform-trace-data.tsx
```diff
@@ -159,6 +159,22 @@ export default function transformTraceData(data: TraceData & { spans: SpanData[]
...
   const traceEmoji = getTraceEmoji(spans);
   const services = Object.keys(svcCounts).map(name => ({ name, numberOfSpans: svcCounts[name] }));
+
+  // Detect orphan spans - spans that reference a parent span ID that doesn't exist in the trace
+  let orphanSpanCount = 0;
+  spans.forEach(span => {
+    if (Array.isArray(span.references) && span.references.length > 0) {
+      const parentRef = span.references[0];
+      if (
+        (parentRef.refType === 'CHILD_OF' || parentRef.refType === 'FOLLOWS_FROM') &&
```
yurishkuro: this condition redundant because it is always true, since these are the only two valid ref types

packages/jaeger-ui/src/model/transform-trace-data.tsx
```diff
@@ -159,6 +159,22 @@ export default function transformTraceData(data: TraceData & { spans: SpanData[]
...
+        !spanMap.has(parentRef.spanID)
+      ) {
+        orphanSpanCount++;
+      }
+    }
+  });
+  const hasOrphanSpans = orphanSpanCount > 0;
```
yurishkuro: redundant

packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageHeader.tsx
```diff
@@ -91,7 +91,23 @@ export const HEADER_ITEMS = [
...
+            <Tooltip title={tooltipText}>
+              <span className="TracePageHeader--incompleteTag">
+                <IoWarning className="TracePageHeader--incompleteIcon" />
+                Incomplete
+              </span>
+            </Tooltip>
+          </span>
+        );
+      }
+      return trace.spans.length;
```
yurishkuro: this is confusing organization, the DOM snippet should be returned only once and the conditional be inside it, e.g.
```
return (
  <span>
    {trace.spans.length}
    (trace.orphanSpans > 0 && (<span>...))
)
```

packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageHeader.css
```diff
@@ -98,3 +98,17 @@ SPDX-License-Identifier: Apache-2.0
   font-size: 1.78em;
   margin-right: 0.15em;
 }
+
+.TracePageHeader--incompleteTag {
+  color: #f2711c;
```
yurishkuro: use theme var

packages/jaeger-ui/src/components/SearchTracePage/SearchForm.jsx
```diff
@@ -708,19 +759,32 @@ export function mapStateToProps(state) {
...
         type: CHANGE_SERVICE_ACTION_TYPE,
         payload: service,
       }),
-    submitFormHandler: fields => submitForm(fields, searchTraces),
+    submitFormHandler: (fields, adjustTimeEnabled) =>
+      submitForm(fields, searchTraces, ownProps.searchAdjustTime, adjustTimeEnabled),
+  };
+}
+
+function mergeProps(stateProps, dispatchProps, ownProps) {
```
yurishkuro: I don't follow what this is doing, is it necessary to have near duplication of mapDispatchToProps ?

xenonnn4w:
@yurishkuro updated, Thanks!

packages/jaeger-ui/src/components/SearchTracePage/SearchForm.jsx
yurishkuro: didn't answer my question - why do we need a new function?

packages/jaeger-ui/src/types/config.tsx
```diff
@@ -118,6 +118,13 @@ export type Config = {
     // maxLimit configures the "search depth" parameter.
     // The interpretation of search depth varies between different backends.
     maxLimit: number;
+
+    // adjustTime shifts the search start/end time back by the specified duration.
```
yurishkuro: > shifts the search start/end time

does it actually do both or just the end time? Let's make sure the description is accurate. If it's only doing end time then the property name should be adjustEndTime

packages/jaeger-ui/src/components/SearchTracePage/SearchForm.jsx
xenonnn4w: I used `mergeProps` because I needed `searchAdjustTime` (from state) inside `submitFormHandler` (in dispatch). Since `mapDispatchToProps` doesn't have access to state props, I thought `mergeProps` was needed to combine them.

xenonnn4w: But you're right, it's simpler to just pass `searchAdjustTime` as an argument when calling the handler from the component

packages/jaeger-ui/src/types/config.tsx
xenonnn4w: umm, it does adjusts the end time. I'll update the comment to be accurate.

xenonnn4w:
Hey, Apologies for this, PR was accidentally closed. I was working on a separate issue in the main jaeger repository and mistakenly pushed those changes to this jaeger-ui branch, then had to revert and force-push. Unfortunately, that caused the PR to close(due to 0 changes).
PR is continued here 
* #3248

yurishkuro:
I can't reopen either, the button is disabled. If you have the changes in a branch locally please create another branch and do a new PR.

PR #3209, opened 2025-12-11
Make update-ui-find backward compatible

ghost:
- Resolves part of #2531

## Description of the changes
- It now accepts both `history` and `navigate`. Components using `history` continue to function properly, also migrated components can start using `useNavigate` without breaking.

## Checklist


yurishkuro (approved):
Is this actually a reusable utility or only called from one place? If the latter might just update the call site to preferred syntax.

ghost:
> Is this actually a reusable utility or only called from one place? If the latter might just update the call site to preferred syntax.

This is used in multiple places (`UiFindInput`, `VirtualizedTraceView`, `TracePage`)

PR #3223, opened 2025-12-14
Convert VirtualizedTraceView component from class to functional component

### LINKED ISSUE #3372: [UI Refactor] Migrate VirtualizedTraceView to Functional Component
This issue is part of the larger effort to modernize the Jaeger UI codebase (Parent Issue #2610). The goal is to refactor the `VirtualizedTraceView` component from a React Class Component to a React Functional Component using Hooks.

**Target Component:** `packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx`

### Acceptance Criteria

| --- | --- | --- |

### Resources

Parship12:
- Resolves part of #2610
- Resolves #3372 

## Description of the changes
- Converted the `VirtualizedTraceView.tsx` component from a class component to a functional component

- Unit tests
- manually verified the ui to check if the component is working or not

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.test.js
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.test.js
@@ -13,12 +13,18 @@ import getLinks from '../../../model/link-patterns';
 import * as linkPatterns from '../../../model/link-patterns';
 import memoizedTraceCriticalPath from '../CriticalPath/index';
 import ListView from './ListView';
+import SpanBarRow from './SpanBarRow';
 import criticalPathTest from '../CriticalPath/testCases/test2';
 
 jest.mock('./SpanTreeOffset');
 jest.mock('../../../utils/update-ui-find');
 
 jest.mock('./ListView', () => jest.fn());
+
+jest.mock('./SpanBarRow', () => {
+  const ActualSpanBarRow = jest.requireActual('./SpanBarRow').default;
+  return jest.fn(props => <ActualSpanBarRow {...props} />);
+});
 
 describe('<VirtualizedTraceViewImpl>', () => {
   let focusUiFindMatchesMock;
@@ -29,6 +35,7 @@ describe('<VirtualizedTraceViewImpl>', () => {
 
   beforeEach(() => {
     ListView.mockClear();
+    SpanBarRow.mockClear();
 
     ListView.mockImplementation(props => {
       const { dataLength, itemRenderer, getKeyFromIndex } = props;
@@ -418,8 +425,11 @@ describe('<VirtualizedTraceViewImpl>', () => {
         <VirtualizedTraceViewImpl {...mockProps} childrenHiddenIDs={childrenHiddenIDs} trace={altTrace} />
       );
       expect(screen.getByTestId('list-view')).toBeInTheDocument();
-      const rows = container.querySelectorAll('.VirtualizedTraceView--row');
-      expect(rows.length).toBeGreaterThan(0);
+      // verify VirtualizedTraceView detected the RPC span and passed rpc prop to SpanBarRow
+      const callsWithRpc = SpanBarRow.mock.calls.filter(
+        call => call[0].rpc !== undefined && call[0].rpc !== null
+      );
+      expect(callsWithRpc.length).toBeGreaterThan(0);
     });
 
     it('renders a SpanBarRow with a client or producer span and no instrumented server span', () => {
@@ -445,8 +455,13 @@ describe('<VirtualizedTraceViewImpl>', () => {
         const altTrace = updateSpan(trace, leafSpanIndex, { tags: tag });
         const { container } = render(<VirtualizedTraceViewImpl {...mockProps} trace={altTrace} />);
         expect(container.querySelector('.VirtualizedTraceView--spans')).toBeInTheDocument();
-        const rows = container.querySelectorAll('.VirtualizedTraceView--row');
-        expect(rows.length).toBe(trace.spans.length);
+
+        // Verify VirtualizedTraceView detected the external uninstrumented service
+        // and passed noInstrumentedServer prop to SpanBarRow
+        const callsWithNoInstrumentedServer = SpanBarRow.mock.calls.filter(
+          call => call[0].noInstrumentedServer !== undefined && call[0].noInstrumentedServer !== null
+        );
+        expect(callsWithNoInstrumentedServer.length).toBeGreaterThan(0);
       });
     });
 
@@ -459,9 +474,14 @@ describe('<VirtualizedTraceViewImpl>', () => {
     });
 
     it('renderSpanDetailRow returns null if detailState is missing', () => {
+      // Simulated by: mockProps.detailStates is initialized as an empty Map() in beforeEach.
+      // This means no spans have an entry in detailStates, so all spans are "not expanded for details".
+      // Expected: Only bar rows (one per span) should be rendered, with no detail rows.
+      // This tests that VirtualizedTraceView correctly skips rendering detail rows when
+      // a span doesn't have a detailState entry.
       const { container } = render(<VirtualizedTraceViewImpl {...mockProps} />);
       expect(container.querySelector('.VirtualizedTraceView--spans')).toBeInTheDocument();
-      // should only render bar rows, no detail rows
+      // Verify only bar rows are rendered (one per span), no detail rows
       const rows = container.querySelectorAll('.VirtualizedTraceView--row');
       expect(rows.length).toBe(trace.spans.length);
     });
@@ -469,6 +489,11 @@ describe('<VirtualizedTraceViewImpl>', () => {
 
   describe('Critical Path rendering', () => {
     it('renders Critical Path segments when row is not collapsed', () => {
+      // childrenHiddenIDs is NOT provided, so it uses the default empty Set from mockProps.
+      // This means all spans are in their normal expanded/uncollapsed state (no rows are collapsed).
+      // Expected: When spans are not collapsed, critical path segments are distributed across
+      // individual spans - each span displays its own portion of the critical path.
+      // This tests VirtualizedTraceView's responsibility to pass critical path data to SpanBarRow.
       const { container } = render(
         <VirtualizedTraceViewImpl
           {...mockProps}
@@ -501,7 +526,10 @@ describe('<VirtualizedTraceViewImpl>', () => {
       const localInstance = container.querySelector('.VirtualizedTraceView--spans');
       expect(localInstance).toBeInTheDocument();
 
-      // Verify critical path segments render when rows collapsed
+      // Verify critical path segments render when rows collapsed.
+      // When a span is expanded and has children, its critical path indicator gets broken up
+      // and parts of it are pushed down to the children. When the span is collapsed, we show
+      // the critical path as belonging to this span alone (i.e. pulled up from children).
       const criticalPathSegments = container.querySelectorAll('[data-testid="SpanBar--criticalPath"]');
       expect(criticalPathSegments.length).toBeGreaterThan(0);
     });
@@ -608,12 +636,16 @@ describe('<VirtualizedTraceViewImpl>', () => {
       expect(mockListView.forceUpdate).toHaveBeenCalled();
     });
 
-    it('handles list resize events when listView is null', () => {
+    it('does not crash when list resize event fires before listView is set', () => {
+      // Using the default ListView mock (from beforeEach) which doesn't call
+      // the ref callback, so listViewRef.current remains null.
+      // Expected: handleListResize checks if listViewRef.current exists before calling forceUpdate,
+      // so the component should handle this gracefully without crashing.
       render(<VirtualizedTraceViewImpl {...mockProps} />);
 
       const event = new Event('jaeger:list-resize');
-      window.dispatchEvent(event);
-      expect(true).toBe(true);
+      // should not throw - the handler checks for null and return early
+      expect(() => window.dispatchEvent(event)).not.toThrow();
     });
 
     it('handles detail measure events with spanID', () => {
@@ -651,12 +683,17 @@ describe('<VirtualizedTraceViewImpl>', () => {
       expect(mockListView.forceUpdate).toHaveBeenCalled();
     });
 
-    it('handles detail measure events when listView is null', () => {
+    it('does not crash when detail measure event fires before listView is set', () => {
+      // Using the default ListView mock (from beforeEach) which doesn't call
+      // the ref callback, so listViewRef.current remains null.
+      // Expected: handleDetailMeasure checks if listViewRef.current exists. When null,
+      // it calls handleListResize which also checks for null and returns early.
+      // The component should handle this gracefully without crashing.
       render(<VirtualizedTraceViewImpl {...mockProps} />);
 
       const event = new CustomEvent('jaeger:detail-measure', { detail: { spanID: 'test-span-123' } });
-      window.dispatchEvent(event);
-      expect(true).toBe(true);
+      // Should not throw - the handler falls back to handleListResize which checks for null
+      expect(() => window.dispatchEvent(event)).not.toThrow();
     });
   });
 });


packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.test.js
```diff
@@ -99,22 +108,83 @@ describe('<VirtualizedTraceViewImpl>', () => {
   }
 
   function createTestInstance(props) {
-    const virtualizedTraceView = new VirtualizedTraceViewImpl(props);
+    // Create helper functions that mimic the component's internal logic
```
yurishkuro: > mimic the component's internal logic

this smells of a bad practice. The tests should not be dependent on internal logic, do we need to fix the tests or make the component more testable?

ghost: ok then we can move the logic into exported helper functions that the component uses, so tests can call the same shared code and will not depend on internal?

Parship12:
Does this approach sound good to you?

yurishkuro:
yes, it's fine, but still not clear why tests specifically need access to these internal functions, especially since you're not testing the helper functions themselves, the component is just calling them.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.test.js
ghost, line 278: @yurishkuro the tests are tesing the helper functions themselves - see lines 300-440. Also, I have found that `ScrollManager` tests mock these functions. The export approach ensures the tests use the actual implementation instead of duplicating it. Should i start implementing it?

yurishkuro: it's fine to export, we do this in other places

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.test.js
```diff
@@ -31,6 +37,10 @@ describe('<VirtualizedTraceViewImpl>', () => {
   let instance;
 
   beforeEach(() => {
+    const ListView = require('./ListView');
```
yurishkuro: I thought require() is frowned upon in modern JS/TS

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.test.js
```diff
@@ -348,32 +385,14 @@ describe('<VirtualizedTraceViewImpl>', () => {
...
-      expect(rowResult.type).toBe('div');
-      expect(rowResult.props.className).toBe('VirtualizedTraceView--row');
-
-      const spanBarRow = rowResult.props.children;
-      expect(spanBarRow.type).toBe(SpanBarRow);
-      expect(spanBarRow.props.span).toBe(trace.spans[1]);
-      expect(spanBarRow.props.isChildrenExpanded).toBe(true);
-      expect(spanBarRow.props.isDetailExpanded).toBe(false);
+      render(<VirtualizedTraceViewImpl {...mockProps} />);
+      expect(screen.getByTestId('list-view')).toBeInTheDocument();
```
yurishkuro, line 402: this seems a lot weaker than the existing assertions.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
```diff
@@ -132,14 +133,60 @@ function generateRowStates(
...
+// export for tests
+export function generateRowStatesFromTrace(
   trace: Trace | TNil,
   childrenHiddenIDs: Set<string>,
   detailStates: Map<string, DetailState | TNil>
 ): RowState[] {
   return trace ? generateRowStates(trace.spans, childrenHiddenIDs, detailStates) : [];
 }
 
+export const testableHelpers = {
```
yurishkuro, line 137: good pattern - why not include generateRowStatesFromTrace here too?

ghost: I missed that :)

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.test.js
ghost: yes, but functional components don't expose a `renderRow()` and we can not access the React element tree without using enzyme (which is deprecated). And also it follows the RTL-style testing ?

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.test.js
```diff
@@ -31,6 +33,9 @@ describe('<VirtualizedTraceViewImpl>', () => {
   let instance;
 
   beforeEach(() => {
+    ListView.mockClear();
```
yurishkuro, line 37: I don't quite understand it - ListView is a production type, why does it have methods that start with `mock`?

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.test.js
yurishkuro: Are you saying that `render(<VirtualizedTraceViewImpl` somehow creates a shallow DOM that is different from prod? If the DOM is still full why can't we locate nested elements? And if the nested elements are actually an implementation detail of virtualized view then they probably should have their own FC that is testable.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.test.js
Parship12: ListView is mocked at the top of the test file. Is not it a standard practice - we mock child components to isolate the component under test and control the mock's behavior in different tests? The alternative would be use require('./ListView') in each test, but it is frowned.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.test.js
ghost: The issue is that ListView is mocked. Should I mock ListView to actually call `itemRenderer('some-key', {}, 1, {})`?

Parship12:
@yurishkuro `ListView` mock now calls `itemRenderer` so `SpanBarRow`/`SpanDetailRow` are actually rendered in the DOM. And updated the tests accordingly.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.test.js
ghost: yes, `ListView` is a production component, but in this test it is replaced with a Jest mock via `jest.mock('./ListView')`. So within the test scope it’s a mock function, so `mockClear` and `mockImplementation` are available.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.test.js
```diff
@@ -473,65 +512,21 @@ describe('<VirtualizedTraceViewImpl>', () => {
...
-        true,
-        trace,
-        spanID,
-        undefined
-      );
-      expect(result).toEqual([]);
+      render(<VirtualizedTraceViewImpl {...mockProps} trace={trace} criticalPath={undefined} />);
+
+      const localInstance = document.querySelector('.VirtualizedTraceView--spans');
+      expect(localInstance).toBeInTheDocument();
```
yurishkuro, line 544: I keep seeing this - how does it in any way validate the intention of the test?

ghost: It just verifies the component renders at all.

yurishkuro: yes but it doesn't verify anything that the unit test was meant to verify, like critical path something something

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
```diff
@@ -293,262 +326,267 @@ export class VirtualizedTraceViewImpl extends React.Component<VirtualizedTraceVi
...
-          color: colorGenerator.getColorByKey(rpcSpan.process.serviceName),
-          operationName: rpcSpan.operationName,
-          serviceName: rpcSpan.process.serviceName,
-          viewEnd: rpcViewBounds.end,
-          viewStart: rpcViewBounds.start,
+      const pathBySpanID = memoizedCriticalPathsBySpanID(criticalPath);
+      return spanID in pathBySpanID ? pathBySpanID[spanID] : [];
+    },
+    []
+  );
```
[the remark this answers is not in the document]
ghost: I don’t think this requires any comment. The empty dependency array is correct by design. It's clearly understandable.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.test.js
```diff
@@ -383,17 +414,12 @@ describe('<VirtualizedTraceViewImpl>', () => {
...
-      const spanBarRow = rowResult.props.children;
-
-      expect(spanBarRow.type).toBe(SpanBarRow);
-      expect(spanBarRow.props.rpc).toBeDefined();
+      const { container } = render(
+        <VirtualizedTraceViewImpl {...mockProps} childrenHiddenIDs={childrenHiddenIDs} trace={altTrace} />
+      );
+      expect(screen.getByTestId('list-view')).toBeInTheDocument();
+      const rows = container.querySelectorAll('.VirtualizedTraceView--row');
+      expect(rows.length).toBeGreaterThan(0);
```
yurishkuro: > `expect(spanBarRow.props.rpc).toBeDefined();`

This assertion is missing. But I have a question about it - the specific rendering of an RPC span sounds like something that should be tested in another component, e.g. SpanBar. Do we understand the motivation why it might have been tested here? One possible reason is someone should be responsible for tagging in individual span row as expanded or collapsed, because the PRC indicator is only displayed when it's collapsed and has a child span.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.test.js
```diff
@@ -417,49 +443,52 @@ describe('<VirtualizedTraceViewImpl>', () => {
...
-
-        const rowResult = instance.renderRow('some-key', {}, leafSpanIndex, {});
-        const spanBarRow = rowResult.props.children;
-
-        expect(spanBarRow.type).toBe(SpanBarRow);
-        expect(spanBarRow.props.noInstrumentedServer).not.toBeNull();
+        const { container } = render(<VirtualizedTraceViewImpl {...mockProps} trace={altTrace} />);
+        expect(container.querySelector('.VirtualizedTraceView--spans')).toBeInTheDocument();
+        const rows = container.querySelectorAll('.VirtualizedTraceView--row');
+        expect(rows.length).toBe(trace.spans.length);
```
yurishkuro: This is not testing what the original test was tesing

> expect(spanBarRow.props.noInstrumentedServer).not.toBeNull();

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.test.js
```diff
@@ -417,49 +443,52 @@ describe('<VirtualizedTraceViewImpl>', () => {
...
     it('renderSpanDetailRow returns null if detailState is missing', () => {
-      const component = new VirtualizedTraceViewImpl(mockProps);
-      const result = component.renderSpanDetailRow(trace.spans[0], 'key', {}, {});
-      expect(result).toBeNull();
+      const { container } = render(<VirtualizedTraceViewImpl {...mockProps} />);
+      expect(container.querySelector('.VirtualizedTraceView--spans')).toBeInTheDocument();
+      // should only render bar rows, no detail rows
+      const rows = container.querySelectorAll('.VirtualizedTraceView--row');
+      expect(rows.length).toBe(trace.spans.length);
     });
```
yurishkuro, line 487: I am not following what this test is testing, how is "if detailState is missing" condition simulated?

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.test.js
```diff
@@ -417,49 +443,52 @@ describe('<VirtualizedTraceViewImpl>', () => {
...
   });
 
   describe('Critical Path rendering', () => {
     it('renders Critical Path segments when row is not collapsed', () => {
```
yurishkuro, line 491: how is "when row is not collapsed'" condition simulated here? (let's add comments into code)

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.test.js
```diff
@@ -468,70 +497,38 @@ describe('<VirtualizedTraceViewImpl>', () => {
...
       );
 
-      const localInstance = document.querySelector('.VirtualizedTraceView--spans');
+      //verify component renders
+      const localInstance = container.querySelector('.VirtualizedTraceView--spans');
       expect(localInstance).toBeInTheDocument();
+
+      // Verify critical path segments render when rows collapsed
+      const criticalPathSegments = container.querySelectorAll('[data-testid="SpanBar--criticalPath"]');
+      expect(criticalPathSegments.length).toBeGreaterThan(0);
```
yurishkuro, line 534: We can add this comment:

> when a span is expanded and it has children then its critical path indicator gets broken up and parts of it are pushed down to the children. When the span is collapsed we show the critical path as belonging to this span allow (i.e. pulled up from children).

And given this description, should the condition be `==1` rather than `>0` ? Also the selector is applies to the whole trace, not to one non-expanded span.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.test.js
```diff
@@ -468,70 +497,38 @@ describe('<VirtualizedTraceViewImpl>', () => {
...
-      const component = new VirtualizedTraceViewImpl(updatedProps);
-      component.listView = {};
-      component.componentDidUpdate(mockProps);
-
-      expect(mockProps.scrollToFirstVisibleSpan).toHaveBeenCalledTimes(1);
-      expect(mockProps.clearShouldScrollToFirstUiFindMatch).toHaveBeenCalledTimes(1);
-    });
-
-    describe('shouldComponentUpdate', () => {
-      it('returns true if props.shouldScrollToFirstUiFindMatch changes to true', () => {
```
yurishkuro, line 500: > `describe('shouldComponentUpdate', ...`

this whole section is deleted, please elaborate

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.test.js
```diff
@@ -596,51 +586,77 @@ describe('<VirtualizedTraceViewImpl>', () => {
...
+      const event = new Event('jaeger:list-resize');
+      window.dispatchEvent(event);
 
       expect(mockListView.forceUpdate).toHaveBeenCalled();
     });
 
     it('handles list resize events when listView is null', () => {
-      const component = new VirtualizedTraceViewImpl(mockProps);
-      component.listView = null;
+      render(<VirtualizedTraceViewImpl {...mockProps} />);
```
yurishkuro, line 644: how do we simulate "when listView is null" here?

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.test.js
```diff
@@ -596,51 +586,77 @@ describe('<VirtualizedTraceViewImpl>', () => {
...
     it('handles detail measure events when listView is null', () => {
-      const component = new VirtualizedTraceViewImpl(mockProps);
-      component.listView = null;
+      render(<VirtualizedTraceViewImpl {...mockProps} />);
 
-      const event = { detail: { spanID: 'test-span-123' } };
-      expect(() => component._handleDetailMeasure(event)).not.toThrow();
+      const event = new CustomEvent('jaeger:detail-measure', { detail: { spanID: 'test-span-123' } });
+      window.dispatchEvent(event);
+      expect(true).toBe(true);
```
yurishkuro: ?

the original tests are poorly named, "handles detail measure events when ..." instead of saying _what_ should be happening. In this case, if listView is null what do we expect to happen?

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.test.js
ghost: Acc to my understanding, `VirtualizedTraceView` is responsible for detecting when a collapsed client span has a child server span and passing that RPC info to `SpanBarRow`. This tests VirtualizedTraceView's logic, not SpanBarRow's rendering.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.test.js
ghost: This tests that VirtualizedTraceView correctly skips rendering detail rows when a span doesn't have a detailState entry.

ADDED COMMENTS.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.test.js
ghost: Is not the selector queries the whole trace (all spans), not just the collapsed span? so I think we should use >0 because critical path segments may be distributed across multiple spans.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.test.js
ghost: `shouldComponentUpdate` doesn't exist in the functional component. The original class component tests verified 3 scenarios about prop changes. This behavior is now handled by React's reconciliation and the useEffect hook.

yurishkuro:
this needs rebase

ghost:
will open a new PR, I tried to resolve the conflicts but tests are failing

PR #3229, opened 2025-12-20
ui: recognize otel.status_code=ERROR as error span

DCchoudhury15:
Resolves jaegertracing/jaeger#7696

## Description of the changes
- Treat spans with `otel.status_code=ERROR` as error spans in the Jaeger UI.
- This ensures error highlighting and filtering work correctly for OpenTelemetry-generated traces.

- Updated unit tests in `TraceTimelineViewer/utils.test.js`
- Ran `npm run test` locally and verified all tests pass.

## Checklist
- [x] I have read the contributing guidelines


yurishkuro:
how does this resolve the original issue, which is about searching traces in the backend?

also the UI is not getting OTLP traces from Jaeger backend, but the original Jaeger model, and translation to that model should be already doing the conversion from span.status to error tag.

DCchoudhury15:
Thank you for the clarification. I understand now that the UI receives the Jaeger model and the translation should happen upstream. I will close this and investigate the OTLP-to-Jaeger translation in the backend repository to ensure otel.status_code is correctly mapped to the error tag during ingestion.

DCchoudhury15:
Closing this based on feedback to address the root cause in the backend OTLP-to-Jaeger translation logic.

PR #3230, opened 2025-12-20
fix: [Bug]: Clicking on Services dropdown's icon opens and closes it immediately

JaviMaligno:
## Summary

Fixes the dropdown arrow click behavior in the Services and Operations comboboxes on the Search page.

## Root Cause

In Ant Design v6, clicking on the dropdown arrow icon (`.ant-select-arrow`) triggers a `mousedown` event that causes the dropdown to immediately close after opening. This happens because the arrow icon's default mousedown behavior conflicts with the dropdown's open/close state management.

## Solution

Added a `handleMouseDown` event handler to `SearchableSelect` that:
1. Detects clicks specifically on the arrow icon element
2. Prevents the default mousedown behavior using `e.preventDefault()` and `e.stopPropagation()`
3. Manually triggers a click on the selector area (`.ant-select-selector`) instead, which properly opens and keeps the dropdown open

This approach targets only the arrow icon clicks without interfering with normal dropdown option selection.

## Testing

- Clicking on the dropdown arrow now opens and keeps the dropdown open
- Clicking on the main body of the dropdown still works as expected
- Selecting options from the dropdown works correctly

---
🤖 Changes prepared with assistance from OSS-Agent

### DIFF
--- a/packages/jaeger-ui/src/components/common/SearchableSelect.tsx
+++ b/packages/jaeger-ui/src/components/common/SearchableSelect.tsx
@@ -58,7 +58,43 @@ export type SearchableSelectProps = SelectProps & {
  */
 const SearchableSelect: FunctionComponent<SearchableSelectProps> = ({ fuzzy, ...props }) => {
   const filterOption = fuzzy ? filterOptionsFuzzy : filterOptionsByLabel;
-  return <Select showSearch filterOption={filterOption} {...props} />;
+
+  // Handle the specific arrow click issue in Ant Design v6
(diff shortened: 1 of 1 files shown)


JaviMaligno:
## 🤖 Automated PR Review




*This review was performed by an automated agent.*

yurishkuro:
the PR description needs to describe what was the root cause and how it's solved, not repeat the issue description.

ghost:
@JaviMaligno I am unable to reproduce this locally (Chrome/Edge on windows). A short before/after screen recording would help confirm the issue and the fix across environments.

PR #3244, opened 2025-12-25
Fix dark theme issues in trace timeline view

jkowall:
## Summary

Fixes visibility issues in the trace timeline view when using dark mode.

## Issues Fixed

1. **Expand/collapse arrows not visible** - The span tree expand/collapse icons were using `#000` which was invisible on dark backgrounds
2. **Timeline header unreadable** - The header row had a light gray background (`#ececec`) that didn't adapt to dark mode
3. **Back button not viewable** - The back button used light backgrounds and borders that blended into dark theme
4. **Other buttons near top not viewable** - Collapse/expand buttons used `rgba(0, 0, 0, 0.5)` which was invisible in dark mode

## Changes Made

Converted hardcoded colors to design tokens from `vars.css`:

| File | Change |
|------|--------|
| `SpanTreeOffset.css` | Icon color, hover background, indent guide colors |
| `TimelineCollapser.css` | Button colors for collapse/expand controls |
| `TimelineHeaderRow.css` | Header background and border |
| `TracePageHeader.css` | Back button, borders, overview items |
| `CanvasSpanGraph.css` | Minimap background |
| `ViewingLayer.css` | Graph border and inactive overlay |
| `TickLabels.css` | Timeline label colors |
| `Ticks.css` | Tick mark background and label colors |
| `Scrubber.css` | Scrubber handle and line colors |
| `GraphTicks.css` | Tick stroke color |

## Testing

- All 142 related tests pass
- Manually verified in both light and dark modes

### DIFF
--- a/packages/jaeger-ui/src/components/App/index.css
+++ b/packages/jaeger-ui/src/components/App/index.css
@@ -18,3 +18,84 @@ a {
 a:hover {
   color: var(--text-link-hover);
 }
+
+/* ============================================
+   Global Ant Design Overrides for Dark Mode
+   ============================================ */
+
+/* Input fields */
+.ant-input {
+  background-color: var(--surface-tertiary);
+  border-color: var(--border-default);
+  color: var(--text-primary);
+}
+
+.ant-input::placeholder {
+  color: var(--text-muted);
+}
+
+.ant-input:hover,
+.ant-input:focus {
+  border-color: var(--interactive-primary);
+}
+
+/* Input wrapper (for inputs with suffix/prefix) */
+.ant-input-affix-wrapper {
+  background-color: var(--surface-tertiary);
+  border-color: var(--border-default);
+}
+
+.ant-input-affix-wrapper:hover,
+.ant-input-affix-wrapper-focused {
+  border-color: var(--interactive-primary);
+}
+
+.ant-input-affix-wrapper .ant-input {
+  background-color: transparent;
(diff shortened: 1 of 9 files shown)


jkowall:
Part of #1911

**Post-merge tasks:**
- Resolve remaining issues in the diff, monitor, and architecture views.
- Implement Material 3 design across style-specific UI components.

packages/jaeger-ui/src/components/App/index.css
```diff
@@ -18,3 +18,84 @@ a {
...
+
+/* ============================================
+   Global Ant Design Overrides for Dark Mode
+   ============================================ */
+
+/* Input fields */
+.ant-input {
+  background-color: var(--surface-tertiary);
+  border-color: var(--border-default);
+  color: var(--text-primary);
```
yurishkuro, line 30: I think this is the wrong direction again. If anything we should be trying to __align__ with AntD theme support, not __work against it__. A change like this ^ undermines AntD's styling.

The ideal situation would be that we do not override so much in `ThemeConfig` in `ThemeProvider.tsx`, but instead:
  1. define minimum set in the theme provider where by setting up the Javascript properties we effectively override standard `var(--ant-...)` variables that are used by all standard AntD components
  2. we define our own variables by referencing the corresponding `--ant-xxx` variables wherever possible
  3. we use our own variables in our own components

PR #3255, opened 2025-12-28
Fix in-trace search

### LINKED ISSUE #3249: [Bug]: Context search is broken
### What happened?

on the main trace view the Find box is no longer working. I don't know when exactly it was broken but it was working in commit e4afea71 which touched `packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageSearchBar.tsx`. It was already not working in commit 5b89ffb8.

### Steps to reproduce

1. Open any trace, e.g. HotROD.
2. Type something like "customer" in the Find bar in the header

### Expected behavior

The up/down buttons to the right of Find input should activate indicating found spans.

### Relevant log output

```shell

```

### Screenshot

_No response_

### Additional context

_No response_

### Jaeger backend version

_No response_

### SDK

_No response_

### Pipeline

_No response_

### Stogage backend

_No response_

### Operating system

_No response_

### Deployment model

_No response_

### Deployment configs

```shell

```

yurishkuro:
Attempt to resolve #3249, AI-generated.

Fix the search functionality regression by unifying the history management in index.tsx.

The issue was caused by a mismatch between the history instance used by the Redux store and the one used by BrowserRouter. By switching to a plain Router with the shared history object, all URL updates are now correctly synchronized.

image

### DIFF
--- a/packages/jaeger-ui/src/index.tsx
+++ b/packages/jaeger-ui/src/index.tsx
@@ -7,12 +7,13 @@
 import './site-prefix';
 
 import React from 'react';
-import { BrowserRouter } from 'react-router-dom';
+import { Router } from 'react-router-dom';
 import { CompatRouter } from 'react-router-dom-v5-compat';
 import { createRoot } from 'react-dom/client';
 
 import JaegerUIApp from './components/App';
 import { context as trackingContext } from './utils/tracking';
+import { history } from './utils/configure-store';
 
 // these need to go after the App import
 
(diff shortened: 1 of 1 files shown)


yurishkuro:
@Parship12 could you review?

Parship12 (approved):
LGTM

We can’t fix this without `history` here.
Several components still rely on `state.router.location`, so navigation must go through the redux-history to keep redux and the URL in sync.
Alternative would be refactoring of these components to stop using `state.router` and to use `useLocation()`, which will be a much larger change for now.
This will be the minimal fix for now.

yurishkuro:
Thanks

PR #3312, opened 2026-01-06
Remove history from ResultItemTitle

ghost:
- Resolves part of #2531 

## Description of the changes
- Remove history ftom `ResultItemTitle.tsx` and `ResultItemTitle.test.js`

- unit tests
- manually verified the behavior in the ui

## Checklist


yurishkuro:
can you elaborate what was history doing in this component in the first place?

ghost:
> can you elaborate what was history doing in this component in the first place?

Just to tell typescript "this prop should be a valid location for navigation". We don't need the history package as a dependency just for types, we can use `React.ComponentProps<typeof Link>['to']` instead.

yurishkuro (approved):

PR #3328, opened 2026-01-07
Enhance TraceStatistics styles for improved theming and hover effects

### LINKED ISSUE #3322: [Bug]: Trace Statistics view in dark mode
### What happened?

When `Color by ...` is enabled text is unreadable

Image

adity1raut:
- Resolve #3322 

## Description of the changes
- Now change the dark mode visibility.
-  In this PR, we fix the dark mode visibility issue in the tracks statistics section on this page.
- The issue occurred when switching to dark mode: the table rows did not change their theme and remained in light mode.
- This PR ensures that the table rows now properly adapt to the selected theme—dark mode applies dark styling, and light mode applies light styling.

- npm run prettier
- npm run lint
- npm test
- npm run build

## Checklist

## Befour

Screenshot From 2026-01-07 09-40-32

##  After

Screenshot From 2026-01-07 09-40-41

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.css
+++ b/packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.css
@@ -16,13 +16,38 @@ SPDX-License-Identifier: Apache-2.0
   color: rgb(153, 153, 153);
   font-style: italic;
   cursor: default;
+  background-color: transparent !important;
 }
 
 .MainTableData--TraceStatistics {
   border-top: 1px solid var(--border-strong);
+  background-color: var(--bg-2) !important;
 }
 
 .MainTableData--TraceStatistics:hover {
-  background: var(--surface-component-background-hover);
+  background-color: var(--bg-3) !important;
   color: var(--text-primary);
+}
+
+/* For table cells */
+.span-table .ant-table-cell {
+  background-color: inherit !important;
+}
+
+/* Default row background that adapts to theme */
+.span-table .ant-table-row {
+  background-color: var(--bg-1);
+}
+
+.span-table .ant-table-row:hover {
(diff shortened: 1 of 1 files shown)


adity1raut:
PTAL @yurishkuro

ghost:
Please add screenshots

ghost (commented):
This is breaking the "Color by" feature. Please check

#### Now:
Screenshot 2026-01-07 170737

#### After the changes:
Screenshot 2026-01-07 183047

ghost:
I think a better fix will be to adjust the color gradient in `TraceStatistics/generateColor.ts`, by making the lightest color little darker for better text contrast.
cc @yurishkuro

adity1raut:
> This is breaking the "Color by" feature. Please check
> 
> #### Now:
> Screenshot 2026-01-07 170737
> #### After the changes:
> Screenshot 2026-01-07 183047

But in the colur feature the text don't display properly

yurishkuro:
I like the more subtle coloring, the full row coloring feels overly obnoxious, in either theme. But coloring just the bottom border feels confusing, since it's between the rows.

Perhaps incorporating relative bars directly into selected column as in #3310 would be better, it's a more common way of showing relative comparison of values. I would extract the relative bar from #3310 into its own component and reuse across both views.

ghost:
> Perhaps incorporating relative bars directly into selected column as in #3310 would be better, it's a more common way of showing relative comparison of values. I would extract the relative bar from #3310 into its own component and reuse across both views.

Agreed, this is a better approach 👍

yurishkuro:
and we should use the precision-reducing function to render numbers (also in that PR), 5-digit precision makes no sense

ghost:
@adity1raut are you still working on this?

Issue #3313, opened 2026-01-06
[🧑‍🎓 LFX]: Upgrading Core Routing and State Management


yurishkuro:
## Modernizing Jaeger-UI: Upgrading Core Routing and State Management

Please read https://github.com/jaegertracing/jaeger/issues/7833 for any application-related questions.

LFX link: https://mentorship.lfx.linuxfoundation.org/project/d13df7c9-06d3-4589-bef8-c13e5c79c2a2

### Description:  

Jaeger is the industry-standard open-source distributed tracing platform used by companies like Uber and Shopify to monitor complex microservices. To ensure long-term maintainability and performance, the Jaeger-UI is undergoing a major modernization effort to move away from legacy patterns and deprecated libraries.  
This project focuses on the final two major "foundational" upgrades for the UI:

1. **Routing Renovation:** Upgrading from legacy react-router patterns to **React Router v7**. This involves migrating from a component-based routing approach to the modern data-driven router architecture.  
2. **State Management Cleanup:** Replacing the deprecated history package and older Redux integration patterns with modern standards like redux-first-history.

As a mentee, you will be the primary architect for these upgrades, diving deep into the internals of a high-traffic React application. You will work on refactoring core navigation logic, improving how the UI handles deep-linking into complex trace data, and ensuring the application remains lightning-fast for developers debugging their systems.

### Expected Outcomes

* A fully migrated Jaeger-UI running on **React Router v7**.  
* Successful removal of the deprecated history v4/v5 dependency in favor of modern state-navigation integration.  
* Refactored functional components that utilize modern hooks (like useNavigate and useParams) instead of legacy HOCs (Higher-Order Components).  
* A robust test suite ensuring no regressions in trace visualization or search functionality during the migration.

### Recommended Skills

* **Javascript/TypeScript & React:** Deep familiarity with hooks and component lifecycles.  
* **State Management:** Experience with Redux or similar libraries.  
* **Routing Knowledge:** Understanding of SPAs (Single Page Applications) and browser history APIs.  
* **Testing:** Experience with Jest/Enzyme or React Testing Library is a plus.

Mentor(s):  
* Yuri Shkuro (@yurishkuro, github@ysh.us)  
* Jonah Kowall (@jkowall, jkowall@kowall.net)  

Upstream Issues:
* #2531 (History upgrade)  
* #2532 (React Router v7 upgrade)
  * May require #2610

gmarav05:
[...]

PR #3341, opened 2026-01-08
fix(deps):  atomic upgrade to React Router v7 and resolve all breaking changes

Harsh-6291:

- Resolves #2529 (Upgrade React Router)
- Contributes to #2532

## Description of the changes
This PR upgrades  to React Router v7 to fix the breaking build issues we've been seeing.

### Why so many changes?
React Router v7 changes a lot of things (like replacing Switch with Routes), so we couldn't just upgrade the package version—we had to update the code at the same time to keep the build working.

 ###  What I did:
- Upgraded react-router-dom to v7 and history to v5.
 Replaced old routing components (Switch, Redirect) with new ones (Routes, Navigate).
- Fixed a bunch of TypeScript errors related to History and Location types.
- Updated matchPath logic to work with the new version.

- npm run build: Success.
- npm run tsc-lint: Passed with 0 errors.
- Unit tests: Verified locally.


Harsh-6291:
hii @yurishkuro and @jkowall , please take a look at this .

yurishkuro:
please resolve merge conflicts

@Parship12 I thought this upgrade would be much bigger deal, am I missing something?

ghost:
PR is incomplete and currently breaking the app.
`v7` is installed, but many part of the codebase still use `v5` APIs (`Switch`, `Redirect`, `exact`, children-based `<Route>`), so the ui is not starting.

A full `v5 -> v7` migration is required across routing, types, and tests (this likely requires additional changes across roughly 20-30 files). Also, the app currently relies on `redux-first-history`, which will also create issues in this upgradation. So my recommendation is to split this into two phases remove `history` first, then upgrade to `v7` which will be much cleaner.

Parship12:
@yurishkuro I forgot to mention, this is the reason why the React router upgradation should be done after history removal:
https://github.com/jaegertracing/jaeger-ui/issues/3313#issuecomment-3728765878

ghost:
> So my recommendation is to split this into two phases remove `history` first, then upgrade to `v7` which will be much cleaner.

sorry, there is a little flaw here: 
two phases will be ->
1. first keep react-router v5 + history + `redux-first-history` and then use existing `react-router-dom-v5-compat` to provide modern hooks.
2. Upgrade to v7 + remove history package + remove `redux-first-history` atomically (will be a breaking change)

little complex architecture: they are basically one upgrade wearing two hats :)

PR #3342, opened 2026-01-08
refactor: convert App/index.tsx file's  class based to functional component 

gulshank0:
part of: #2610 

## Description of the changes
1. Removed class syntax - Changed from `class JaegerUIApp extends Component<{}>` to function `JaegerUIApp()`.
2. Updated imports - Changed `import React, { Component } from 'react'` to `import { useEffect } from react (React import not needed with modern JSX transform)`.
3. Converted constructor to useEffect - The initialization logic that was in the constructor: 
  `(JaegerAPI as any).apiRoot = DEFAULT_API_ROOT;`
  `processScripts();`
4. Removed render method - The JSX is now directly returned from the function body

  The test file has a test that references the class constructor: it `('should handle constructor with props', () => { expect(() => new JaegerUIApp({})).not.toThrow(); })`. Since we've converted to a functional component, this test file updated:

## Tested 
passes : 
1. npm run prettier 
2.  npm run lint 
3.  npm test 
4.  npm run build

## Checklist

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,3 +1,9 @@
+# Copyright (c) The Jaeger Authors
+# SPDX-License-Identifier: Apache-2.0
+
+# Validate config via:
+#   curl -X POST --data-binary @.codecov.yml https://codecov.io/validate
+
 codecov:
   notify:
     require_ci_to_pass: yes
@@ -11,9 +17,7 @@ coverage:
   status:
     project:
       default:
-        # ABSOLUTE FLOOR: The status will fail if overall coverage is below 95%.
-        # This is enforced *after* the relative check passes.
-        target: 95%
+        target: auto  # Prevents Plexus from tanking the global status
 
         # RELATIVE PROTECTION: This is the maximum allowed drop *below* the base commit's coverage.
         # Setting this to 0% means the project coverage cannot drop at all,
@@ -25,6 +29,19 @@ coverage:
         if_not_found: success
         if_ci_failed: error
         only_pulls: false
+
+      jaeger-ui:
+        flags:
+          - jaeger-ui
+        target: 95%  # The "Absolute Floor" for main UI
+        threshold: 0%
+
+      plexus:
+        flags:
+          - plexus
+        target: auto  # Allows Plexus to start at 0% and grow
+        threshold: 0%
+
     patch:
       default:
         only_pulls: true
@@ -36,6 +53,17 @@ coverage:
         if_no_uploads: error
         if_not_found: success
         if_ci_failed: error
+
+flags:
+  jaeger-ui:
+    paths:
+      - "packages/jaeger-ui/src/"
+    carryforward: true  # see the breakdown in the PR
+
+  plexus:
+    paths:
+      - "packages/plexus/src/"
+    carryforward: true
 
 ignore:
   - "packages/jaeger-ui/src/utils/helpers/ReactShallowRenderer.js"
--- a/.github/workflows/lint-build.yml
+++ b/.github/workflows/lint-build.yml
@@ -32,7 +32,7 @@ jobs:
       - name: Ensure PR is not on main branch
         uses: jaegertracing/jaeger/.github/actions/block-pr-from-main-branch@main
 
-  validate-renovate-config:
+  validate-3rd-party-configs:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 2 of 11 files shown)


yurishkuro (dismissed):

packages/jaeger-ui/src/components/App/index.tsx
```diff
@@ -32,65 +32,62 @@ import './index.css';
 import { store } from '../../utils/configure-store';
 import ThemeProvider from './ThemeProvider';
 
-export default class JaegerUIApp extends Component<{}> {
-  constructor(props: {}) {
-    super(props);
+export default function JaegerUIApp() {
+  useEffect(() => {
```
yurishkuro: why is this appropriate place to do the initialization logic, especially processScripts(), which is not idempotent?

gulshank0: hey I think the solution, is not to initialization logic in  inside the useEffect but out side like this : 
```JaegerAPI.apiRoot = DEFAULT_API_ROOT;
processScripts();

export default function JaegerUIApp() {
  // No useEffect
  return 
  );
}

Comment (gulshank0) [reply packages/jaeger-ui/src/components/App/index.tsx]: and declaring this: ```JaegerAPI.apiRoot = DEFAULT_API_ROOT;```
 outside as well.

Comment (yurishkuro) [reply packages/jaeger-ui/src/components/App/index.tsx]: what kind of testing can you do to validate?

Comment (gulshank0) [reply packages/jaeger-ui/src/components/App/index.tsx]: For the Testing we have to change the test file little bit now : 
The initialization tests were updated to verify module-level initialization:
- should initialize API and process scripts at module load time - confirms JaegerAPI.apiRoot is set and processScripts() is called exactly once when the module loads
- should not re-initialize on component re-renders - confirms initialization doesn't repeat on re-renders.

1)Why module-level initialization is correct :
a). processScripts() appends <script> elements to the DOM and is not idempotent. Using useEffect would cause issues with:
a). React StrictMode (effects run twice in dev)
b). Component remounting
-Module-level initialization guarantees exactly-once execution

And the test performing are : 
`npm test`  (passes)
`npm run build`  (passes)

**Comment by gulshank0** [comment]
Hi @yurishkuro, please let me know if there’s anything else I should take care of. Thanks!

**Comment by yurishkuro** [comment]
@Parship12 please review, this feels like a risky change

**Comment by ghost** [comment]
@yurishkuro I first thought about adding an idempotency guard. Before doing that, i ran HMR testing in the browser.
Result: no duplicate scripts were added (with or without a guard). Based on this testing, the current approach appears safe.

HMR testing in browser:
- started dev server
- opened browser Devtools and counted script elements in `document.body`
- made multiple edits to `App/index.tsx` to trigger HMR repeatedly
- same script count after each HMR update

Also I don’t see any other issues with the overall component conversion. Everything appears to be working as expected. please let me know if there is any other testing approach you’d recommend

File: packages/jaeger-ui/src/components/App/index.tsx [hunk 2]
```diff
@@ -32,65 +31,60 @@ import './index.css';
...
 import ThemeProvider from './ThemeProvider';
 
-export default class JaegerUIApp extends Component<{}> {
-  constructor(props: {}) {
-    super(props);
-    (JaegerAPI as any).apiRoot = DEFAULT_API_ROOT;
-    processScripts();
-  }
+JaegerAPI.apiRoot = DEFAULT_API_ROOT;
+processScripts();
```
yurishkuro, line 37: please add comments explaining why this is done at module level

gulshank0: The comment is done:   Initialize API configuration and process configuration scripts at module level to ensure they run once when the application is loaded, before any components are rendered

ghost: I think we should also add in the comment:
- why not useEffect
- and about HMR testing confirmation

@yurishkuro what do you think?

gulshank0: hey @Parship12, I think these comments works fine if needed:
=>  This is intentionally not in useEffect because:
 1. These configurations must be set before any component renders or makes API calls
 2. useEffect runs after render, which would be too late for initial API requests
 3. Module-level execution guarantees single initialization even with React StrictMode's double-mounting.
=>  Verified: HMR works correctly - module re-execution on hot reload does not cause issues

PR #3348, opened 2026-01-09
Fix Trace Statistics dark mode readability issues (#3322)

### LINKED ISSUE #3322: [Bug]: Trace Statistics view in dark mode

chinmay3012:
Resolves #3322 

## Description of the changes
-*Relative Bar Visualization*: Replaced the legacy background-color heatmap logic with a dedicated <RelativeBar /> component. This provides significantly better readability across both light and dark modes and eliminates text contrast issues.

-*Performance Optimization*: Optimized the rendering of rows in Trace Statistics by precomputing the maximum column values once per render cycle, preventing $O(N^2)$ overhead on large traces.

-*Search Highlighting*: Switched to var(--trace-emphasis-highlight) for search matches, ensuring they are vibrant and visible in both light and dark themes.

-*SQL Popup & Detail Rows*: Modernized the SQL Tag popup and subgroup rows to use CSS variables and transparent backgrounds, fixing the "white box" and theme-inconsistent styling issues in dark mode.

-*UI State Consistency*: Initialized 'Color by' (Bars) to be enabled by default and refactored the header as a controlled component to ensure the UI state and visualization are always synchronized.

- Manual inspection of the color generation logic and verification of CSS variable mapping. The use of `rgba` was chosen as a universal solution that preserves the heatmap utility in light mode while ensuring contrast safety in dark mode.

## Checklist
- [ done ] I have read https://github.com/jaegertracing/jaeger/blob/master/CONTRIBUTING_GUIDELINES.md
- [ done ] I have signed all commits
- [ done ] I have added unit tests for the new functionality
- [ done ] I have run lint and test steps successfully


yurishkuro:
we already have a PR where we discussed a different approach

https://github.com/jaegertracing/jaeger-ui/pull/3328#issuecomment-3719786941

chinmay3012:
Hi @yurishkuro,
I've addressed all the feedback from your review:
 **Extracted relative bars into a reusable component** ([RelativeBar.tsx](cci:7://file:///Users/chinmaymehrotra/Desktop/GSOC/jaegertracing/jaeger/jaeger-ui/packages/jaeger-ui/src/components/common/RelativeBar.tsx:0:0-0:0))  
   - Now used in both [TraceStatistics](cci:2://file:///Users/chinmaymehrotra/Desktop/GSOC/jaegertracing/jaeger/jaeger-ui/packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx:111:0-409:1) and [TraceSpanView](cci:2://file:///Users/chinmaymehrotra/Desktop/GSOC/jaegertracing/jaeger/jaeger-ui/packages/jaeger-ui/src/components/TracePage/TraceSpanView/index.tsx:37:0-312:1)
   - Provides subtle visual comparison directly within the selected column
**Reduced number precision**  
   - Applied [formatDurationCompact](cci:1://file:///Users/chinmaymehrotra/Desktop/GSOC/jaegertracing/jaeger/jaeger-ui/packages/jaeger-ui/src/utils/date.ts:221:0-256:1) utility for readable formatting
   - Shows compact values (e.g., `1.23ms`) with tooltips for precise values on hover
**Removed full-row coloring**  
   - Replaced the "obnoxious" pink/red backgrounds with subtle relative bars
   - Bars only appear in the column selected via the "Color by" dropdown
 **Theme-aware styling**  
   - Proper dark mode support with adequate contrast
   - Uses CSS variables for consistency
All tests pass (56/56 ✓). The implementation follows the pattern you suggested from #3310.

yurishkuro:
Please post screenshots

@Parship12 can you review?

chinmay3012:
Here are the screenshots highlighting the updates:
1. **Trace Statistics — Cleaner Relative Bars**
You’ll now see subtle relative bars only in the column chosen in the “Color by” dropdown.
This replaces the old full-row color fill and makes comparisons easier without overwhelming the UI.
2. **Multiple “Color By” Options**
Switching between Count, Total, or Self Time updates the relative bars instantly.
This makes it easy to view whichever metric matters most in the moment.
3. **Full Dark Mode Support**
The new styling adapts to dark mode with proper contrast, so the bars and text stay clear and consistent across themes.
4. **Shared Component Across Trace Views**
The RelativeBar component is now reused in the Trace Spans View as well, making the visuals feel more unified throughout the product.
5. **Precise Tooltips**
Cells show compact values for quick scanning, but hovering reveals the exact numbers for deeper analysis — the best of both worlds.

Screenshot 2026-01-12 at 2 36 22 PM
Screenshot 2026-01-12 at 2 36 44 PM
Screenshot 2026-01-12 at 2 37 13 PM

yurishkuro:
if the values get right-aligned when the bar is in the column, why not always display them right aligned?

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.css
```diff
@@ -26,3 +26,28 @@ SPDX-License-Identifier: Apache-2.0
   background: var(--surface-component-background-hover);
   color: var(--text-primary);
 }
+
+.RelativeBar {
```
yurishkuro: Why do we need to override the styles for RelativeBar here? The style should be identical between statistics view and span table view

packages/jaeger-ui/src/components/TracePage/TraceStatistics/PopupSql.css
yurishkuro, line 1: Just curious, why do we have this "popup sql" style? Why "SQL", is it somehow singled out in the code?

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -196,15 +202,15 @@ export default class TraceStatistics extends Component<Props, State> {
     uiFind: string | null | undefined
   ) => {
     const allTableSpansChange = allTableSpans;
-    const yellowSearchCollor = 'rgb(255,243,215)';
-    const defaultGrayCollor = 'rgb(248,248,248)';
+    const yellowSearchColor = 'var(--trace-emphasis-highlight)';
+    const defaultGrayColor = 'var(--surface-primary)';
```
yurishkuro: Since you're making these changes, can we clean this up? The code should not be dealing with "colors" and using colors for business logic like `record.searchColor !== 'transparent'`. The component properties should reflect business functions, like having a boolean property `searchMatch` which controls a class added to the styles, and then the color considerations should be all in .css based on the class names.

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -299,8 +303,41 @@ export default class TraceStatistics extends Component<Props, State> {
...
+        if (val.suffix === 'ms') {
+          const microseconds = ((cell as number) * 1000) as any;
+          const compactValue = formatDurationCompact(microseconds);
+          const preciseValue = `${cell}ms`;
+          displayValue = <Tooltip title={preciseValue}>{compactValue}</Tooltip>;
+        } else if (val.suffix === '%') {
+          displayValue = `${cell}%`;
+        }
+
+        if (isColored) {
```
yurishkuro:
```suggestion
        if (includeBars) {
```

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -299,8 +303,41 @@ export default class TraceStatistics extends Component<Props, State> {
               {cell}
             </span>
           );
-        return `${cell}${val.suffix}`;
+
+        const isColored = val.attribute === this.state.colorByAttribute;
+        let displayValue: React.ReactNode = cell;
+
+        if (val.suffix === 'ms') {
```
yurishkuro: this seems fragile. The timestamp suffix could be `s`, `ms`, `us`, etc. Instead of reading the suffix we should add a field to column metadata indicating which type of value that is (enum: count | percent | time)

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -299,8 +303,41 @@ export default class TraceStatistics extends Component<Props, State> {
...
+            <div
+              style={{
+                display: 'flex',
+                alignItems: 'center',
+                justifyContent: 'space-between',
+                width: '100%',
+              }}
+            >
+              <RelativeBar value={cell as number} maxValue={max} />
+              <div style={{ whiteSpace: 'nowrap', fontFamily: 'monospace', fontSize: '12px' }}>
```
yurishkuro: style should be in .css

packages/jaeger-ui/src/components/common/RelativeBar.css
```diff
@@ -0,0 +1,24 @@
+/*
+Copyright (c) 2024 The Jaeger Authors.
```
ghost:
```suggestion
Copyright (c) 2026 The Jaeger Authors.
```

packages/jaeger-ui/src/components/common/RelativeBar.css
```diff
@@ -0,0 +1,24 @@
...
+SPDX-License-Identifier: Apache-2.0
+*/
+
+.RelativeBar {
```
ghost, line 6: why duplicate definitions of `.RelativeBar` and `.RelativeBar--fill` classes: `index.css`?

packages/jaeger-ui/src/components/common/RelativeBar.tsx
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) 2024 The Jaeger Authors.
```
ghost:
```suggestion
// Copyright (c) 2026 The Jaeger Authors.
```

packages/jaeger-ui/src/components/TracePage/TraceStatistics/PopupSql.css
ghost: I think because sql queries can be very long

chinmay3012:
@yurishkuro @Parship12 I have updated the changes needed and resolved the queries . Thank you for the feedback . 

Files changed
index.tsx - Main component logic refactored
index.css - Added new CSS classes, removed duplicates
types.ts - Updated interface with searchMatch boolean
tableValues.ts - Updated initialization code
index.test.js - Updated test assertions
RelativeBar.tsx - Added CSS import
RelativeBar.css - Updated copyright year

chinmay3012:
Hi @yurishkuro @Parship12 Can you please review the updated code and suggest any feedback if needed ?

chinmay3012:
Hey @Parship12 can you please review the PR ? 
I have already updated the codes 3 days ago , feedback is appreciated .

chinmay3012:
hey @yurishkuro requesting for review of the updated PR .

chinmay3012:
Screenshot 2026-01-23 at 11 11 16 AM
Screenshot 2026-01-23 at 11 11 23 AM

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -196,15 +216,13 @@ export default class TraceStatistics extends Component<Props, State> {
     uiFind: string | null | undefined
   ) => {
     const allTableSpansChange = allTableSpans;
-    const yellowSearchCollor = 'rgb(255,243,215)';
-    const defaultGrayCollor = 'rgb(248,248,248)';
+    // Initialize all entries as not matching search
     for (let i = 0; i < allTableSpansChange.length; i++) {
+      // Parent rows with subgroups should not be highlighted
       if (!allTableSpansChange[i].isDetail && allTableSpansChange[i].hasSubgroupValue) {
-        allTableSpansChange[i].searchColor = 'transparent';
```
yurishkuro, line 203: it seems that previously this variable had 3 states: gray, yellow, and transparent. What did transparent mean? How is that use case handled now that the variable was converted to a Boolean?

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.ts
```diff
@@ -273,9 +273,9 @@ function valueFirstDropdown(selectedAttributeKey: string, trace: IOtelTrace, use
         selfMax: resultValue.selfMax,
         percent: resultValue.percent,
         color: '',
-        searchColor: 'transparent',
+        searchMatch: false,
         parentElement: '',
-        colorToPercent: 'rgb(248,248,248)',
+        colorToPercent: 'transparent',
```
yurishkuro: the semantics are hard to understand here. And is this field even used now?

packages/jaeger-ui/src/components/TracePage/TraceStatistics/TraceStatisticsHeader.tsx
```diff
@@ -49,11 +49,14 @@ export default class TraceStatisticsHeader extends Component<Props, State> {
...
+    const checkboxStatus = false;
+    const colorBy = this.getValue(checkboxStatus);
     this.props.handler(
       getColumnValues(serviceName, this.props.trace, this.props.useOtelTerms),
       getColumnValues(serviceName, this.props.trace, this.props.useOtelTerms),
       serviceName,
-      null
+      null,
+      colorBy
     );
```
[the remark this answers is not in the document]
yurishkuro: don't know about the proposed solution but the concern seems valid.

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
chinmay3012: Previously, searchColor used three values: yellow (match), gray (default), and transparent. The transparent state was being used for parent rows with subgroups to ensure they did not get a background highlight even if they matched a search.

In the new searchMatch boolean logic, I've simplified this:

Matches get the TraceStatistics--searchMatch CSS class via onCell.

Non-matches (and the old 'gray'/default rows) get no class, keeping the theme background.
The 'transparent' case is explicitly handled by setting searchMatch to false for parent rows, ensuring they never receive the highlight class. This makes the logic much easier to follow...

packages/jaeger-ui/src/components/TracePage/TraceStatistics/tableValues.ts
chinmay3012: You're absolutely right. With the introduction of the RelativeBar component for data visualization, colorToPercent became dead code.

I have just pushed a commit that:
Removes the colorToPercent field from the ITableSpan interface.
Deletes the generateColor.ts helper and its tests entirely, as they are no longer used.
Cleans up all remaining assignments and test data throughout the component. This makes the semantics much  cleaner...

chinmay3012:
Screenshot 2026-02-03 at 12 30 04 AM

Notice how the white text is perfectly readable against that gold highlight in Dark Mode when searched .
Code identifies that parent rows should remain "transparent" to keep the UI clean, even if the text matches.

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
yurishkuro: hm, but shouldn't the condition be `|| allTableSpansChange[i].hasSubgroupValue` instead of `&&`?

yurishkuro (commented):
please respond to all open comments

yurishkuro:
@chinmay3012 please make sure all commits are signed.

Also, did you verify that sub-group rows are displayed correctly? The above screenshot shows them collapsed, but on main if I expand them I get incorrect bright background:

image

chinmay3012:
Hi @yurishkuro , sorry i was unable to update due to university exams ongoing . 

<li>Fixed the sub-groups display error {Refer Screenshot 1 attached below}  . 

<li>Also did an update in which <b>Expansion</b> is prevented if all children are "Without Tag/Attribute" entries, as these are redundant when they represent 100% of the parent group's spans.{Refer Screenshot 2 attached}.

<li>Signed off the commit . 

<h3>Screenshot 1</h3>
Screenshot 2026-02-06 at 10 53 30 AM
<br></br><br></br>

<H3> Expansion of rows prevented when they are not having any tags definition </H3>
<p>For example , below is the screenshot where only <b>mysql</b> is getting expanded when sub-group db.type is selected. </p>

Screenshot 2026-02-06 at 10 53 40 AM

Ps : Signed off all the commits .

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
chinmay3012: > hm, but shouldn't the condition be `|| allTableSpansChange[i].hasSubgroupValue` instead of `&&`?

Using `||` instead of `&&` as you suggested would have incorrectly suppressed highlights for almost every row in the table (since most rows have a subgroup value), effectively breaking the search feature.

I've refactored the searchInTable logic to be more perfect.
I used `&&` in a final cleanup loop specifically to target rows that are both parents (`!isDetail`) and contain subgroup values. This will ensure the group headers to stay transparent while their matching child spans receive the highlight, which keeps the visual focus on the specific data points.

I also added a check to ensure this 'transparency' is only enforced when subgroups are actually present. This fixes an issue where rows wouldn't highlight at all when only a single aggregation (e.g., just Service Name) was selected.
Let me know if you have any further thoughts!

package-lock.json
yurishkuro, line 1: is this needed for this PR?

packages/jaeger-ui/src/components/TracePage/TraceStatistics/TraceStatisticsHeader.tsx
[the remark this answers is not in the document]
yurishkuro: ?

package-lock.json
chinmay3012: Thanks for the review , you're right. Those changes were unintended and likely caused by a local environment mismatch. I have reverted package.json and package-lock.json to match the main branch and pushed a clean commit

packages/jaeger-ui/src/components/TracePage/TraceStatistics/TraceStatisticsHeader.tsx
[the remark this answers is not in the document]
yurishkuro: please address this

yurishkuro (commented):
please address comments in packages/jaeger-ui/src/components/TracePage/TraceStatistics/TraceStatisticsHeader.tsx

packages/jaeger-ui/src/components/TracePage/TraceStatistics/TraceStatisticsHeader.tsx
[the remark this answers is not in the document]
chinmay3012: The concern about the default state mismatch was definitely valid. I've updated the TraceStatisticsHeader to align its initial state with the parent component by setting checkboxStatus to true by default.

I also fixed the initialization order in the constructor to ensure state is assigned before being accessed, and updated the Checkbox to be a controlled component. This ensures the 'Relative Bars' are visible by default and the state remains consistent across components. Thanks for the feedback!"

I have already modified the files in your local workspace to include this fix. You can now review the changes in TraceStatisticsHeader.tsx .

packages/jaeger-ui/src/components/TracePage/TraceStatistics/TraceStatisticsHeader.tsx
```diff
@@ -49,19 +49,20 @@ export default class TraceStatisticsHeader extends Component<Props, State> {
...
+      checkboxStatus: true,
     };
+    const colorBy = this.getValue(this.state.checkboxStatus);
+    this.props.handler(
+      getColumnValues(serviceName, this.props.trace, this.props.useOtelTerms),
+      getColumnValues(serviceName, this.props.trace, this.props.useOtelTerms),
+      serviceName,
+      null,
+      colorBy
+    );
```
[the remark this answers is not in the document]
chinmay3012: "Good catch! I shifted from the rgba heatmap approach to the <RelativeBar /> visualization because it provides much better readability in both light and dark modes. I'll update the PR description to accurately reflect this upgrade.

packages/jaeger-ui/src/components/TracePage/TraceSpanView/index.tsx
```diff
@@ -177,26 +178,7 @@ export default class TraceSpanView extends Component<Props, State> {
...
-                      height: '100%',
-                      background: 'var(--interactive-primary)',
-                      borderRadius: '2px',
-                    }}
-                  />
-                </div>
+                <RelativeBar value={span.duration} maxValue={this.state.maxDuration} />
                 <div
                   style={{
                     whiteSpace: 'nowrap',
```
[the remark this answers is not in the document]
chinmay3012: Definitely a valid concern. Calculating the max via a spread operator inside the cell renderer was $O(N^2)$ and would definitely cause lag on large traces. I've optimized this by precalculating the activeMax once per render cycle outside the column mapping. This is fixed in the latest commit!"

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -299,17 +320,42 @@ export default class TraceStatistics extends Component<Props, State> {
...
+          displayValue = <Tooltip title={preciseValue}>{compactValue}</Tooltip>;
+        } else if (val.valueType === 'percent') {
+          displayValue = `${cell}%`;
+        }
+
+        if (includeBars) {
+          const max = Math.max(...this.state.tableValue.map(r => (r as any)[val.attribute] as number), 1);
+          return (
+            <div className="TraceStatistics--valueContainer">
+              <RelativeBar value={cell as number} maxValue={max} />
```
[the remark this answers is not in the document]
chinmay3012: the percentage variable became redundant after the refactor to use RelativeBar. I've removed the dead code to keep things clean. Fixed!"

chinmay3012:
> please address comments in packages/jaeger-ui/src/components/TracePage/TraceStatistics/TraceStatisticsHeader.tsx

Just pushed those changes for TraceStatisticsHeader.tsx. Fixed that state initialization order to kill off those race conditions, and the 'Color by' toggle is now properly controlled (and defaults to on).

Also did a bit of campsite cleaning in the parent component: precomputing max values now so it doesn't lag on heavy traces, and the search logic is a bit cleaner/more functional. Tests are passing—let me know what you think!

chinmay3012:
@yurishkuro please review if its safe to merge , already responded to the comments a week ago

packages/jaeger-ui/src/components/TracePage/TraceStatistics/TraceStatisticsHeader.tsx
```diff
@@ -49,19 +49,20 @@ export default class TraceStatisticsHeader extends Component<Props, State> {
...
-
     this.state = {
       valueNameSelector1: serviceName,
       valueNameSelector2: null,
       valueNameSelector3: 'Count',
-      checkboxStatus: false,
+      checkboxStatus: true,
     };
+    const colorBy = this.getValue(this.state.checkboxStatus);
+    this.props.handler(
```
yurishkuro: * can we keep this code in the original position?

packages/jaeger-ui/src/components/TracePage/TraceStatistics/TraceStatisticsHeader.tsx
```diff
@@ -49,19 +49,20 @@ export default class TraceStatisticsHeader extends Component<Props, State> {
...
-    );
-
     this.state = {
       valueNameSelector1: serviceName,
       valueNameSelector2: null,
       valueNameSelector3: 'Count',
-      checkboxStatus: false,
+      checkboxStatus: true,
     };
+    const colorBy = this.getValue(this.state.checkboxStatus);
```
yurishkuro: why do we need to read this back from state that we just set above?

packages/jaeger-ui/src/components/TracePage/TraceStatistics/TraceStatisticsHeader.tsx
```diff
@@ -49,19 +49,20 @@ export default class TraceStatisticsHeader extends Component<Props, State> {
...
-      serviceName,
-      null
-    );
-
     this.state = {
       valueNameSelector1: serviceName,
       valueNameSelector2: null,
       valueNameSelector3: 'Count',
-      checkboxStatus: false,
+      checkboxStatus: true,
```
yurishkuro: why is this flipping to true?

package-lock.json
yurishkuro: this is still in the pr, please restore `git checkout main <filename>`

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -119,12 +136,11 @@ export default class TraceStatistics extends Component<Props, State> {
       wholeTable: [],
       valueNameSelector1: getServiceName(),
       valueNameSelector2: null,
+      colorByAttribute: 'count',
     };
 
     this.handler = this.handler.bind(this);
     this.togglePopup = this.togglePopup.bind(this);
-
-    this.searchInTable(this.props.uiFindVertexKeys!, this.state.tableValue, this.props.uiFind);
```
yurishkuro, line 127: why is it safe to delete?

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -138,12 +154,8 @@ export default class TraceStatistics extends Component<Props, State> {
   }
 
   changeTableValueSearch() {
-    this.searchInTable(this.props.uiFindVertexKeys!, this.state.tableValue, this.props.uiFind);
-    // reload the componente
-    const tableValueState = this.state.tableValue;
     this.setState(prevState => ({
-      ...prevState,
-      tableValue: tableValueState,
+      tableValue: this.searchInTable(this.props.uiFindVertexKeys!, prevState.tableValue, this.props.uiFind),
```
yurishkuro: is searchInTable() a pure function (computes results without side effects)?

yurishkuro:
@chinmay3012 this was pretty close to completion. Can you resolve merge conflicts and address comments?

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
chinmay3012: It's completely safe (and necessary) to delete for two main reasons:
1. **It was dead code:** `searchInTable` does not mutate the original array in-place; it maps over the array, creates a new one with updated `searchMatch` properties, and returns it. In the old constructor, this method was being called but its return value was completely ignored, meaning it had absolutely zero effect on the component's state.
2. **Redundant on initialization:** Inside the constructor, `this.state.tableValue` is initially defined as just an empty array (`[]`). Attempting to run a search over an empty array on component initialization is completely unnecessary. The actual table data is correctly populated and searched later when `TraceStatisticsHeader` triggers the `handler()` method.

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
chinmay3012: Yes, exactly! `searchInTable()` is a pure function. 

It relies entirely on the arguments passed into it (`uiFindVertexKeys`, `allTableSpans`, and `uiFind`) and doesn't reach outside its scope. Inside, it uses `.map(s => ({ ...s, searchMatch: false }))` to create a fresh copy of the array and its objects, guaranteeing it doesn't mutate the original array passed in. It computes the new `searchMatch` flags on that copy and returns it without causing any side effects. 
This is also why the previous implementation was flawed—it was calling a pure function and throwing away the returned array instead of using it to update the state!

If you look at the very first line inside searchInTable, it does this:
`const allTableSpansChange = allTableSpans.map(s => ({ ...s, searchMatch: false }));`
`map()` creates a brand new array.
`{ ...s }` uses the spread operator to create a brand new object in memory for every single row, copying the properties from the old object.
Because it creates a completely new array full of newly cloned objects, all the subsequent loops where it does `allTableSpansChange[i].searchMatch = true` are only mutating the clones. It never touches or mutates the original allTableSpans array that was passed into it.
It also never directly references or alters this.state or this.props—it strictly uses the inputs provided in its arguments.

chinmay3012 (commented):
.

chinmay3012:
> @chinmay3012 this was pretty close to completion. Can you resolve merge conflicts and address comments?

Hi , I have resolved the conflicts and updated the branch as well. <br>
Summary of final updates:
1. Refactored Visualization: Successfully migrated from the legacy heatmap logic to the semantic <RelativeBar /> component, resolving the dark mode readability issues.
2. Cleanup: Removed the now-obsolete generateColor utility and its associated tests to keep the codebase lean.
3. Code Quality: Performed a project-wide format fix and ensured all unit tests (including the Trace Statistics suite) are passing.
4. DCO Compliance: All commits have been squashed and signed-off to meet project requirements.

yurishkuro:
## Code Review

### Positive Changes

- **`searchColor: string` → `searchMatch: boolean`** is a clean semantic improvement. Separating "does this row match?" from "what color to use?" is the right abstraction.
- **Deletion of `generateColor.ts`** is correct. The function had O(N²) behavior (inner loop for max-finding) and mutated the input array in-place rather than returning a new one (`const tableValue2 = tableValue` is not a copy).
- **CSS variables migration** in `PopupSql.css` and `index.css` is correct and necessary.
- **`searchInTable` now returns a new array** via `.map()` instead of mutating the input — better.
- **`changeTableValueSearch` fix** — the original code read state and immediately used it (stale read); the new version using `prevState` in `setState` is correct.
- **Controlled checkbox** (`checked={checkboxStatus}`) fixes the uncontrolled/controlled mismatch bug.

---

### Issues

#### Critical

**1. Pagination removed**

```tsx
pagination={false}
```

The original had configurable pagination (10/20/50/100 rows). Removing it entirely is a performance regression. Jaeger UI is regularly tested with traces up to 80k spans — with two-level grouping (service + operation) the row count can grow very large. Please restore pagination or explain why it was removed.

---

#### Medium

**2. `!important` overrides**

```css
.span-view-table .ant-table-tbody > tr:hover > td {
  background-color: var(--surface-component-background-hover) !important;
}
.TraceStatistics--searchMatch {
  background: var(--trace-emphasis-highlight) !important;
  border-color: var(--trace-emphasis-highlight) !important;
}
```

`!important` to fight Ant Design's specificity is fragile. Also, `border-color` has no visible effect on `<td>` elements — antd controls table borders at the component level. The intended "highlighted border" effect likely doesn't render. Consider `outline` or `box-shadow` if a border effect is desired.

**3. Parent-row highlight suppression is post-hoc**

```tsx
// First, marks parent rows as searchMatch=true in the loops above...

// Then immediately un-marks them:
const hasDetails = allTableSpansChange.some(s => s.isDetail);
if (hasDetails) {
  for (...) {
    if (!row.isDetail && row.hasSubgroupValue) row.searchMatch = false;
  }
}
```

If a parent group's own name matches the search term (e.g., the user searches for a service name), the parent row is silently de-highlighted while its children are highlighted — confusing UX. Better to guard against setting `searchMatch = true` on those rows in the first place.

**4. Missing test coverage for bar rendering**

`generateColor.test.js` is correctly deleted, but no tests cover the new `RelativeBar`-based cell rendering. There are no tests that the bar appears in the right column or that `activeMax` is computed correctly.

---

#### Minor

**5. `key` based on joined names can collide**

```tsx
key={groupedAndSubgroupedSpanData.map(r => r.name).join(',')}
```

If a service name contains a comma, this produces key collisions (e.g., rows `["a,b", "c"]` and `["a", "b,c"]` yield the same key). The old length-based key was crude but collision-free.

**6. Default sort direction change is undocumented**

```tsx
// was: 'ascend'
defaultSortOrder: 'descend'
```

This is arguably the right default (highest count first), but it's a behavior change not mentioned in the PR description.

**7. `percent` column bars scale relative, not absolute**

When `colorByAttribute === 'percent'`, `RelativeBar` renders `value / activeMax` where `activeMax` is the max percent value in the table rather than 100. Bars are scaled relative to each other rather than as absolute percentages. For a column that already represents a ratio this is arguably misleading — consider using `maxValue={100}` for the percent column specifically.

chinmay3012:
Here is the issue-by-issue markdown reply addressing each of the author's review comments:

***

### 1. Pagination removed (Critical)
 *Review Comment: Restoring configurable pagination (10/20/50/100 rows) in the `<Table>` component.*

**Response**: Restored configurable pagination in the statistics `<Table>` component with page size options of `10`, `20`, `50`, and `100` (defaulting to `10`). This ensures optimal rendering performance when loading traces with large numbers of unique groupings.

---

### 2. Specificity and `!important` overrides (Medium)
 *Review Comment: Avoid using `!important` to override Ant Design defaults and fix the cell border styling (as `border-color` on `<td>` has no effect).*

**Response**: 
* Prefixing CSS selectors with `.span-view-table.ant-table-wrapper` increased specificity, allowing for clean styling overrides without any `!important` flags.
* Replaced `border-color` on `.TraceStatistics--searchMatch` with an inset box shadow (`box-shadow: inset 0 0 0 1px var(--trace-emphasis-highlight);`), which correctly highlights the borders of matching cells.

---

### 3. Parent-row highlight suppression is post-hoc (Medium)
 *Review Comment: Implement the parent highlight suppression logic directly in the search loops instead of un-marking them post-hoc. Ensure parent rows whose own names match the search term are highlighted.*

**Response**: 
* Removed the post-hoc cleanup loop from `searchInTable`.
* Added a helper function `isParentWithSubgroup` to guard against setting `searchMatch = true` for parent rows *only* during match propagation from child rows. 
* Direct search matches on parent row names (e.g. searching for a service name directly) will now correctly highlight the parent row.
* Modified the table `key` to include the active search term and subgroup selection. This forces a table remount when search query or grouping changes, allowing the table to correctly apply expansions and display matching child rows immediately.

---

### 4. Missing test coverage for bar rendering (Medium)
*Review Comment: Add unit tests in `index.test.jsx` covering the relative bar cell rendering and scaling.*

**Response**: Added two new unit tests to `index.test.jsx`:
1. `renders RelativeBar inside the active column and scales them correctly`: Asserts that `RelativeBar` width scales proportionally based on the column's computed `activeMax` (e.g., `50%` and `100%` widths).
2. `scales percent column relative to 100 rather than activeMax`: Asserts that percentage bars scale to absolute percentages (e.g. `80%` and `40%` width) rather than scaling relative to the highest percentage in the table.

---

### 5. Key based on joined names can collide (Minor)
 *Review Comment: Use a collision-free representation for the table `key` instead of a comma-separated join of names.*

**Response**: Replaced the comma-separated join key with a stringified array representation using `JSON.stringify()`, which is collision-free and safely forces updates when columns or row lists change.

---

### 6. Default sort order (Minor)
 *Review Comment: Confirm or clarify why the count column sort direction changed to `descend`.*

**Response**: Changing the default sort direction of the Count column to `descend` (highest count first) was done as a deliberate UX improvement. In trace analysis, identifying the most frequent operations is the most common path, so starting with a descending sort saves the user an extra click.

---

### 7. Percent column bars scale relative (Minor)
*Review Comment: Ensure the relative bar inside the `percent` column uses `maxValue={100}` instead of `activeMax` to reflect absolute percentages.*

**Response**: Updated the `RelativeBar` instantiation in `index.tsx` to set `maxValue={val.attribute === 'percent' ? 100 : activeMax}`, ensuring the `percent` column renders absolute percentage widths.

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -299,64 +324,61 @@ export default class TraceStatistics extends Component<Props, State> {
...
+    const groupedAndSubgroupedSpanData: ITableSpan[] = this.state.tableValue
+      .filter(s => !s.isDetail)
+      .map((parent, i) => ({
+        ...parent,
+        key: i.toString(),
+        children: this.state.tableValue
+          .filter(s => s.isDetail && s.parentElement === parent.name)
+          .map((child, j) => ({
+            ...child,
+            key: `${i}-${j}`,
```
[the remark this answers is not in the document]
chinmay3012: While using stable, deterministic keys derived from row identity is generally a React best practice, the current implementation is stable and not prone to expansion glitches or incorrect row identity mismatches during sorting/filtering.

This is because the parent <Table> component uses a dynamic key prop:
key={`${this.state.valueNameSelector1}-${this.state.valueNameSelector2}-${this.props.uiFind}-${JSON.stringify(
  groupedAndSubgroupedSpanData.map(r => r.name)
)}`}

This key configuration ensures that the entire `Table` component is completely unmounted and remounted from scratch whenever:
The sorting or metric column selection changes (valueNameSelector1, valueNameSelector2).
A search filter is applied or cleared (uiFind).
The underlying row list or its ordering changes (groupedAndSubgroupedSpanData.map(r => r.name)).
Because the component remounts on every structural/sorting transition, the table starts with clean state and applies expansion states fresh from the recalculated defaultExpandedRowKeys. Thus, there is no risk of React preserving incorrect row expansion states or mismatching DOM nodes.

packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx
```diff
@@ -299,64 +324,61 @@ export default class TraceStatistics extends Component<Props, State> {
...
+          })),
+      }));
```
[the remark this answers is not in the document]
chinmay3012: Pre-grouped all detail child rows into a `Map<string, ITableSpan[]>` in a single O(N) pass, then attached the children to their parent rows in a second O(N) pass.

jkowall:
Thanks for working on this. The underlying dark-mode bug in #3322 is valid, but this PR has accumulated too much unrelated scope and is now substantially behind main. It changes search behavior, table keys and remounting, sorting, pagination and expansion, duration formatting, and SQL popup styling in addition to the contrast fix, with several review concerns still unresolved.

We are going to close this PR and address #3322 in a fresh, narrowly scoped change that preserves the existing table behavior and only makes the Color by visualization theme-safe. The shared RelativeBar work also landed separately in #3855.

Thank you for the effort and the many iterations here.

PR #3360, opened 2026-01-13
feat(ui): add configurable 12h/24h time format

### LINKED ISSUE #552: Feature request: option to display times in 24 hour clock

Most of our logs, alerts, etc. show times using a 24h clock. When looking using the Jaeger UI to browse traces, it's harder to scan through times as a human due to the difference in format.

Jaeger currently displays time in 12 hour format with a trailing am/pm, and with a leading zero, which makes it look like a 24 hour time before seeing the suffix.

Adding an option to UI configuration to swap between alternate time formats (swapping between 12h and 24h time is what I'm after, but could theoretically be extended to further formats depending on complexity).

Time show in spans/charts would then be `hh:mm:ss` instead of `hh:mm:ss [am|pm]`, e.g. `19:40:00` instead of `07:40:00 pm`.

Harshdev098:
-  fixes #552 

## Description of the changes
- Added a new timeFormat UI configuration option ('12h' | '24h')
- Updated timestamp formatting logic to respect the configured format

-  npm run test

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ResultItem.tsx
+++ b/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ResultItem.tsx
@@ -126,6 +126,9 @@ export default function ResultItem({
           <Col span={4} className="ub-p3 ub-tx-right-align">
             {formatRelativeDate(startTime / 1000)}
             <Divider vertical />
+            {/* For 12h format: slice off the trailing " am"/" pm" (including the preceding space),
+            keep only the suffix ("am"/"pm"), and re-join with &nbsp; for non-breaking layout.
+            For 24h: no suffix exists → use full timeStr without slicing to avoid corruption. */}
             {is24h ? timeStr : `${timeStr.slice(0, -3)}&nbsp;${timeStr.slice(-2)}`}
             <br />
             <small>{fromNow}</small>
--- a/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ResultItem.tsx
+++ b/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ResultItem.tsx
@@ -126,7 +126,7 @@ export default function ResultItem({
           <Col span={4} className="ub-p3 ub-tx-right-align">
             {formatRelativeDate(startTime / 1000)}
             <Divider vertical />
-            {is24h ? timeStr : `${timeStr.slice(0, -3)}&nbsp;${timeStr.slice(-3)}`}
+            {is24h ? timeStr : `${timeStr.slice(0, -3)}&nbsp;${timeStr.slice(-2)}`}
             <br />
             <small>{fromNow}</small>
           </Col>
--- a/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ResultItem.tsx
+++ b/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ResultItem.tsx
@@ -20,6 +20,8 @@ import { formatRelativeDate } from '../../../utils/date';
 import { IOtelTrace, StatusCode } from '../../../types/otel';
 
 import './ResultItem.css';
+import { getTimeFormat } from '../../../utils/time-format';
+import { getConfigValue } from '../../../utils/config/get-config';
 
 dayjs.extend(relativeTime);
 
@@ -50,10 +52,11 @@ export default function ResultItem({
   const [numErredSpans, setNumErredSpans] = React.useState(0);
   const [timeStr, setTimeStr] = React.useState('');
   const [fromNow, setFromNow] = React.useState<string | boolean>('');
+  const is24h = getConfigValue('timeFormat') === '24h';
 
   React.useEffect(() => {
     const startTimeDayjs = dayjs(startTime / 1000);
-    setTimeStr(startTimeDayjs.format('h:mm:ss a'));
+    setTimeStr(startTimeDayjs.format(getTimeFormat(true)));
     setFromNow(startTimeDayjs.fromNow());
 
     const errored = new Set<string>();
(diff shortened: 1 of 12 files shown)


packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ResultItem.tsx
```diff
@@ -123,7 +126,7 @@ export default function ResultItem({
           <Col span={4} className="ub-p3 ub-tx-right-align">
             {formatRelativeDate(startTime / 1000)}
             <Divider vertical />
-            {timeStr.slice(0, -3)}&nbsp;{timeStr.slice(-2)}
+            {timeStr}
```
yurishkuro: what is happening here? Why 12/24h format affects this specific code?

packages/jaeger-ui/src/utils/time-format.ts
```diff
@@ -0,0 +1,11 @@
+import { getConfigValue } from './config/get-config';
+
+export function getTimeFormat(withSeconds = true) {
```
yurishkuro: define two separate methods instead of using Boolean argument, `getTimeFormat(false)` is not intuitive at all

packages/jaeger-ui/src/components/TracePage/TraceSpanView/index.tsx
```diff
@@ -222,7 +223,7 @@ export default class TraceSpanView extends Component<Props, State> {
 
           return (
             <Tooltip
-              title={`${dayjs(span.startTime / 1000).format('DD MMM YYYY hh:mm:ss A')} (${preciseValue})`}
+              title={`${dayjs(span.startTime / 1000).format(`DD MMM YYYY ${getTimeFormat(true)}`)} (${preciseValue})`}
```
yurishkuro: it doesn't make sense that this format is different from formats defined in date.ts, they should be all standardized to a single form.

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ResultItem.tsx
```diff
@@ -50,10 +52,11 @@ export default function ResultItem({
   const [numErredSpans, setNumErredSpans] = React.useState(0);
   const [timeStr, setTimeStr] = React.useState('');
   const [fromNow, setFromNow] = React.useState<string | boolean>('');
+  const is24h = getConfigValue('timeFormat') === '24h';
```
yurishkuro, line 55: where is this used?

yurishkuro (changes requested):

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ResultItem.tsx
Harshdev098: Because 12-hour time have the suffix and it was slicing and adding `&nbsp` while in 24-hour time doesn't have suffix so the slicing can break.
Btw have updated it to use both time format!standardize time formatting

packages/jaeger-ui/src/utils/time-format.ts
Harshdev098: agree!

packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ResultItem.tsx
```diff
@@ -123,7 +126,7 @@ export default function ResultItem({
           <Col span={4} className="ub-p3 ub-tx-right-align">
             {formatRelativeDate(startTime / 1000)}
             <Divider vertical />
-            {timeStr.slice(0, -3)}&nbsp;{timeStr.slice(-2)}
+            {is24h ? timeStr : `${timeStr.slice(0, -3)}&nbsp;${timeStr.slice(-2)}`}
```
yurishkuro, line 132: a) this line could use a comment explaining what exactly is being done with these slice() operations & why
b) why is this slicing dependent on 12/24 format?

packages/jaeger-ui/src/constants/default-config.ts
```diff
@@ -116,6 +116,7 @@ const defaultConfig: Config = {
   themes: {
     enabled: true,
   },
+  timeFormat: '24h',
```
yurishkuro: the issue said

> Jaeger currently displays time in 12 hour format

so why is 24h the default?

packages/jaeger-ui/src/utils/date.ts
```diff
@@ -100,7 +101,7 @@ export function formatTime(duration: number): string {
  * ```
  */
 export function formatDatetime(duration: number): string {
-  return dayjs(duration / ONE_MILLISECOND).format(STANDARD_DATETIME_FORMAT);
+  return dayjs(duration / ONE_MILLISECOND).format(`MMMM D YYYY, ${getTimeFormatWithSeconds()}.SSS`);
```
yurishkuro, line 103: Is something else using STANDARD_DATETIME_FORMAT? Don't leave unused constants laying around.

packages/jaeger-ui/src/utils/time-format.ts
yurishkuro, line 1: new file needs new tests

packages/jaeger-ui/src/constants/default-config.ts
Harshdev098: @yurishkuro  switching to 12h format causes test failures in `SearchForm.test.js`
```
- Expected: `"03:13"`    Received: `"03:13 am"`
- Expected: `"10:00"`    Received: `"10:00 am"`
```
previously the search form time formatting was hardcoded directly to `HH:mm`, with '12h' default, we now get `"03:13 am"`, breaking exact string matches.

So, have used hardcoded that SearchForm.ts to use hardcoded one instead of 12h /24h formats for inputs

PR #3402, opened 2026-01-14
Convert SvgEdges to functional component

### LINKED ISSUE #3397: [UI Refactor] Migrate SvgEdges to Functional Component
This issue is part of the larger effort to modernize the Jaeger UI codebase (Parent Issue #2610). The goal is to refactor the `SvgEdges` component from a React Class Component to a React Functional Component using Hooks.

**Target Component:** `packages/plexus/src/Digraph/SvgEdges.tsx`

### Acceptance Criteria

### Plexus-Specific Guidance

| --- | --- | --- |

| --- | --- | --- |

### Resources

thc1006:

Resolves #3397

## Description of the changes

Converted the `SvgEdges` class component to a functional component using React.memo with custom comparator.

- Added unit tests for SvgEdges component
- Run `npm run test` in packages/plexus directory

## Checklist

### DIFF
--- a/packages/plexus/src/Digraph/SvgEdges.test.js
+++ b/packages/plexus/src/Digraph/SvgEdges.test.js
@@ -1,4 +1,4 @@
-// Copyright (c) 2026 The Jaeger Authors.
+// Copyright (c) 2026 Uber Technologies, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
 import React from 'react';
--- a/packages/plexus/src/Digraph/SvgEdges.tsx
+++ b/packages/plexus/src/Digraph/SvgEdges.tsx
@@ -1,4 +1,4 @@
-// Copyright (c) 2026 The Jaeger Authors.
+// Copyright (c) 2019 Uber Technologies, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
 import * as React from 'react';


yurishkuro:
unit tests are failing

thc1006:
> unit tests are failing

Done!

thc1006:
Dear @yurishkuro,

Following up on the previous fixes, I have addressed the additional feedback:

- Added complete memoization test coverage for all props in arePropsEqual (getClassName, renderUtils, markerStartId, setOnEdge)
- Restored the original Uber copyright header in SvgEdges.tsx to maintain consistency with other plexus files
- Updated test file copyright to match the plexus package convention

All 21 tests pass with 100% coverage for SvgEdges.tsx.

Regarding the latest Copilot suggestion about testing setOnEdge as an array: the current tests verify memoization behavior for function references, which covers the primary use case. The array comparison logic is an implementation detail of isSamePropSetter utility. I can add those tests if you think they are necessary for this PR.

Ready for review when you have a chance. Thank you.

packages/plexus/src/Digraph/SvgEdges.test.js
```diff
@@ -0,0 +1,326 @@
...
+            getClassName={mockGetClassName}
+            layoutEdges={mockLayoutEdges}
+            renderUtils={mockRenderUtils}
+            setOnEdge={newSetOnEdge}
+          />
+        </svg>
+      );
+
+      expect(mockRenderCount).toBeGreaterThan(initialRenderCount);
+    });
```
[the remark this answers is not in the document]
thc1006: There is no any this kinda array `isSamePropSetter` testing in whole plexus Package.
if really need may ask dear rela maintainer !

packages/plexus/src/Digraph/SvgEdges.tsx
```diff
@@ -17,32 +17,39 @@ type TProps<T = {}> = {
...
-        layoutEdge={edge}
-        markerEndId={markerEndId}
-        markerStartId={markerStartId}
-        renderUtils={renderUtils}
-        setOnEdge={setOnEdge}
-        label={edge.edge.label}
-      />
-    ));
-  }
+// Comparison function that mirrors the original shouldComponentUpdate logic
```
yurishkuro, line 20: Why can't we simply rely on React doing the comparison? That is literally the reason why React was created.

thc1006: Sorry for the delayed reply, I was at KubeCon EU these past two weeks.

Thank you for the review. You're right, the custom comparator shouldn't be here. I raised the same point on #3403 and offered to simplify, but should have just gone ahead and done it. The isSamePropSetter call on setOnEdge is handling something the caller should stabilize with useMemo, not something the child should work around.

I'll switch to plain React.memo() here and on #3403. Both need a rebase against the toolchain changes anyway, so I'll handle them together.

thc1006:
Apologies for the back and forth here. After taking a closer look at main, I realized SvgEdges.tsx was already converted to a functional component in #3518 by @hharshhsaini. I should have checked before replying to the arePropsEqual comment earlier.

Thank you for the feedback regardless, it clarified the right approach for Nodes (#3403) as well.

One thing that might still have value: this branch includes a 326-line test file for SvgEdges covering rendering, memoization behavior, and edge cases. There don't appear to be tests for this component on main yet.

Would a standalone test PR (migrated to Vitest) be welcome? Happy to put one together if so.

PR #3420, opened 2026-01-14
refactor(plexus): convert SvgDefEntry from class to functional component

### LINKED ISSUE #3395: [UI Refactor] [plexus] Migrate SvgDefEntry to Functional Component
This issue is part of the larger effort to modernize the Jaeger UI codebase (Parent Issue #2610). The goal is to refactor the `SvgDefEntry` component from a React Class Component to a React Functional Component using Hooks.

**Target Component:** `packages/plexus/src/Digraph/SvgDefEntry.tsx`

### Acceptance Criteria

### Plexus-Specific Guidance

| --- | --- | --- |

| --- | --- | --- |

### Resources

thc1006:
## Summary
- Convert the SvgDefEntry component from class-based to functional component
- Use React.memo for shallow comparison (equivalent to PureComponent)
- Add comprehensive test suite with 17 tests

## Changes
- Replace class component with arrow function component
- Preserve generic type parameters with trailing comma syntax
- Preserve renderDefaultMarker helper function unchanged

## Test Coverage
- Basic rendering and custom renderEntry support
- Global ID generation via renderUtils
- className application
- setOnEntry prop handling
- Default marker rendering with scaling
- renderEntry callback parameter passing
- React.memo verification

Closes #3395

Signed-off-by: thc1006 <84045975+thc1006@users.noreply.github.com>

### DIFF
--- a/packages/plexus/src/Digraph/SvgDefEntry.test.js
+++ b/packages/plexus/src/Digraph/SvgDefEntry.test.js
@@ -1,4 +1,4 @@
-// Copyright (c) 2026 Uber Technologies, Inc.
+// Copyright (c) 2026 The Jaeger Authors
 // SPDX-License-Identifier: Apache-2.0
 
 import React from 'react';


jkowall:
@thc1006 please review the CI errors and copilot feedback. Thanks!

packages/plexus/src/Digraph/SvgDefEntry.test.js
```diff
@@ -0,0 +1,279 @@
...
+        </svg>
+      );
+      expect(customRender.mock.calls[0][2]).toBe('global-entry-id');
+    });
+  });
+
+  describe('React.memo behavior', () => {
+    it('is wrapped with React.memo for performance', () => {
+      expect(SvgDefEntry.$$typeof).toBeDefined();
+    });
```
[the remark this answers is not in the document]
thc1006: This suggestion has already been applied.

**Current code (line 276):**
\
**Technical clarification:** The statement that "\ property exists on all React elements" is not entirely accurate:

| Component Type | \ Value |
|----------------|---------------------|
| Regular function component | \ |
| \ | \ |
| \ | \ |

The \ property only exists on special React wrapper types (memo, forwardRef, lazy, etc.), not on plain function components. However, using the specific symbol check is indeed the better approach as it:
1. Ensures it's specifically \ (not accidentally \)
2. Provides clearer test intent

Test verified passing: 1 passed, 16 skipped (other tests filtered out).

packages/plexus/src/Digraph/SvgDefEntry.tsx
```diff
@@ -47,16 +47,17 @@ function renderDefaultMarker(
...
+  const entryProps = assignMergeCss(
+    {
+      className: getClassName('DefEntry'),
+    },
+    getProps(setOnEntry, graphState)
+  );
+  return renderEntry(graphState, entryProps, id);
+};
+
+// React.memo provides shallow comparison equivalent to PureComponent
```
Parship12: it explains the "what", not a non-obvious constraint. Please remove it.

ghost: Redundant comment

packages/plexus/src/Digraph/SvgDefEntry.test.js
```diff
@@ -0,0 +1,279 @@
...
+            <SvgDefEntry {...defaultProps} graphState={graphState2} />
+          </defs>
+        </svg>
+      );
+      const marker2 = c2.querySelector('marker');
+      const height2 = marker2.getAttribute('markerHeight');
+
+      // Both should have some height value (scaling logic is mocked)
+      expect(height1).toBeTruthy();
+      expect(height2).toBeTruthy();
```
ghost, line 234: `getValueScaler: () => () => 0.5,`
As `getValueScaler` is mocked to return 0.5, so these heights will be identical. We need to un-mock `getValueScaler` (or use the real implementation) to actually test that.

packages/plexus/src/Digraph/SvgDefEntry.test.js
```diff
@@ -0,0 +1,298 @@
+// Copyright (c) 2026 Uber Technologies, Inc.
```
yurishkuro:
```suggestion
// Copyright (c) 2026 The Jaeger Authors
```

yurishkuro:
what is the story with two open comments?

thc1006:
Dear @yurishkuro ,
Sorry, should have resolved these earlier. Both are addressed in the code.

Parship12's comment about getValueScaler being mocked to a constant was fixed in e3928c22, switching to actual.getValueScaler so the scaling test exercises real logic. (Parship12 approved after the fix.)

The Copilot TDZ suggestion is a false positive. mockAssignMergeCssResults and mockGetPropsResults are only read during test execution (component render path), not at module load time, so no TDZ issue. All 17 tests pass.

I'll resolve both threads. Let me know if you'd rather I apply the Copilot suggestion anyway.

thc1006:
Just noticed that SvgDefEntry.tsx was already converted in #3528 by @hharshhsaini. Closing this since the conversion is already on main.

@Parship12 thank you for taking the time to review and approve this. The feedback on the getValueScaler mock and the redundant comment both made the tests better, and I'll carry those lessons forward.

Similar to #3402, this branch has a 298-line test file for SvgDefEntry covering rendering, gradient calculations, and edge cases. There don't seem to be tests for this component on main yet.

If a standalone test PR (migrated to Vitest) would be useful, I'm happy to submit one. Let me know.

PR #3427, opened 2026-01-14
Migrate TraceGraph to functional component

aaryan359:
Migrate TraceGraph to a functional component #3369

## Description of the changes
- Converted TraceGraph from React Class Component to Functional Component using hooks
- Replaced state management: this.state → useState, lifecycle methods → useEffect, instance variables → useRef
- Wrapped with React.memo() to preserve PureComponent optimization behavior

- All 18 existing unit tests pass without modification to test logic
- Added new tests for lifecycle cleanup and state persistence
- Verified no visual regressions - component behavior remains identical
image

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceGraph/TraceGraph.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceGraph/TraceGraph.tsx
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import * as React from 'react';
+import { useState, useEffect, useRef, useMemo } from 'react';
 import { Card, Button, Tooltip } from 'antd';
 import { IoClose, IoHelpCircleOutline } from 'react-icons/io5';
 import cx from 'classnames';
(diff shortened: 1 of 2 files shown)


packages/jaeger-ui/src/components/TracePage/TraceGraph/TraceGraph.tsx
```diff
@@ -104,162 +101,167 @@ export const getHelpContent = (useOtelTerms: boolean) => (
...
+      if (layoutManagerRef.current) {
+        layoutManagerRef.current.stopAndRelease();
+      }
+    };
+  }, []);
 
-  toggleNodeMode(newMode: string) {
-    this.setState({ mode: newMode });
-  }
+  const toggleNodeMode = (newMode: string) => {
```
yurishkuro, line 136: this function is pointless. Since it's used in a lambda form anyway

onClick={() => toggleNodeMode(MODE_SERVICE)}

You can just directly use setMode from there.

Similar comment to the other two const functions, they provide no benefit.

yurishkuro:
already done

PR #3428, opened 2026-01-14
migrate graph component to functional

aaryan359:
migrate graph component to functional #3378

## Description of the changes
- convert to a functional component

- All  existing unit tests pass without modification to the test logic
- Added new tests for lifecycle cleanup and state persistence
image


yurishkuro:
please provide evidence of testing with the real UI, not just unit tests

Issue #3379, opened 2026-01-14
[UI Refactor] Migrate DeepDependencies to Functional Component


jkowall:
This issue is part of the larger effort to modernize the Jaeger UI codebase (Parent Issue #2610). The goal is to refactor the `DeepDependencies` component from a React Class Component to a React Functional Component using Hooks.

**Target Component:** `packages/jaeger-ui/src/components/DeepDependencies/index.tsx`

### Acceptance Criteria

- [ ] The component is converted to a Functional Component.
- [ ] `this.state` is replaced with `useState` or `useReducer`.
- [ ] Lifecycle methods (componentDidMount, etc.) are replaced with `useEffect`.
- [ ] If Redux is used, the `connect` HOC is replaced with `useSelector` and `useDispatch` hooks (optional but preferred).
- [ ] All existing PropType validations are preserved.
- [ ] **Performance:** If the original component was a `PureComponent`, the new one is wrapped in `React.memo`.
- [ ] **Visual Verification:** Attach "Before" and "After" screenshots to the PR to prove no visual regression.
- [ ] **Tests:** Existing unit tests must pass.
- [ ] **Note:** If tests rely on Enzyme's `state()` or `instance()`, they must be refactored to check the DOM or Props instead.

### Migration Guide (Cheat Sheet)

| Feature | Old (Class) | New (Functional) |
| --- | --- | --- |
| **State** | `this.state = { isOpen: false }` | `const [isOpen, setIsOpen] = useState(false);` |
| **Side Effects** | `componentDidMount()` | `useEffect(() => { ... }, [])` |
| **Props** | `this.props.traceId` | `props.traceId` (or destructure: `const { traceId } = props;`) |
| **Default Props** | `static defaultProps = { name: 'Guest' }` | `function Component({ name = 'Guest' }) { ... }` |
| **Performance** | `class MyComp extends React.PureComponent` | `export default React.memo(MyComp)` |
| **Context** | `static contextType = MyContext` | `const val = useContext(MyContext)` |
| **Refs** | `this.myRef = React.createRef()` | `const myRef = useRef(null)` |

### Common Pitfalls (Read Carefully!)

- **Stale Closures:** Be careful when using `useEffect`. If you use a variable inside the effect that comes from props or state, it must be in the dependency array. If you omit it, the effect will use the old value.
[...]

PR #3433, opened 2026-01-15
Refactor Deependecies component to functional component

aaryan359:
- Migrate DeepDependencies to Functional Component #3379 

## Description of the changes
- Converted TraceGraph, Graph, DdgNodeContent, and DeepDependencyGraphPageImpl from React Class Components to Functional Components
- Replaced this.state with useState, lifecycle methods with useEffect, instance variables with useRef, and Redux connect HOC with useDispatch/useSelector hooks
- Wrapped all components with React.memo() to preserve PureComponent optimization behavior

- All existing unit tests pass with refactored tests that verify DOM state instead of internal class methods
-  Added new tests for useEffect cleanup, state persistence across re-renders, and useMemo dependency tracking
- Verified no visual regressions - all components maintain identical rendering and behavior

### DIFF
--- a/packages/jaeger-ui/src/components/DeepDependencies/index.test.js
+++ b/packages/jaeger-ui/src/components/DeepDependencies/index.test.js
@@ -2,15 +2,17 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import * as React from 'react';
-import { render, screen } from '@testing-library/react';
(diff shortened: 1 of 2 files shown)


yurishkuro:
why are there merge conflicts? Is this component already upgraded?

PR #3438, opened 2026-01-15
refactor tracespan view to functional component

aaryan359:
Refactor tracespan view to a functional component #3383 

## Description of the changes
- Refactored TraceSpanView from a class-based component to a functional component using React hooks, without altering any UI or behavior.
- Replaced this.state and class methods with useState, useMemo, and useCallback while preserving all existing rendering logic and styles.
- Wrapped the component with React.memo to maintain the same performance characteristics as the original implementation.

- Ran the existing unit tests for TraceSpanView to confirm all tests pass without modification to expected behavior.
- Manually verified the Trace Tabular View in the Jaeger UI and compared before/after screenshots to ensure there is no visual regression.
- Interacted with the Service and Operation filters and Reset button in the browser to confirm filtering and table updates work as before.

### before: 
Screenshot from 2026-01-15 21-32-04

### After: 
image


yurishkuro:
Copilot comments not addressed

PR #3440, opened 2026-01-15
refactor(ViewingLayer): convert class component to functional component

gulshank0:
 part of: #2610 

 [UI Refactor] Migrate ViewingLayer to Functional Component #3371

## Description of the changes
- The component is converted to a Functional Component.
- Converted `React.PureComponent` class to a functional component with React.memo for equivalent performance optimization.
- Added `React.forwardRef` to support ref forwarding for testing.
- Replaced `this.state` = `{ preventCursorLine: false }` with `useState(false)` hook.
- `this._root → useRef<Element | TNil>()
this._draggerReframe`, `this._draggerStart`, `this._draggerEnd` → `useRef<DraggableManager | null>()`.
- Added `viewRangeRef`, `updateNextViewRangeTimeRef`, `updateViewRangeTimeRef` to avoid stale closure issues.
- All class methods `(_handleReframeMouseMove, _handleScrubberDragEnd, etc.)` converted to `useCallback` hooks for memoization.
- `componentWillUnmount()` → `useEffect()` cleanup function for disposing DraggableManagers.
- Testing Support
1).Added useImperativeHandle to expose internal methods (_setRoot, _getDraggingBounds, _handleScrubberEnterLeave, etc.) and state for test compatibility
2).Exported ViewingLayerHandle interface for TypeScript suppor.

#Test File 
1. Added `act()` wrappers for async state updates in scrubber tests

2. Updated `_root` access from direct assignment `(ref.current._root = null)` to method call `(ref.current._setRoot(null))`

3. Changed loop structure from `forEach` to `for...of` for proper `async/await`  handling in drag end tests

- 

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TracePageHeader/SpanGraph/ViewingLayer.test.js
+++ b/packages/jaeger-ui/src/components/TracePage/TracePageHeader/SpanGraph/ViewingLayer.test.js
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import React from 'react';
-import { render, fireEvent } from '@testing-library/react';
+import { render, fireEvent, act } from '@testing-library/react';
 import '@testing-library/jest-dom';
 import ViewingLayer, { dragTypes } from './ViewingLayer';
 import { EUpdateTypes } from '../../../../utils/DraggableManager';
@@ -39,7 +39,7 @@ describe('<SpanGraph>', () => {
     });
 
     it('throws if _root is not set', () => {
-      ref.current._root = null;
+      ref.current._setRoot(null);
       expect(() => ref.current._getDraggingBounds(dragTypes.REFRAME)).toThrow();
     });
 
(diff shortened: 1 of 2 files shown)


gulshank0:
After the recent changes here is the pic:  
Screenshot_12-Mar_01-38-16_24283


https://github.com/user-attachments/assets/05d316eb-886f-4ae7-925d-f045ba30ab01


 @yurishkuro  @jkowall  review this PR and let me know if anything more is required. Thanks.

yurishkuro:
Thanks for taking this on! The end goal (functional component, part of #2610) is right, but I think this needs a preliminary step before it's mergeable, and the size of the diff is the signal.

Most of the complexity added here exists only to keep the *existing* tests passing, not to improve the component. The `ViewingLayerHandle` + `useImperativeHandle` machinery makes the functional component impersonate a class instance — exposing `state`, `setState`, and every `_private` method through a ref — purely because the current tests reach into class internals:

```js
ref.current._setRoot(null);
ref.current.setState({ preventCursorLine: true });
ref.current._handleScrubberDragEnd(update);
expect(ref.current.state.preventCursorLine).toBe(false);
```

Those are tests of implementation details, and this PR preserves that coupling rather than removing it. The test diff actually moves the wrong way — it replaces a real DOM event with a direct internal-method call:

```diff
- fireEvent.mouseEnter(container.querySelectorAll('[data-testid="scrubber"]')[0]);
+ await act(async () => { ref.current._handleScrubberEnterLeave({ type: EUpdateTypes.MouseEnter }); });
```

The `fireEvent` version is the better test and would work fine against a functional component. It's the direct-method poking that *forces* `useImperativeHandle` to exist at all.

Suggested sequence:

1. **First PR — refactor the tests to be behavior-based** against the current class component. Drive everything through the DOM (`fireEvent` on the scrubbers / the `svg`, assert on rendered output and on `updateViewRangeTime` / `updateNextViewRangeTime` calls). No `ref.current._method`, no `.state`. This lands green with zero production changes and proves behavior is unchanged.
2. **Second PR — the migration.** With tests no longer touching internals, the component can drop `forwardRef`, `useImperativeHandle`, the `ViewingLayerHandle` type, and the `_setRoot`/`state`/`setState` shims. The diff shrinks to the genuinely interesting parts.

A couple of secondary notes (mostly downstream of the same issue):

- The three prop-mirror refs (`viewRangeRef`, `updateNextViewRangeTimeRef`, `updateViewRangeTimeRef` + their sync `useEffect`s) are a stale-closure workaround needed only because every `useCallback` has an empty dependency array. Once the tests don't require stable identities for the handle, this can likely be simplified.
- `draggerReframeRef.current ??= new DraggableManager(...)` in the render body is lazy init as a render side effect. It works, but a `useMemo`/init `useEffect` would read more clearly.

Could you split out the test-refactor-first PR? That'll make this one small and clean.

gulshank0:
@yurishkuro  oh! i will update soon . thanks for the update.

PR #3471, opened 2026-01-27
Enable react-hooks/exhaustive-deps linter rule

### LINKED ISSUE #3445: [chore]: Can we enable react-hooks/exhaustive-deps linter?
### Requirement

It seems like a very useful linter especially prior to converting the rest of the components to functional.

The scope here would be:
  * [ ] publish code to actually enable the linter, but not commit yet
  * [ ] run the linter to see how many breaks in the existing code we have
  * [ ] decide if we want to fix them all in the same PR, or create independent PRs fixing dependencies first and only then enable the linter

taanvi2205:
Closes #3445

- closes #3445

## Description of the changes
- Add eslint-plugin-react-hooks dependency
- Fix 12 violations across 6 files by adding missing hook dependencies
- Add eslint-disable comment for intentional mount-only useEffect in SearchTracePage

- npm run eslint passes with 0 errors
- npm run test passes

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/SearchTracePage/index.tsx
+++ b/packages/jaeger-ui/src/components/SearchTracePage/index.tsx
@@ -109,6 +109,8 @@ export function SearchTracePageImpl(props: SearchTracePageImplProps) {
     if (needForDiffs.length) {
       fetchMultipleTraces(needForDiffs);
     }
+    // Intentionally run only on mount, we only want to trigger the initial search
+    // and fetch diff traces once when the component loads, not on every state change.
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 


packages/jaeger-ui/src/components/DeepDependencies/index.test.js
```diff
@@ -91,7 +91,7 @@ describe('DeepDependencyGraphPage', () => {
         state: fetchedState.DONE,
         viewModifiers: new Map(),
       },
-      navigate: jest.fn(),
+      history: { push: jest.fn() },
```
yurishkuro: how is this change related to the linter?

taanvi2205:
@yurishkuro  The ~10k lines are from package-lock.json (adding the eslint-plugin-react-hooks dependency). I've removed the unrelated DeepDependencies changes.

yarn.lock
yurishkuro, line 1: we do not use yarn

yurishkuro:
all commits must be signed

yarn.lock
taanvi2205: Sorry, I accidentally ran yarn instead of npm

taanvi2205:
@yurishkuro please review it now

packages/jaeger-ui/src/components/Monitor/ServicesView/index.tsx
```diff
@@ -232,17 +233,12 @@ export function MonitorATMServicesViewImpl(props: TProps) {
...
       fetchMetrics();
     }
-  }, [services]);
+  }, [services, fetchMetrics]);
 
   useEffect(() => {
     calcGraphXDomain();
   }, [selectedTimeFrame, calcGraphXDomain]);
 
-  useEffect(() => {
```
yurishkuro, line 241: why is this removed?

packages/jaeger-ui/src/components/SearchTracePage/index.tsx
```diff
@@ -109,9 +109,7 @@ export function SearchTracePageImpl(props: SearchTracePageImplProps) {
     if (needForDiffs.length) {
       fetchMultipleTraces(needForDiffs);
     }
-    // This may require "eslint-disable-next-line react-hooks/exhaustive-deps"
-    // in the future if we enable this linter.
-    // https://github.com/jaegertracing/jaeger-ui/issues/3445
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
yurishkuro, line 114: disabling linter must be done with a comment explaining why

yurishkuro (commented):
mostly LGTM, just two questions

packages/jaeger-ui/src/components/Monitor/ServicesView/index.tsx
taanvi2205: This was a Copilot's suggestion, it was causing duplicate fetchMetrics calls. The useEffect on line 232-236 already has fetchMetrics in its dependency array, so both effects were firing when fetchMetrics changed.

yurishkuro (approved):

PR #3474, opened 2026-01-27
[fix] Fix the dark mode for TraceDiff nodes

gulshank0:
## Description of the changes
- This PR improves nodes ( Text ) readability in dark mode, helping reduce eye strain and enhance overall UX.
Before :
Screenshot_28-Jan_02-16-45_21996

After:
Screenshot_28-Jan_02-17-16_3123

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/TraceDiff/TraceDiffGraph/renderNode.css
+++ b/packages/jaeger-ui/src/components/TraceDiff/TraceDiffGraph/renderNode.css
@@ -5,8 +5,8 @@ SPDX-License-Identifier: Apache-2.0
 
 .DiffNode {
   background: var(--surface-component-background);
-  border: 1px solid #777;
-  box-shadow: 0 0px 3px 1px rgba(0, 0, 0, 0.2);
+  border: var(--border-default);
+  box-shadow: var(--shadow-sm);
   cursor: pointer;
   white-space: nowrap;
 }


packages/jaeger-ui/src/components/TraceDiff/TraceDiffGraph/renderNode.css
```diff
@@ -9,15 +9,17 @@ SPDX-License-Identifier: Apache-2.0
   box-shadow: 0 0px 3px 1px rgba(0, 0, 0, 0.2);
   cursor: pointer;
   white-space: nowrap;
+  color: #333;
 }
 
 .DiffNode.is-changed {
-  color: var(--text-primary);
+  color: #333;
```
yurishkuro: this is a change in the opposite direction of what we want. We want to use design tokens, not hardcode colors.

yurishkuro: .

gulshank0: Hi @yurishkuro, I had doubt about the CSS. Noticed that in  few places, hardcoded colors are being used instead of design tokens. Just wanted to know if this is intentional for now or something that will be improved as things move forward. Thanks!

ghost: Yes there are few parts where hardcoded colors are still used. The plan is to fix the themes to use the existing design tokens and see how that settles. After that, we can look at the themes again and decide whether overrides/hardcoded colors are needed or not.

gulshank0: Got it, thanks for the clarification. I’ve used a hardcoded color for the text in this case please let me know if that’s okay, or if you’d prefer using a design token instead.

ghost: I think we can't use design tokens here to make text darker. But i think you can give a try of using design tokens for node backgrounds to improve contrast. Before committing any changes, please share screenshots here so we can confirm how it looks.

gulshank0:
here is the recording : 

https://github.com/user-attachments/assets/4852314c-1270-4f5b-829d-120406a7cc18

ghost:
What changes did you make in the code?

gulshank0:
The change adds  color: #333; (dark gray text color) to improve text readability in the dark mode only in the Graph Nodes  :)

ghost:
> I think we can't use design tokens here to make text darker. But i think you can give a try of using design tokens for node backgrounds to improve contrast. Before committing any changes, please share screenshots here so we can confirm how it looks.

.

ghost:
If we can't find any other way then i think it's better to use hardcoded colors here for now

gulshank0:
Oh, understood. Instead of using hardcoded text colors, we can make the background match dark mode. Something like this :

Screenshot_29-Jan_16-01-53_11085

gulshank0:
hi @Parship12 look at this , I think it is the desired solution : 


https://github.com/user-attachments/assets/ad67ee1f-8e70-4a0a-a5a8-c586df116370

packages/jaeger-ui/src/components/TraceDiff/TraceDiffGraph/renderNode.css
yurishkuro: @Parship12 why can't we use design tokens here to make text darker?

ghost: I have tried to use design tokens but they did not work here, because i think they are designed to work with the expected background color of the current theme. They assuming: Light theme = light backgrounds -> dark text & Dark theme = dark backgrounds -> light text
but the diff nodes use hardcoded light backgrounds that don't change with the theme, breaking this assumption.

yurishkuro: Nothing should be using hard coded colors for anything, that's the point. I don't see why graph is any different from other views.

ghost: yes, then we have to update it to remove all hardcoded colors.

ghost: can we use here: `var(--feedback-success)` for light green, `var(--feedback-error)` for red and more colors like this...?

yurishkuro: For the diffs that sounds reasonable. In trace graph we have a heat map color coding, which would be different.

ghost: > For the diffs that sounds reasonable. In trace graph we have a heat map color coding, which would be different.

cc: @gulshank0

gulshank0: the recent changes used ` --critical-path-color` it is black in color to avoid hardcoded colors

ghost:
@yurishkuro I have a question. What's the preferred approach for semantic token usage? Should tokens be used strictly according to their documented purpose, or is it acceptable to reuse them anywhere based on needs?
like `var(--feedback-success)` for green, `var(--feedback-error)` for red

yurishkuro:
tokens are used for their semantic meaning, not for their color. Please read the ADR in docs/ for theme approach.

ghost:
Yes, then using `--critical-path-color` for diff nodes is semantically incorrect. @gulshank0

gulshank0:
In the  semantic token  there is no declaration for black color Sementicaly for this type of uses other than this ?  is we have to declare there or use something like `--span-color-15`  for now ?

gulshank0:
hey changed the approach now Instead of using hardcoded text colors, we can make the background match dark mode with design tokens. here is the video for that : 


https://github.com/user-attachments/assets/d15ed3b8-1a62-4d18-87ee-4e819889098c

ghost:
LGTM 👍

packages/jaeger-ui/src/components/TraceDiff/TraceDiffGraph/renderNode.css
```diff
@@ -4,7 +4,7 @@ SPDX-License-Identifier: Apache-2.0
 */
 
 .DiffNode {
-  background: #bbb;
+  background: var(--surface-component-background);
   border: 1px solid #777;
```
yurishkuro: why is the border hardcoded? don't we have a token for it?

gulshank0: Its already there and due to minimal changes I leave as it is . and for it we have ` --border-default` .

gulshank0: should I update ?

packages/jaeger-ui/src/components/TraceDiff/TraceDiffGraph/renderNode.css
yurishkuro, line 20: what about these? or is this in the category of red/green nodes?

PR #3475, opened 2026-01-28
fix: guard missing labels and metricPoints in ops metrics reducer

Purav001:
- Fixes #7423

## Description of the changes
- Added defensive guards for missing `labels` and `metricPoints` in the operations metrics reducer.
- Skip malformed metric entries instead of throwing a runtime error.
- Added a regression unit test to ensure the reducer does not crash on incomplete metrics payloads.

- Added unit test in `metrics.test.js` for missing labels/metricPoints.
- Ran TypeScript type check: `npm run tsc-lint`
- Ran full test suite: `npm test` (2418 tests passed)

## Checklist


Purav001:
Thanks for the review! I’ve added DCO sign-off.  
Waiting for CI approval from maintainers.

ghost:
the issue added in the description(https://github.com/jaegertracing/jaeger/issues/7423), already fixed in #3460

packages/jaeger-ui/src/reducers/metrics.ts
```diff
@@ -186,12 +186,19 @@ function fetchOpsMetricsDone(
...
-                opsName = label.value;
-              }
-            });
-
+            if (Array.isArray(metricDetails.labels)) {
+              metricDetails.labels.forEach((label: { name: string; value: string }) => {
+                if (label.name === 'operation') {
+                  opsName = label.value;
+                }
+              });
```
yurishkuro, line 194: unnecessary looping, should use find()

```
const opsLabel = metricDetails.labels.find((label: { name: string; value: string }) => label.name === 'operation');
const opsName: string | null = opsLabel ? opsLabel.value : null;
```

PR #3483, opened 2026-01-31
Test(ui): Remove unused mocked history from TopNav tests

Popie52:
Removes an unused mocked history object from the TopNav test suite.
The existing tests already validate TopNav behavior using router context, so the mocked history was unnecessary and potentially confusing.

## Description of the changes
- Removes unused mocked history from TopNav tests.
- Cleans up related imports.
- Keeps all existing router-based tests intact (BrowserRouter + CompatRouter).
- No production code changes.
- No behavior changes.

- Ran the Jaeger-UI unit test suite locally.
- Verified the new TopNav test passes successfully:
```bash
        npm install --ignore-scripts
        npm test -- TopNav
```

## Checklist
- [*] I have read https://github.com/jaegertracing/jaeger/blob/master/CONTRIBUTING_GUIDELINES.md
- [*] I have signed all commits
- [*] I have added unit tests for the new functionality
- [*] I have run lint and test steps successfully

### DIFF
--- a/packages/jaeger-ui/.gitignore
+++ b/packages/jaeger-ui/.gitignore
@@ -7,9 +7,3 @@ stats.html
 # Local UI config files for development (see README.md for details)
 jaeger-ui.config.js
 jaeger-ui.config.json
-
-# npm / windows garbage
-AppData/
-npm-cache
-**/_cacache
-node_modules


yurishkuro:
I am not sure I follow what this gives us. cc @Parship12

Popie52:
@yurishkuro
“I am not sure I follow what this gives us.”

Thanks for the question—happy to clarify.

The primary value of this test is to lock in a migration-safe contract for TopNav behavior as routing internals change. Today, some tests rely on manually mocked history, which works but obscures whether navigation is actually derived from router context.

This test intentionally avoids mocked history and instead asserts that TopNav derives its navigation state from the router itself (via CompatRouter/MemoryRouter). The goal is not to test new behavior but to document and protect existing behavior so that future router upgrades (e.g., removing legacy history patterns during the v7 migration) don’t silently change how navigation links are resolved.

Without a test like this, routing refactors can appear safe at compile time but subtly alter link generation or navigation assumptions. This regression test provides an early signal if a migration breaks that contract.

If this feels redundant or misaligned with the migration goals, I’m happy to adjust or drop it—feedback is welcome.

packages/jaeger-ui/src/components/App/TopNav.test.js
ghost: > Ensures navigation state is derived from the router context (CompatRouter) rather than mocked history.

all existing tests already do this. The mocked history in `jest.mock('../../utils/configure-store')` is never used.

ghost:
the existing tests already validates that `TopNav` is workingg correctly with router context. I think it's unecessary to add another test that does the same thing with slightly different router setup.

ghost:
Thanks for the contribution!
if you want to you can use this PR to remove the unused mocked history since it's not needed and update the title, commit msg and description accordingly.

Popie52:
Thanks for the suggestion! I’ve updated the PR to remove the unused mocked history from the TopNav tests, cleaned up the imports, and updated the title, commit message, and description accordingly. All tests continue to pass.

packages/jaeger-ui/src/components/App/TopNav.test.js
```diff
@@ -46,31 +38,6 @@ jest.mock('antd', () => {
   return { ...actual, Menu, Dropdown };
 });
 
-jest.mock('../../utils/config/get-config', () => {
```
ghost, line 49: why was this part removed?

Popie52: my bad! Restored the get-config mock and scoped the PR to just removing the unused history.

packages/jaeger-ui/.gitignore
```diff
@@ -7,3 +7,9 @@ stats.html
 # Local UI config files for development (see README.md for details)
 jaeger-ui.config.js
 jaeger-ui.config.json
+
+# npm / windows garbage
```
ghost: why these changes included in this PR?

Popie52: They were included automatically in the latest commit while resolving linting issues.

Popie52:
Removed unrelated .gitignore changes.

ghost:
lgtm

yurishkuro:
> `[*] I have read`

@Popie52 Tip: use `x` not `*`, then it properly renders `[]` as checkboxes

PR #3490, opened 2026-02-02
refactor(TraceGraph): migrate from class to functional component

### LINKED ISSUE #3369: [UI Refactor] Migrate TraceGraph to Functional Component
This issue is part of the larger effort to modernize the Jaeger UI codebase (Parent Issue #2610). The goal is to refactor the `TraceGraph` component from a React Class Component to a React Functional Component using Hooks.

**Target Component:** `packages/jaeger-ui/src/components/TracePage/TraceGraph/TraceGraph.tsx`

### Acceptance Criteria

| --- | --- | --- |

### Resources

taanvi2205:
Convert TraceGraph from React.PureComponent to a functional component using React hooks as part of the UI modernization effort (#3369).

Resolves #3369

## Description of the changes
- Replace class component with function wrapped in React.memo
- Convert this.state to useState hooks (showHelp, mode)
- Replace componentWillUnmount with useEffect cleanup
- Use useMemo for LayoutManager to maintain instance across renders
- Update tests to verify behavior via DOM assertions instead of setState spies

- All 10 existing unit tests pass
- Build completes successfully with no type errors
- UI working with no changes

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceGraph/TraceGraph.test.js
+++ b/packages/jaeger-ui/src/components/TracePage/TraceGraph/TraceGraph.test.js
@@ -90,8 +90,7 @@ describe('<TraceGraph>', () => {
     expect(screen.getByText('No trace found')).toBeInTheDocument();
   });
 
-  it('switches node mode when clicking mode buttons - with state verification', async () => {
-    const setStateSpy = jest.spyOn(TraceGraph.prototype, 'setState');
+  it('switches node mode when clicking mode buttons', async () => {
     render(<TraceGraph {...props} />);
 
     // Initial mode should be service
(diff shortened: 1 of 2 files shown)


emphor11:
@taanvi2205 

This is a really solid refactor — thanks for the careful work here 

I’ve gone through the component and the updated tests in detail. The migration from React.PureComponent to a functional component is clean and preserves behavior well. State and lifecycle logic have been mapped correctly to hooks, LayoutManager handling via useMemo + cleanup is spot-on, and wrapping with React.memo maintains the original performance characteristics.

The test updates are especially well done — moving away from class internals to DOM-based assertions and user interactions makes the tests more robust and future-proof. Coverage looks good and all checks passing gives high confidence in this change.

Just one thing to tell that the issue checklist mentions before/after screenshots for visual verification. It might be nice to add those to the PR description for completeness, but this doesn’t block approval from my side.

Overall, great refactor — this fits very well with the UI modernization effort.

yurishkuro:
I believe this is done already

Issue #3466, opened 2026-01-25
[Bug] : Low Color Contrast in Search Sidebar and UI Refactor


Harshul23:
### What happened?

While investigating the Jaeger UI for modernization, I identified a critical accessibility (WCAG 2.1) violation in the Search Sidebar. When a time-range option is selected, the contrast ratio between the text and the background highlight is insufficient, rendering the text nearly invisible in Dark Mode.

Related issues
#3322 
#3457

### Steps to reproduce

1. Open Jaeger UI in Dark Mode.
2. Navigate to the Search tab in the left sidebar.
3. Click on the time-range dropdown (e.g., "Last Hour").
4. Observe the "Selected" state of the list item.

### Expected behavior

The text should maintain a contrast ratio of at least 4.5:1 against its background to ensure readability.

### Relevant log output

```shell

```

### Screenshot

<img width="1233" height="630" alt="Image" src="https://github.com/user-attachments/assets/8b215f5f-c759-4f30-a4a2-f743db97d34d" />

<br />
<br />

<img width="1226" height="761" alt="Image" src="https://github.com/user-attachments/assets/bdd837a2-fbbb-4c5b-9dc8-c579653ba8d5" />

### Operating system

macOS Tahoe 26.2

### Deployment configs

```shell

```

samar-703:
hey @Harshul23 , unable to reproduce on main, selected option has good contrast in my setup. 

could you please specify which browser/commit are you testing?

<img width="1919" height="967" alt="Image" src="https://github.com/user-attachments/assets/aa2a98ef-a67a-4596-bef2-d2ee6c6a12bf" />

greedy-wudpeckr:
hey @Harshul23 , i understood this issue will make a PR shortly.

aryunewaskar77-art:
I’ve opened a PR #3491 to fix this by improving Dark Mode contrast for the selected time-range option.

yurishkuro:
I don't see any contrast issues in the latest main
<img width="482" height="392" alt="image" src="https://github.com/user-attachments/assets/c950e4dd-38bb-4e14-83fb-086162743d73" />

PR #3491, opened 2026-02-02
fix(UI): improve contrast for selected time-range in Dark Mode

aryunewaskar77-art:
Issue #3466

Fixes a WCAG 2.1 accessibility issue in the Jaeger UI Search Sidebar.
Improves contrast for the selected time-range option in Dark Mode to ensure readable text and meet the 4.5:1 contrast requirement.

Testing

Ran Jaeger UI locally (yarn start)

Verified selected time-range visibility in Dark Mode

Confirmed no Light Mode regression

### DIFF
--- a/packages/jaeger-ui/src/components/App/TopNav.css
+++ b/packages/jaeger-ui/src/components/App/TopNav.css
@@ -5,7 +5,7 @@ SPDX-License-Identifier: Apache-2.0
 
 .Dropdown--icon-container {
   display: flex;
-  color: var(--text-inverse);
+  color: #fff;
   margin-right: 30px;
   flex-direction: row;
   align-items: center;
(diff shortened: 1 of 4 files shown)


packages/jaeger-ui/src/components/App/TopNav.css
```diff
@@ -5,28 +5,28 @@ SPDX-License-Identifier: Apache-2.0
 
 .Dropdown--icon-container {
   display: flex;
-  color: var(--text-inverse);
+  color: #fff;
```
ghost, line 8: why were the tokens removed and hardcoded colors added? see: [vars.css](https://github.com/jaegertracing/jaeger-ui/blob/main/packages/jaeger-ui/src/components/common/vars.css) & [ADR 0001](https://github.com/jaegertracing/jaeger-ui/blob/main/docs/adr/0001-design-token-based-theming.md)

aryunewaskar77-art: Thanks for the review.

I avoided --text-inverse here intentionally since it’s a flipping token, while the TopNav is a static dark surface. Using it caused contrast issues in this context.

yurishkuro:
I don't see any contrast issues in the latest main
image

PR #3495, opened 2026-02-03
#3494 ref: component gets location from Redux state

### LINKED ISSUE #3494: [Bug]:  Page.tsx still uses withRouteProps
### What happened?

Remove unnecessary withRouteProps from Page.tsx - component gets location from Redux state
- follwoing [#2531, #2532 , #3313 ] 

### Steps to reproduce

1- update the single line

### Expected behavior

same as before 

### Relevant log output

```shell

```

### Screenshot

_No response_

### Additional context

_No response_

### Jaeger backend version

_No response_

### SDK

_No response_

### Pipeline

_No response_

### Stogage backend

_No response_

### Operating system

_No response_

### Deployment model

_No response_

### Deployment configs

```shell

```

Amr-Shams:
- Resolves #3494 
## Description of the changes
- remove unnecessary withRouteProps 

- rebuild the app, and check it works fine

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/DeepDependencies/index.tsx
+++ b/packages/jaeger-ui/src/components/DeepDependencies/index.tsx
@@ -42,7 +42,6 @@ import { localeStringComparator } from '../../utils/sort';
 
 import './index.css';
 import { ApiError } from '../../types/api-error';
-import withRouteProps from '../../utils/withRouteProps';
 import { useServices, useSpanNames } from '../../hooks/useTraceDiscovery';
 
 interface IDoneState {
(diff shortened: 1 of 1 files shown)


emphor11:
This looks good to me and fits well with the incremental routing cleanup approach.

In #3496, I intentionally scoped the TopNav change to avoid touching Page.tsx as part of the same migration, so it’s nice to see this follow-up handled separately and conservatively. Using Redux-provided location here makes sense for a top-level layout component and keeps the refactor low-risk.

One small process note: the checklist mentions added unit tests, but I didn’t notice test changes in the diff — if this was manually verified only, it might be worth updating that for accuracy.

Otherwise, the change looks solid

Amr-Shams:
> This looks good to me and fits well with the incremental routing cleanup approach.
> 
> In #3496, I intentionally scoped the TopNav change to avoid touching Page.tsx as part of the same migration, so it’s nice to see this follow-up handled separately and conservatively. Using Redux-provided location here makes sense for a top-level layout component and keeps the refactor low-risk.
> 
> One small process note: the checklist mentions added unit tests, but I didn’t notice test changes in the diff — if this was manually verified only, it might be worth updating that for accuracy.
> 
> Otherwise, the change looks solid

thank you for the feedback, yep u are right, i have changed the pr body

packages/jaeger-ui/src/components/App/Page.tsx
```diff
@@ -54,4 +54,4 @@ export function mapStateToProps(state: ReduxState) {
   return { embedded, pathname, search };
 }
 
-export default connect(mapStateToProps)(withRouteProps(PageImpl));
+export default connect(mapStateToProps)(PageImpl);
```
yurishkuro, line 57: why is this safe to do?

Amr-Shams: The component doesn't use history, params, or location object directly , only the two string fields that Redux already supplies.

PR #3503, opened 2026-02-05
refactor(ui): replace legacy history usage in TracePage

aryunewaskar77-art:
This PR addresses part of the routing modernization effort by removing reliance on a stale routing dependency.

Specifically, it updates TracePage/index.tsx to stop using the legacy history object (introduced via withRouteProps) and instead rely on maintained React Router navigation APIs. This aligns with the guidance in #2531 to gradually move away from deprecated dependencies and modernize routing one file at a time.

What changed:
1. Removed legacy routing dependency usage in TracePage
2. No behavior or UI changes

Why this change:
The goal is to avoid stale or deprecated dependencies in the codebase and rely on actively maintained APIs compatible with modern React.

How this was tested:
1. Ran the app locally
2. Verified TracePage loads correctly
3. Confirmed URL updates and browser navigation behave the same as before

Related issue

Part of #2531


emphor11:
Hi @aryunewaskar77-art 

I noticed one thing that in TracePage/index.tsx, navigation has been migrated to useNavigate, which aligns well with the stated goal of moving away from legacy routing dependencies.
But however, that the component is still wrapped with withRouteProps, and route data (params, location) is still sourced from that HOC (e.g. mapStateToProps relying on ownProps.params).

aryunewaskar77-art:
Thanks for pointing that out. You're right — I migrated the navigation to useNavigate, but the component is still wrapped with withRouteProps and using route data from it.
I was taking it step by step, but I can refactor this further to remove the HOC and switch to the v6 hooks if that’s preferred.

PR #3512, opened 2026-02-06
Add environment checks to prepare-release target

jkowall:
This PR adds a  target to the  that validates the Node.js version and ensures  exists before running . This addresses issues where running the release preparation in an unconfigured environment causes problems.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,8 +10,12 @@ draft-release:
 	chmod 755 ./scripts/draft-release.py
 	./scripts/draft-release.py --title "Jaeger UI" --repo jaeger-ui
 
+.PHONY: check-env
+check-env:
+	nvm use
+
 .PHONY: prepare-release
-prepare-release:
+prepare-release: check-env
 	@test $(VERSION) || (echo "VERSION is not set. Use 'make prepare-release VERSION=vX.Y.Z'"; exit 1)
 	python3 scripts/prepare-release.py --version $(VERSION)
 


Makefile
```diff
@@ -10,8 +10,24 @@ draft-release:
...
+.PHONY: check-env
+check-env:
+	@NODE_VERSION=$$(node -v | cut -d 'v' -f 2); \
+	NVMRC_VERSION=$$(cat .nvmrc | tr -d '[:space:]'); \
+	if [ "$$NODE_VERSION" != "$$NVMRC_VERSION" ] && [ "$${NODE_VERSION%%.*}" != "$$NVMRC_VERSION" ]; then \
+		echo "Error: Node.js version mismatch."; \
+		echo "Current: v$$NODE_VERSION, Required (from .nvmrc): v$$NVMRC_VERSION"; \
+		echo "Please run 'nvm use'"; \
+		exit 1; \
+	fi
```
yurishkuro: instead of checking will be simpler to just run `nvm use`, it's quick if already on the good version.

yurishkuro:
```suggestion
	nvm use
```

jkowall:
Addressed your feedback — simplified `check-env` to just run `nvm use` instead of the manual version comparison. Also rebased on latest main and dropped the unrelated version bump/CHANGELOG commits so this PR is now Makefile-only.

Makefile
jkowall: Updated `check-env` to use `nvm use`, and the PR now only carries the `Makefile` change. I also validated the branch locally after refreshing dependencies.

PR #3524, opened 2026-02-08
fix: make spanKind optional in v3 operations schema

### LINKED ISSUE #3521: [Bug]: Operations fail to load in 2.15.0 due to strict Zod validation of spanKind
### Summary

Upgrading to Jaeger 2.15.0 caused the operations dropdown to stay empty for some users. This is caused by strict Zod validation in the UI's v3 API client.

### Details

The `OperationsResponseSchema` in `jaeger-ui/src/api/v3/generated-client.ts` requires a `spanKind` field for each operation. However, the backend may omit this field in some configurations (e.g. when using Elasticsearch), causing the entire JSON response to fail validation and results to be discarded.

### Root Cause

Strict validation of `OperationSchema` requires `spanKind: z.string()`.

### Proposed Fix

Make `spanKind` optional: `spanKind: z.string().optional()`.

### Link to related issue
<https://github.com/jaegertracing/jaeger/issues/7989#issuecomment-3863451243>

Sapthagiri777:

Resolves #3521
Relates to jaegertracing/jaeger#7989

The Jaeger UI v2.15.0 fails to load operations for a selected service. The `/api/v3/operations` endpoint returns operation objects without `spanKind` in certain storage configurations (e.g. Elasticsearch), but the UI's Zod schema strictly requires `spanKind` as a string. This causes validation to reject the entire response, and the operations dropdown never loads.

## Description of the changes

- Makes `spanKind` optional with a safe default in the Zod schema (`z.string().optional().default('')`) so the inferred TypeScript type remains `string` and no downstream code changes are needed
- Adds a regression test verifying `fetchSpanNames()` works correctly when `spanKind` is missing from the backend response

**Note:** `generated-client.ts` is auto-generated via `npm run generate:api-types`. This change should ideally be preserved in the generation pipeline or upstream OpenAPI spec across regenerations.

- Added a new unit test `handles operations without spanKind from backend` that mocks a backend response without `spanKind` and verifies the schema defaults it to empty string
- All 19 existing tests continue to pass
- Ran `npx jest --testPathPatterns="api/v3/client.test"` — all passed

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/api/v3/client.test.ts
+++ b/packages/jaeger-ui/src/api/v3/client.test.ts
@@ -172,6 +172,28 @@ describe('JaegerClient', () => {
         'Failed to fetch span names for service "my-service": 500 Internal Server Error'
       );
     });
+
+    it('handles operations without spanKind from backend', async () => {
+      // Regression test: v2.15.0 backends may omit spanKind (jaeger#7989)
+      const mockOperations = [
+        { name: 'GET /api/users' },
+        { name: 'POST /api/orders' },
+      ];
+      mockFetch.mockResolvedValue({
+        ok: true,
+        json: async () => ({ operations: mockOperations }),
+      });
+
+      const promise = client.fetchSpanNames('test-service');
+      jest.runAllTimers();
+      const result = await promise;
+
+      expect(result).toEqual([
+        { name: 'GET /api/users', spanKind: '' },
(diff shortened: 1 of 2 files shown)


Sapthagiri777:
Closing this — per @yurishkuro's feedback, the fix should be on the backend side to always populate spanKind in the query service response, not by relaxing the UI validation.

PR #3544, opened 2026-02-13
feat(ui): fetch storage capabilities from backend during Vite dev mode

### LINKED ISSUE #3541: [chore]: Improve dev mode by loading backend settings dynamically
# Fetch storage capabilities from backend during Vite dev mode

## Problem

During development (`npm start`), the UI is served by Vite, not the Jaeger backend. The backend injects config and capabilities into `index.html` via search-replace at serve time, but Vite serves its own copy of `index.html` with hardcoded defaults:

```javascript
// index.html — always returns defaults during npm start
function getJaegerStorageCapabilities() {
  const DEFAULT_STORAGE_CAPABILITIES = { "archiveStorage": false };
  const JAEGER_STORAGE_CAPABILITIES = DEFAULT_STORAGE_CAPABILITIES;
  return JAEGER_STORAGE_CAPABILITIES;
}
```

This means `archiveStorage` is always `false` during development, and the proposed `metricsStorage` (#3540) would also be `false`, hiding the Monitor tab even when the local backend has metrics storage configured.

The existing `jaeger-ui.config.js` override mechanism (handled by the Vite plugin) only covers `getJaegerUiConfig()` — it does not cover capabilities or version.

## Proposed Solution

Extend the existing `jaegerUiConfigPlugin` in `vite.config.mts` to dynamically fetch capabilities from the running backend during HTML transformation.

### Prerequisites: Backend API endpoint

Add an endpoint to the Jaeger query service that returns the same data it currently injects into HTML:

```
GET /api/ui/config
```

Response:
```json
{
  "uiConfig": { ... },
  "storageCapabilities": { "archiveStorage": true, "metricsStorage": true },
  "version": { "gitCommit": "abc123", "gitVersion": "1.60.0", "buildDate": "2025-01-15" }
}
```

This endpoint returns the exact same values the server would inject via search-replace into `index.html`. It requires no authentication (the existing UI config is already public — it's embedded in the HTML served to browsers).

#### 1. Extend the Vite plugin (`packages/jaeger-ui/vite.config.mts`)

The `jaegerUiConfigPlugin` already has a `transformIndexHtml` hook that runs in Node.js on every page load during dev mode. Extend it to fetch capabilities from the backend:

```typescript
function jaegerUiConfigPlugin() {
  const jsConfigPath = path.resolve(__dirname, 'jaeger-ui.config.js');
  const jsonConfigPath = path.resolve(__dirname, 'jaeger-ui.config.json');

  // Cache the backend config to avoid fetching on every request.
  // Invalidated on config file changes (existing watcher) or periodically.
  let cachedBackendConfig: any = null;
  let cacheTimestamp = 0;
  const CACHE_TTL_MS = 30_000; // re-fetch every 30s

  async function fetchBackendConfig() {
    const now = Date.now();
    if (cachedBackendConfig && now - cacheTimestamp < CACHE_TTL_MS) {
      return cachedBackendConfig;
    }
    try {
      const response = await fetch('http://localhost:16686/api/ui/config');
      if (response.ok) {
        cachedBackendConfig = await response.json();
        cacheTimestamp = now;
        console.log('[jaeger-ui-config] Fetched config from backend');
      }
    } catch {
      // Backend not running — use defaults silently
    }
    return cachedBackendConfig;
  }

  return {
    name: 'jaeger-ui-config',
    configureServer(server) { /* ... existing watcher code ... */ },
    transformIndexHtml: {
      order: 'pre' as const,
      async handler(html: string) {
        // ... existing JS/JSON config injection logic (unchanged) ...

        // Fetch capabilities and version from the running backend
        const backendConfig = await fetchBackendConfig();
        if (backendConfig?.storageCapabilities) {
          html = html.replace(
            'const JAEGER_STORAGE_CAPABILITIES = DEFAULT_STORAGE_CAPABILITIES;',
            `const JAEGER_STORAGE_CAPABILITIES = ${JSON.stringify(backendConfig.storageCapabilities)};`
          );
        }
        if (backendConfig?.version) {
          html = html.replace(
            'const JAEGER_VERSION = DEFAULT_VERSION;',
            `const JAEGER_VERSION = ${JSON.stringify(backendConfig.version)};`
          );
        }
        // Optionally also inject uiConfig from the backend if no local
        // jaeger-ui.config.js/json file exists, so dev mode fully mirrors production.

        return html;
      },
    },
  };
}
```

Key design points:
- **Async fetch in `transformIndexHtml`** — this hook supports returning a Promise, so the async fetch works naturally.
- **Caching with TTL** — avoids fetching on every page load; 30s TTL keeps capabilities reasonably fresh while avoiding unnecessary requests.
- **Silent fallback** — if the backend is not running, the fetch fails silently and defaults remain. This allows working on the UI without a running backend.
- **Same search-replace pattern** — uses the same placeholder strings the Go server uses, so the injection is consistent.

#### 2. No changes needed to `get-config.ts`

The browser-side code (`getCapabilities()`, `getConfig()`) remains unchanged. It reads `window.getJaegerStorageCapabilities()` as before — the only difference is that during dev mode, the Vite plugin now injects real values instead of defaults.

#### 3. The `index.html` placeholders remain unchanged

The existing placeholder pattern in `index.html` continues to serve as both the production injection target (for the Go server) and the dev injection target (for the Vite plugin). No changes needed.

## Benefits

- **Dev mode matches production** — capabilities reflect the actual backend configuration.
- **No manual config files needed** — developers don't need to create `jaeger-ui.config.js` to enable features like Monitor or Archive.
- **Backward compatible** — if the backend doesn't have the `/api/ui/config` endpoint (older version), the fetch fails silently and defaults are used.
- **Extends naturally** — any new capability or config the server reports is automatically available in dev mode.

## Verification

1. Start the Jaeger backend with metrics storage configured on `localhost:16686`.
2. Run `npm start` in the UI.
3. Open the browser — the Monitor tab should appear if the backend reports `metricsStorage: true`.
4. Stop the backend, refresh — the Monitor tab should disappear (defaults applied).
5. Restart the backend, wait 30s (cache TTL), refresh — the Monitor tab should reappear.

aryunewaskar77-art:
Fixes  #3541

This PR improves the Jaeger UI development experience by making the Vite dev server automatically fetch configuration from the running Jaeger backend. Previously, developers had to manually tweak local config files to test features like the Monitor tab or Archive storage. With this update, the UI now reflects the backend’s actual capabilities by default.

Key Changes
Automatic Backend Sync: Fetches storageCapabilities (archive/metrics), version, and uiConfig from http://localhost:16686/api/ui/config.

Smart Config Merging:
Highest priority: jaeger-ui.config.js — full manual override for advanced dev scenarios.
Next: jaeger-ui.config.json — shallow merge on top of backend defaults for easy tweaks.
Fallback: backend defaults automatically used if no local config exists.
Resilient Fallback: Added a 1-second fetch timeout and graceful failover. If the backend isn’t running, the UI still works with default settings.
Better Developer Feedback: Clear console logs show which config source is currently active, making debugging easier.

Why This Is Needed
Before this change, dev mode could be misleading: feature flags and tabs like Monitor wouldn’t appear unless a developer manually created a JSON config. Now, the UI behaves almost exactly like production out-of-the-box, reducing friction and guesswork during development.

Testing Done
✅ Backend running: window.getJaegerStorageCapabilities() correctly reflects backend settings (e.g., archiveStorage: true).
✅ Backend down: Dev server starts without errors and defaults are used.
✅ Local JSON overrides: Confirmed jaeger-ui.config.json merges correctly with backend values.
✅ UI behavior: Monitor tab and Archive buttons appear/disappear according to backend state.


emphor11:
Hi @aryunewaskar77-art 

I’ve gone through the changes in this PR and the updated files. One thing I wanted to know that in packages/jaeger-ui/vite.config.mts, the plugin is always enabled, and transformIndexHtml also runs during vite build.
Since the PR description mentions this being a dev-mode improvement, does this mean the backend fetch could also run during production builds?

aryunewaskar77-art:
It does not run during the production build. While the plugin logic itself is present in this configuration file, it is designed to be triggered only when a developer run the development server (npm start / vite).

PR #3546, opened 2026-02-17
Fix (plexus): Self-loop edge label positioning [SvgEdge]

### LINKED ISSUE #3500: [Bug]: [UI] System Architecture Graph: Edge label text is obscured/clipped on self-referencing nodes
### What happened?

I encountered a visual bug in the **System Architecture** tab (DAG view) where the request count label for a self-referencing edge is partially obscured by the service node icon.

Specifically, the edge loop for the `frontend` service has a request count of **228**, but it renders visually as **28** (or similar), with the leading digit clipped or hidden behind the node circle.

### Steps to reproduce

1. Open the **System Architecture** tab in the Jaeger UI.
2. Locate a service node that has a self-referencing dependency (a loop where the service calls itself).
3. Observe the request count label on the loop edge.
4. Notice that the text label overlaps with the node icon, making the number difficult or impossible to read correctly.

### Expected behavior

The edge label text should be positioned clearly outside the node's boundary or have a higher z-index/background to ensure it is fully legible. It should not clip behind the service icon.

### Relevant log output

```shell

```

### Screenshot

I inspected the element in the browser DevTools to confirm the data is present but visually hidden.
- **Element:** `<text>` inside the `plexus-Digraph--SvgEdgesLayer`.
- **Value:** `228`
- **Observation:** The `x` and `y` coordinates for the text seem to place it too close to the node center for this specific graph layout.

Image

Image

### Additional context

**Actual Behavior**
The text label overlaps with the node. In my specific case:
- **Actual Value:** `228` (verified via DOM inspector).
- **Rendered Value:** Looks like `28` because the first `2` is obscured.

### Jaeger backend version

_No response_

### SDK

_No response_

### Pipeline

_No response_

### Stogage backend

_No response_

### Operating system

Linux

### Deployment model

Docker

### Deployment configs

```shell

```

riteshdavv:
Fixes #3500

## Problem

In the Trace Graph / System Architecture DAG views, edges rendered by the plexus [SvgEdge](file:///e:/Desktop/jaeger-ui/packages/plexus/src/Digraph/SvgEdge.tsx#67-107) component display labels (e.g., span counts). For **self-loop edges** (a service calling itself), the label was positioned at the midpoint of the path's start and end points — which for a self-loop lands directly on top of the node circle, making the label invisible.

## Changes

**File:** [packages/plexus/src/Digraph/SvgEdge.tsx](https://github.com/jaegertracing/jaeger-ui/blob/main/packages/plexus/src/Digraph/SvgEdge.tsx)

### 1. Self-loop detection via graph data

Self-loop edges are detected using `layoutEdge.edge.from === layoutEdge.edge.to`, which reliably identifies when an edge connects a node to itself. Coordinate-based detection was considered but rejected because self-loop bezier paths start and end at different points on the node's perimeter, making positional comparison unreliable.

### 2. Centroid-based label positioning for self-loops

For self-loop edges, the label is positioned at the **centroid of all bezier control points**, which places it at the visual center of the oval loop curve:

```tsx
if (isSelfLoop && pathPoints.length > 2) {
  let sumX = 0;
  let sumY = 0;
  for (const [px, py] of pathPoints) {
    sumX += px;
    sumY += py;
  }
  return {
    labelX: sumX / pathPoints.length - xOffset / 2,
    labelY: sumY / pathPoints.length,
  };
}
```

This approach is O(n) where n = number of path points (typically 4 for a cubic bezier), so there is no performance concern.

### 3. Normal edge positioning unchanged

For non-self-loop edges, the existing midpoint-based positioning between start and end points is preserved.

## Before / After

| Scenario | Before | After |
|----------|--------|-------|
| Self-loop label | Hidden behind node circle | Centered inside the oval loop |
| Normal edge label | Midpoint between nodes | Unchanged |

DDG

## Testing

Verified manually using HotROD demo traces in the Jaeger UI:
- **Trace Graph** view (individual trace → dropdown → "Trace Graph")
- **System Architecture** view (top navigation)

Self-loop edges (e.g., `frontend → frontend`) now display their span count labels clearly inside the loop curve.

### DIFF
--- a/packages/plexus/src/Digraph/SvgEdge.tsx
+++ b/packages/plexus/src/Digraph/SvgEdge.tsx
@@ -37,11 +37,27 @@ function makePathD(points: [number, number][]) {
   return dArr.join(' ');
 }
 
-function computeLabelCoord(pathPoints: [number, number][], label?: string | undefined) {
+function computeLabelCoord(pathPoints: [number, number][], isSelfLoop: boolean, label?: string | undefined) {
+  const xOffset = (label?.length ?? 0) * 5;
+
+  if (isSelfLoop && pathPoints.length > 2) {
+    // For self-loops, position label at the center of the loop curve
+    // Average all path/control points to find the centroid of the oval
+    let sumX = 0;
+    let sumY = 0;
+    for (const [px, py] of pathPoints) {
+      sumX += px;
+      sumY += py;
+    }
+    return {
+      labelX: sumX / pathPoints.length - xOffset / 2, // center the text
+      labelY: sumY / pathPoints.length,
+    };
+  }
+
+  // Normal edge: midpoint between start and end
   const [startX, startY] = pathPoints[0];
   const [endX, endY] = pathPoints[pathPoints.length - 1];
-
-  const xOffset = (label?.length ?? 0) * 5;
   const labelX = (startX + endX) / 2 - xOffset;
   const labelY = (startY + endY) / 2;
 
@@ -65,7 +81,8 @@ export default class SvgEdge<U = {}> extends React.PureComponent<TProps<U>> {
       getProps(setOnEdge, layoutEdge, renderUtils)
     );
 
-    const { labelX, labelY } = computeLabelCoord(pathPoints, label);
+    const isSelfLoop = layoutEdge.edge.from === layoutEdge.edge.to;
+    const { labelX, labelY } = computeLabelCoord(pathPoints, isSelfLoop, label);
 
     return (
       <g>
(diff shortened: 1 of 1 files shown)


ghost (commented):
Please sign your commits
and what about the copilot's suggestion?

riteshdavv:
@Parship12 Done! I've signed the commit and also applied the Copilot suggestion. Thanks for the review!

ghost (commented):
What about the copilot suggestions?
Also, can you please resolve the merge conflicts?

Issue #3540, opened 2026-02-11
[refactor]: Simplify ATM configuration


yurishkuro:
### Problem

Three overlapping mechanisms currently control Monitor/SPM visibility in the UI:

1. **`monitor.menuEnabled`** — a static config flag (defaults `true`) that controls whether the Monitor tab appears in the nav. Redundant because the server already knows at startup whether metrics storage is configured.
2. **`isATMActivated`** — a runtime probe that detects HTTP 501 responses from the metrics API and permanently disables the Monitor page for the session. Fragile because 501 can also indicate transient storage failures, and there is no retry logic.
3. **`StorageCapabilities`** — server-reported capabilities injected into the UI at serve time (currently only reports `archiveStorage`). This is the right mechanism but doesn't yet cover metrics storage.

These three mechanisms are redundant and sometimes contradictory. A cleaner design uses a single source of truth: the server reports whether metrics storage is available via `StorageCapabilities`, and the UI trusts that.

### Target Design

- **Extend `StorageCapabilities`** with `metricsStorage: boolean` — the server sets this based on its configuration.
- **Derive Monitor tab visibility** from `storageCapabilities.metricsStorage` instead of `monitor.menuEnabled`.
- **Remove `isATMActivated`** from the UI state — no more runtime 501 probing to detect feature availability.
- **Treat HTTP 501 as a normal API error** — existing per-metric error fields already capture it; display errors rather than hiding the feature.
- **Keep the "No data yet!" alert** in ServicesView for configured-but-empty scenarios.

ghost:
@yurishkuro Should I take a look at this issue?

yurishkuro:
@Parship12 help with code reviews would be much more valuable to the project, we have a huge backlog.

ghost:
Okay, sure.

PR #3547, opened 2026-02-17
[refactor]: Simplified ATM configuration 

gulshank0:
 Resolves: #3540

### Description of the changes
Replaced three overlapping mechanisms that control Monitor/SPM tab visibility (monitor.menuEnabled, isATMActivated, and the absence of metricsStorage in StorageCapabilities) with a single source of truth: StorageCapabilities.metricsStorage.

- Before this change, three different things decided whether the Monitor tab showed up:

1). A config flag (monitor.menuEnabled) — on by default, even if no metrics backend existed.
2) A runtime check (isATMActivated) — watched for HTTP 501 errors and permanently hid the page for the session, with no way to retry.
3)  StorageCapabilities — the server already told the UI about archive storage, but said nothing about metrics storage.

- These three didn't always agree, making the behavior confusing and fragile.

- Now there's just one: the server tells the UI whether metrics storage is available through StorageCapabilities.metricsStorage, and the UI uses that single value to show or hide the Monitor tab. Simple.

### What was removed:

- isATMActivated from Redux state and the 501-sniffing logic in the reducer. A 501 error is now handled like any other API error — it shows up as an error message instead of silently killing the feature.
monitor.menuEnabled config flag — no longer needed.
- The MonitorATMEmptyState component and its config, styles, test, and image — hiding the tab is cleaner than showing a "not configured" page.
### What was kept:

- The "No data yet!" alert still shows up when metrics storage is configured but empty, with a link to the setup docs.
- Safe default: metricsStorage defaults to false, so the Monitor tab stays hidden until the server says it's ready. Better to hide it than show something broken.

- npm run build
- npm test
Screenshot_18-Feb_02-12-36_11674

## Checklist


packages/jaeger-ui/src/constants/default-config.ts
```diff
@@ -77,22 +78,6 @@ const defaultConfig: Config = {
   },
   linkPatterns: [],
   monitor: {
-    menuEnabled: true,
```
yurishkuro, line 80: why is this removed? The server capability may indicate if SPM is supported, but the rest of the config is UI-related

gulshank0: as it is mentioned to remove in Step-3. let me check

gulshank0: as it is mentioned to remove in Step-3 in the issue description. it is used here in the UI to control both the visibility of the feature and what users see when no monitoring data is available yet, and to show the official Jaeger SPM documentation. And removing this part leads not to display the "monitor" endpoint  in the Navbar.

gulshank0:
Hi @yurishkuro, please let me know if any change is required, I will update accordingly. Thanks.

yurishkuro:
I already said which changes are required - you are completely destroying the `monitor` configuration, which was not what the ticket was about.

yurishkuro:
> `Resolves #3540`

how does this resolve the issue if step 1 (server side) is not implemented?

yurishkuro:
already solved

PR #3550, opened 2026-02-21
Fix: white hover line overflow on critical path segments

ghost:
### Before:
https://github.com/user-attachments/assets/10977639-2641-4c0d-9e38-b5b184f47359

### After:
https://github.com/user-attachments/assets/075522fe-d50c-4a25-8b94-ec034480727f

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBar.css
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBar.css
@@ -38,8 +38,7 @@ SPDX-License-Identifier: Apache-2.0
   top: 45%;
   height: 11%;
   z-index: 2;
-  overflow: visible;
-  /* Changed from hidden to show shadow */
+  overflow: hidden;
   box-shadow:
     0 -0.5px 0 var(--critical-path-outline),
     0 0.5px 0 var(--critical-path-outline);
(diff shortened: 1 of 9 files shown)


packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBar.css
```diff
@@ -38,8 +38,7 @@ SPDX-License-Identifier: Apache-2.0
   top: 45%;
   height: 11%;
   z-index: 2;
-  overflow: visible;
-  /* Changed from hidden to show shadow */
```
yurishkuro, line 42: what shadow was the comment referring to, and why it is ok not to have it?

ghost: Sorry I forgot to mention, it was added in #3306 
Maybe it was assumed that due to `overflow: hidden`, the shadow might get clipped. So it was switched to `overflow: visible` to make sure it shows.
But i think `overflow: hidden` never clips shadow. Box shadows are painted outside element's boundary and that should be unaffected by `overflow`.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanBar.css
```diff
@@ -38,8 +38,7 @@ SPDX-License-Identifier: Apache-2.0
...
+  overflow: hidden;
   box-shadow:
```
ghost, line 42: this is the box-shadow i am referring to, I am not getting why this was added as it is rarely visible - just two thin white gap between the edges of the critical path bar. Maybe it was added so that we can tell the critical path bar is a different layer overlaid on the span??

PR #3557, opened 2026-03-05
Add commit-msg hook to always attach sign-off

yurishkuro:
When using Graphite CLI it does not support flags like `-s`, and having this as part of a hook actually makes it easier for new contributors.

### DIFF
--- a/.husky/commit-msg
+++ b/.husky/commit-msg
@@ -17,7 +17,7 @@ fi
 SIG="Signed-off-by: $NAME <$EMAIL>"
 
 # Check if it's already there to avoid duplicates
-if ! grep -q "$SIG" "$1"; then
+if ! grep -Fq -- "$SIG" "$1"; then
     # Append a newline and the signature in a portable way
     printf '\n%s\n' "$SIG" >> "$1"
 fi


.husky/commit-msg
```diff
@@ -0,0 +1,34 @@
...
+if [ -z "$NAME" ] || [ -z "$EMAIL" ]; then
+    IDENT=$(git var GIT_AUTHOR_IDENT 2>/dev/null || true)
+    if [ -n "$IDENT" ]; then
+        # IDENT format: "Name With Spaces <email@example.com> 1234567890 +0000"
+        NAME=$(printf '%s\n' "$IDENT" | sed -e 's/^[[:space:]]*//' -e 's/[[:space:]]*<.*//')
+        EMAIL=$(printf '%s\n' "$IDENT" | sed -n 's/.*<\([^>]*\)>.*/\1/p')
+    fi
+fi
+
+# Abort if we still don't have a valid NAME or EMAIL
```
yurishkuro:
```suggestion
# Abort if we don't have a valid NAME or EMAIL
```

PR #3565, opened 2026-03-06
Phase 2: Migrate route definitions to React Router v6 syntax

ghost:
- Resolves part of #3313 

## Description of the changes
- This PR converts the top-level route definitions from React Router v5 syntax to v6-compatible syntax using the `react-router-dom-v5-compat`. `<Router history={history}>` + `<CompatRouter>` setup in `src/index.tsx` is there, so both the v5 and v6 contexts available to all components.

- **Blocker 1: Traces were not opening**
  - **Reason:** `withRouteProps.tsx` was importing `useLocation` and `useParams` from `react-router-dom`(v5). After `App/index.tsx` was switched to `<Routes>` from `react-router-dom-v5-compat`, route params (like :id) are stored in the v6 route context. The v5 `useParams()` know nothing about that context, so it returned an empty object - making `params.id` undefined and prevent `TracePage` from fetching any trace.
  - **Fix:** changed the import in `withRouteProps.tsx` to `react-router-dom-v5-compat`

- **Blocker 2: Compare page not working**
  - **Reason:** React Router v6 automatically escapes special characters like `\`, `.`, and `?` when it converts a route path into a regex for matching URLs. Because of this, the `TraceDiff` route `/trace/:a?\.\.\.:b?` was compiled into a pattern that expected literal backslashes and dots (`\.\.\.`) in the URL. However, real browser URLs like `/trace/abc...def` do not contain backslashes, so the route never matched. As a result, the request fell through to the more general route `/trace/:id`, which captured the entire string `abc...def` as a single trace ID. This creating a 404 error.
  - **Fix:** Added `TraceRouter.tsx`, so that both normal trace URLs and trace comparison URLs use the same route `/trace/:id`. `TraceRouter` checks whether the id contains "...". If it does, it renders `TraceDiff`; otherwise it renders `TracePage`. Additionally, `TraceDiff.tsx` was updated so that `mapStateToProps` splits `params.id` on "..." to extract the two trace IDs (a and b).

- Other blockers were in `DeepDependencies` and `Monitor` tests that are easily fixed.

- Verify all routes working properly
- all tests passing

## Checklist

- [x] **Moderate**: AI helped with code generation or debugging specific parts **(helped add unit tests and give suggestions to fix blocker 2 (Compare page not working))**

### DIFF
--- a/packages/jaeger-ui/src/components/App/index.test.js
+++ b/packages/jaeger-ui/src/components/App/index.test.js
@@ -4,8 +4,7 @@
 import React from 'react';
 import { render } from '@testing-library/react';
 import '@testing-library/jest-dom';
-import { MemoryRouter } from 'react-router-dom';
-import { CompatRouter } from 'react-router-dom-v5-compat';
+import { MemoryRouter } from 'react-router-dom-v5-compat';
 
 jest.mock('./NotFound', () => () => <div data-testid="not-found" />);
 jest.mock('./Page', () => ({ children }) => <div data-testid="page">{children}</div>);
@@ -13,16 +12,16 @@ jest.mock('../DependencyGraph', () => () => <div data-testid="dependency-graph" 
 jest.mock('../DeepDependencies', () => () => <div data-testid="deep-dependencies" />);
 jest.mock('../QualityMetrics', () => () => <div data-testid="quality-metrics" />);
 jest.mock('../SearchTracePage', () => () => <div data-testid="search-trace" />);
-jest.mock('../TraceDiff', () => () => <div data-testid="trace-diff" />);
-jest.mock('../TracePage', () => () => <div data-testid="trace-page" />);
+jest.mock('../TraceDiff', () => ({ __esModule: true, default: () => <div data-testid="trace-diff" /> }));
+jest.mock('../TracePage', () => ({ __esModule: true, default: () => <div data-testid="trace-page" /> }));
 jest.mock('../Monitor', () => () => <div data-testid="monitor" />);
 
 jest.mock('../DependencyGraph/url', () => ({ ROUTE_PATH: '/dependencies' }));
 jest.mock('../DeepDependencies/url', () => ({ ROUTE_PATH: '/deep-dependencies' }));
 jest.mock('../QualityMetrics/url', () => ({ ROUTE_PATH: '/quality-metrics' }));
 jest.mock('../SearchTracePage/url', () => ({ ROUTE_PATH: '/search' }));
-jest.mock('../TraceDiff/url', () => ({ ROUTE_PATH: '/trace-diff' }));
-jest.mock('../TracePage/url', () => ({ ROUTE_PATH: '/trace' }));
+jest.mock('../TraceDiff/url', () => ({ ROUTE_PATH: '/trace-diff', getUrl: jest.fn(), matches: jest.fn() }));
+jest.mock('../TracePage/url', () => ({ ROUTE_PATH: '/trace/:id' }));
 jest.mock('../Monitor/url', () => ({ ROUTE_PATH: '/monitor' }));
 
 jest.mock('../../api/jaeger', () => ({
@@ -78,9 +77,7 @@ const renderWithPath = pathname => {
   mockHistory = createMockHistory(pathname);
   return render(
     <MemoryRouter initialEntries={[pathname]}>
-      <CompatRouter>
-        <JaegerUIApp />
-      </CompatRouter>
+      <JaegerUIApp />
     </MemoryRouter>
   );
 };
@@ -108,7 +105,8 @@ describe('JaegerUIApp', () => {
 
   const routes = [
     ['/search', 'search-trace'],
-    ['/trace-diff', 'trace-diff'],
+    // TraceDiff is now routed under /trace/:id - when id contains "...", TraceRouter renders TraceDiff
+    ['/trace/abc...def', 'trace-diff'],
     ['/trace/123', 'trace-page'],
     ['/dependencies', 'dependency-graph'],
     ['/deep-dependencies', 'deep-dependencies'],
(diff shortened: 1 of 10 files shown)


packages/jaeger-ui/src/components/TraceDiff/TraceDiff.tsx
```diff
@@ -154,7 +154,12 @@ export function TraceDiffImpl({
 
 // TODO(joe): simplify but do not invalidate the URL
 export function mapStateToProps(state: ReduxState, ownProps: TOwnProps) {
-  const { a, b } = ownProps.params;
+  let { a, b, id } = ownProps.params;
+  if (!a && id && id.includes('...')) {
+    const parts = id.split('...');
+    a = parts[0] || undefined;
+    b = parts[1] || undefined;
+  }
```
ghost, line 170: when a user opens the compare page, the URL is `/trace/abc...def`. The route only has one param, id = 'abc...def'. So we parse that string to get the two trace ids and assign them to a and b. The rest of `TraceDiff` still expects a and b, this way we support both the old params and the new id format.

ghost:
@yurishkuro ptal

packages/jaeger-ui/src/components/Monitor/index.test.js
```diff
@@ -3,6 +3,7 @@
 
 import React from 'react';
 import { MemoryRouter } from 'react-router-dom';
```
yurishkuro, line 5: should we not kill this import?

packages/jaeger-ui/src/components/TraceDiff/TraceDiff.tsx
yurishkuro: nit: you should put this as comment in the code.

It's not clear to me what you mean by old and new format. Currently I see the URL for compare page looks like this (the IDs repeated twice): `http://localhost:5173/trace/73b4e476e9c287603ee96ed7019ab8d4...9c18cb9d26e6d931fa54a1e6ad8fbb06?cohort=73b4e476e9c287603ee96ed7019ab8d4&cohort=9c18cb9d26e6d931fa54a1e6ad8fbb06`

packages/jaeger-ui/src/components/Monitor/index.test.js
ghost: Yes, there are many other imports also from react-router-dom in other files. We will remove them step by step. For this file we can't remove this now because it's connected to `ServicesView/index.tsx`, so it will break the tests. This import will be removed maybe in Phase 5, when ServicesView is updated to use <Link> from `react-router-dom-v5-compat` instead.

packages/jaeger-ui/src/components/TraceDiff/TraceDiff.tsx
ghost: Sorry for the confusion, the URL format in the browser has not changed at all. Actually, in v5, the route pattern `/trace/:a?\.\.\.":b?` tells the router to split the path segment at `...` automatically, giving the component two separate params.
But in v6, that regex-style pattern is not supported, so the route never matched. We replaced it with `/trace/:id`, which gives a single param: `params.id = '73b4e476...9c18cb9d'`. This code (in `TraceDiff.tsx`) then manually splits that string on `...` to get `a` and `b`, so the rest of the component continue to work properly without any changes.

ghost: > nit: you should put this as comment in the code.

done

packages/jaeger-ui/src/components/App/index.test.js
```diff
@@ -4,25 +4,24 @@
...
+jest.mock('../TracePage', () => ({ __esModule: true, default: () => <div data-testid="trace-page" /> }));
 jest.mock('../Monitor', () => () => <div data-testid="monitor" />);
 
 jest.mock('../DependencyGraph/url', () => ({ ROUTE_PATH: '/dependencies' }));
 jest.mock('../DeepDependencies/url', () => ({ ROUTE_PATH: '/deep-dependencies' }));
 jest.mock('../QualityMetrics/url', () => ({ ROUTE_PATH: '/quality-metrics' }));
 jest.mock('../SearchTracePage/url', () => ({ ROUTE_PATH: '/search' }));
-jest.mock('../TraceDiff/url', () => ({ ROUTE_PATH: '/trace-diff' }));
-jest.mock('../TracePage/url', () => ({ ROUTE_PATH: '/trace' }));
+jest.mock('../TraceDiff/url', () => ({ ROUTE_PATH: '/trace-diff', getUrl: jest.fn(), matches: jest.fn() }));
```
yurishkuro, line 23: if the diff page uses `/trace/id...id` format where is `/trace-diff` coming from?

packages/jaeger-ui/src/components/App/index.tsx
```diff
@@ -60,43 +58,20 @@ export default function JaegerUIApp() {
...
-              </Route>
+              <Route path="/" element={<Navigate to={searchPath} replace />} />
+              <Route path={prefixUrl()} element={<Navigate to={searchPath} replace />} />
+              <Route path={prefixUrl('/')} element={<Navigate to={searchPath} replace />} />
 
-              <Route>
-                <NotFound error="Page not found" />
-              </Route>
-            </Switch>
+              <Route path="*" element={<NotFound error="Page not found" />} />
```
yurishkuro, line 73: this is nice, I think previously any unrecognized url would silently bounce to search.

yurishkuro (approved):

packages/jaeger-ui/src/components/App/index.test.js
ghost: yeah, this mock is stale.

packages/jaeger-ui/src/components/App/index.tsx
ghost: yes

PR #3574, opened 2026-03-07
Phase 3.1: Replace state.router.location with useLocation()

ghost:
- Resolves part of #3313

## Description of the changes
- Components were reading the current URL from Redux like this: 
`state.router.location.search -> e.g. "?service=frontend&limit=20"`
This means the URL was being stored in Redux and components were reading it from there.
- The new way in this PR: components use a React Router hook:
`const { search } = useLocation(); -> e.g. "?service=frontend&limit=20"`
This means ask React Router directly what the current URL search is. No Redux needed.

#### Why make this change? 
Because in our next phases, we will remove the `history` and `redux-first-history` packages. Those packages are what put the URL into Redux in the first place. So, before we can remove them, every component that reads from `state.router.location` must be updated to use hooks.

#### Every changes in this PR using one of these three patterns:
- Hook directly in component (used in `Page.tsx`, `UiFindInput.tsx`)
- Wrapper component (used in `SearchForm.tsx`, `DdgNodeContent.tsx`, `DetailsPanel.tsx`)
- Already-injected prop (used in `TraceDiff.tsx`)

- Unit tests
- Components working properly

## Checklist

- [x] **Light**: AI provided minor assistance (formatting, simple suggestions) **[gave suggestions and fixing the failed unit tests, lint errors]**

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageSearchBar.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageSearchBar.tsx
@@ -76,11 +76,7 @@ export function TracePageSearchBarFn(props: TracePageSearchBarProps & { forwarde
     <div className="TracePageSearchBar">
       {/* style inline because compact overwrites the display */}
       <Input.Group className="ub-justify-end" compact style={{ display: 'flex' }}>
-        <UiFindInput
-          inputProps={uiFindInputInputProps}
-          forwardedRef={forwardedRef}
-          trackFindFunction={trackFilter}
-        />
+        <UiFindInput inputProps={uiFindInputInputProps} ref={forwardedRef} trackFindFunction={trackFilter} />
         <Tooltip
           arrow={{ pointAtCenter: true }}
           placement="bottomLeft"
--- a/packages/jaeger-ui/src/components/common/UiFindInput.tsx
+++ b/packages/jaeger-ui/src/components/common/UiFindInput.tsx
@@ -7,7 +7,6 @@ import { IoClose } from 'react-icons/io5';
 import { useNavigate, useLocation } from 'react-router-dom-v5-compat';
 import _debounce from 'lodash/debounce';
 import _isString from 'lodash/isString';
-import { connect } from 'react-redux';
 
 import updateUiFind from '../../utils/update-ui-find';
 import { TNil, ReduxState } from '../../types/index';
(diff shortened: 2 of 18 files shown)


packages/jaeger-ui/src/components/common/UiFindInput.tsx
```diff
@@ -33,15 +29,20 @@ export const UnconnectedUiFindInput = React.forwardRef<InputRef, TProps>((props,
...
     ...defaultProps,
     ...props,
   };
 
   const navigate = useNavigate();
   const location = useLocation();
+
+  // derive uiFind from the URL when not provided as a prop.
+  const { uiFind: uiFindFromUrl } = parseQuery(location.search);
+  const uiFindFromLocation = Array.isArray(uiFindFromUrl) ? uiFindFromUrl.join(' ') : uiFindFromUrl;
```
yurishkuro: this is the same parsing logic as in extractUiFindFromState, let's move to a helper function

ghost: done

packages/jaeger-ui/src/components/common/UiFindInput.tsx
```diff
@@ -109,10 +110,14 @@ export const UnconnectedUiFindInput = React.forwardRef<InputRef, TProps>((props,
...
 UnconnectedUiFindInput.displayName = 'UnconnectedUiFindInput';
 
+export type TExtractUiFindFromStateReturn = {
+  uiFind: string | undefined;
+};
+
 export function extractUiFindFromState(state: ReduxState): TExtractUiFindFromStateReturn {
   const { uiFind: uiFindFromUrl } = parseQuery(state.router.location.search);
   const uiFind = Array.isArray(uiFindFromUrl) ? uiFindFromUrl.join(' ') : uiFindFromUrl;
   return { uiFind };
```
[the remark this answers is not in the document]
ghost: will be done in next part of phase 3

packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageSearchBar.tsx
```diff
@@ -76,11 +76,7 @@ export function TracePageSearchBarFn(props: TracePageSearchBarProps & { forwarde
     <div className="TracePageSearchBar">
       {/* style inline because compact overwrites the display */}
       <Input.Group className="ub-justify-end" compact style={{ display: 'flex' }}>
-        <UiFindInput
-          inputProps={uiFindInputInputProps}
-          forwardedRef={forwardedRef}
-          trackFindFunction={trackFilter}
-        />
+        <UiFindInput inputProps={uiFindInputInputProps} ref={forwardedRef} trackFindFunction={trackFilter} />
```
yurishkuro, line 79: I don't see changes in UiFindInput that would account for switching from `forwardedRef` to `ref`. Are those React standard props? What is the reason for such change?

ghost: `UiFindInput` was already using `React.forwardRef` correctly. We just fixed the two call sites (`TracePageSearchBar.tsx` and `DeepDependencies/Header/index.tsx`) that were passing the `ref` as a custom prop instead of standard `ref`, which is what `React.forwardRef` actually listen for.

PR #3590, opened 2026-03-11
fix(api): resolve infinite loading on trace/service fetch errors #3079

### LINKED ISSUE #3079: [Bug]: UI Shows Infinite Loading for Non-Existent Trace ID

Champbreed:
This PR resolves the infinite loading state observed when the Jaeger v3 API returns an error (both 4xx/5xx and 200 OK with an error body).

Changes:
- Modified fetchWithTimeout to explicitly throw an Error when !response.ok, ensuring React Query transitions to an error state.
- Implemented internal error checking in fetchServices and fetchSpanNames to catch "hidden" errors in the JSON body (closing the blind spot identified in #3161).
- Verified: Confirmed with a 404 response for non-existent Trace IDs (now displays error UI instead of spinning).

Fixes #3079.

### DIFF
--- a/packages/jaeger-ui/src/api/v3/client.ts
+++ b/packages/jaeger-ui/src/api/v3/client.ts
@@ -16,7 +16,8 @@ export class JaegerClient {
 
   /**
    * Helper to handle Jaeger v3 specific error arrays in 200 OK responses.
-   * Graphite Fix: Throws raw message; wrapping happens in public methods.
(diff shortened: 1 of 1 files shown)


Champbreed:
Can this be reopened

yurishkuro:
> Can this be reopened

copilot feedback was not addressed, this is rightly closed.

PR #3593, opened 2026-03-11
refactor(TracePage): extract keyboard shortcut setup into dedicated module

thc1006:
## Summary
- Extract keyboard shortcut registration/teardown from `componentDidMount`/`componentWillUnmount` into a new `keyboard-setup.ts` module
- Pure code movement with no behavior changes — the same shortcuts are registered in the same order
- `shortcutConfig` and `makeShortcutCallbacks` move to the new module and are re-exported from `index.tsx` for backward compatibility

This is the second in a series of incremental PRs preparing TracePage for conversion to a functional component (#3381). The extracted `setupTracePageShortcuts()` function will map directly to a `useEffect` in the final conversion.

Follows the approach agreed upon in #3435:
> "agreed, let's make smaller changes ideally starting with just those that don't change the logic but simply move things around." — @yurishkuro

## Changes
- **New:** `keyboard-setup.ts` — exports `setupTracePageShortcuts()` (returns cleanup fn), `shortcutConfig`, `makeShortcutCallbacks`
- **Modified:** `index.tsx` — `componentDidMount` calls `setupTracePageShortcuts()`, `componentWillUnmount` calls the returned cleanup. Removed `_mapValues` import (no longer needed here).

## Test plan
- [x] All 66 existing TracePage tests pass unchanged
- [x] `makeShortcutCallbacks` tests still work via re-export
- [x] Keyboard shortcut setup test still verifies `mergeShortcuts` is called with correct callbacks
- [x] Unmount test still verifies `resetShortcuts` and `cancelScroll` are called
- [x] Pre-commit hooks (eslint, prettier) pass

Depends on: #3592 (can be reviewed independently but should merge after)
Closes partially #3381

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/index.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/index.tsx
@@ -7,7 +7,6 @@ import { useNormalizeTraceId } from './useNormalizeTraceId';
 import { Location, History as RouterHistory } from 'history';
 import _clamp from 'lodash/clamp';
 import _get from 'lodash/get';
-import _mapValues from 'lodash/mapValues';
 import _memoize from 'lodash/memoize';
 import { connect } from 'react-redux';
 import { bindActionCreators, Dispatch } from 'redux';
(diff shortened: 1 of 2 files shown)


thc1006:
Closing — this work was superseded by #3609 which was merged on 2026-03-14. The same keyboard shortcut extraction was done as a class method _setupKeyboardShortcuts() instead of a separate module.

PR #3594, opened 2026-03-11
Fix Dependabot security alerts

jkowall:
## Purpose
This PR resolves open Dependabot security issues in jaeger-ui and related workspace dependencies.

## Summary
- bump react-router-dom-v5-compat from 6.30.0 to 6.30.3
- add root overrides for svgo, rollup, and qs
- refresh lockfile via npm install + npm audit fix

## Dependabot alerts addressed
- svgo DoS (high)
- rollup path traversal arbitrary file write (high)
- qs DoS advisories (medium/low)
- react-router unexpected external redirect (medium)
- @remix-run/router open redirect/XSS path (high)
- serialize-javascript RCE advisory (high)

## Validation
- npm run prettier
- npm run lint
- npm test
- npm run build

### DIFF
--- a/.github/workflows/check_bundle.yml
+++ b/.github/workflows/check_bundle.yml
@@ -18,9 +18,13 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@395ad3262231945c25e8478fd5baf05154b1d79f # v6.1.0
         with:
-          node-version-file: '.nvmrc'
+          node-version-file: 'package.json'
           cache: 'npm'
-          
+      - name: Upgrade npm
+        run: |
+          NPM_VERSION=$(node -p "require('./package.json').engines.npm.replace(/[^0-9.]/g, '')")
+          npm install -g npm@$NPM_VERSION
+      - run: npm --version
       - name: Install dependencies
         run: npm ci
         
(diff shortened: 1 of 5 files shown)


yurishkuro:
conflicts

yurishkuro:
> add root overrides for svgo, rollup, and qs

What's the motivation for these? Overrides are generally technical debt.

jkowall:
The lockfile needs to change here for two separate reasons:

1. The intended dependency update in this PR is not just `package.json` metadata. The root `overrides` for `qs`, `svgo`, and `rollup`, plus the `react-router-dom-v5-compat` bump, require the resolved dependency tree in `package-lock.json` to move to the patched versions.

2. CI runs `npm ls --all` with Node 24 / npm 11, and npm 11 is stricter about invalid trees. With only the original override changes, CI failed because `packages/plexus` still depended on `webpack-dev-server@5.2.2`, which pulled an older `express` / `body-parser` chain that conflicted with the `qs` override. Bumping `webpack-dev-server` to `5.2.3` updates that chain so the lockfile and the installed tree are consistent again.

So the lockfile diff is required both to carry the security fixes through to the actual resolved tree and to make the npm 11 CI validation pass.

PR #3605, opened 2026-03-13
fix: replace unsafe _get() with optional chaining in embedded reducer

zamadye:
hey, i saw the issue about removing unsafe _get() calls and started with the embedded reducer. this one was straightforward - just replaced _get(window, 'location.search') with window?.location?.search. let me know if you want me to tackle more of these!


yurishkuro:
commits not signer, PR description not following the template

zamadye:
thanks for the feedback! 👍

PR #3613, opened 2026-03-15
Final Part of Phase 3.2 (task 1): refactor(TracePage): convert class component to functional

### LINKED ISSUE #3381: [UI Refactor] Migrate TracePage to Functional Component
This issue is part of the larger effort to modernize the Jaeger UI codebase (Parent Issue #2610). The goal is to refactor the `TracePage` component from a React Class Component to a React Functional Component using Hooks.

**Target Component:** `packages/jaeger-ui/src/components/TracePage/index.tsx`

### Acceptance Criteria

| --- | --- | --- |

### Resources

ghost:
- Resolves #3381 
- Part of #3313 

## Description of the changes
- Convert `TracePage/index.tsx` from class component to functional component.

- Unit tests
- Verified that the component working properly in the ui

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/index.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/index.tsx
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import * as React from 'react';
-import { useCallback, useEffect, useRef, useState } from 'react';
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { InputRef } from 'antd';
 import { useNormalizeTraceId } from './useNormalizeTraceId';
 import { Location, History as RouterHistory } from 'history';
@@ -173,9 +173,16 @@ export function TracePageImpl(props: TProps) {
   const [viewType, setViewType] = useState<ETraceViewType>(ETraceViewType.TraceTimelineViewer);
   const [viewRange, setViewRange] = useState<IViewRange>({ time: { current: [0, 1] } });
 
+  const traceDagEV = useMemo(
+    () =>
+      viewType === ETraceViewType.TraceGraph && trace?.data
+        ? calculateTraceDagEV(trace.data.asOtelTrace())
+        : null,
+    [trace, viewType]
+  );
+
   const searchBarRef = useRef<InputRef>(null);
   const headerElmRef = useRef<HTMLElement | TNil>(null);
-  const traceDagEVRef = useRef<TEv | TNil>(null);
   const viewRangeRef = useRef(viewRange);
   viewRangeRef.current = viewRange;
   const prevIdRef = useRef(id);
@@ -240,6 +247,7 @@ export function TracePageImpl(props: TProps) {
     shortcutCallbacks.scrollToPrevVisibleSpan = scrollToPrevVisibleSpan;
     shortcutCallbacks.clearSearch = () => clearSearchRef.current();
     shortcutCallbacks.searchSpans = () => focusOnSearchBarRef.current();
+    resetShortcuts();
     mergeShortcuts(shortcutCallbacks);
 
     return () => {
@@ -277,11 +285,13 @@ export function TracePageImpl(props: TProps) {
     headerElmRef.current = elm;
     if (elm) {
       setHeaderHeight(elm.clientHeight);
-      const resizeObserver = new ResizeObserver(() => {
-        setHeaderHeight(elm.clientHeight);
-      });
-      resizeObserver.observe(elm);
-      headerResizeObserverRef.current = resizeObserver;
+      if (typeof ResizeObserver !== 'undefined') {
+        const resizeObserver = new ResizeObserver(() => {
+          setHeaderHeight(elm.clientHeight);
+        });
+        resizeObserver.observe(elm);
+        headerResizeObserverRef.current = resizeObserver;
+      }
     } else {
       setHeaderHeight(null);
     }
@@ -294,15 +304,9 @@ export function TracePageImpl(props: TProps) {
     });
   }, []);
 
-  const setTraceView = useCallback(
-    (newViewType: ETraceViewType) => {
-      if (trace && trace.data && newViewType === ETraceViewType.TraceGraph) {
-        traceDagEVRef.current = calculateTraceDagEV(trace.data.asOtelTrace());
-      }
-      setViewType(newViewType);
-    },
-    [trace]
-  );
+  const setTraceView = useCallback((newViewType: ETraceViewType) => {
+    setViewType(newViewType);
+  }, []);
 
   const archiveTrace = useCallback(() => {
     archiveTraceProp(id);
@@ -350,7 +354,7 @@ export function TracePageImpl(props: TProps) {
   let spanFindMatches: Set<string> | null | undefined;
   if (uiFind) {
     if (viewType === ETraceViewType.TraceGraph) {
-      graphFindMatches = getUiFindVertexKeys(uiFind, _get(traceDagEVRef.current, 'vertices', []));
+      graphFindMatches = getUiFindVertexKeys(uiFind, _get(traceDagEV, 'vertices', []));
       findCount = graphFindMatches ? graphFindMatches.size : 0;
     } else {
       spanFindMatches = filterSpansMemo(uiFind, _get(trace, 'data.spans'));
@@ -418,7 +422,7 @@ export function TracePageImpl(props: TProps) {
     view = (
       <TraceGraph
         headerHeight={headerHeight}
-        ev={traceDagEVRef.current}
+        ev={traceDagEV}
         uiFind={uiFind}
         uiFindVertexKeys={graphFindMatches}
         traceGraphConfig={traceGraphConfig}
--- a/packages/jaeger-ui/src/components/TracePage/index.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/index.tsx
@@ -207,11 +207,11 @@ export function TracePageImpl(props: TProps) {
 
   const clearSearch = useCallback(() => {
     updateUiFind({ history, location, trackFindFunction: trackFilter });
-    if (searchBarRef.current) searchBarRef.current.blur();
+    searchBarRef.current?.blur();
   }, [history, location]);
 
   const focusOnSearchBar = useCallback(() => {
-    if (searchBarRef.current) searchBarRef.current.focus();
+    searchBarRef.current?.focus();
   }, []);
 
   const clearSearchRef = useRef(clearSearch);
@@ -252,11 +252,6 @@ export function TracePageImpl(props: TProps) {
   useEffect(() => {
     scrollManagerRef.current.setTrace(trace && trace.data ? trace.data.asOtelTrace() : undefined);
   }, [trace]);
-
-  // Mount: reset view range to full (matches class component componentDidMount behavior)
-  useEffect(() => {
-    updateViewRangeTime(0, 1);
-  }, [updateViewRangeTime]);
 
   useEffect(() => {
     if (prevIdRef.current !== id) {
--- a/packages/jaeger-ui/src/components/TracePage/index.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/index.tsx
@@ -214,6 +214,11 @@ export function TracePageImpl(props: TProps) {
     if (searchBarRef.current) searchBarRef.current.focus();
   }, []);
 
+  const clearSearchRef = useRef(clearSearch);
+  clearSearchRef.current = clearSearch;
+  const focusOnSearchBarRef = useRef(focusOnSearchBar);
+  focusOnSearchBarRef.current = focusOnSearchBar;
+
   const adjustViewRange = useCallback(
     (startChange: number, endChange: number, trackSrc: string) => {
       const [viewStart, viewEnd] = viewRangeRef.current.time.current;
@@ -233,8 +238,8 @@ export function TracePageImpl(props: TProps) {
     shortcutCallbacks.scrollPageUp = scrollPageUp;
     shortcutCallbacks.scrollToNextVisibleSpan = scrollToNextVisibleSpan;
     shortcutCallbacks.scrollToPrevVisibleSpan = scrollToPrevVisibleSpan;
-    shortcutCallbacks.clearSearch = clearSearch;
-    shortcutCallbacks.searchSpans = focusOnSearchBar;
+    shortcutCallbacks.clearSearch = () => clearSearchRef.current();
+    shortcutCallbacks.searchSpans = () => focusOnSearchBarRef.current();
     mergeShortcuts(shortcutCallbacks);
 
     return () => {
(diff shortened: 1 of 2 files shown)


packages/jaeger-ui/src/components/TracePage/index.tsx
```diff
@@ -125,348 +142,327 @@ export function makeShortcutCallbacks(adjRange: (start: number, end: number) =>
...
-    this.ensureTraceFetched();
-    this.updateViewRangeTime(0, 1);
-    this._setupKeyboardShortcuts();
-  }
-
-  componentDidUpdate({ id: prevID }: TProps) {
-    const { id, trace } = this.props;
-
-    this._scrollManager.setTrace(trace && trace.data ? trace.data.asOtelTrace() : undefined);
+export function TracePageImpl(props: TProps) {
```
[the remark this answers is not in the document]
ghost: not necessary as it will add very little value

ghost:
@yurishkuro ptal

yurishkuro:
some review comments from Claude Code:

## Issues

### 1. `updateViewRangeTime(0, 1)` not called on mount when trace is already loaded

**Severity: Bug**

In the class component, `componentDidMount` unconditionally called `this.updateViewRangeTime(0, 1)`.
In the functional component, the id/trace effect only calls `updateViewRangeTime(0, 1)` when either
the id changed or `!trace`:

```tsx
// index.tsx ~line 1042
useEffect(() => {
  if (prevIdRef.current !== id) {
    ...
    updateViewRangeTime(0, 1);
  } else if (!trace) {
    fetchTrace(id);
    updateViewRangeTime(0, 1);   // ← only called when trace is absent
  }
}, [id, trace, fetchTrace, updateViewRangeTime, clearSearch]);
```

If `trace` is already loaded when the component mounts (e.g., navigating back to a previously
loaded trace), `updateViewRangeTime(0, 1)` is never called. This could leave the view range in
a stale state from a previous mount. Fix: add an explicit mount-only effect or call
`updateViewRangeTime(0, 1)` unconditionally in the first branch of the id-change check on initial
render.

---

### 2. Stale closure in keyboard shortcuts (mount effect)

**Severity: Bug (latent)**

The mount effect captures `clearSearch` and `focusOnSearchBar` at initial render time:

```tsx
// index.tsx ~line 1001
useEffect(() => {
  ...
  shortcutCallbacks.clearSearch = clearSearch;
  shortcutCallbacks.searchSpans = focusOnSearchBar;
  mergeShortcuts(shortcutCallbacks);
  ...
  // eslint-disable-next-line react-hooks/exhaustive-deps
}, []);
```

`clearSearch` closes over `history` and `location` props. If those props change during the
component's lifetime (which is plausible in a routed app), the keyboard shortcut will silently
invoke the stale version. The eslint-disable comment acknowledges this but doesn't fix it.

A safe pattern is to forward via a ref:

```tsx
const clearSearchRef = useRef(clearSearch);
clearSearchRef.current = clearSearch;   // kept fresh every render

useEffect(() => {
  ...
  shortcutCallbacks.clearSearch = () => clearSearchRef.current();
  shortcutCallbacks.searchSpans = () => focusOnSearchBarRef.current();
  mergeShortcuts(shortcutCallbacks);
  ...
}, []);   // no stale captures
```

---

### 3. Header height not updated when header content changes

**Severity: Regression**

In the class component, `componentDidUpdate` called `this.setHeaderHeight(this._headerElm)` on
every update, so the measured `clientHeight` was always fresh (e.g., after toggling slim view).

In the functional component there are two mechanisms:
- `headerRefCallback` — called only when the element mounts/unmounts (correct for initial height)
- `useEffect([headerHeight])` — re-runs only when `headerHeight` itself changes (circular)

Neither mechanism captures a height change caused by a re-render (e.g., the header shrinks after
`toggleSlimView()`). A `ResizeObserver` on `headerElmRef.current` would be the correct React
replacement. At a minimum, a comment explaining the intentional deviation from the class
component's behavior would be helpful.

---

### 4. Missing `React.memo` wrapper

**Severity: Minor / Perf regression**

`TracePageImpl` was previously a `React.PureComponent`, meaning it performed shallow-prop
comparisons before re-rendering. The functional equivalent is `React.memo`. Without it, every
parent render causes `TracePageImpl` to re-render regardless of whether props changed.

`TracePageImpl` is exported directly (used in tests and potentially other call sites), so
wrapping it in `React.memo` would restore the original optimization:

```tsx
export const TracePageImpl = React.memo(function TracePageImpl(props: TProps) { ... });
```

This was noted in the existing Copilot review comment but dismissed. Given the explicit mention
in the issue's acceptance criteria (#3381), it is worth addressing.

---

### 5. `updateViewRangeTime(0, 1)` called unnecessarily in `else if (!trace)` branch

**Severity: Nit**

The state starts at `[0, 1]` already (from `useState({ time: { current: [0, 1] } })`). Calling
`updateViewRangeTime(0, 1)` on first mount when there is no trace is a no-op (it sets `[0, 1]`
again) and also triggers an unnecessary `trackRange` side-effect path (guarded by `trackSrc`, so
harmless here, but confusing).

---

### 6. `idRef` and `viewRangeRef` mutations during render

**Severity: Nit**

```tsx
viewRangeRef.current = viewRange;   // line 927
idRef.current = id;                 // line 930
```

Mutating refs during the render phase is a side effect and technically violates React's render
purity contract (even though it works in practice because React renders synchronously in the
common case). Prefer updating these in a `useLayoutEffect` with no deps (runs synchronously after
each commit) or accept them as intentional imperative escapes and add a comment.

---

## Positive Observations

- **Test quality is significantly improved.** The old tests relied on `new TracePage(props)`
  instantiation, direct instance method calls, and prototype patching — all of which are fragile
  and bypass React's actual rendering. The new tests render the component for real and interact
  through captured props, which is much closer to how the code runs in production.

- **`computeAdjustedRange` extraction** is a good separation of a pure function that was
  previously buried in a method. The corresponding test change to test it directly (rather than
  indirectly through `_adjustViewRange`) is cleaner.

- **`useCallback` usage is consistent and appropriate** — no missing or excessive memoization.

- **`filterSpansMemo` via `useRef`** correctly avoids re-creating the memoized function on every
  render while keeping the `id` cache key up to date through `idRef`.

- **Scroll manager lifecycle** is correctly handled: created once in `useRef`, updated via a
  `useEffect([trace])`, and destroyed in the mount effect's cleanup.

---

## Summary

The conversion is largely correct and the test improvements are a clear win. Three issues
warrant fixes before merging:

1. `updateViewRangeTime(0, 1)` not called on initial mount when trace is already present
2. Stale `clearSearch` / `focusOnSearchBar` captured in keyboard shortcut mount effect
3. Header height not re-measured after content-triggered height changes (slim view toggle)

Items 4–6 are lower priority.

ghost:
@yurishkuro Issues 1-3 are fixed

packages/jaeger-ui/src/components/TracePage/index.tsx
```diff
@@ -125,348 +142,329 @@ export function makeShortcutCallbacks(adjRange: (start: number, end: number) =>
...
-    } else if (this.state.headerHeight) {
-      this.setState({ headerHeight: null });
+      updateViewRangeTime(0, 1);
+      clearSearch();
+    } else if (!trace) {
+      fetchTrace(id);
+      updateViewRangeTime(0, 1);
     }
-  };
+  }, [id, trace, fetchTrace, updateViewRangeTime, clearSearch]);
```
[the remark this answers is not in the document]
ghost: Yes there is double call but i think it's harmless

packages/jaeger-ui/src/components/TracePage/index.tsx
```diff
@@ -125,348 +142,329 @@ export function makeShortcutCallbacks(adjRange: (start: number, end: number) =>
...
-  componentWillUnmount() {
-    resetShortcuts();
-    cancelScroll();
-    this._scrollManager.destroy();
-    this._scrollManager = new ScrollManager(undefined, {
-      scrollBy,
-      scrollTo,
-    });
-  }
+  const updateNextViewRangeTime = useCallback((update: ViewRangeTimeUpdate) => {
```
yurishkuro, line 211: why is it called "next"? "update" already implies a change vs. previous state, by adding "next" it sounds like it's referring to a _different data structure_ compared to just the "view range time".

packages/jaeger-ui/src/components/TracePage/index.tsx
```diff
@@ -125,348 +142,329 @@ export function makeShortcutCallbacks(adjRange: (start: number, end: number) =>
...
+  }, [history, location]);
+
+  const focusOnSearchBar = useCallback(() => {
+    if (searchBarRef.current) searchBarRef.current.focus();
+  }, []);
+
+  const clearSearchRef = useRef(clearSearch);
+  clearSearchRef.current = clearSearch;
+  const focusOnSearchBarRef = useRef(focusOnSearchBar);
+  focusOnSearchBarRef.current = focusOnSearchBar;
```
yurishkuro, line 227: this looks very convoluted. Since both clearSearch and focusOnSearchBar are already stable function pointers, why do you need to wrap them in useRef?

packages/jaeger-ui/src/components/TracePage/index.tsx
```diff
@@ -125,348 +142,329 @@ export function makeShortcutCallbacks(adjRange: (start: number, end: number) =>
...
-      scrollTo,
-    });
-  }
+  const updateNextViewRangeTime = useCallback((update: ViewRangeTimeUpdate) => {
+    setViewRange(prev => ({ ...prev, time: { ...prev.time, ...update } }));
+  }, []);
+
+  const clearSearch = useCallback(() => {
+    updateUiFind({ history, location, trackFindFunction: trackFilter });
+    if (searchBarRef.current) searchBarRef.current.blur();
```
yurishkuro:
```suggestion
    searchBarRef.current?.blur();
```

packages/jaeger-ui/src/components/TracePage/index.tsx
```diff
@@ -125,348 +142,329 @@ export function makeShortcutCallbacks(adjRange: (start: number, end: number) =>
...
+  }, [history, location]);
+
+  const focusOnSearchBar = useCallback(() => {
+    if (searchBarRef.current) searchBarRef.current.focus();
```
yurishkuro:
```suggestion
    searchBarRef.current?.focus();
```

packages/jaeger-ui/src/components/TracePage/index.tsx
```diff
@@ -125,348 +142,329 @@ export function makeShortcutCallbacks(adjRange: (start: number, end: number) =>
...
+    const adjViewRange = (a: number, b: number) => adjustViewRange(a, b, 'kbd');
     const shortcutCallbacks = makeShortcutCallbacks(adjViewRange);
     shortcutCallbacks.scrollPageDown = scrollPageDown;
     shortcutCallbacks.scrollPageUp = scrollPageUp;
     shortcutCallbacks.scrollToNextVisibleSpan = scrollToNextVisibleSpan;
     shortcutCallbacks.scrollToPrevVisibleSpan = scrollToPrevVisibleSpan;
-    shortcutCallbacks.clearSearch = this.clearSearch;
-    shortcutCallbacks.searchSpans = this.focusOnSearchBar;
+    shortcutCallbacks.clearSearch = () => clearSearchRef.current();
+    shortcutCallbacks.searchSpans = () => focusOnSearchBarRef.current();
```
yurishkuro, line 249: why can these not refer directly to the callbacks?

```suggestion
    shortcutCallbacks.clearSearch = clearSearch;
    shortcutCallbacks.searchSpans = focusOnSearchBar;
```

packages/jaeger-ui/src/components/TracePage/index.tsx
```diff
@@ -125,348 +142,329 @@ export function makeShortcutCallbacks(adjRange: (start: number, end: number) =>
...
-        this.setState({ headerHeight: elm.clientHeight });
+    return () => {
+      resetShortcuts();
+      cancelScroll();
+      sm.destroy();
+    };
+  }, [adjustViewRange]);
+
+  useEffect(() => {
+    scrollManagerRef.current.setTrace(trace && trace.data ? trace.data.asOtelTrace() : undefined);
```
yurishkuro: simplify?
```suggestion
    scrollManagerRef.current.setTrace(trace?.data?.asOtelTrace());
```

packages/jaeger-ui/src/components/TracePage/index.tsx
```diff
@@ -125,348 +142,329 @@ export function makeShortcutCallbacks(adjRange: (start: number, end: number) =>
...
+  useEffect(() => {
+    if (prevIdRef.current !== id) {
+      prevIdRef.current = id;
+      if (!trace) {
+        fetchTrace(id);
       }
-    } else if (this.state.headerHeight) {
-      this.setState({ headerHeight: null });
+      updateViewRangeTime(0, 1);
+      clearSearch();
```
yurishkuro, line 271: any idea why clearSearch() is not called in the `else if` part?

packages/jaeger-ui/src/components/TracePage/index.tsx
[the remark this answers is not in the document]
yurishkuro: could we simply call it once at the end of this larger useEffect() (not inside each of the if branches)?

packages/jaeger-ui/src/components/TracePage/index.tsx
ghost: This is used in many files. This change belongs in a separate cleanup PR.

packages/jaeger-ui/src/components/TracePage/index.tsx
ghost: `clearSearch` is not stable. It depends on `history`, `location` and `location` changing on every URL navigation.

packages/jaeger-ui/src/components/TracePage/index.tsx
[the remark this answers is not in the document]
ghost: > call it once at the end of this larger useEffect()

but it would cause the view range to reset every time trace finish the loading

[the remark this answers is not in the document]
ghost: so i just removed the duplicate useEffect()

packages/jaeger-ui/src/components/TracePage/index.tsx
ghost: `clearSearch()` is only called when the user navigates to a different trace id (so the old search filter is cleared)

packages/jaeger-ui/src/components/TracePage/index.tsx
ghost: https://github.com/jaegertracing/jaeger-ui/pull/3613#discussion_r2950579309 same reason

packages/jaeger-ui/src/components/TracePage/index.tsx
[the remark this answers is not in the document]
yurishkuro: > but it would cause the view range to reset every time trace finish the loading

is that the wrong thing to do? Previous view range doesn't make sense for a completely new trace.

packages/jaeger-ui/src/components/TracePage/index.tsx
yurishkuro: ok what about focusOnSearchBar? It has no dependencies

yurishkuro (approved):

packages/jaeger-ui/src/components/TracePage/index.tsx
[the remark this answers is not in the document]
ghost: It's not wrong but on a trace page, location changes on every keystroke in the "Find" search bar. So this causes the effect to re-run and if `updateViewRangeTime(0, 1)` is at the end, and it will reset the user's zoom every time they type. That's why i kept it inside the branches.

This is the issue:
https://github.com/user-attachments/assets/e355d70d-c319-436e-a616-b91ad95b1365

packages/jaeger-ui/src/components/TracePage/index.tsx
ghost: yes focusOnSearchBar is stable

PR #3618, opened 2026-03-20
Part of Phase 4 & 5: Migrate Link to react-router-dom-v5-compat

ghost:
- Resolves part of #3313 

## Description of the changes
- Moving all `Link` imports from `react-router-dom` (v5) to `react-router-dom-v5-compat`. So preparing for a future switch to `react-router-dom` v7.
- `CompatRouter` must sit under v5 `MemoryRouter` / `BrowserRouter` from `react-router-dom`. Using compat’s `MemoryRouter` there will break `CompatRouter` (missing location) or cause any other errors. So tests that use `MemoryRouter` + `CompatRouter` still import routers from `react-router-dom` like in:
```
TraceIDSearchInput.test.js
TopNav.test.js
Monitor/index.test.js
SearchTracePage/index.test.js
SearchResults/index.test.js
```

- Unit tests
- manually verify the UI

## Checklist
- [ ] I have read https://github.com/jaegertracing/jaeger/blob/main/CONTRIBUTING_GUIDELINES.md

### DIFF
--- a/packages/jaeger-ui/src/components/Monitor/ServicesView/index.test.js
+++ b/packages/jaeger-ui/src/components/Monitor/ServicesView/index.test.js
@@ -4,7 +4,7 @@
 import React from 'react';
 import { render, screen, cleanup, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
-import { MemoryRouter } from 'react-router-dom';
+import { MemoryRouter } from 'react-router-dom-v5-compat';
 import '@testing-library/jest-dom';
 import {
   MonitorATMServicesViewImpl as MonitorATMServicesView,
--- a/packages/jaeger-ui/src/components/App/Branding.tsx
+++ b/packages/jaeger-ui/src/components/App/Branding.tsx
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import React from 'react';
-import { Link } from 'react-router-dom';
+import { Link } from 'react-router-dom-v5-compat';
 import prefixUrl from '../../utils/prefix-url';
 import JaegerLogo from '../../img/jaeger-logo.svg';
 
--- a/packages/jaeger-ui/src/components/App/NotFound.test.js
+++ b/packages/jaeger-ui/src/components/App/NotFound.test.js
@@ -3,7 +3,7 @@
 
 import React from 'react';
 import { render, screen } from '@testing-library/react';
-import { MemoryRouter } from 'react-router-dom';
+import { MemoryRouter } from 'react-router-dom-v5-compat';
 import '@testing-library/jest-dom';
 import NotFound from './NotFound';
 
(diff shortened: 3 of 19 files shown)


packages/jaeger-ui/src/components/Monitor/ServicesView/index.test.js
```diff
@@ -4,7 +4,7 @@
 import React from 'react';
 import { render, screen, cleanup, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
-import { MemoryRouter } from 'react-router-dom';
+import { MemoryRouter } from 'react-router-dom-v5-compat';
```
yurishkuro, line 7: I am not sure I follow the need for doing this - is there a different behavior in the new MemoryRouter? The final goal is to still stay on react-router-dom v7, is that correct? In which case we'd have to do another sweeping change to move from "-compat" to clean library name?

yurishkuro (approved):

yurishkuro:
looking at the code there are still about a dozen places importing `from 'react-router-dom'`. I am curious if instead of rewriting to import from compat module we could just do a build-level override / substitution like "any import from react-router-dom actually means import from compat". This way we don't have to keep changing the code back and forth.

packages/jaeger-ui/src/components/Monitor/ServicesView/index.test.js
ghost: > I am not sure I follow the need for doing this - is there a different behavior in the new MemoryRouter?

Yes, in practice: `react-router-dom` v5 `MemoryRouter` implement the v5 router (`history`+ v5 context) but `react-router-dom-v5-compat` re-export router pieces built on React router v6+

ghost: > The final goal is to still stay on react-router-dom v7, is that correct? In which case we'd have to do another sweeping change to move from "-compat" to clean library name?

Yes the usual plan is v5 -> v5-compat -> drop compat and history & use v7. And yes when we will do the final step, we will have to do another sweep.

ghost: the “sweep” is like replace v5 test wrappers with v7 ones and fix any mocks that still pointed to v5 or compat.

ghost:
> looking at the code there are still about a dozen places importing `from 'react-router-dom'`. I am curious if instead of rewriting to import from compat module we could just do a build-level override / substitution like "any import from react-router-dom actually means import from compat". This way we don't have to keep changing the code back and forth.

Yes but if we alias all react-router-dom -> compat, Is not it like we are pretending the whole app is on one module?? but root wiring still using v5’s Router from react-router-dom in index.tsx - that’s intentional today. Forcing that through an alias to compat is exactly where maybe different bugs or issues will show up.

ghost:
and we can do the bulk import changes through commands and then will check for any issues (`find` + `sed`)

Issue #3657, opened 2026-03-22
Tracking: ADR 0004 Migrate from Redux to Zustand + TanStack React Query


Parship12:
## Summary

Track the **incremental migration** away from classic **Redux** (`redux`, `redux-actions`, `connect`) toward the architecture described in [ADR 0004: State Management Strategy](https://github.com/jaegertracing/jaeger-ui/blob/main/docs/adr/0004-state-management-strategy.md) and [ADR 0008: Target state management architecture](https://github.com/jaegertracing/jaeger-ui/blob/main/docs/adr/0008-target-state-management-architecture.md).

### Phase 0: Foundations
- [x] #3662 -> Shared TanStack Query client, Zustand + class bridge, Type alignment

### Phase 1: Zustand for client UI state (Redux still holds server data)
- [x] #3702  -> Trace compare (traceDiff)
- [x] #3715  -> Archive notifier (archive)
- [x] #3719 -> layout prefs (trace timeline)
- [x] #3721 -> collapse/expand and detail state (trace timeline)
- [x] #3750 -> DDG view modifiers
- [x] #3761 -> Embedded / Page / TopNav flags

TRIVENI206:
Hi, I’d like to contribute to this migration.

After going through ADR 0004, I understand that the goal is to move away from classic Redux toward a clearer separation of concerns using Zustand for client/UI state and React Query for server state.

I also understand that this migration is intended to be incremental, with small focused PRs that improve maintainability and performance without introducing large changes at once.

I’d be happy to start contributing by picking a small part of the existing Redux-based logic and migrating it in line with this architecture.

Please let me know if there’s a recommended area to begin with.

ghost:
Hi @TRIVENI206, thanks for your interest. At the moment, this migration is not in a state where we can collaborate externally. If we need help at a later stage, i will definitely reach out. However, you can help us with the related PR reviews, and feel free to reach out on Slack anytime if needed.

ghost:
@yurishkuro can you please reopen it?

PR #3662, opened 2026-03-24
Phase 0: feat: add Zustand and query foundation for state migration

ghost:
- Resolves part of #3657

## Description of the changes
- This PR is for some foundational works. No Redux removal and no feature migration yet.
- `src/query/app-query-client.ts`: One place for `QueryClient` instead of defining it inline in `App`.
- `src/utils/zustand-class-bridge.tsx`: avoid blocking Zustand on a large class -> FC rewrite.

- Unit tests

## Checklist


packages/jaeger-ui/src/query/app-query-client.ts
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) 2026 The Jaeger Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+import { QueryClient } from '@tanstack/react-query';
+
+// Default TanStack Query client for Jaeger UI. Centralize the cache timing
+//  so new query hooks stay consistent.
+export function createAppQueryClient(): QueryClient {
+  return new QueryClient({
```
yurishkuro: we are already making queries from useTraceDiscovery, isn't this a repetition of the query object?

ghost: Yes, but we will add many queries. So shared defaults will avoid copy pasting.

packages/jaeger-ui/src/utils/zustand-class-bridge.tsx
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) 2026 The Jaeger Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+import React from 'react';
+import type { StoreApi, UseBoundStore } from 'zustand';
+
+export function createStoreConnector<T, P extends Record<string, unknown>>(
```
yurishkuro, line 35: please write documentation to the purpose of new functions like these

packages/jaeger-ui/src/components/App/index.tsx
```diff
@@ -54,7 +45,7 @@ const PlexusDemoPage = import.meta.env.DEV ? React.lazy(() => import('../PlexusD
 
 export default function JaegerUIApp() {
   return (
-    <QueryClientProvider client={queryClient}>
+    <QueryClientProvider client={appQueryClient}>
```
yurishkuro: this looks like an abstraction leak, why can't the util module simply provide `<AppQueryClientProvider>`? Currently this code needs to know about both tanstack QueryClientProvider AND appQueryClient

ghost: Fair point. So should we add a seperate file for `AppQueryClientProvider` that will wrap `QueryClientProvider` + `appQueryClient` so that `App` doesn’t depend on both Tanstack’s provider and the singleton??

packages/jaeger-ui/src/utils/zustand-class-bridge.tsx
```diff
@@ -0,0 +1,39 @@
...
+      private unsubscribe?: () => void;
+
+      state: { slice: P } = {
+        slice: selector(useStore.getState()),
+      };
+
+      componentDidMount() {
+        this.unsubscribe = useStore.subscribe(nextState => {
+          this.setState({ slice: selector(nextState) });
+        });
```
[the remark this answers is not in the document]
Parship12: I think it's not necessary as it’s just a small performance concern

packages/jaeger-ui/src/utils/zustand-class-bridge.tsx
```diff
@@ -0,0 +1,46 @@
...
```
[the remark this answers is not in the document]
ghost: fixed

packages/jaeger-ui/src/utils/zustand-class-bridge.tsx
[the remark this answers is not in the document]
ghost: fixed

packages/jaeger-ui/src/utils/zustand-class-bridge.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+      componentWillUnmount() {
+        this.unsubscribe?.();
+      }
+
+      render() {
+        const { slice } = this.state;
+        return <Wrapped {...(this.props as Record<string, unknown>)} {...slice} />;
+      }
+    };
```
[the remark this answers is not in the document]
ghost: fixed

packages/jaeger-ui/src/query/query-keys.ts
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2026 The Jaeger Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+export const queryKeys = {
+  services: () => ['services'] as const,
```
yurishkuro: why is it beneficial to extract these keys here, far from where they are used and they are not reusable anywhere else? I think the original place was better, local to the actual query which knows what caching policy is required (and the corresponding TTL for such queries)

packages/jaeger-ui/src/components/App/index.tsx
yurishkuro: I would define the component in the same file as appQueryClient singleton. The singleton itself probably does not need to be exported.

packages/jaeger-ui/src/utils/zustand-class-bridge.tsx
yurishkuro: docs should be next to the symbol.

```suggestion
// createStoreConnector allows to inject Zustand state into legacy class components
// (hooks only work in function components). The connector is a function component
// that uses useStoreWithEqualityFn + shallow so React integrates with the store
// correctly and re-renders only when the selected slice changes.
export function createStoreConnector<T, P extends Record<string, unknown>>(
```

Any usage example? It's hard to judge this function that is not used anywhere.

ghost: There is no usage in app code yet

packages/jaeger-ui/src/components/App/index.tsx
ghost: done

packages/jaeger-ui/src/utils/zustand-class-bridge.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+import React from 'react';
+import { shallow } from 'zustand/shallow';
+import { useStoreWithEqualityFn } from 'zustand/traditional';
+import type { StoreApi, UseBoundStore } from 'zustand';
+
+// createStoreConnector allows to inject Zustand state into legacy class components
+// (hooks only work in function components). The connector is a function component
+// that uses useStoreWithEqualityFn + shallow so React integrates with the store
+// correctly and re-renders only when the selected slice changes.
```
yurishkuro:
```suggestion
/**
 * A Higher-Order Component (HOC) factory that bridges Zustand state to legacy 
 * class components by injecting selected state slices as props.
 *
 * This utility allows incremental migration from Redux to Zustand without 
 * refactoring existing class components into functional components.
 *
 * @template T - The full state type of the Zustand store.
 * @template P - The type of the props being injected (the 'slice').
 *
 * @param boundStore - The Zustand store hook (e.g., `useTraceStore`).
 * @param selector - A function to pick specific state properties and map them to props.
 *
 * @returns An HOC that wraps a component and injects the selected state.
 *
 * @example
 * // 1. Create a specific connector
 * const connect = createStoreConnector(useAppStore, (state) => ({
 *   services: state.services,
 *   loading: state.loading,
 * }));
 *
 * // 2. Apply it to a class component
 * class ServiceList extends React.Component<P & MyOtherProps> { ... }
 * export default connect(ServiceList);
 */
```

packages/jaeger-ui/src/query/query-keys.ts
ghost: I added this because during our migration i think we will add many `useQuery` hooks. So this is to avoid scattering raw tuples like `['services']` everywhere so we don’t get typos or slightly different keys for the same data. Should i revert this change?

packages/jaeger-ui/src/hooks/useTraceDiscovery.ts
```diff
@@ -28,7 +29,7 @@ export function useSpanNames(
   spanKind?: string
 ): UseQueryResult<{ name: string; spanKind: string }[]> {
   return useQuery({
-    queryKey: ['spanNames', service],
+    queryKey: queryKeys.spanNames(service),
     queryFn: () => jaegerClient.fetchSpanNames(service!),
```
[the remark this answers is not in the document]
ghost: not strictly needed given enabled: `!!service`
line 34: `enabled: !!service, // Only fetch when service is selected`

packages/jaeger-ui/src/query/query-keys.ts
yurishkuro: But why would we use raw `useQuery` hooks? `useTraceDiscovery.ts` already exposes domain-specific useServices/useSpanNames functions and encapsulates the knowledge of how the queries are mapped to react-query. I would want to use the same pattern and consolidate the querying capability all in one place, not disperse it throughout the code.

ghost: Ok, got it, reverting this now

yurishkuro (approved):

Issue #3480, opened 2026-01-30
[Bug]: Compare page ui is broken on the mobile device


aaryan359:
### What happened?

I noticed that the Trace Comparison page UI is broken on mobile (layout overflows / content not usable).  
Before starting work, I wanted to check:

- I’m proposing a **CSS-only fix** focused on mobile responsiveness
- No logic changes
- No class → functional component migration
- No refactor of existing components

Given the ongoing functional component migration, would a scoped CSS-only PR for mobile usability be acceptable, or is this area already planned for a rewrite?

Happy to proceed (or hold off) based on your guidance.  
Thanks!

### Steps to reproduce

1. Go to the comparison page
2. Open this page in mobile mode

### Expected behavior

It should be responsive on mobile or smaller screen devices.

### Relevant log output

```shell

```

### Screenshot

<img width="385" height="843" alt="Image" src="https://github.com/user-attachments/assets/44ac0161-61ec-4ba7-b0a0-7e54656bb98f" />

### Deployment configs

```shell

```

aaryan359:
@yurishkuro @jkowall  
I can work on this issue. The website is not responsive on mobile and smaller screens. There are many such issues, and I am interested in working on them as well.

yurishkuro:
I don't think I ever opened Jaeger UI on a mobile screen. If it's a one-line change I don't mind fixing, but if it turns into a whole project then it's not worth it, mobile is not really the target form factor for a tracing UI.

aaryan359:
@yurishkuro Thanks for the clarification! That makes sense.

I’ll take a quick look to see if there’s any **very small, obvious CSS fix** (something truly minimal).  
If it looks like more than that, I’ll drop it and look for other issues to contribute to instead.

swetalin-10:
Hi, I checked the issue and was able to reproduce it on mobile view.

It looks like the layout is not adapting properly due to fixed widths / spacing issues. I’ll focus on small CSS adjustments (like width, alignment, and responsiveness) to improve the layout without making major changes.

I’ll share the approach shortly before opening a PR.

PR #3665, opened 2026-03-26
Add some responsive css for mobile views

lpmi-13:
- Addresses https://github.com/jaegertracing/jaeger-ui/issues/3480

## Description of the changes
- this is intentionally a very minimal targeted change to update the CSS for mobile views.

it used to look like this:

Screenshot from 2026-03-26 11-19-29

and with this commit it looks like this:

Screenshot from 2026-03-26 11-53-31

- just loading locally

## Checklist
- [ ] I have added unit tests for the new functionality (N/A)

- [x] **Heavy**: AI generated most or all of the code changes

### DIFF
--- a/packages/jaeger-ui/src/components/SearchTracePage/index.tsx
+++ b/packages/jaeger-ui/src/components/SearchTracePage/index.tsx
@@ -141,13 +141,13 @@ export function SearchTracePageImpl(props: SearchTracePageImplProps) {
   return (
     <Row className="SearchTracePage--row">
       {!embedded && (
-        <Col xs={24} lg={6} className="SearchTracePage--column">
+        <Col xs={24} sm={6} className="SearchTracePage--column">
           <div className="SearchTracePage--find">
             <Tabs size="large" items={tabItems} />
           </div>
         </Col>
       )}
-      <Col xs={24} lg={!embedded ? 18 : 24} className="SearchTracePage--column">
+      <Col xs={24} sm={!embedded ? 18 : 24} className="SearchTracePage--column">
         {showErrors && (
           <div className="js-test-error-message">
             <h2>There was an error loading traces: </h2>


packages/jaeger-ui/src/components/SearchTracePage/index.tsx
```diff
@@ -141,13 +141,13 @@ export function SearchTracePageImpl(props: SearchTracePageImplProps) {
...
       {!embedded && (
-        <Col span={6} className="SearchTracePage--column">
+        <Col xs={24} lg={6} className="SearchTracePage--column">
           <div className="SearchTracePage--find">
             <Tabs size="large" items={tabItems} />
           </div>
         </Col>
       )}
-      <Col span={!embedded ? 18 : 24} className="SearchTracePage--column">
+      <Col xs={24} lg={!embedded ? 18 : 24} className="SearchTracePage--column">
```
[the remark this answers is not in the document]
lpmi-13: I had considered this but deliberately tried to introduce the smallest change possible. I'd love to also make this change in the PR if it would be acceptable.

packages/jaeger-ui/src/components/SearchTracePage/index.tsx
```diff
@@ -141,13 +141,13 @@ export function SearchTracePageImpl(props: SearchTracePageImplProps) {
   return (
     <Row className="SearchTracePage--row">
       {!embedded && (
-        <Col span={6} className="SearchTracePage--column">
+        <Col xs={24} lg={6} className="SearchTracePage--column">
```
yurishkuro: not sure how exactly the xs/lg get translated into other sizes but with this setting making the window just a tad smaller on the laptop already transforms to the mobile view. Assuming the settings use the rule "explicitly specified size and above" you need to set `{6}` to the next size after `xs`, not to `lg`.

lpmi-13: Oh right...sorry, I thought I had tested that on my own laptop. I'll fix it today.

lpmi-13: css is so confusing, and I always think it's like "xs == this size and below", but that is clearly not the case :laughing:

lpmi-13:
alright, pushed a new commit with the change so that it won't snap to a mobile nav until it hits 575 pixels, which should be enough to not trigger this on laptop screens.

Screenshot from 2026-03-30 13-14-54

Issue #3701, opened 2026-04-03
Tracking: ADR-007 Vite+ toolchain migration


yurishkuro:
This issue tracks the completion of [ADR-0007: Migrate to Vite+ (Full Vite Toolchain)](https://github.com/jaegertracing/jaeger-ui/blob/main/docs/adr/0007-vite-plus-migration.md).

## Summary

Replaced the fragmented Jest + Babel + ESLint + Prettier toolchain with a unified Vite+ stack:

- **Linting**: ESLint → Oxlint (Rust-based, 50–100× faster, no `@typescript-eslint` lag)
- **Formatting**: Prettier → Oxfmt
- **Testing**: Jest + Babel → Vitest (shares the Vite transform pipeline)
- **TypeScript**: upgraded to 6.0.2; `moduleResolution: "bundler"`
- **plexus build**: dropped (package is consumed via workspace source alias only)

## PRs

- [x] #3677 — Drop plexus library build; simplify plexus tsconfig (step A)
- [x] #3680 — Simplify plexus tsconfig to noEmit (step A continued)
- [x] #3681 — Delete legacy ESLint configs (step B)
- [x] #3682 — Add Oxlint in parallel with ESLint (step C1 phase 1)
- [x] #3684 — Replace ESLint with Oxlint (step C1 phase 2)
- [x] #3686 — Replace Prettier with Oxfmt (step C2)
- [x] #3688 — Upgrade TypeScript to 6.0.2 (step D)
- [x] #3689 — Consolidate jaeger-ui tsconfigs (step E)
- [x] #3690 — Migrate packages/plexus tests from Jest to Vitest (step F)
- [x] #3691 — Rename JSX test files from .js to .jsx (step H1)
- [x] #3692 — Replace require() in test bodies with static imports (step H2a)
- [x] #3693 — Replace arrow ctor mocks with regular functions (step H2b)
- [x] #3694 — Add mockDefault helper for default-export mocks (step H2c)
- [x] #3695 — Migrate packages/jaeger-ui tests from Jest to Vitest (step H3)
- [x] #3699 — Remove `__esModule: true` from mock factories (step I1)
- [x] #3700 — Close ADR-007: docs, delete orphaned Jest files (step G)

## Status

✅ **Fully implemented.** All 226 test files pass under Vitest (2603 tests).

jkowall:
Nice job @yurishkuro !

PR #3702, opened 2026-04-03
Part of Phase 1 (1a): refactor: migrate trace diff state from Redux to Zustand

ghost:
- Resolves part of #3657 

## Description of the changes
- New file -> `trace-diff-store.ts` with the same behavior as the `TraceDiff` duck: cohort add/remove, diffSetA / diffSetB, and forceState for URL <-> store sync.
- Removed the traceDiff slice from `configure-store.ts` and from `types/index.ts`
- `TraceDiff.tsx`: reads/writes compare state via Zustand; only `fetchMultipleTraces` stays on Redux dispatch.
- Search page uses the store for cohort actions and derives `diffCohort` from Redux `trace` + Zustand `cohort` (`traceForDiff` + `stateTraceDiffXformer` exported for tests).
- TopNav builds the Compare link from Zustand (`useShallow` on a, b, cohort).
- Tests: new `trace-diff-store.test.ts`, updated `TraceDiff`, `SearchTracePage`, and `TopNav` tests; removed `duck.test.js`.

- Unit tests
- Manually: add traces to compare cohort on Search, open compare, change A/B, reload; nav “Compare” URL should reflect cohort.

## Checklist


packages/jaeger-ui/src/components/SearchTracePage/index.tsx
```diff
@@ -48,7 +49,7 @@ interface ISearchTracePageImplOwnProps {
 // Props from mapStateToProps
 interface IStateProps {
   queryOfResults: IQueryOfResults | null;
-  diffCohort: FetchedTrace[];
+  traceForDiff: ReduxState['trace'];
```
yurishkuro:
```suggestion
  // passed as-is from Redux; cohort lookup happens in the component where Zustand is accessible
  tracesInRedux: ReduxState['trace'];
```

packages/jaeger-ui/src/components/SearchTracePage/index.tsx
```diff
@@ -91,6 +88,15 @@ export function SearchTracePageImpl(props: SearchTracePageImplProps) {
     traces,
   } = props;
 
+  const { cohortAddTrace, cohortRemoveTrace } = useTraceDiffStore(
```
yurishkuro, line 107: I am trying to understand this change, here's Claude explanation
```
  The store starts empty (a: null, b: null, cohort: []). State flows in from the URL via TraceDiff.tsx's syncStates mechanism:

  1. URL is the source of truth. When you navigate to /trace/diff?a=abc&b=def&cohort=abc,def,xyz, the URL parameters are parsed by
  mapStateToProps in TraceDiff.tsx and passed to TraceDiffImpl as props (a, b, cohort).
  2. processProps syncs URL → store. On mount and on prop changes, TraceDiffImpl calls:
  syncStates({ a, b, cohort }, traceDiffState, useTraceDiffStore.getState().forceState);
  2. syncStates compares the URL-derived props against the current store state. If they differ, it calls forceState(...) which writes
  the URL values into the Zustand store.
  3. Store is downstream of the URL. The store is essentially a cached copy of the URL state, kept in sync by TraceDiff while it's
  mounted. Other components (TopNav, SearchTracePage) read from the store as a convenience, rather than each independently parsing the
  URL.

  So the flow is:
  URL params
    → TraceDiff mapStateToProps (Redux)
      → TraceDiffImpl props
        → syncStates → forceState
          → Zustand store
            → TopNav reads cohort/a/b to build Compare link
            → SearchTracePage reads cohort for diff badge

  The design issue you're pointing at is real: TopNav and SearchTracePage are reading from the store, but the store only has correct
  values while TraceDiff is mounted. If you're on the Search page (without TraceDiff rendered), the store reflects whatever was last
  synced — not the current URL. That seems like a bug or at least a design smell.
```

yurishkuro (commented):
Did you write an ADR for the migration? Can we follow the ADR process, similar to what I did in #3701.

ghost:
> Did you write an ADR for the migration? Can we follow the ADR process, similar to what I did in #3701.

Yes, I will update ADR-004 and also update the issue.

packages/jaeger-ui/src/components/SearchTracePage/index.tsx
ghost: On Search: when we click “add to compare” / remove, we write that list straight into the store. So the list is not only updated when the tracediff is open - search updates it itself.

On compare page: the browser URL is what we trust for that view. When that screen loads, we compare URL vs store and, if they differ, we copy the URL into the store so everything matches.

So in normal use: search fills the list from clicks; compare corrects the list from the URL when we open it. TopNav reads the store to build the Compare link - same list search is using. I know having both URL and store can sound like two sources of truth. The idea is: URL wins when we are on compare; while we are on Search, the list in the store is whatever we set with the ui actions, not something that only updates when we open the compare page.

yurishkuro: can we add this as code comment? This is somewhat convoluted behavior which we can probably clean up in the future, but for now let's record the current state.

packages/jaeger-ui/src/components/TraceDiff/TraceDiff.test.jsx
```diff
@@ -87,78 +80,62 @@ describe('TraceDiff', () => {
...
       renderWithRouter(<TraceDiffImpl {...defaultProps} b={newBValue} />);
 
-      expect(forceStateMock).toHaveBeenLastCalledWith({
-        a: defaultProps.a,
-        b: newBValue,
-        cohort: defaultProps.cohort,
-      });
+      expect(useTraceDiffStore.getState().a).toBe(defaultProps.a);
+      expect(useTraceDiffStore.getState().b).toBe(newBValue);
+      expect(useTraceDiffStore.getState().cohort).toEqual(defaultProps.cohort);
```
yurishkuro, line 110: why not just replace the original check?

expect(useTraceDiffStore.getState()).toBe(...)

yurishkuro (approved):

PR #3704, opened 2026-04-04
docs(adr): add state management implementation plan (0008)

ghost:
Add ADR 0008 as the implementation / phased migration companion to [ADR 0004](https://github.com/jaegertracing/jaeger-ui/blob/main/docs/adr/0004-state-management-strategy.md)


yurishkuro:
the research part was valuable, you now just provide decisions / claims without backing. I suggest writing a new ADR that focuses on implementation plan and leaving this one alone as the motivation / reference.

ghost:
@yurishkuro ptal

docs/adr/0008-state-management-implementation-plan.md
```diff
@@ -0,0 +1,171 @@
...
+- **Product context** (large immutable traces, virtualization, open-source ergonomics).
+- **Alternatives considered** (Context vs Redux/RTK vs Zustand + Query) and **out-of-scope** boundaries.
+
+ADR 0008 does **not** repeat that rationale; it only lists **what to do, in what order**, and how to **roll back** or **verify** progress. If a phase conflicts with 0004, **0004 wins** until 0004 is superseded.
+
+---
+
+## TL;DR
+
+Phased migration: **Phase 0** foundations → **Phase 1** move UI slices to Zustand → **Phase 2** move server fetches to TanStack Query → **Phase 3** derived selectors → **Phase 4** remove Redux. Details below; strategic framing remains in [ADR 0004](./0004-state-management-strategy.md).
```
yurishkuro: hard to read, convert to bullet list

docs/adr/0008-state-management-implementation-plan.md
```diff
@@ -0,0 +1,171 @@
...
+**Goal**: Dependencies, shared Query defaults, typing hygiene, and a class-component path to Zustand.
+
+#### ✅ 0a. Shared TanStack Query client
+
+- Implement in `packages/jaeger-ui/src/query/app-query-client.tsx`: `createAppQueryClient()`, a singleton `QueryClient`, and `AppQueryClientProvider` wrapping the app shell from `components/App/index.tsx`. Shared defaults include `staleTime` and `retry` for queries.
+- **Follow-up (optional)**: central **query key** constants or helpers. Discovery hooks still use **inline** `queryKey` arrays in `hooks/useTraceDiscovery.ts` (`['services']`, `['spanNames', service]`); extracting them keeps new hooks consistent.
+
+#### ✅ 0b. Zustand + class bridge
+
+- Add **Zustand**; implement **`createStoreConnector`** for legacy class components that cannot use hooks.
```
yurishkuro: I would like more details here, may be a mermaid diagram of components and data movements.

docs/adr/0008-state-management-implementation-plan.md
```diff
@@ -0,0 +1,171 @@
...
+
+- Remove stale `ReduxState` fields that are not in the real `combineReducers` shape (e.g. legacy `services` slice type if absent from the store).
+
+---
+
+### Phase 1 - Zustand for client UI state (Redux still holds server data)
+
+**Goal**: Move **view / interaction** slices off Redux; keep fetches on Redux or Query until Phase 2.
+
+#### ⬜ 1a. Trace compare (`traceDiff`)
```
yurishkuro: there's very little detail here. What data specifically is moving from Redux to "store"? Just the IDs?

docs/adr/0008-state-management-implementation-plan.md
```diff
@@ -0,0 +1,171 @@
...
+- **Removed** from Redux: `traceDiff` reducer.
+- **Wiring**: `TraceDiff.tsx`, `SearchTracePage` (cohort actions + derived `diffCohort`), `TopNav` (Compare URL).
+
+#### ⬜ 1b. Archive notifier (`archive` duck)
+
+- Zustand (or colocated store) for archive UX; `TracePage` stops dispatching archive actions via Redux.
+
+#### ⬜ 1c. Trace timeline UI (`traceTimeline` duck)
+
+- Largest and most performance-sensitive slice; **split across multiple PRs** (e.g. layout/prefs vs collapse/detail vs `uiFind`-related behavior).
```
yurishkuro: needs more details

docs/adr/0008-state-management-implementation-plan.md
```diff
@@ -0,0 +1,171 @@
...
+- Rewire or replace Redux **tracking middleware** hooks that listen to timeline action types.
+
+#### ⬜ 1d. Deep Dependencies **client-only** modifiers
+
+- Move **view modifier** actions from Redux `ddg` slice into Zustand (or a dedicated small store).
```
yurishkuro: it would be good to have a section that explains how URL state is mapped to UI selections. Is there a common pattern for that?

yurishkuro (commented):
When I look at ADR-4 it has research / comparison / decision, and it has a high level migration strategy. But it does not have the final design (similar to ADR-5 with the current design).

Perhaps your ADR-8 should focus on being the representation of the current state which keeps morphing into future / desired state so that eventually it could serve as a claer architectural document of "this is how we do things". Meanwhile updates regadring migration steps should go back to ADR-4, since it already has that but the steps could be expanded with more details that you tried adding here.

ghost:
> Perhaps your ADR-8 should focus on being the representation of the current state which keeps morphing into future / desired state so that eventually it could serve as a claer architectural document of "this is how we do things". Meanwhile updates regadring migration steps should go back to ADR-4, since it already has that but the steps could be expanded with more details that you tried adding here.

Ok, I will update it.

docs/adr/0008-target-state-management-architecture.md
```diff
@@ -0,0 +1,216 @@
...
+| :--- | :--- |
+| **[ADR 0004](./0004-state-management-strategy.md)** | **Decision and migration**: why we chose Zustand + TanStack Query, alternatives considered, and the **phased migration checklist** (Phase 0–4) with rollback and testing notes. |
+| **[ADR 0005](./0005-current-state-management-architecture.md)** | **Transition snapshot**: how Redux, Query, URL, and local state interact **today** while migration is incomplete. |
+| **This ADR (0008)** | **Target architecture**: where each kind of state **should** live when migration is complete, and how data flows between layers. Update this document as the codebase catches up so it stays the canonical "how we do things" reference. |
+
+If this ADR conflicts with **0004** on strategic direction, **0004 wins** until it is superseded. Execution order and checkmarks belong in **0004 → Migration Path**, not here.
+
+---
+
+## TL;DR
```
yurishkuro, line 20: Love the table! There's just one thing that gives me pause - why have a split brain situation between Zustand state and URL state? Is this a transient migration artifact or long term state? Why not consolidate everything in the URL state, which makes shared URLs much more descriptive / usable?

ghost: let's imagine 2 types of things:
1. which span is my mouse hovering over right now? as we move our mouse across the trace timeline, this changes...say 60 times per second. If we try to save that in the URL, Will not the URL would be changing 60 times per second? browsers would crash.

2. Which of the 10,000 spans did I manually expand? A span id looks like: a3f2b1c4d5e6f7a8. If we expand 200 spans, we have to stuff 200 of those into the URL. The URL would become thousands of characters long.

yurishkuro (approved):

PR #3726, opened 2026-04-12
fix(scripts): handle missing binaries in prepare-release

jkowall:
## Summary
- catch `FileNotFoundError` in `scripts/prepare-release.py` after the move to `subprocess.run(..., shell=False)`
- preserve the script's normal error reporting for missing tools such as `git`, `make`, and `npm`
- keep sensitive-command output suppressed as before

## Why this fix is needed
The earlier hardening changed command execution from shell strings to argv lists with `shell=False`, which is the right security fix. That change also altered the failure mode for missing executables: Python now raises `FileNotFoundError` instead of `CalledProcessError`.

`run_command()` only handled `CalledProcessError`, so missing binaries started surfacing as uncaught tracebacks instead of the script's normal user-facing error output. This change restores that behavior without undoing the security improvement.

## Validation
- `npm run fmt`
- `npm run lint`
- `npm run build`
- `npm test` still reports pre-existing unrelated failures in `packages/plexus`

Related: https://github.com/jkowall/jaeger-ui-F/pull/26

### DIFF
--- a/scripts/prepare-release.py
+++ b/scripts/prepare-release.py
@@ -10,37 +10,48 @@ import tempfile
 import time
 from datetime import date
 
-def run_command(command, cwd=None, capture_stdout=True, capture_stderr=True):
+def run_command(command, cwd=None, capture_stdout=True, capture_stderr=True, sensitive=False):
+    if isinstance(command, str):
+        raise TypeError(f"command must be a list of arguments, not a string: {command!r}")
     try:
         result = subprocess.run(
             command,
             cwd=cwd,
             check=True,
-            shell=True,
+            shell=False,
             text=True,
             stdout=subprocess.PIPE if capture_stdout else None,
             stderr=subprocess.PIPE if capture_stderr else None
         )
         return result.stdout.strip() if capture_stdout and result.stdout else ""
     except subprocess.CalledProcessError as e:
-        print(f"Error running command: {command}")
-        if capture_stdout and e.stdout:
-            print(f"STDOUT: {e.stdout}")
-        if capture_stderr and e.stderr:
-            print(f"STDERR: {e.stderr}")
+        if sensitive:
+            print("Error running sensitive command (output suppressed)")
+        else:
+            print(f"Error running command: {command}")
+            if capture_stdout and e.stdout:
+                print(f"STDOUT: {e.stdout}")
+            if capture_stderr and e.stderr:
+                print(f"STDERR: {e.stderr}")
+        raise
+    except FileNotFoundError:
+        if sensitive:
+            print("Error: Sensitive command binary not found (output suppressed)")
+        else:
+            print(f"Error: Command binary not found for: {command}")
         raise
 
 def check_dependencies():
     try:
-        run_command("gh --version")
-    except:
+        run_command(["gh", "--version"])
+    except (subprocess.CalledProcessError, FileNotFoundError):
         print("Error: 'gh' CLI is not installed or not in PATH.")
         sys.exit(1)
 
 def get_gh_token():
     try:
-        return run_command("gh auth token")
-    except:
+        return run_command(["gh", "auth", "token"], sensitive=True)
+    except (subprocess.CalledProcessError, FileNotFoundError):
         print("Error: Could not retrieve GitHub token using 'gh auth token'. Please login via 'gh auth login'.")
         sys.exit(1)
 
@@ -50,7 +61,7 @@ def validate_version(version):
         sys.exit(1)
 
 def check_git_status():
-    status = run_command("git status --porcelain")
+    status = run_command(["git", "status", "--porcelain"])
     if status:
         print("Error: Git working directory is not clean. Please commit or stash changes.")
         sys.exit(1)
@@ -61,14 +72,14 @@ def create_branch(version, dry_run=False):
         print(f"[Dry Run] Would create branch {branch_name}")
     else:
         print(f"Creating branch {branch_name}...")
-        run_command(f"git checkout -b {branch_name}")
+        run_command(["git", "checkout", "-b", branch_name])
     return branch_name
 
 def generate_release_notes():
     print("Generating release notes via 'make changelog'...")
     # Run make -s (silent) to suppress echoing commands, capturing only the script output
     # Stream stderr (capture_stderr=False) to show progress bars from release-notes.py
-    return run_command("make -s changelog", capture_stderr=False)
+    return run_command(["make", "-s", "changelog"], capture_stderr=False)
 
 def update_changelog(version, notes, dry_run=False):
     print("Updating CHANGELOG.md...")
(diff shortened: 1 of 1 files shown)


yurishkuro:
merge history is messed up, lots of unrelated changes

jkowall:
Sorry I had this issue before and fixed it, but apparently the wrong changes were pushed via the PR. Fixed now.

yurishkuro:
merge conflicts

jkowall:
> merge conflicts

resolved now.

scripts/prepare-release.py
```diff
@@ -10,37 +10,48 @@
...
+            if capture_stdout and e.stdout:
+                print(f"STDOUT: {e.stdout}")
+            if capture_stderr and e.stderr:
+                print(f"STDERR: {e.stderr}")
+        raise
+    except FileNotFoundError:
+        if sensitive:
+            print("Error: Sensitive command binary not found (output suppressed)")
+        else:
+            print(f"Error: Command binary not found for: {command}")
```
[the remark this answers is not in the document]
jkowall: Addressed in `b1502c62`. `run_command()` now captures `FileNotFoundError as e`, uses `e.filename`, and distinguishes a missing working directory from other missing paths while keeping the sensitive-output behavior intact.

scripts/prepare-release.py
```diff
@@ -10,37 +10,48 @@
...
-        if capture_stderr and e.stderr:
-            print(f"STDERR: {e.stderr}")
+        if sensitive:
+            print("Error running sensitive command (output suppressed)")
+        else:
+            print(f"Error running command: {command}")
+            if capture_stdout and e.stdout:
+                print(f"STDOUT: {e.stdout}")
+            if capture_stderr and e.stderr:
+                print(f"STDERR: {e.stderr}")
```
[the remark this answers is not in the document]
jkowall: Addressed in `b1502c62`. I added `format_command_for_display()` with `shlex.join(...)` and now use that formatted command string in the error messages while still executing with `shell=False`.

scripts/prepare-release.py
```diff
@@ -27,12 +33,42 @@ def run_command(command, cwd=None, capture_stdout=True, capture_stderr=True, sen
...
+                    'not found (output suppressed)'
+                )
+        else:
+            if cwd_str and missing_path == cwd_str:
+                print(f"Error: Working directory not found for command {formatted_command}: {cwd_str}")
+            elif missing_path:
+                print(
+                    'Error: Required path not found while running command '
+                    f'{formatted_command}: {missing_path}'
+                )
```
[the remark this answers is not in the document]
jkowall: Addressed in `7f99db30`. The `FileNotFoundError` handling now detects `missing_path == str(command[0])` and reports that case as an executable missing from PATH, while keeping the working-directory case separate.

scripts/prepare-release.py
```diff
@@ -27,12 +33,42 @@ def run_command(command, cwd=None, capture_stdout=True, capture_stderr=True, sen
...
+            else:
+                print(f"Error: Required path not found while running command: {formatted_command}")
+        raise
```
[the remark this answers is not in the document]
jkowall: Addressed in `7f99db30`. I added a `print_errors` flag to `run_command()` and disabled internal printing in `check_dependencies()` and `check_gh_auth()`, so those callers can emit their own single user-facing message without duplicate output.

scripts/prepare-release.py
```diff
@@ -27,16 +41,60 @@ def run_command(command, cwd=None, capture_stdout=True, capture_stderr=True, sen
...
+                print(f"STDOUT: {e.stdout}")
+            if not sensitive and capture_stderr and e.stderr:
+                print(f"STDERR: {e.stderr}")
+        raise
+    except FileNotFoundError as e:
+        missing_path = getattr(e, 'filename', None)
+        cwd_str = str(cwd) if cwd is not None else None
+        is_missing_cwd = cwd_str is not None and missing_path == cwd_str
+        is_missing_executable = executable is not None and missing_path == executable
+        if print_errors:
```
yurishkuro: I think this is a massive overkill. The prepare-release is run locally - even if it prints a GH token, big deal. But having this 50 LOC handing of printing an error is completely unnecessary for a helper script.

jkowall: I agree with you on that one. I will simplify it, but I do think that this is still something we should address, because it can be a vector that is used to compromise your local machine, which can then be used to change the code in the repository. It should be hardened a little bit more.

jkowall: Agreed. I simplified this down in `576880f9` to keep the context proportional to a local helper script: dropped the path/PATH classification and command-formatting helpers, kept only the small `FileNotFoundError` path plus a top-level exit to avoid raw tracebacks, and left the `shell=False` command execution intact.

scripts/prepare-release.py
```diff
@@ -178,4 +191,7 @@ def main():
         git_commit_and_pr(version, branch_name)
 
 if __name__ == "__main__":
-    main()
+    try:
```
yurishkuro, line 194: any uncaught exception from main would result in a stacktrace and exit(1) - why do you need to catch them?

scripts/prepare-release.py
```diff
@@ -47,7 +59,8 @@ def check_gh_auth():
             ["gh", "auth", "status"],
             capture_stdout=False,
             capture_stderr=False,
-            sensitive=True
+            sensitive=True,
```
yurishkuro, line 62: what is sensitive about `auth status`? It obfuscates the token
```
$ gh auth status
github.com
  ✓ Logged in to github.com account yurishkuro (keyring)
  - Active account: true
  - Git operations protocol: ssh
  - Token: gho_************************************
  - Token scopes: 'gist', 'notifications', 'read:audit_log', 'read:discussion', 'read:enterprise', 'read:gpg_key', 'read:org', 'read:project', 'read:public_key', 'read:repo_hook', 'read:ssh_signing_key', 'read:user', 'repo'
```

yurishkuro (commented):
Sorry I don't get the point of the PR. The first check we do is `gh --version` which was already catching `FileNotFoundError`:
```
except (subprocess.CalledProcessError, FileNotFoundError):
```
So the reason for change stated in the description is not valid - we would've already caught the missing `gh` binary. Meanwhile all other changes are not even needed because the whole point of using `gh` binary was that it handles authentication automatically and securely without requiring us to do anything like reading files or env vars for tokens.

jkowall:
> Sorry I don't get the point of the PR.

The main point is that I can easily inject malicious code into this script via various vectors, which would essentially compromise the release. I believe there are valid areas where we could harden the script, but I will close out the PR.

yurishkuro:
When you say "I can inject" are you referring to non-maintainer injecting and maintainer unwillingly executing? 'Cause otherwise it's a maintainer doing it to themselves.

jkowall:
With #3717, we are better covered from most aspects; however, since we are implementing cross-repo inclusion without pinning, that's still not a best practice, but it's on paper and not realistic for our current situation.

PR #3329, opened 2026-01-07
[otel migration] Phase 3.1: add JaegerClient v3 and use for services / operations

yurishkuro:
Part of #3265

# Task List: Milestone 3.1 - Trace Discovery Integration

## Infrastructure & APIs
- [x] Setup QueryClient and QueryClientProvider in App 
- [x] Add unit tests for `JaegerClient` class
- [x] Create `JaegerClient` class (Renamed from TraceDiscoveryClient/OtlpApiClient) 
- [x] **Address Feedback (JaegerClient Refactor)**
    - [x] Rename `OtlpApiClient` to `JaegerClient`
    - [x] Drop `Otlp` prefix from API types
    - [x] Update `fetchSpanNames` to return `{ name: string; spanKind: string }[]`
    - [x] Rename `useOtlpMetadata.ts` to `useTraceDiscovery.ts` and update hook names

## Search Form & Page
- [x] Integrate Trace Discovery Hooks into `SearchForm`
    - [x] Initial integration
    - [x] Handle `{ name, spanKind }` in Span Names dropdown (via wrapper projection)
- [x] Optimize Loading UX
    - [x] Remove top-level blocking indicator in `SearchTracePage`
    - [x] Implement granular loading state in `SearchFormImpl`
- [x] **Clean up Redux Redundancy**
    - [x] Remove `mapStateToProps` service mapping in `SearchTracePage`
    - [x] Remove `fetchServices` calls in `SearchTracePage`

## Testing & Verification
- [x] Fix failing tests in `SearchForm.test.js`
- [x] Fix failing tests in `SearchTracePage/index.test.js`
    - [x] Add `QueryClientProvider` to test context
    - [x] Remove legacy Redux-based state/action assertions
    - [x] Restore behavioural tests: Verify `jaegerClient` discovery is triggered
    - [x] Restore behavioural tests: Verify granular loading state UX
- [x] Fix TypeScript error TS2554 in `SearchFormWithOtlpMetadata.tsx`
- [x] Verify all 73 unit tests in affected files pass

## Documentation
- [x] Update ADR 0005 for `useConfig`
- [x] Update ADR 0005 terminology to "Trace Discovery"

PR #3734, opened 2026-04-15
fix(ui): sort services alphabetically #3733

bender316:
## Which problem is this PR solving?
fixes #3733

## Description of the changes
- sort services in search form alphabetically

## How was this change tested?
- running jaeger UI locally against development storage and verify sorting
- added unit tests

## Checklist
- [x] I have read https://github.com/jaegertracing/jaeger/blob/main/CONTRIBUTING_GUIDELINES.md
- [x] I have signed all commits
- [x] I have added unit tests for the new functionality
- [x] I have run lint and test steps successfully: `make lint test`

## AI Usage in this PR (choose one)
See [AI Usage Policy](https://github.com/jaegertracing/jaeger/blob/main/CONTRIBUTING_GUIDELINES.md#ai-usage-policy).
- [ ] **None**: No AI tools were used in creating this PR
- [ ] **Light**: AI provided minor assistance (formatting, simple suggestions)
- [x] **Moderate**: AI helped with code generation or debugging specific parts
- [ ] **Heavy**: AI generated most or all of the code changes

PR #3740, opened 2026-04-16
fix(SearchForm): sort services alphabetically in the dropdown

### LINKED ISSUE #3733: [Bug]: Services are no longer sorted alphabetically
### What happened?

I guess due to the refactoring in #3329 (and related changes) the services in the search form are no longer sorted alphabetically.

### Steps to reproduce

1. Have multiple services
2. List is sorted after order in storage (e.g. time of creation)

### Expected behavior

I think it's best to have them still sorted alphabetically. Despite have a filter for the list you might not always know what you're looking for.

### Relevant log output

```shell

```

### Screenshot

_No response_

### Additional context

_No response_

### Jaeger backend version

_No response_

### SDK

_No response_

### Pipeline

_No response_

### Stogage backend

_No response_

### Operating system

_No response_

### Deployment model

_No response_

### Deployment configs

```shell

```

tmchow:
## Summary

Restore alphabetical ordering for services across the UI. After the v3 client migration in #3329, `useServices()` returns them in the backend's storage order, which is typically creation time rather than alphabetical.

## Approach

Sort inside the `useServices()` hook via React Query's `select` transform, so every consumer gets the sorted list. `SearchForm`, `Monitor/ServicesView`, `DeepDependencies`, and `QualityMetrics` all benefit from one fix.

The sort uses a case-insensitive `localeCompare` with `sensitivity: 'base'` so mixed-case service names group the way users expect. The transform returns a copy (`[...data].sort(...)`) so React Query's cached array is not mutated.

### Note on `Monitor/ServicesView`

`ServicesView` picks `services[0]` as the default selection when no `lastAtmSearchService` is saved. With alphabetical sort applied inside the hook, that default now lands on the alphabetically-first service instead of whatever the backend happened to return first. This is the behavior requested in review but worth flagging explicitly.

## Test plan

```
cd packages/jaeger-ui
npx vitest run src/hooks/useTraceDiscovery.test.ts src/components/SearchTracePage/SearchForm.test.jsx
# 77 tests pass

npx vitest run src/components/Monitor/ src/components/QualityMetrics/ src/components/DeepDependencies/
# 400 tests pass
```

Fixes #3733

This contribution was developed with AI assistance (Claude Code).

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)

### DIFF
--- a/packages/jaeger-ui/src/components/SearchTracePage/SearchForm.tsx
+++ b/packages/jaeger-ui/src/components/SearchTracePage/SearchForm.tsx
@@ -366,8 +366,15 @@ export const SearchFormImpl: React.FC<ISearchFormImplProps> = ({
     resultsLimit: initialValues?.resultsLimit,
   }));
 
-  // Fetch services using React Query
-  const { data: services = [], isLoading: isLoadingServices, error: servicesError } = useServices();
+  // Fetch services using React Query. The API may return them in storage
+  // order, which the search form has historically displayed alphabetically
+  // (issue #3733). Sort here rather than in useServices so other consumers
+  // (Monitor, DeepDependencies, QualityMetrics) keep the API-provided order.
+  const { data: rawServices = [], isLoading: isLoadingServices, error: servicesError } = useServices();
+  const services = useMemo(
+    () => [...rawServices].sort((a, b) => a.localeCompare(b, undefined, { sensitivity: 'base' })),
+    [rawServices]
+  );
 
   // Fetch span names for the currently selected service
   const currentService = formData.service;
--- a/packages/jaeger-ui/src/components/SearchTracePage/SearchForm.tsx
+++ b/packages/jaeger-ui/src/components/SearchTracePage/SearchForm.tsx
@@ -366,15 +366,8 @@ export const SearchFormImpl: React.FC<ISearchFormImplProps> = ({
     resultsLimit: initialValues?.resultsLimit,
   }));
 
-  // Fetch services using React Query. The API may return them in storage
-  // order, which the search form has historically displayed alphabetically
-  // (issue #3733). Sort here rather than in useServices so other consumers
-  // (Monitor, DeepDependencies, QualityMetrics) keep the API-provided order.
-  const { data: rawServices = [], isLoading: isLoadingServices, error: servicesError } = useServices();
-  const services = useMemo(
-    () => [...rawServices].sort((a, b) => a.localeCompare(b, undefined, { sensitivity: 'base' })),
-    [rawServices]
-  );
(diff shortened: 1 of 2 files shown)


packages/jaeger-ui/src/components/SearchTracePage/SearchForm.tsx
```diff
@@ -366,8 +366,15 @@ export const SearchFormImpl: React.FC<ISearchFormImplProps> = ({
     resultsLimit: initialValues?.resultsLimit,
   }));
 
-  // Fetch services using React Query
-  const { data: services = [], isLoading: isLoadingServices, error: servicesError } = useServices();
+  // Fetch services using React Query. The API may return them in storage
+  // order, which the search form has historically displayed alphabetically
+  // (issue #3733). Sort here rather than in useServices so other consumers
+  // (Monitor, DeepDependencies, QualityMetrics) keep the API-provided order.
+  const { data: rawServices = [], isLoading: isLoadingServices, error: servicesError } = useServices();
```
yurishkuro: the fix should be in useServices() so that it works for all components

packages/jaeger-ui/src/hooks/useTraceDiscovery.ts
```diff
@@ -14,6 +17,7 @@ export function useServices(): UseQueryResult<string[]> {
     queryFn: () => jaegerClient.fetchServices(),
     staleTime: 60 * 1000, // 1 minute
     refetchOnWindowFocus: true,
+    select: data => [...data].sort((a, b) => a.localeCompare(b, undefined, { sensitivity: 'base' })),
```
yurishkuro, line 20: Define a function at module level to avoid memoization issues with lambda function.

tmchow:
Thanks @yurishkuro. Moved the sort into `useServices()` via React Query's `select` transform - pushed in the latest commit. Same case-insensitive `localeCompare` logic, just applied once in the hook so `Monitor`, `DeepDependencies`, `QualityMetrics`, and `SearchForm` all get the sorted list.

One behavior change worth flagging: `Monitor/ServicesView` uses `services[0]` as the default metrics selection when no `lastAtmSearchService` is saved. That default now lands on the alphabetically-first service instead of the backend's first response. Assuming that's the outcome you want; happy to revert if not.

Also addressed @Copilot's note - dropped the local `useMemo` on `rawServices` in `SearchForm` (the select transform returns a stable array from React Query's cache so it's not needed anymore).

All tests pass: 77 in the hook/SearchForm suites, 400 in Monitor/QualityMetrics/DeepDependencies.

yurishkuro:
I merged another PR #3734

Issue #3746, opened 2026-04-17
[chore]: Fix linter warnings


yurishkuro:
### Requirement

There are approximately these oxlint warnings:
```
  ┌───────┬──────────────────────────────────────────────┬───────────────────────────────────────────┐
  │ Count │                     Rule                     │                 Category                  │
  ├───────┼──────────────────────────────────────────────┼───────────────────────────────────────────┤
  │   228 │ typescript-eslint(no-explicit-any)           │ TypeScript — any type usage               │
  ├───────┼──────────────────────────────────────────────┼───────────────────────────────────────────┤
  │    37 │ eslint(no-unused-vars)                       │ Unused imports/variables                  │
  ├───────┼──────────────────────────────────────────────┼───────────────────────────────────────────┤
  │    29 │ eslint-plugin-jest(no-conditional-expect)    │ Test quality — conditional assertions     │
  ├───────┼──────────────────────────────────────────────┼───────────────────────────────────────────┤
  │    26 │ eslint-plugin-jest(expect-expect)            │ Test quality — tests with no assertions   │
  ├───────┼──────────────────────────────────────────────┼───────────────────────────────────────────┤
  │    25 │ eslint-plugin-jest(require-to-throw-message) │ Test quality — .toThrow() with no message │
  ├───────┼──────────────────────────────────────────────┼───────────────────────────────────────────┤
  │    18 │ eslint-plugin-jest(valid-title)              │ Test quality — invalid describe/it titles │
  ├───────┼──────────────────────────────────────────────┼───────────────────────────────────────────┤
  │    13 │ eslint-plugin-unicorn(no-new-array)          │ Code style — new Array(n) ambiguity       │
  ├───────┼──────────────────────────────────────────────┼───────────────────────────────────────────┤
  │     9 │ eslint(no-shadow)                            │ Variable shadowing                        │
  ├───────┼──────────────────────────────────────────────┼───────────────────────────────────────────┤
  │     9 │ eslint-plugin-jest(valid-expect)             │ Test quality — misused expect()           │
  ├───────┼──────────────────────────────────────────────┼───────────────────────────────────────────┤
  │     5 │ eslint(no-use-before-define)                 │ Usage before declaration                  │
  ├───────┼──────────────────────────────────────────────┼───────────────────────────────────────────┤
  │     4 │ eslint(no-redeclare)                         │ Duplicate declarations                    │
[...]

PR #3760, opened 2026-04-20
fix(lint): Fix no-use-before-define warnings, promote to error in TS

mateenali66:
reorders declarations in four files so each identifier is defined before it's used, then promotes no-use-before-define from warn to error in the TS override.

- UiFindInput.tsx: parseUiFind was defined after the component that calls it. moved the utility functions and related type above the component.
- transformTracesToPaths.ts: walkPaths called processPath before it was declared. swapped the two arrow-function consts.
- SearchTracePage/index.tsx: the component used stateTraceDiffXformer before the export const. moved the memoized selector and its interface above the component.
- vitest-setup.ts: the MessageChannel mock had port1.postMessage referencing port2 inside its closure before port2 was declared. initialized both port shells first, then attached postMessage.

all behavior is identical. closures still capture correctly, the memoized selector is still exported with the same name, tests pass unchanged.

Refs #3746

### DIFF
--- a/packages/jaeger-ui/src/components/SearchTracePage/index.tsx
+++ b/packages/jaeger-ui/src/components/SearchTracePage/index.tsx
@@ -70,6 +70,18 @@ interface IDispatchProps {
 
 type SearchTracePageImplProps = ISearchTracePageImplOwnProps & IStateProps & IDispatchProps;
 
+interface IStateTraceDiff {
+  cohort: string[];
+}
+
+export const stateTraceDiffXformer = memoizeOne(
+  (stateTrace: ReduxState['trace'], stateTraceDiff: IStateTraceDiff) => {
+    const { traces } = stateTrace;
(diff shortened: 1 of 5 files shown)


mateenali66:
addressed the Copilot inline comment — updated the TS override comment to note that no-use-before-define is intentionally kept as error while the rest are downgraded to warn.

yurishkuro (changes requested):
https://github.com/jaegertracing/jaeger-ui/pull/3755#issuecomment-4276329644

mateenali66:
closing this — didn't notice #3755 was already up with the same scope and files when I claimed this rule on the tracking issue. @yurishkuro's comment on #3755 is also the right direction: disable the rule rather than enforce ordering, since tsc catches the real cases. deferring to @Pasta-coder.

Pasta-coder:
looking into it , thanks.

PR #3761, opened 2026-04-20
Part of Phase 1 (1e): migrate embed flags off Redux

ghost:
- Resolves part of #3657

## Description of the changes
- Implements ADR 0004 Phase 1e (embedded UI flags)
- Problem: embedded was a Redux slice with no actions, only an initializer reading `window.location.search` once.
- Added `packages/jaeger-ui/src/stores/embedded-store.ts`
- Removed: `reducers/embedded.ts`, its test, and embedded from `combineReducers` and `ReduxState`.
- Tests: `embedded-store.test.ts`; `Page.test.jsx`, `SearchTracePage/index.test.jsx`, `TracePage/index.test.jsx` updated.

- Unit tests
- Use a single `?` then `&uiEmbed=v0&...` (a second ? breaks parsing), e.g. `/search?...&uiEmbed=v0&uiSearchHideGraph=1`.
- load a DDG graph, hover nodes/edges, change service

## Checklist


packages/jaeger-ui/src/components/App/Page.tsx
```diff
@@ -18,13 +17,15 @@ import withRouteProps from '../../utils/withRouteProps';
 
 type TProps = {
   children: React.ReactNode;
-  embedded: EmbeddedState;
+  embedded?: EmbeddedState | null;
```
yurishkuro, line 20: so null and undefined have different meaning here?

packages/jaeger-ui/src/stores/embedded-store.ts
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) 2026 The Jaeger Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+// URL-driven embedded UI flags (`?uiEmbed=v0&...`). Parsed once per page load - same
+// behavior as the former Redux `embedded` reducer (no actions, no updates).
+
+import { EmbeddedState } from '../types/embedded';
+import { getEmbeddedState } from '../utils/embedded-url';
+
+let cached: EmbeddedState | null | undefined;
```
yurishkuro, line 10: what is the semantic difference between null and undefined? Can we use just `?EmbeddedState`?

packages/jaeger-ui/src/components/App/Page.tsx
Parship12: we only distinguish “prop not passed” vs “prop passed as null”:
- embedded omitted (undefined) -> read flags from the URL
- embedded={null} -> explicit “not embedded” for this render(mostly for tests), without depend on the URL.

So null is not the same as “missing”: missing means “use URL”; null means “override to off” That’s why `EmbeddedState?` alone isn’t enough... we need a way to force non-embedded when the URL might still carry `uiEmbed=v0`

packages/jaeger-ui/src/stores/embedded-store.ts
Parship12: undefined -> cache not initialize yet (first call).
null -> initialized; URL says we are not in `uiEmbed=v0`
EmbeddedState -> initialized; embedded mode on.

packages/jaeger-ui/src/components/App/Page.tsx
Parship12: so here null is just for tests

yurishkuro: Seems overly convoluted. (a) I am not a fan of relaxing domain types because tests are too lazy to initialize a value as needed. (b) why does using URL depend on what is passed via props? Especially on the Page component, when would anyone ever pass EmbeddedState via props and not via URL?

Parship12: ok so i think we can refactor embed mode so Page / SearchTracePage / TracePage no longer accept an embedded prop, they always read embed state from the URL with `getEmbeddedFromUrl()`, and tests use the real query string or a mock of that helper?

packages/jaeger-ui/src/stores/embedded-store.ts
Parship12: Here we still need three ideas
we cannot merge those three meanings into `EmbeddedState?` alone without losing information. We can make the code look clear may be by creating a union.

packages/jaeger-ui/src/components/App/Page.tsx
Parship12: > when would anyone ever pass EmbeddedState via props and not via URL?

yes, this prop existed just for tests, not for a real user path

ghost:
@yurishkuro ptal

packages/jaeger-ui/src/components/App/Page.tsx
yurishkuro: I consulted Claude and it agreed with me:

## Problem

The PR introduces `embedded?: EmbeddedState | null` in `Page`, `SearchTracePage`, and `TracePage`. This creates a three-state pattern:

- `undefined` — prop not passed; component falls back to reading the URL via `getEmbeddedFromUrl()`
- `null` — explicitly not embedded
- `EmbeddedState` — embedded with config

The intent is to support dual-source reading (props or URL), motivated by testability (inject via props in tests). But deciding *where to get state from* is not a component's responsibility.

## Recommendation

The PR already introduces `stores/embedded-store.ts` with `getEmbeddedFromUrl()` as a cached singleton — no additional Zustand store wrapper is needed. The fix is simply to remove `embedded` from all component prop types and always call `getEmbeddedFromUrl()` directly inside the component body.

The PR is already doing this on the `undefined` branch:

```ts
const embedded = props.embedded === undefined ? getEmbeddedFromUrl() : props.embedded;
```

Remove the conditional entirely — just call `getEmbeddedFromUrl()` unconditionally:

```ts
const embedded = getEmbeddedFromUrl();
```

- Drop `embedded` from `TProps` in `Page`, `IStateProps` in `SearchTracePage`, and `TProps` in `TracePage`
- Remove it from `mapStateToProps` in `SearchTracePage` and `TracePage` (already done in the PR)
- The three-state trick disappears because components no longer accept `embedded` as a prop at all
- Tests use `resetEmbeddedFromUrlCacheForTests()` (already in the PR) to control the URL-parsed value, rather than injecting via props

packages/jaeger-ui/src/stores/embedded-store.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export function getEmbeddedFromUrl(): EmbeddedState | null {
```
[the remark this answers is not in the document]
ghost: @yurishkuro what do you think we should do here?

packages/jaeger-ui/src/stores/embedded-store.ts
```diff
@@ -0,0 +1,28 @@
...
+  }
+  const search = window.location.search;
+  cached = search ? getEmbeddedState(search) : null;
+  return cached;
+}
+
+// Resets the lazy cache (Vitest only).
+export function resetEmbeddedFromUrlCacheForTests(): void {
+  cached = undefined;
+}
```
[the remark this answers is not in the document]
ghost: Not required. It's for tests, already mentioned in the name

packages/jaeger-ui/src/components/App/Page.tsx
```diff
@@ -4,27 +4,25 @@
...
-  embedded: EmbeddedState;
 };
 
 const { Header, Content } = Layout;
 
 // export for tests
-export const PageImpl: React.FC<TProps> = ({ children, embedded }) => {
+export const PageImpl: React.FC<TProps> = props => {
+  const embedded = getEmbeddedFromUrl();
+  const { children } = props;
```
yurishkuro:
```suggestion
  const { children } = props;
  const embedded = getEmbeddedFromUrl();
```

packages/jaeger-ui/src/components/App/Page.tsx
```diff
@@ -4,27 +4,25 @@
...
-  embedded: EmbeddedState;
 };
 
 const { Header, Content } = Layout;
 
 // export for tests
-export const PageImpl: React.FC<TProps> = ({ children, embedded }) => {
+export const PageImpl: React.FC<TProps> = props => {
+  const { children } = props;
+  const embedded = getEmbeddedFromUrl();
```
yurishkuro: the naming is unconventional, should it be something like `useEmbeddedState()`

packages/jaeger-ui/src/stores/embedded-store.ts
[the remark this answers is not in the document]
yurishkuro: I had the same comment above - why not call this useEmbeddedState()?  Also #3780

yurishkuro (approved):

PR #3768, opened 2026-04-21
fix: prevent timeline tick labels from overlapping when column is narrow

### LINKED ISSUE #3767: [Bug]: Timeline view time scale labels are overlapping
When timeline column is narrowed the tick labels can start overlapping

Image

aabhinavvvvvvv:
- Resolves #3767

## Description of the changes
- Added a `ResizeObserver` in the `Ticks` component to measure the timeline column width on mount and resize
- Intermediate tick labels are skipped using a power-of-2 step when they would overlap, so remaining labels stay evenly spaced
- The first and last labels are always shown regardless of column width

- Added unit tests for the `Ticks` component covering label rendering, `isEndAnchor` class, and no-label case
- Manually verified in the browser by narrowing the timeline column (screenshots below)

## Checklist

## Before
Screenshot 2026-04-21 130622

## After
Screenshot 2026-04-21 130759

Screenshot 2026-04-21 130831

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/Ticks.test.jsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/Ticks.test.jsx
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import React from 'react';
-import { render } from '@testing-library/react';
+import { render, act } from '@testing-library/react';
 
 import Ticks from './Ticks';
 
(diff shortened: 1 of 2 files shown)


yurishkuro:
I think if you are removing a label you should also remove the vertical guideline corresponding to it, otherwise the display looks confusing and in your last screenshot it's actually hard to say which line the 65us actually belongs to.

We also need to fix the precision of the labels, 234.23us makes no sense, that's like 0.01% precision. Probably this one found by AI

> packages/jaeger-ui/src/utils/number.ts:19 — toFloatPrecision. It computes significant digits relative to the integer part, so toFloatPrecision(324.23, 2) gives 324 (2 decimal digits of precision on the fractional side, but normalised by the magnitude of the number).

aabhinavvvvvvv:
@yurishkuro I've fixed the tick lines to be hidden along with their labels and for the precision fix, I've switched the tick label formatter to formatDurationCompact which rounds sub-ms values to whole microseconds

yurishkuro (approved):
looks good!

PR #3772, opened 2026-04-22
refactor(TraceTimelineViewer): decouple SpanTreeOffset from span model

### LINKED ISSUE #3770: refactor(timeline): Simplify SpanTreeOffset to accept precomputed ancestor data instead of full span
A follow-up to #3765.

## Problem

`SpanTreeOffset` currently accepts a full `IOtelSpan` and walks the `parentSpan` linked list
internally to build the ancestor chain for rendering indent guides. This means any caller that
wants to render tree lines for a synthetic or non-real span (e.g. `PrunedSpanRow`) must construct
a fully linked chain of fake `IOtelSpan` objects, including prototype tricks to satisfy the
last-child check:

```ts
const syntheticParent = Object.create(parentSpan) as IOtelSpan;
(syntheticParent as unknown as { childSpans: IOtelSpan[] }).childSpans = [...parentSpan.childSpans, fake];
(fake as { parentSpan: IOtelSpan }).parentSpan = syntheticParent;
```

The root cause is that `SpanTreeOffset` is doing tree traversal when it should be a pure rendering
widget.

## Proposed Fix

The caller computes two things by walking `span.parentSpan` once:

```ts
// SpanTreeOffset props (sketch)
type SpanTreeOffsetProps = {
  spanID: string;
  hasChildren: boolean;
  childCount: number;
  // One entry per ancestor depth, ordered from outermost to immediate parent.
  // color = draw a vertical bar at that depth; null = no bar (that ancestor has no more siblings below it).
  // The last entry (immediate parent) must never be null — its color is always needed for the
  // horizontal connector. isLastChild controls whether that connector is a T-junction or a bottom-left angle.
  ancestorColors: (string | null)[];
  isLastChild: boolean;
  color: string;  // own dot/box color
  // ... hover/click handlers unchanged
};
```

`SpanTreeOffset` becomes a pure stripe renderer — no tree traversal, no span model knowledge.
Root spans pass an empty `ancestorColors` array and `isLastChild: false`; the component renders
only the dot/box with no indent guides, which is the correct behavior today.
Callers (`SpanBarRow`) walk `span.parentSpan` once to compute the array. `PrunedSpanRow` does
the same from `parentSpan` directly — no fake span construction, no prototype tricks.

## Impact

- `SpanTreeOffset` and `SpanBarRow` need updating.
- `PrunedSpanRow` simplifies significantly — no fake span, no prototype hacks.
- `SpanTreeOffset` tests become trivial — no span graph setup needed, just arrays.
- No user-visible behavior change.

vibhor-5:
Unifies the tree hierarchy calculation in a central utility function and makes SpanTreeOffset a pure rendering component. This eliminates the need for complex prototype hacks to render synthetic spans.

Resolves #3770

## Description of the changes
refactored the SpanTreeOffset component to be a pure rendering widget and eliminated the internal tree traversal and dependency on the IOtelSpan model

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanTreeOffset.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanTreeOffset.tsx
@@ -1,7 +1,7 @@
 // Copyright (c) 2017 Uber Technologies, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-import React, { useMemo } from 'react';
+import React from 'react';
 import cx from 'classnames';
 import _get from 'lodash/get';
 import { connect } from 'react-redux';
@@ -9,14 +9,18 @@ import { bindActionCreators, Dispatch } from 'redux';
 
 import { actions } from './duck';
 import { ReduxState } from '../../../types';
-import { IOtelSpan } from '../../../types/otel';
-import colorGenerator from '../../../utils/color-generator';
 
 import './SpanTreeOffset.css';
 
 type TDispatchProps = {
   addHoverIndentGuideId: (spanID: string) => void;
   removeHoverIndentGuideId: (spanID: string) => void;
+};
+
+export type SpanTreeOffsetAncestor = {
+  spanID: string;
+  color: string;
+  isTerminated: boolean;
 };
 
 type TProps = {
@@ -27,8 +31,13 @@ type TProps = {
   onClick?: () => void;
   showChildrenIcon?: boolean;
   isDetailRow?: boolean;
-  span: IOtelSpan;
+  spanID: string;
+  hasChildren: boolean;
+  childCount: number;
+  ancestors: SpanTreeOffsetAncestor[];
+  isLastChild: boolean;
   color: string;
+  parentColor: string | null;
 };
 
 export const UnconnectedSpanTreeOffset: React.FC<TProps> = ({
@@ -36,22 +45,16 @@ export const UnconnectedSpanTreeOffset: React.FC<TProps> = ({
   onClick = undefined,
   showChildrenIcon = true,
   isDetailRow = false,
-  span,
+  spanID,
+  hasChildren,
+  childCount,
+  ancestors,
+  isLastChild,
+  color,
+  parentColor,
   addHoverIndentGuideId,
   removeHoverIndentGuideId,
-  color,
 }) => {
-  // Build ancestor chain directly from span.parentSpan
-  const ancestors = useMemo(() => {
-    const chain: IOtelSpan[] = [];
-    let current = span.parentSpan;
-    while (current) {
-      chain.unshift(current);
-      current = current.parentSpan;
-    }
-    return chain;
-  }, [span]);
-
   /**
    * If the mouse leaves to anywhere except another span with the same ancestor id, this span's ancestor id is
    * removed from the set of hoverIndentGuideIds.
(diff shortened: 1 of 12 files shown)


packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanTreeOffset.tsx
```diff
@@ -19,6 +17,12 @@ type TDispatchProps = {
   removeHoverIndentGuideId: (spanID: string) => void;
 };
 
+export type SpanTreeOffsetAncestor = {
+  spanID: string;
```
yurishkuro: why is span ID relevant for ancestors in this component?

vibhor-5: Its mostly cuz of the vertical line hover effect.
When you hover over an indent guide line, the component fires off the ancestor's spanID to Redux. That way, all the other rows in the tree that share that same ancestor know to highlight their piece of the vertical line too.
image

yurishkuro:
Thanks for taking this on — making `SpanTreeOffset` a pure renderer and dropping the `Object.create` hacks in `PrunedSpanRow` are real wins, please keep both.

One concern though: `buildSpanTreeOffsetState(span)` still walks `parentSpan` upward, so it's O(depth) per call. Calling it inline from `SpanBarRow` / `SpanDetailRow` makes that O(n · depth) across a trace render, and it's not memoized (`PrunedSpanRow` does `useMemo`, the others don't). The original code at least cached the walk inside `SpanTreeOffset`'s `useMemo`, so as it stands this PR moves the cost from one place to another rather than removing it. With ~80k-span traces being a real target, I'd like to address the underlying shape before merging rather than as a follow-up.

The fix that matches the spirit of issue #3770: compute the tree-offset state for every span in **one DFS pass over the trace**, keyed by spanID. Spans are already in pre-order DFS (`generateRowStates.ts` relies on that), and trace normalization already builds `spanMap` / `childSpans` — adding a per-span `treeOffsetState` fits naturally there. Row components then read `treeOffsetMap.get(spanID)` — O(1) per row, no walk.

Two observations make the DFS cheap:

1. `ancestors[i].isTerminated` is purely structural ("is `chain[i+1]` the last child of `chain[i]`?") — it doesn't depend on the rendered span. So all siblings at the same depth share one `ancestors` array; allocate once per parent during DFS, reuse for all its children.
2. `ancestors[last].isTerminated` is dead — the render only reads `isTerminated` when `!isLastAncestor`. So the only per-span variation is `isLastChild` (one bool) and `parentColor` (one lookup). The heavy array is shared.

Sketch:

```ts
// run once per trace during normalization
function buildTreeOffsetMap(spans: ReadonlyArray<IOtelSpan>): Map<string, SpanTreeOffsetState> {
  const result = new Map();
  const stack: SpanTreeOffsetAncestor[] = [];   // one entry per depth
  let cachedAncestorsForDepth: SpanTreeOffsetAncestor[][] = []; // shared across siblings
  for (const span of spans) {
    // pop stack down to span.depth, invalidate cache below
    // snapshot stack into cachedAncestorsForDepth[span.depth] if not already cached for this parent
    // isLastChild = parent.childSpans[last]?.spanID === span.spanID
    // parentColor = stack[span.depth - 1]?.color ?? null
    // result.set(span.spanID, { ancestors: cachedAncestorsForDepth[span.depth], isLastChild, parentColor })
    // push self with isTerminated derived from "am I last child of my parent?"
  }
  return result;
}
```

`SpanTreeOffset`'s prop shape and the `PrunedSpanRow` simplification stay exactly as you have them. The only change is that callers read from a precomputed map instead of calling a walk-the-parents helper. Net effect: O(n) per trace, O(1) per row render, and the helper's awkward shape (taking a span just to walk its ancestors) goes away.

vibhor-5:
> Two observations make the DFS cheap:
> 
> 1. `ancestors[i].isTerminated` is purely structural ("is `chain[i+1]` the last child of `chain[i]`?") — it doesn't depend on the rendered span. So all siblings at the same depth share one `ancestors` array; allocate once per parent during DFS, reuse for all its children.
> 2. `ancestors[last].isTerminated` is dead — the render only reads `isTerminated` when `!isLastAncestor`. So the only per-span variation is `isLastChild` (one bool) and `parentColor` (one lookup). The heavy array is shared

Thanks for the detailed sketch it made the direction very clear.

Made some changes now, `buildTreeOffsetMap` now does a single O(n) pass over the pre-order span array. Siblings at the same depth share one `ancestors` array (allocated once per parent via `cachedAncestors[depth]`), so total allocation is O(max_depth), not O(n·depth). Row components do a plain `map.get(spanID)` — no walks at render time. The map is memoized in VirtualizedTraceView via memoizeOne so it's built once per trace object.

vibhor-5:
@yurishkuro Made some changes to my approach - 
- `buildTreeOffsetMap(spans)` — single DFS pass over the pre-order span array, keyed by spanID. Siblings under the same parent share one ancestors array (allocated once, reused). Row components call treeOffsetMap.get(spanID) — O(1) per row
- `VirtualizedTraceView` — instance-level cache rebuilds the map only when trace.spans reference changes.
- `PrunedSpanRow` — Object.create prototype hack gone; derives state directly from the parent's map entry.
- `SpanTreeOffset` — pure renderer, no internal traversal.

Net: O(n) build per trace, O(1) per row render.

One note on the stack.map() snapshot: I tried the incremental/prototype-chain approach to avoid O(depth) per snapshot but it carries stale `isTerminated` values for descendants of the last sibling . The snapshot is only taken once per parent group so the build is O(n) for real traces; a linear-chain worst case is O(n·depth) for the build only , not per render.

yurishkuro:
Thank you for the PR — the overall approach is sound and the `buildTreeOffsetMap` O(n) single-pass idea is exactly the optimization the issue is after. A few things to address before merge:

## Bug: `is-last` class never fires (visual regression)

In `SpanTreeOffset.tsx` the render condition for the immediate-parent guide is:

```tsx
'is-last': isLastAncestor && !isDetailRow && !shouldTerminate && isLastChild,
```

`shouldTerminate` comes from `ancestor.isTerminated`, and in `buildTreeOffsetMap` you set `selfEntry.isTerminated = isLastChild`. So when a span **is** the last child, `shouldTerminate = true` and the whole expression evaluates to `false` — `is-last` is never applied. The half-height vertical connector is never drawn for any last-child span.

The fix is to drop the `!shouldTerminate` guard from this specific class; `isTerminated` should only suppress the **full-height** bar for non-immediate ancestors:

```tsx
'is-last': isLastAncestor && !isDetailRow && isLastChild,
```

## Redundant `parentColor` prop

`parentColor` can always be derived as `ancestors.at(-1)?.color ?? color`. Removing it simplifies the API and eliminates one source of inconsistency (callers can pass the wrong value).

## Fragile mutable-snapshot invariant in `buildTreeOffsetMap`

The logic that mutates `parentEntry.isTerminated` on the live stack after siblings have already been snapshotted is correct but brittle. A future reader (or a later refactor of the traversal order) can easily break it. Consider adding a targeted comment explaining the invariant: "the snapshot for depth D is rebuilt when we enter the first child of a new subtree, so mutations to the live stack entry are always visible before the snapshot is taken."

## Minor: `getTreeOffsetMap` constructor placement

The `getTreeOffsetMap` method is defined before the constructor, which is an unusual ordering for a class. Move it after the constructor or, better, make `_treeOffsetMap` and `_treeOffsetMapSpans` a single private `memoizeOne`-wrapped field to keep the caching logic in one place.

Overall this is the right architectural direction. Once the `is-last` regression is fixed and `parentColor` is removed the component API will match the spirit of #3770 cleanly.

vibhor-5:
@yurishkuro addresed the review

PR #3773, opened 2026-04-22
test: lighten react-router-dom mocks in unit tests

### LINKED ISSUE #3698: test: replace vi.mock('react-router-dom') spread pattern with MemoryRouter
## Problem

15 test files mock `react-router-dom` by loading the entire actual module and spreading it, just to override one or two hooks (typically `useLocation` or `useNavigate`):

```js
vi.mock('react-router-dom', async () => ({
  ...(await vi.importActual('react-router-dom')),
  useLocation: () => mockLocation,
}));
```

This is unnecessarily heavy — the whole module is loaded and re-exported just to patch a single export.

## Affected files

- `src/components/App/TraceIDSearchInput.test.jsx`
- `src/components/QualityMetrics/index.test.jsx`
- `src/components/QualityMetrics/url.test.js`
- `src/components/common/UiFindInput.test.jsx`
- `src/components/DeepDependencies/index.test.jsx`
- `src/components/DeepDependencies/SidePanel/DetailsPanel.test.jsx`
- `src/components/DeepDependencies/url.test.js`
- `src/components/DeepDependencies/Graph/DdgNodeContent/index.test.jsx`
- `src/components/DeepDependencies/traces.test.jsx`
- `src/components/TraceDiff/url.test.js`
- `src/components/TraceDiff/TraceDiff.test.jsx`
- `src/components/TracePage/useNormalizeTraceId.test.ts`
- `src/components/TracePage/index.test.jsx`
- `src/components/SearchTracePage/SearchResults/index.test.jsx`
- `src/components/SearchTracePage/url.test.js`

## Preferred alternatives

**1. Wrap with `MemoryRouter` (React Router-recommended approach):**
```jsx
render(
  <MemoryRouter initialEntries={['/path?foo=bar']}>
    <ComponentUnderTest />
  </MemoryRouter>
);
```
No mock needed — the component runs with a real router and a controlled URL.

**2. `vi.spyOn` for targeted hook override:**
```js
import * as ReactRouterDom from 'react-router-dom';

beforeEach(() => {
  vi.spyOn(ReactRouterDom, 'useNavigate').mockReturnValue(mockNavigate);
});
```

The right choice depends on the test — `MemoryRouter` is better when the component uses routing deeply; `spyOn` is a quick drop-in when only the return value matters.

ghost:
## Summary
- Resolves #3698 
- Replace `vi.mock('react-router-dom', async () => ({ ...(await vi.importActual('react-router-dom')), ... }))` patterns with minimal mocks that only stub the symbols each test needs (`matchPath`, `useNavigate`, `useLocation`, `useParams`). Where tests still wrap ui in a router, pull in only the real router shell (`MemoryRouter` / `BrowserRouter`) via `vi.importActual`.
- **Why not `MemoryRouter` everywhere?**
  - Many tests using fake navigate function (e.g. “`navigate` was called with this path”). With only `MemoryRouter` and no mock, `useNavigate` is the real navigate - we don’t get a simple `jest.fn()` to assert on unless we change how we test. So for those files we kept controlled fakes for the hooks.
- **Why not `vi.spyOn` on the real module?**
  - In this setup, some exports can’t be replaced from the outside, so Vitest errors. Mocking the whole module with `vi.mock` avoids that.
- So now tests still checking the same behavior, but `react-router-dom` is mocked in a thinner way, only the pieces each file needs - instead of loading/spreading the whole library.

### DIFF
--- a/packages/jaeger-ui/src/components/DeepDependencies/Graph/DdgNodeContent/index.test.jsx
+++ b/packages/jaeger-ui/src/components/DeepDependencies/Graph/DdgNodeContent/index.test.jsx
@@ -13,12 +13,10 @@ vi.mock('./calc-positioning', async () =>
 // Mutable object so individual tests can control the location without re-creating the mock.
 const mockLocation = { search: '' };
 
-vi.mock('react-router-dom', async () => ({
-  ...(await vi.importActual('react-router-dom')),
+vi.mock('../../../../model/path-agnostic-decorations', () => mockDefault(jest.fn(() => ({}))));
+vi.mock('react-router-dom', () => ({
   useLocation: () => mockLocation,
 }));
-
-vi.mock('../../../../model/path-agnostic-decorations', () => mockDefault(jest.fn(() => ({}))));
 
 import React from 'react';
 import { render, screen, fireEvent } from '@testing-library/react';
--- a/packages/jaeger-ui/src/components/App/TraceIDSearchInput.test.jsx
+++ b/packages/jaeger-ui/src/components/App/TraceIDSearchInput.test.jsx
@@ -11,8 +11,9 @@ import TraceIDSearchInput from './TraceIDSearchInput';
 
 const mockNavigate = jest.fn();
 vi.mock('react-router-dom', async () => {
+  const { MemoryRouter: ActualMemoryRouter } = await vi.importActual('react-router-dom');
   return {
-    ...(await vi.importActual('react-router-dom')),
+    MemoryRouter: ActualMemoryRouter,
     useNavigate: () => mockNavigate,
   };
 });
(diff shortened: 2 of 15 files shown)


packages/jaeger-ui/src/components/DeepDependencies/Graph/DdgNodeContent/index.test.jsx
```diff
@@ -13,13 +13,11 @@ vi.mock('./calc-positioning', async () =>
 // Mutable object so individual tests can control the location without re-creating the mock.
 const mockLocation = { search: '' };
 
-vi.mock('react-router-dom', async () => ({
-  ...(await vi.importActual('react-router-dom')),
+vi.mock('../../../../model/path-agnostic-decorations', () => mockDefault(jest.fn(() => ({}))));
```
yurishkuro, line 16: is this not the same as just `vi.mock('../../../../model/path-agnostic-decorations');` ?

yurishkuro (approved):

PR #3782, opened 2026-04-28
fix: preserve embed params on back-to-search navigation and consolida…

### LINKED ISSUE #3780: [refactor]: Consolidate embedded-url utils into embedded-store and fix lost embed params in search URL
## Background

Jaeger UI supports an "embedded mode" where it is rendered inside a host application (e.g. an iframe). The host passes URL query parameters like `uiEmbed=v0`, `uiSearchHideGraph=1`, `uiTimelineHideMinimap=1`, etc. to configure which UI chrome is shown.

The feature was first introduced in PR #263 (commit `f50592b3`). It was revised in PR #286 (commit `41124825`), which among other things introduced `utils/embedded-url.ts` with `getEmbeddedState()` (URL string parser) and `stripEmbeddedState()` (strips `ui*` keys from a query-param object), and moved embed state into Redux. PR #3761 (commit `eb157ad7`) later migrated embed flags off Redux into a Zustand-based store (`stores/embedded-store.ts`), which now calls `getEmbeddedState()` internally.

`getEmbeddedState()` is today only called from `embedded-store.ts`. `stripEmbeddedState()` is only called from `SearchResults/index.tsx`.

## Problem 1: `stripEmbeddedState` drops embedding config from navigation URLs

`SearchResults/index.tsx` computes a single `searchUrl` via `getUrl(stripEmbeddedState(queryOfResults))` and uses it for **two distinct purposes**:

1. The `fromSearch` value stored in React Router location state, which `TracePage` reads back as `toSearch` to render the "Back to Search" button.
2. A standalone link — a `<NewWindowIcon>` link rendered in the search results header that opens the same search in a new tab **outside** the embedding context (i.e. without the `ui*` params). This link is only rendered when `showStandaloneLink` is true, which is set to `Boolean(embedded)` in `SearchTracePage/index.tsx` — so it is only visible when Jaeger is already running in embedded mode (`uiEmbed=v0`). The link uses `getTargetEmptyOrBlank()`, which returns `_blank` normally but returns `''` (same frame) when the `forbidNewPage` user config option is set.

`stripEmbeddedState` was introduced in PR #286 (commit `41124825`, Dec 2018) with the explicit intent of powering the standalone link — stripping `ui*` params so the new window is not itself embedded. The PR #286 commit message says: *"Sequester parsing and stripping query parameters for the embed mode to a util"*.

The bug is that the same stripped URL is reused for `fromSearch`. When a user clicks a trace result and then clicks "Back to Search", the `ui*` embedding params (`uiEmbed=v0`, `uiSearchHideGraph=1`, etc.) are gone, so the embedding constraints are silently dropped on the return navigation. Since both paths are only active in embedded mode, the bug is always triggered whenever a user navigates search → trace → back-to-search inside an embedded Jaeger instance.

**Fix:** Use two separate URLs — preserve `ui*` params in `fromSearch` (for back-navigation), and keep stripping them only for the standalone link. Additionally, when `forbidNewPage` is true the standalone link should not be rendered at all: its sole purpose is to escape the embedding context into a new tab, and rendering it as a same-frame navigation to a stripped URL actively breaks the embedding config.

## Problem 2: `utils/embedded-url.ts` is an unnecessary indirection

`utils/embedded-url.ts` exports two functions:

- `getEmbeddedState(search)` — parses a query string into an `EmbeddedState`. Used only by `embedded-store.ts`.
- `stripEmbeddedState(state)` — strips `ui*` keys from a query-param object. Used only by `SearchResults/index.tsx` (and should be removed per Problem 1).

After fixing Problem 1, `getEmbeddedState` has only one caller (`embedded-store.ts`) and `stripEmbeddedState` has only one caller (`SearchResults/index.tsx`, for the standalone link URL). Both functions belong conceptually with the embedded store rather than in a generic utils file. They can be moved into `stores/embedded-store.ts`, and `utils/embedded-url.ts` and its test file deleted entirely.

## Proposed changes

1. Move `getEmbeddedState` and `stripEmbeddedState` from `utils/embedded-url.ts` into `stores/embedded-store.ts`, then delete `utils/embedded-url.ts` and `utils/embedded-url.test.js`. Move the tests for both functions into `stores/embedded-store.test.ts`.
2. In `SearchResults/index.tsx`, split the single `searchUrl` into two: a plain URL (no stripping) for `fromSearch` back-navigation, and a stripped URL for the standalone `<Link>`. Update the `stripEmbeddedState` import to come from `stores/embedded-store`.
3. In `SearchResults/index.tsx`, hide the standalone link when `forbidNewPage` is true (i.e. change the render condition from `showStandaloneLink` to `showStandaloneLink && !forbidNewPage`), so it is never rendered as a same-frame navigation to a stripped URL.
4. Update `SearchResults` tests accordingly.

## Files affected

- `packages/jaeger-ui/src/utils/embedded-url.ts` — delete
- `packages/jaeger-ui/src/utils/embedded-url.test.js` — delete
- `packages/jaeger-ui/src/stores/embedded-store.ts` — inline `getEmbeddedState` logic
- `packages/jaeger-ui/src/stores/embedded-store.test.ts` — add tests for `getEmbeddedState` and `stripEmbeddedState` (moved from deleted util); remove mock of `embedded-url`
- `packages/jaeger-ui/src/components/SearchTracePage/SearchResults/index.tsx` — split `searchUrl` into back-nav URL (no stripping) and standalone URL (stripped); import `stripEmbeddedState` from `stores/embedded-store`; hide standalone link when `forbidNewPage` is true

aditya86-id:
Resolves #3780

## Description

While tracing the embedded mode flow, I noticed that navigating 
Search → Trace → Back to Search silently drops uiEmbed=v0 and related 
params from the return URL.

The root cause: SearchResults/index.tsx computed a single searchUrl via 
stripEmbeddedState() and reused it for both fromSearch back-navigation 
and the standalone new-tab link — even though only the latter should 
have embed params stripped.

Fix: split into two explicit URLs:
- backNavUrl: keeps ui* params intact so "Back to Search" returns to 
  the same embedded context
- standaloneUrl: strips ui* params for the standalone link (original 
  intent from PR #286)

Also hid the standalone link when forbidNewPage=true. Its only purpose 
is opening search in a new tab outside the embed context — forcing 
same-frame navigation to a stripped URL actively breaks the config.

On the refactor side, getEmbeddedState() and stripEmbeddedState() each 
had exactly one caller, both embed-related. Moved both into 
stores/embedded-store.ts, deleted utils/embedded-url.ts, and moved the 
tests with them. The vi.mock of embedded-url in embedded-store.test.ts 
is replaced with tests against the real implementations.

## Checklist
- [x] I have read CONTRIBUTING_GUIDELINES.md

## AI Usage
- [x] Light: AI provided minor assistance (formatting, simple suggestions)


ghost (commented):
can you please share a before and after screen recording like https://github.com/jaegertracing/jaeger-ui/issues/3780#issuecomment-4324235065 ? I was looking into problem 1, but I was not able to reproduce it.

yurishkuro:
please rebase

yurishkuro:
Thanks for the clean PR description and for moving the tests to real implementations — that's a strictly better approach.

A few issues and suggestions:

**1. Missing test for the core bug fix**

The PR's main fix is that `fromSearch` now preserves embed params. There's no test asserting this. Given that the bug is the main motivation, a test that verifies `backNavUrl` (passed as `fromSearch`) still contains `uiEmbed=v0` and related params would significantly increase confidence. The new `forbidNewPage` test is good, but it covers the refactor side, not the fix.

**2. Redundant `= undefined` default**

In the new `stripEmbeddedState` in `embedded-store.ts`:
```typescript
const { uiEmbed = undefined, ...rv } = state;
```
Destructuring a missing key already yields `undefined`. The `= undefined` is a no-op and adds visual noise.

**3. Mixing store and utility concerns in `embedded-store.ts`**

The original split — `utils/embedded-url.ts` for pure functions, `stores/embedded-store.ts` for the caching layer — was clean. After this PR, `embedded-store.ts` becomes a hybrid: it exports both URL-parsing utilities and a cached singleton. The "each had only one caller" rationale is debatable; caller count isn't a strong reason to co-locate unrelated concerns. A `stores/` module that also exports stateless URL helpers will be surprising to future readers. I'd suggest either keeping a thin `utils/embedded-url.ts` (or renaming it `utils/embedded-params.ts`) and having `embedded-store.ts` import from there — which is what the original code did — or at minimum exporting the two utility functions from a barrel that makes their home feel intentional.

**4. Copyright attribution dropped**

`utils/embedded-url.ts` carried `Copyright (c) 2018 Uber Technologies, Inc.` which documents the Uber origins of this code. That line is gone in the new `embedded-store.ts` header. The SPDX license identifier is preserved, but the attribution history is silently dropped. Please consider retaining it.

PR #3784, opened 2026-04-29
Refactor: Simplify SpanTreeOffset to use precomputed ancestor data

### LINKED ISSUE #3770: refactor(timeline): Simplify SpanTreeOffset to accept precomputed ancestor data instead of full span

farhanxansari:

* Fixes #3770

## Description of the changes

* Refactored `SpanTreeOffset` to remove internal parent traversal logic
* Replaced `span: IOtelSpan` with precomputed props:

  * `ancestorEntries`, `isLastChild`, `spanID`, `hasChildren`, `childCount`
* Introduced `span-tree-utils.ts` with helper functions:

  * `computeAncestorEntries`
  * `computeIsLastChild`
* Updated dependent components (`SpanBarRow`, `SpanDetailRow`) to compute and pass data using `useMemo`
* Simplified `PrunedSpanRow` by removing synthetic span construction and prototype manipulation
* Cleaned up logic to make the component purely presentational and deterministic

* Added new unit tests for `span-tree-utils` (10 test cases)
* Rewrote `SpanTreeOffset` tests to use plain data instead of linked spans
* Updated all affected component tests to match new API
* Verified:

  * 65 tests passing across modified/new test files
  * Full test suite: 2819 passing (6 unrelated pre-existing timeouts)
  * `tsc-lint` passed
  * `fmt-lint` passed
  * `oxlint`  passed (no new issues)

## Checklist

* [x] I have added unit tests for the new functionality

* [x] **Moderate**: AI helped with code structuring and debugging specific parts

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -322,6 +322,7 @@
       "integrity": "sha512-CGOfOJqWjg2qW/Mb6zNsDm+u5vFQ8DxXfbM09z69p5Z6+mE1ikP2jUXw+j42Pf1XTYED2Rni5f95npYeuwMDQA==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.29.0",
         "@babel/generator": "^7.29.0",
@@ -651,6 +652,7 @@
         }
       ],
       "license": "MIT",
+      "peer": true,
       "engines": {
         "node": ">=20.19.0"
       },
@@ -699,6 +701,7 @@
         }
       ],
       "license": "MIT",
+      "peer": true,
       "engines": {
         "node": ">=20.19.0"
       }
@@ -709,6 +712,7 @@
       "integrity": "sha512-yq6OkJ4p82CAfPl0u9mQebQHKPJkY7WrIuk205cTYnYe+k2Z8YBh11FrbRG/H6ihirqcacOgl2BIO8oyMQLeXw==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@emnapi/wasi-threads": "1.2.1",
         "tslib": "^2.4.0"
@@ -720,6 +724,7 @@
       "integrity": "sha512-ewvYlk86xUoGI0zQRNq/mC+16R1QeDlKQy21Ki3oSYXNgLb45GV1P6A0M+/s6nyCuNDqe5VpaY84BzXGwVbwFA==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "tslib": "^2.4.0"
       }
@@ -771,7 +776,6 @@
       "integrity": "sha512-EriSTlt5OC9/7SXkRSCAhfSxxoSUgBm33OH+IkwbdpgoqsSsUg7y3uh+IICI/Qg4BBWr3U2i39RpmycbxMq4ew==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": "^12.0.0 || ^14.0.0 || >=16.0.0"
       }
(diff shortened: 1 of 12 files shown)


yurishkuro:
Thank you for the PR — the decoupling is clean, `SpanTreeOffset` is now a proper pure renderer, and the `null`-color encoding for terminated branches is a nice fit for the issue's `(string | null)[]` proposal. A few issues to address:

## Performance: not O(n) globally (issue's stated goal)

`computeAncestorEntries` walks `span.parentSpan` per span, and it is called inside `useMemo` in each of `SpanBarRow`, `SpanDetailRow`, and `PrunedSpanRow`. `useMemo` prevents recomputation on re-renders but the **initial render** of a visible row still costs O(depth). For deeply nested traces this adds up.

Issue #3770 specifically asks for the caller to walk `span.parentSpan` **once** and produce the array. The intended pattern is a single pre-pass over all spans (e.g. in `VirtualizedTraceView` or `generateRowStates`) that builds a `Map<spanID, AncestorEntry[]>`. Each row then does an O(1) lookup. Consider adopting that structure.

## Circular type dependency

`span-tree-utils.ts` imports `AncestorEntry` from `./SpanTreeOffset`:

```ts
import type { AncestorEntry } from './SpanTreeOffset';
```

This makes a utility file depend on a component, which is the wrong direction. Move the `AncestorEntry` type into `span-tree-utils.ts` (or a shared types file) and have `SpanTreeOffset` import from there.

## `PrunedSpanRow` still needs caller knowledge

`PrunedSpanRow` calls `computeAncestorEntries(parentSpan)` and then manually appends `parentSpan` itself as the last entry. This "append-self-as-immediate-parent" pattern is not part of the utility's API contract and isn't documented. Either add a second utility (e.g. `computeAncestorEntriesForChild(parentSpan)`) that does this correctly, or have the pre-pass map handle it uniformly.

## Noisy `package-lock.json` diff

The PR includes hundreds of unrelated `"peer": true` flag changes in `package-lock.json`. Please revert those to keep the diff reviewable. Run `git checkout HEAD -- package-lock.json` and regenerate only if you actually changed dependencies.

## Misplaced JSDoc

There are two `/** ... */` block comments above `computeIsLastChild` — the first one describes `computeAncestorEntries`. Move/merge them so each function has exactly one doc comment.

The component-level decoupling here is correct and the `null`-color API is cleaner than carrying a separate `isTerminated` boolean. With the pre-pass optimization and the type-dependency inversion this would be in good shape.

PR #3787, opened 2026-05-01
fix(SpanDetail): Format complex attribute values in summary and copy

### LINKED ISSUE #8447: [Bug]: Span Detail attribute summary displays "[object Object]" for complex OTel attributes
### What happened?

**Describe the bug**
In the `SpanDetail` component, when the "Attributes" (or "Tags") accordion is collapsed, a summary of the key-value pairs is displayed to provide a quick preview. 

Currently, the `AttributesSummary` component renders these values using `String(item.value)`. Since OpenTelemetry (OTel) attributes support complex data types like nested objects and arrays, this approach causes the UI to display the non-descriptive string `"[object Object]"` for any non-primitive attribute value. 

Additionally, the "Copy" functionality in the `AttributesTable` also falls back to `String(value)`, resulting in `[object Object]` being copied to the clipboard instead of the actual data.

**To Reproduce**
1. Load a trace containing spans with complex/nested attributes (common in OTel-native instrumentation).
2. Expand a span to view the `SpanDetail` view.
3. Keep the "Attributes" (or "Tags") section collapsed.
4. Observe that any complex attribute shows as `key=[object Object]` in the summary list.
5. Expand the section and click the "Copy" icon for a complex attribute; observe that `[object Object]` is copied to the clipboard.

**Expected behavior**
*   **Summary**: Should show a meaningful preview, such as `[N items]` for arrays or `{...}` for objects.
*   **Copying**: Should copy the actual value (e.g., as a JSON string) instead of the string representation of the object type.

**Actual behavior**
The UI displays and copies `[object Object]`.

**Proposed Solution**
1.  Introduce a utility function to format `AttributeValue` types for the summary view.
2.  Update `AccordionAttributes.tsx` to use this utility in `AttributesSummary`.
3.  Update `AttributesTable.tsx` to use `JSON.stringify` for complex types when copying values to the clipboard.

**Affected Files**
- `packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/AccordionAttributes.tsx`
- `packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/AttributesTable.tsx`

### Steps to reproduce

1. Open a trace in the Jaeger UI that contains spans with complex attribute values (arrays or objects).
2. Click on a span in the timeline to open the **SpanDetail** view.
3. Keep the **Attributes** (or **Tags**) section collapsed and observe the summary preview next to the section header.
4. Notice that complex attributes are rendered as `key=[object Object]`.
5. Expand the section and click the **Copy** icon on a complex attribute.
6. Paste the clipboard content and observe that it is `[object Object]` instead of the formatted JSON data.

### Expected behavior

- The attribute summary should provide a human-readable preview for complex types. 
    - For arrays, it should show a count (e.g., `[3 items]`).
    - For objects, it should show a placeholder (e.g., `{...}`) or truncated JSON.
- The "Copy" button for complex attributes should copy the actual JSON value to the clipboard, allowing users to paste and inspect the data elsewhere.

### Relevant log output

```shell

```

### Screenshot

_No response_

### Additional context

_No response_

### Jaeger backend version

_No response_

### SDK

_No response_

### Pipeline

_No response_

### Stogage backend

_No response_

### Operating system

_No response_

### Deployment model

_No response_

### Deployment configs

```shell

```

sakshar2303:
### Problem

In the `SpanDetail` component, complex attribute values (objects and arrays) were rendered as the non-descriptive string `[object Object]` in two places:
- The **collapsed accordion summary** preview
- The **"Copy value"** clipboard feature

This bug is particularly impactful for traces using OpenTelemetry GenAI semantic conventions, where attributes like `gen_ai.content.prompt` and tool inputs/outputs carry complex nested objects , making readable summary display critical for GenAI observability workflows.

---

### Changes

- ✅ Created a `formatAttributeValue` utility to handle complex types:
  - Arrays now show as `[N items]` in the summary
  - Objects now show as `{...}` in the summary
- ✅ Updated `AccordionAttributes` to use the new formatter for the summary preview
- ✅ Improved `AttributesTable` copy logic to use `JSON.stringify` for complex attributes, ensuring actual data is copied to the clipboard
- ✅ Added unit tests in `AccordionAttributes.test.js` and `AttributesTable.test.js` to verify correct formatting and copying of nested structures

---

### Linked Issue

Fixes jaegertracing/jaeger#8447

---

### Context

This fix is directly relevant to the upcoming GenAI Observability work in Jaeger , the `SpanDetail` side panel will need to render complex `gen_ai.*` attributes correctly as part of the LFX Summer 2026 GenAI Observability project. Having this foundation in place ensures the existing attribute rendering is solid before the richer Markdown and JSON tree rendering is added on top.

---

### Testing

- ✅ All existing unit tests passing
- ✅ New unit tests added for `formatAttributeValue` covering arrays, objects, primitives, and null values
- ✅ Tested locally against traces with complex nested OTel attributes


yurishkuro:
Still a valid fix — but it needs a rebase and some reconciliation with work that merged after it was opened.

### Still needed
The two bugs are live on `main`: `AccordionAttributes.tsx` renders the collapsed summary as `String(item.value)` (→ `[object Object]`), and `AttributesTable`'s "Copy value" button copies `String(row.value)` (→ `[object Object]`). Nothing merged fixes these — the value *display* in `AttributesTable` was already fine (it uses the `react-json-view-lite` tree), so this PR's scope is correctly limited to the summary and the copy button.

### Reconcile with #4171 (merged)
#4171 fixed the same class of bug in the span-search path (`filter-spans.ts`) using the identical guard:
```ts
typeof value === 'object' && !(value instanceof Uint8Array) ? JSON.stringify(value) : String(value)
```
This PR inlines that same guard again in `AttributesTable` (and its test), so the idiom would then exist in three places. Please extract a single shared "stringify an attribute value" helper that both `filter-spans` and the copy button consume, rather than a third copy. Keep `formatAttributeValue` for the lossy *summary* rendering (`[N items]` / `{...}`) — that's a genuinely separate concern.

### Correctness gap: circular references
Crucially, #4171 wrapped its `JSON.stringify` in a `try/catch` (with a dedicated test) precisely because a circular attribute value throws. This PR's copy path is a bare `JSON.stringify(row.value, null, 2)` with no guard — so a circular value would throw during render. Reusing #4171's helper fixes this for free.

### Copilot comments — please address
1. **`utils.ts` null/undefined branch** returning `''` silently hides data and diverges from the prior `String(value)`. Since `AttributeValue` doesn't currently include null/undefined, either drop the branch or return an explicit placeholder.
2. **No focused test for `formatAttributeValue`** — it's only exercised indirectly. Add a unit test covering arrays, objects, primitives, `Uint8Array`, and whatever you decide for null/undefined.
3. **Render-time `JSON.stringify`** runs for every object/array row even without a copy click — real cost for large GenAI attributes. This reinforces the shared-helper direction: defer stringification to click (e.g. a `getCopyText()` callback on `CopyIcon`) or memoize it. Worth doing given the timeline is expected to handle very large traces.

Net: keep it open; rebase, route the copy formatting through a shared helper that carries #4171's circular-ref guard, add the focused test, and resolve the null/undefined and lazy-stringify comments.

PR #3822, opened 2026-05-07
feat(trace-view): add decorative icons to span tree view

### LINKED ISSUE #3821: [Feature]: Decorative icons in the tree view
### Requirement

When a user views a trace they want to have visual indications of the types of spans, such as a database call or GenAI step like MCP tool call. 

### Problem

The trace view shows span hierarchy in a tree view, but no decorations for the spans.

### Proposal

Devise a way to introduce icons into the tree view and a framework for assigning the icons based on some trace heuristics (e.g. looking for specific attribute name patterns).

### Open questions

_No response_

vibhor-5:
Description
  This PR introduces visual icons in the trace tree view based on span attributes, addressing the requirement in #3821. This enhancement allows users to quickly identify the type of operation (e.g., database call, GenAI step, messaging) at a glance without expanding span details.

  Key Features
   - Flexible Matching Engine: Supports a rule-based system where spans are matched against specific tag/attribute key-value patterns.
   - Built-in OTel Heuristics: Provides sensible defaults out-of-the-box based on standard OpenTelemetry semantic conventions:
       - 🗄️ Database: db.system matches -> io.Database token.
       - 📬 Messaging: messaging.system matches -> io.PaperPlane token.
       - ↔️ RPC: rpc.system matches -> io.Swap token.
       - 🌐 HTTP: http.method matches -> io.Globe token.
       - 🤖 GenAI: gen_ai.system matches -> io.HardwareChip token.
       - 🔍 Elasticsearch: db.system matches -> si.Elasticsearch token.
       - 🕸️ GraphQL: graphql.operation matches -> si.Graphql token.
   - Configurable: Users can extend or override these icons via the spanDecorations field in the UI configuration (jaeger-ui.config.json).
   - Seamless Integration: Icons are rendered inline within the SpanBarRow, maintaining the existing tree view layout, alignment, and theme compatibility.

  Screenshot
  
image

  Testing Plan
   - Unit Tests: Added comprehensive test cases packages/jaegerui/src/components/TracePage/TraceTimelineViewer/SpanBarRow.test.jsx to verify the attribute matching and icon selection logic.
   - Manual Verification: Verified using a local Jaeger backend with OTLP-injected traces for various semantic conventions (PostgreSQL, OpenAI, etc.).
   - Regression: Confirmed that tree view offsets, error icons, and RPC arrows remain correctly aligned when decorations are present.

  Checklist
   - [x] I have signed off my commits (DCO).
   - [x] I have added unit tests for my changes.
   - [x] All existing tests pass.
   - [x] I have updated the documentation/examples if necessary.

  Closes #3821

### DIFF
--- a/packages/jaeger-ui/docs/span-decorations.md
+++ b/packages/jaeger-ui/docs/span-decorations.md
@@ -1,0 +1,67 @@
+# Span Decorations
+
+Span decorations allow visual icons to be displayed next to spans in the trace timeline view based on span attributes (tags). This helps users quickly identify different types of operations, such as database calls, messaging, or AI-related spans.
+
+## Configuration
+
+Span decorations are configured via the `spanDecorations` property in the Jaeger UI configuration.
+
+```json
+{
+  "spanDecorations": [
+    {
+      "entries": [
+        { "key": "db.system", "value": ".*" }
+      ],
+      "icon": "io.Database",
+      "tooltip": "Database Call"
+    },
+    {
+      "entries": [
+        { "key": "http.method", "value": "GET|POST" }
+      ],
+      "icon": "https://example.com/icons/http.svg",
+      "tooltip": "HTTP Request"
+    }
+  ]
+}
+```
+
+### Rule Evaluation
+
+The span decoration engine follows these rules:
+
+1.  **Order of Evaluation:** Rules are evaluated in the order they appear in the `spanDecorations` array.
+2.  **First Match Wins:** The first rule that matches a span is used, and subsequent rules are ignored.
+3.  **All Criteria Must Match (AND):** A rule matches a span only if **all** of its `entries` match.
+4.  **Regex Matching:** Both the `key` and `value` in an entry are treated as regular expressions.
+5.  **Attribute Search:** The engine searches for matches in both span attributes and resource attributes.
+
+### Icons
+
+The `icon` property can be one of:
+
+1.  **Built-in Tokens:** Pre-defined icons bundled with Jaeger UI (e.g., `io.Database`, `io.Server`, `io.Cloud`).
+2.  **External URLs:** A full URL starting with `http://` or `https://` pointing to an image (SVG, PNG, etc.).
+3.  **Local Paths:** A path starting with `/` pointing to an asset served by the Jaeger UI server.
+
+#### Supported Built-in Tokens
+
+- `io.Server`: General server operation.
+- `io.Database`: Database operations.
+- `io.Cloud`: Cloud/External services.
+- `io.Globe`: Web/HTTP requests.
+- `io.PaperPlane`: Messaging/PubSub.
+- `io.Swap`: RPC/Inter-service communication.
+- `io.HardwareChip`: AI/Compute intensive tasks.
+- `io.Shield`: Security/Auth operations.
+- `io.Terminal`: CLI/System commands.
+- `io.Settings`: Configuration/Internal tasks.
+
+## Extensibility
+
+Users can extend the span decorations purely via configuration by:
+- Adding new rules to match custom attribute patterns.
+- Using external URLs for custom icons not included in the built-in set.
+
+For more complex heuristics, the Jaeger UI can be built with custom icons added to the internal registry in `IconProvider.tsx`.
--- a/packages/jaeger-ui/docs/span-decorations.md
+++ b/packages/jaeger-ui/docs/span-decorations.md
@@ -6,21 +6,36 @@ Span decorations allow visual icons to be displayed next to spans in the trace t
 
 Span decorations are configured via the `spanDecorations` property in the Jaeger UI configuration.
 
+### Where to define the configuration
+
+The exact location of your configuration depends on how you are running the Jaeger UI:
+
+- **Production (Jaeger Backend):** When running the Jaeger Query or Jaeger All-in-One binaries, you can provide a UI configuration file using the `--query.ui-config` command-line flag. This file must be valid JSON.
+  ```bash
+  jaeger-all-in-one --query.ui-config=/path/to/jaeger-ui.config.json
+  ```
+
+- **Local Development (Vite):** If you are running the UI standalone for frontend development (e.g., using `npm start`), the Vite dev server automatically reads from a `jaeger-ui.config.json` file located in the `packages/jaeger-ui/` directory. You can copy the provided `jaeger-ui.config.example.json` to create your own configuration.
+
+### Configuration Schema
+
+Once you have located or created your configuration file, you can add or modify the `spanDecorations` array at the top level:
+
 ```json
 {
   "spanDecorations": [
     {
-      "entries": [
-        { "key": "db.system", "value": ".*" }
+      "attributes": [
+        { "key": "^(?:db\\.system)$", "value": ".*" }
       ],
       "icon": "io.Database",
       "tooltip": "Database Call"
     },
     {
-      "entries": [
-        { "key": "http.method", "value": "GET|POST" }
+      "attributes": [
+        { "key": "^(?:http\\.method)$", "value": "GET|POST" }
       ],
-      "icon": "https://example.com/icons/http.svg",
+      "icon": "https://example.com/icons/http.png",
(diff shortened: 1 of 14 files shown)


yurishkuro:
please show realistic trace (from HotROD) with multiple spans

vibhor-5:
@yurishkuro 



Screenshot 2026-05-08 at 12 42 25



does this look alright?

yurishkuro (commented):
I would like to see some design doc & user documentation. What is the "rule engine" look like? How does one extend it? How can a user extend it via config?

On the last point, if a user can extend it purely via config, then why are the icon classes being imported at compile time?

vibhor-5:
> I would like to see some design doc & user documentation. What is the "rule engine" look like? How does one extend it? How can a user extend it via config?

okay I'll add documentation 

> On the last point, if a user can extend it purely via config, then why are the icon classes being imported at compile time?

My bad I made it only part-configurable with limited icons which users can match to based on rules they provide therefore the limited icons could be imported at compile time itself 
I'll improve the implementation to a more configurable approach for users

packages/jaeger-ui/src/components/common/IconProvider.tsx
```diff
@@ -0,0 +1,82 @@
...
+  IoSettings,
+  IoSearch,
+  IoCodeWorking,
+  IoBug,
+  IoCube,
+  IoFlash,
+} from 'react-icons/io5';
+
+const ICON_MAP: Record<string, React.ElementType> = {
+  // Legacy aliases (for backward compatibility with the first version of PR)
```
yurishkuro: why do you need backwards compatibility within the same PR?

packages/jaeger-ui/src/components/common/IconProvider.tsx
```diff
@@ -0,0 +1,82 @@
...
+  IoHardwareChip,
+
+  // Formal built-in tokens
+  'io.Server': IoServer,
+  'io.PaperPlane': IoPaperPlane,
+  'io.Swap': IoSwapHorizontal,
+  'io.Globe': IoGlobeOutline,
+  'io.HardwareChip': IoHardwareChip,
+  'io.Cloud': IoCloud,
+  'io.Database': IoServer, // Fallback to Server if Database is missing in this version
```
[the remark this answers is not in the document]
yurishkuro: what does "if Database is missing in this version" mean? It's statically compiled.

packages/jaeger-ui/src/components/common/IconProvider.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+const IconProvider: React.FC<Props> = ({ icon, className, tooltip }) => {
+  // 1. Check built-in tokens
+  const BuiltInIcon = ICON_MAP[icon];
+  if (BuiltInIcon) {
+    return <BuiltInIcon className={className} title={tooltip} />;
+  }
+
+  // 2. Check for image URL
+  if (icon.startsWith('http://') || icon.startsWith('https://') || icon.startsWith('/')) {
```
yurishkuro, line 81: this check should be made before you check the default map

packages/jaeger-ui/docs/span-decorations.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Configuration
+
+Span decorations are configured via the `spanDecorations` property in the Jaeger UI configuration.
+
+```json
+{
+  "spanDecorations": [
+    {
+      "entries": [
```
yurishkuro: if this is matching against span attributes it should be called `attributes`

packages/jaeger-ui/docs/span-decorations.md
```diff
@@ -0,0 +1,67 @@
...
+
+### Icons
+
+The `icon` property can be one of:
+
+1.  **Built-in Tokens:** Pre-defined icons bundled with Jaeger UI (e.g., `io.Database`, `io.Server`, `io.Cloud`).
+2.  **External URLs:** A full URL starting with `http://` or `https://` pointing to an image (SVG, PNG, etc.).
+3.  **Local Paths:** A path starting with `/` pointing to an asset served by the Jaeger UI server.
+
+#### Supported Built-in Tokens
```
yurishkuro, line 63: What were your grounds for this specific set? Did you research how competitor tools do it? I've seen icons specific to the source code language of a service, for instance.

packages/jaeger-ui/docs/span-decorations.md
```diff
@@ -0,0 +1,67 @@
...
+  ]
+}
+```
+
+### Rule Evaluation
+
+The span decoration engine follows these rules:
+
+1.  **Order of Evaluation:** Rules are evaluated in the order they appear in the `spanDecorations` array.
+2.  **First Match Wins:** The first rule that matches a span is used, and subsequent rules are ignored.
```
yurishkuro: What is your proposal for dealing with user-config vs. build-it mapping?

packages/jaeger-ui/docs/span-decorations.md
```diff
@@ -0,0 +1,67 @@
...
+
+The `icon` property can be one of:
+
+1.  **Built-in Tokens:** Pre-defined icons bundled with Jaeger UI (e.g., `io.Database`, `io.Server`, `io.Cloud`).
+2.  **External URLs:** A full URL starting with `http://` or `https://` pointing to an image (SVG, PNG, etc.).
+3.  **Local Paths:** A path starting with `/` pointing to an asset served by the Jaeger UI server.
+
+#### Supported Built-in Tokens
+
+- `io.Server`: General server operation.
```
yurishkuro: The list in `packages/jaeger-ui/src/components/common/IconProvider.tsx` is larger than here. I would prefer the docs just point to the source of truth.

packages/jaeger-ui/docs/span-decorations.md
vibhor-5: I did try to look into a few other tracing services
- zipkin and a few other services dont have icons , some I couldnt access the trace page or trace page screenshots so not sure
- took a bit of inspiration from New Relic in their trace page they use similar icons for most of these 
- also similar to the icons in datadog except that they also do have source language specific icons 
I will integrate the source language and tool specific icons as well

packages/jaeger-ui/docs/span-decorations.md
vibhor-5: okay will change that

packages/jaeger-ui/src/components/common/IconProvider.tsx
vibhor-5: Sorry that's lazy of me , will fix that

packages/jaeger-ui/docs/span-decorations.md
vibhor-5: I think maybe we can prepend the user config to default config to keep the users config first as for the first match wins

sambhavi0:
Following this discussion one alternative approach worth considering: instead of prepending, provide a mode flag in the config like "spanDecorations": { "mode": "override" | "extend", "rules": [...] }. This makes user intent explicit, override replaces defaults entirely, extend prepends. It also avoids the ambiguity of the current array prepending approach when a user only wants to add one new rule. Happy to discuss tradeoffs

vibhor-5:
@sambhavi0 I am not sure how adding `mode` will make it more explicit as prepending the custom config by default will do `override` in case of exisiting rules as well as if the rule or intent for a type doesn't exist it will by default due to prepending be working as `extend`, I feel adding mode is only redundant.

sambhavi0:
> @sambhavi0 I am not sure how adding `mode` will make it more explicit as prepending the custom config by default will do `override` in case of exisiting rules as well as if the rule or intent for a type doesn't exist it will by default due to prepending be working as `extend`, I feel adding mode is only redundant.

that's fair, prepending naturally handles both cases without an explicit flag. The ambiguity I was thinking about is more of a UX concern than a technical one: a user reading the config documentation might not immediately understand that prepending gives them partial override behavior. But if the documentation makes this behavior explicit, the mode flag becomes redundant.

yurishkuro:
thanks for contributing. The feature was implemented in #3832

PR #3828, opened 2026-05-08
feat(SearchResults): add compact table view as alternative to list view

### LINKED ISSUE #3816: [Feature]: Compact table view as alternative to list view
## Problem

`SearchResults/index.tsx` renders search results as a vertical list of `ResultItem` cards:

```tsx
<ul className="ub-list-reset">
  {traces.map(trace => (
    <li key={trace.traceID}>
      <ResultItem
        durationPercent={trace.duration / maxTraceDuration}
        isInDiffCohort={/* cohort logic */}
        linkTo={/* link object */}
        toggleComparison={/* handler */}
        trace={trace}
        disableComparision={disableComparisons}
      />
    </li>
  ))}
</ul>
```

`ResultItem` uses an antd `Row/Col` grid and renders each trace as a tall card. When a search returns 20 or more results, users must scroll through 20 tall cards to find the trace they want. There is no compact alternative and no view toggle.

For GenAI workflows this is especially limiting: a developer running repeated agent evaluations may have dozens of traces from the same service in the result set and needs to scan across columns (error count, duration, model used) rather than reading each card top to bottom.

Relates to jaegertracing/jaeger#8401 (Deliverable 5 - Table of Traces).

---

### 1. Create `SearchResults/TraceTable.tsx`

A new component using the antd `Table` component. Props:

```typescript
type TraceTableProps = {
  traces: IOtelTrace[];
  maxTraceDuration: number;
  onRowClick: (traceID: string) => void;
};
```

Columns: Trace Name (linkable to trace page), Services, Span Count, Error Count, Duration, Start Time.

Error count computed as `trace.spans.filter(s => s.status.code === 'ERROR').length`.

The table's `onChange` callback must call the existing `handleSortChange` from `SearchResults` so the scatter plot sort and the table sort stay in sync. Using antd `Table`'s built-in column sort in isolation would bypass `handleSortChange` and cause inconsistency between the two views.

### 2. Add view toggle to `SearchResults/index.tsx`

Add `viewMode: 'list' | 'table'` to local state, defaulting to `'list'`. Add a segmented control or two icon buttons to the header row. Conditionally render `TraceTable` or the existing `ul` based on `viewMode`.

The default must stay `'list'` so existing users see no change.

---

## Required Tests

| Test case | Expected |
|---|---|
| `TraceTable` renders correct row count for a 5-trace fixture | 5 rows |
| `TraceTable` shows correct error count for traces with errored spans | matches filtered count |
| View toggle switches between list and table rendering | each mode renders its component |
| Sorting via table column header calls `handleSortChange` | sort handler invoked |

---

## Regression Risks

Low. The existing list view is entirely unchanged. `TraceTable` is a new file. The only modification to `SearchResults/index.tsx` is adding local state and a toggle button. The toggle defaults to `'list'`, so the feature is invisible until a user explicitly opts in.

---

## Related

- jaegertracing/jaeger#8401 - umbrella issue (Deliverable 5)
- #3813 - GenAI detection utility (table can show a GenAI indicator column when `trace.isGenAITrace` is true)

swetalin-10:
Adds a table view option to the search results page alongside the existing list view. Part of jaegertracing/jaeger#8401 (D5), resolves #3816.

The list view is the default. A `Radio.Group` toggle in the results header lets users switch to a compact table showing Trace Name, Services, Spans, Errors, Duration, and Start Time. Sorting works on Spans, Duration, and Start Time columns via Ant Design Table built-ins. Clicking any row navigates to the trace page using the same `goToTrace` handler the list view uses.

The table is only shown when `traceResultsView` is true, so the DDG view is unaffected. No changes to existing list rendering.

Tested with 3 unit tests in `TraceTable.test.tsx`: row count, error count rendering, and row click callback.


swetalin-10:
Hi! All Copilot review comments from all rounds have been addressed and all tests are passing. Could a maintainer please add a `changelog:skip` label so the `check-label` check passes? Ready for review whenever you get a chance. Thanks!

yurishkuro:
Thank you @swetalin-10 for the groundwork on this! Your PR was used as the starting point and the implementation was completed in #3980, which will close this issue on merge. I really appreciate the effort you put in.

PR #3836, opened 2026-05-09
fix(search): Surface per-trace errors in multi-trace search

### LINKED ISSUE #51: Multi-trace traceID search does not show 404s

WasThatRudy:
## Problem

Multi-trace search (`?traceID=a,b,c`) silently drops any traceID the backend can't return. The query API answers with `{ data, errors: [{msg, traceID}] }` so the per-trace failure is on the wire, but `searchDone` in `reducers/trace.ts` only iterates `payload.data`, so failed IDs never make it into `state.traces` or `state.search.results` and the UI shows no indication. `fetchMultipleTracesDone` in the same file already ingests `payload.errors`, so this is essentially a parity gap.

## Fix

Extend `searchDone` to walk `payload.errors`, store each entry as `{id, error, state: ERROR}` in the trace map, and push the failed IDs into `search.results` so downstream selectors see them. `stateTraceXformer` in `SearchTracePage/index.tsx` now splits the IDs into successful traces and errored ones, exposing a new `traceErrors` prop that flows into `SearchResults`. The list view renders an error row per failed ID alongside the successful results, using a small new `ResultItemError` component rather than overloading `ResultItem`. Empty-state handling is updated so a search that returns only errors shows the error rows instead of "No trace results."

The trace-diff cohort path (`fetchMultipleTraces`) already handled errors and is left alone. `rawTraces` (used for the JSON download) continues to contain only successful results.

## Testing

- Reducer test in `trace.test.js` for a `searchDone` payload that mixes `data` and `errors`.
- `mapStateToProps` test in `SearchTracePage/index.test.jsx` asserting the selector returns errored IDs in `traceErrors` and excludes them from `traces`.
- `SearchResults/index.test.jsx` renders error rows for each `traceErrors` entry, including the case where no successful traces are present.
- Unit test for `ResultItemError` covering the label, trace ID, and optional message.

`npm run lint`, `npm run tsc-lint`, `npm test`, `npm run build` green locally on Node 24.

Fixes #51

## AI Usage in this PR

- [x] **Specific parts**: AI assisted with code generation on parts of the change, reviewed and verified before submission.

### DIFF
--- a/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ResultItemError.css
+++ b/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ResultItemError.css
@@ -1,0 +1,28 @@
+/*
+Copyright (c) 2026 The Jaeger Authors.
+SPDX-License-Identifier: Apache-2.0
+*/
+
+.ResultItemError {
+  border: 1px solid var(--feedback-error);
+  background: color-mix(in srgb, var(--feedback-error) 6%, var(--surface-primary));
+  padding: 0.75rem 1rem;
+}
+
+.ResultItemError--label {
+  color: var(--feedback-error);
+  font-weight: 600;
+  margin-right: 0.4rem;
+}
+
+.ResultItemError--traceID {
+  color: var(--text-primary);
+  font-family: monospace;
+  word-break: break-all;
+}
+
+.ResultItemError--message {
+  color: var(--text-secondary);
+  margin-top: 0.25rem;
+  font-size: 0.9em;
+}
--- a/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ResultItemError.test.jsx
+++ b/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ResultItemError.test.jsx
@@ -10,7 +10,7 @@ import ResultItemError from './ResultItemError';
 describe('<ResultItemError>', () => {
   it('renders the trace ID and error label', () => {
     render(<ResultItemError traceID="abc123" message="trace not found" />);
-    expect(screen.getByText('Trace not found:')).toBeInTheDocument();
+    expect(screen.getByText('Trace fetch failed:')).toBeInTheDocument();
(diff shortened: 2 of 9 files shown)


WasThatRudy:
Pushed d29f4c9 addressing the review:

- Reducer now stores `new Error(msg)` so the displayed message is just the backend reason, without the redundant `Error:` prefix and inlined trace ID.
- Card label is `Trace fetch failed` rather than `Trace not found`, since `errors[].msg` can cover failures beyond 404.
- Error rows are lifted out of the `traceResultsView` gate, so they stay visible when the user switches to the DDG view.
- Header now appends `({n} failed)` next to the trace count so the count matches what is rendered.

Tests added for the DDG-view case and the header count.

yurishkuro:
We are planning to switch the UI to use api/v3/ for querying. I am not sure if that API also returns "not found" for individual trace IDs.

WasThatRudy:
> We are planning to switch the UI to use api/v3/ for querying. I am not sure if that API also returns "not found" for individual trace IDs.

checked v3, tracesdata is just resourceSpans, no error envelope, no multiID endpoint. so under v3 multiID lookups would presumably fan out to /traces/:traceId and surface 404s per call.if that's the plan, the UI rendering and selector split here carry over. only the reducer change is throwaway. happy to land this for the legacy path now and drop the reducer bit when v3 lands, or hold if you'd rather wait.

WasThatRudy:
@yurishkuro any thoughts on the v3 question above? happy to land this for the legacy path now or hold for the v3 migration, whichever you'd prefer.

yurishkuro:
the search is already migrated to v3

PR #3843, opened 2026-05-10
feat(SearchResults): add compact table view with list/table toggle

### LINKED ISSUE #3816: [Feature]: Compact table view as alternative to list view

bhuvan-somisetty:
Adds a compact table view to the search results page as an alternative to the existing card list. Users can switch between the two views using a List/Table toggle added to the results header.

The table shows one row per trace with six columns: trace name (linked to the detail page), service count, span count, error count (shown as a red tag when non-zero), duration, and start time. All columns except trace name are sortable. Pagination kicks in at 100 results.

The existing list view is completely unchanged. The toggle defaults to List so there is no visible difference for existing users until they opt in. The TraceTable component is self-contained and does not touch any shared state.

Covered by seven unit tests: row count, span count, error tag rendering, no-error case, service count, trace link href, and traceID fallback when traceName is empty. All 178 SearchTracePage tests pass with no regressions. TypeScript reports no errors.

Closes #3816

### DIFF
--- a/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/TraceTable.css
+++ b/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/TraceTable.css
@@ -1,0 +1,3 @@
+.TraceTable {
+  margin-top: 1rem;
+}
--- a/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/TraceTable.test.tsx
+++ b/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/TraceTable.test.tsx
@@ -1,0 +1,140 @@
+// Copyright (c) 2026 The Jaeger Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+import * as React from 'react';
+import { render, screen } from '@testing-library/react';
+import { MemoryRouter } from 'react-router-dom';
+
+import TraceTable from './TraceTable';
+import { IOtelTrace, StatusCode, SpanKind } from '../../../types/otel';
+
+vi.mock('react-router-dom', async () => {
+  const { MemoryRouter: ActualMemoryRouter } = await vi.importActual('react-router-dom');
+  return {
+    MemoryRouter: ActualMemoryRouter,
+    Link: ({ to, children }: any) => {
+      const href = typeof to === 'string' ? to : `${to.pathname}${to.search || ''}`;
+      return <a href={href}>{children}</a>;
+    },
+  };
+});
+
+function makeTrace(overrides: Partial<IOtelTrace> = {}): IOtelTrace {
+  return {
+    traceID: 'abc123',
+    traceName: 'test-trace',
+    spans: [],
+    services: [],
+    duration: 1000000 as IOtelTrace['duration'],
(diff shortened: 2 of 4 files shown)


yurishkuro:
Thank you @bhuvan-somisetty for working on this! The feature has been completed in #3980 (built on top of #3828), which will close this issue on merge. Your effort and interest in contributing is appreciated.

Issue #3820, opened 2026-05-07
[feature] GenAI Observability: Specialized Trace Visualization


yurishkuro:
### Requirement

This is a part of https://github.com/jaegertracing/jaeger/issues/8401, for easier house-keeping in the UI project.

PR #3856, opened 2026-05-11
feat(utils): add GenAI span classification and isGenAITrace detection

### LINKED ISSUE #3813: feat(OtelTraceFacade): add GenAI span classification and isGenAITrace detection utility
## Problem

No `gen_ai.*` attribute detection exists anywhere in the jaeger-ui codebase today.

`OtelTraceFacade` constructs the span map but performs no classification pass:

```typescript
// src/model/OtelTraceFacade.ts (constructor, simplified)
this._spans = trace.spans.map(s => new OtelSpanFacade(s));
this._spanMap = new Map(this._spans.map(s => [s.spanID, s]));
this._rootSpans = trace.roots.map(id => this._spanMap.get(id)).filter(Boolean) as IOtelSpan[];
```

Without a shared classification utility, every downstream component that needs to know whether a span is an LLM call, a tool call, an agent invocation, or a standard infrastructure span must independently scan `span.attributes` for `gen_ai.*` keys. This means:

- **O(n × attributes) duplicated work** — each component re-scans the same spans on every render cycle.
- **Divergent heuristics** — if the detection logic changes (e.g. a new `gen_ai.operation.name` value is added to the OTel semconv), every component must be updated independently.
- **No shared cache** — the UI cannot gate any GenAI-specific behaviour (icons, logical view toggle, rich-media rendering) on a single authoritative boolean.

Relates to jaegertracing/jaeger#8401 (Deliverable 1 — Automatic GenAI Mode Detection).

---

### 1. Create `src/utils/genai/detect.ts`

```typescript
import type { IOtelSpan, IOtelTrace } from '../../types/otel';

export type GenAISpanKind =
  | 'LLM_CALL'
  | 'TOOL_CALL'
  | 'AGENT'
  | 'RETRIEVAL'
  | 'UNKNOWN_GENAI'
  | 'STANDARD';

const OPERATION_NAME_MAP: Record<string, GenAISpanKind> = {
  chat: 'LLM_CALL',
  text_completion: 'LLM_CALL',
  execute_tool: 'TOOL_CALL',
  invoke_agent: 'AGENT',
  retrieval: 'RETRIEVAL',
};

export function classifySpan(span: IOtelSpan): GenAISpanKind {
  const attrs = span.attributes ?? [];
  const opName = attrs.find(a => a.key === 'gen_ai.operation.name')?.value;
  if (opName) {
    return OPERATION_NAME_MAP[String(opName)] ?? 'UNKNOWN_GENAI';
  }
  const hasGenAI = attrs.some(a => a.key.startsWith('gen_ai.'));
  return hasGenAI ? 'UNKNOWN_GENAI' : 'STANDARD';
}

export function isGenAITrace(trace: IOtelTrace): boolean {
  return trace.spans.some(s => classifySpan(s) !== 'STANDARD');
}
```

### 2. Expose `isGenAITrace` on `OtelTraceFacade`

Add a `readonly isGenAITrace: boolean` property computed once at construction time (or lazily via a getter backed by a private field). This makes the result available to `SpanBarRow`, `VirtualizedTraceView`, `TraceViewSettings`, and `SearchResults` without any of them importing `isGenAITrace` directly and re-scanning spans.

```typescript
// src/model/OtelTraceFacade.ts
import { isGenAITrace } from '../utils/genai/detect';

export class OtelTraceFacade implements IOtelTrace {
  // ... existing fields
  readonly isGenAITrace: boolean;

  constructor(trace: Trace) {
    this._spans = trace.spans.map(s => new OtelSpanFacade(s));
    this._spanMap = new Map(this._spans.map(s => [s.spanID, s]));
    this._rootSpans = trace.roots.map(id => this._spanMap.get(id)).filter(Boolean) as IOtelSpan[];
    this.isGenAITrace = isGenAITrace(this); // single O(n) pass
  }
}
```

---

## Required Tests

| Test case | Expected result |
|---|---|
| `classifySpan` with `gen_ai.operation.name = 'chat'` | `'LLM_CALL'` |
| `classifySpan` with `gen_ai.operation.name = 'text_completion'` | `'LLM_CALL'` |
| `classifySpan` with `gen_ai.operation.name = 'execute_tool'` | `'TOOL_CALL'` |
| `classifySpan` with `gen_ai.operation.name = 'invoke_agent'` | `'AGENT'` |
| `classifySpan` with `gen_ai.operation.name = 'retrieval'` | `'RETRIEVAL'` |
| `classifySpan` with unknown `gen_ai.operation.name` but other `gen_ai.*` attrs | `'UNKNOWN_GENAI'` |
| `classifySpan` with no `gen_ai.*` attrs | `'STANDARD'` |
| `isGenAITrace` with mixed trace (at least one GenAI span) | `true` |
| `isGenAITrace` with all-STANDARD trace | `false` |
| `OtelTraceFacade.isGenAITrace` matches `isGenAITrace(trace)` for same input | identical |

---

## Regression Risks

**Low.** `detect.ts` is a new file with no side effects. The `OtelTraceFacade` change adds one readonly property and does not modify any existing method or getter. The O(n) classification pass runs once at construction time — the same point where `_spans` and `_spanMap` are already being built.

---

## Downstream Dependents

Once this utility is merged, the following issues can reference `classifySpan` and `OtelTraceFacade.isGenAITrace` directly:

- #3810 — Agentic span icons in `SpanBarRow`
- #3808 — Rich-media attribute rendering in `SpanDetail`
- Logical view toggle in `VirtualizedTraceView` and `TraceViewSettings`
- GenAI-aware content routing in `SpanDetailSidePanel`

swetalin-10:

Part of #3820, closes #3813

Other GenAI PRs need a way to know whether a span or trace involves GenAI work. This adds that utility in one place so each component does not have to independently scan attributes.

## Description of the changes

`src/utils/genai/detect.ts` exports three functions:

- `isGenAISpan(span)` — returns `true` if the span has any `gen_ai.*` attribute.
- `classifySpan(span)` — maps `gen_ai.operation.name` to a `GenAISpanKind`. Known operations (`chat`, `text_completion`, `generate_content`, `embeddings`, `execute_tool`, `invoke_agent`, `create_agent`, `invoke_workflow`, `retrieval`) get specific kinds; anything else with a `gen_ai.*` attribute gets `UNKNOWN_GENAI`; spans with no `gen_ai.*` attributes get `STANDARD`.
- `isGenAITrace(spans)` — returns `true` if any span is not `STANDARD`.

Both functions accept a narrow `{ attributes: ReadonlyArray<IAttribute> }` shape rather than the full `IOtelSpan`, so they work with plain objects in tests without any type casting.

`OtelTraceFacade` now exposes a `readonly isGenAITrace: boolean` property computed once in the constructor after the spans array is built. This means downstream components can read `trace.isGenAITrace` without triggering repeated O(n) attribute scans on each render.

27 tests across two files:
- `detect.test.ts` — each operation name mapping, the unknown-name fallback, the generic attribute fallback, `isGenAISpan` true/false cases, mixed traces, all-standard traces, and empty span list.
- `OtelTraceFacade.test.ts` — `isGenAITrace` is `false` for a plain trace and `true` when at least one span carries a `gen_ai.*` tag.

`npm run lint`, `npm test` pass.

## Checklist


packages/jaeger-ui/src/utils/genai/detect.ts
```diff
@@ -0,0 +1,26 @@
...
+  invoke_agent: 'AGENT',
+  retrieval: 'RETRIEVAL',
+};
+
+export function classifySpan(span: IOtelSpan): GenAISpanKind {
+  const opAttr = span.attributes.find(a => a.key === 'gen_ai.operation.name');
+  if (opAttr) {
+    return OPERATION_TO_KIND[String(opAttr.value)] ?? 'UNKNOWN_GENAI';
+  }
+  return span.attributes.some(a => a.key.startsWith('gen_ai.')) ? 'UNKNOWN_GENAI' : 'STANDARD';
```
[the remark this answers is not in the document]
yurishkuro: I agree with this suggestion, we need to be conscious of performance of these functions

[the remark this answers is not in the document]
swetalin-10: Thanks for the feedback! This is already addressed in the latest commit -replaced the two-pass `find()` + `some()` with a single `for...of` loop that does both checks in one pass: early-returns when it hits `gen_ai.operation.name`, and tracks the `hasGenAI` flag along the way.

Also addressed the other Copilot points while at it:
- `OPERATION_TO_KIND` tightened to `Partial<Record<string, GenAISpanKind>>` so the `??` fallback is properly type-safe
- Parameter narrowed from `IOtelSpan` to `{ attributes: ReadonlyArray<IAttribute> }` - that's all these functions actually read
- `as unknown as IOtelSpan` double-cast removed from the test helper; the narrower parameter type makes a plain object literal work directly

All 12 tests pass.

packages/jaeger-ui/src/utils/genai/detect.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export type GenAISpanKind = 'LLM_CALL' | 'TOOL_CALL' | 'AGENT' | 'RETRIEVAL' | 'UNKNOWN_GENAI' | 'STANDARD';
+
+const OPERATION_TO_KIND: Record<string, GenAISpanKind> = {
+  chat: 'LLM_CALL',
+  text_completion: 'LLM_CALL',
+  execute_tool: 'TOOL_CALL',
+  invoke_agent: 'AGENT',
+  retrieval: 'RETRIEVAL',
+};
```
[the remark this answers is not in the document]
swetalin-10: Already addressed — OPERATION_TO_KIND is now typed as Partial<Record<string, GenAISpanKind>> (see line 10), so index access correctly returns GenAISpanKind | undefined and the ?? 'UNKNOWN_GENAI' fallback is type-safe.

packages/jaeger-ui/src/utils/genai/detect.test.ts
```diff
@@ -0,0 +1,75 @@
+// Copyright (c) 2026 The Jaeger Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+import { classifySpan, isGenAITrace } from './detect';
+import type { IOtelSpan } from '../../types/otel';
+
+function makeSpan(attrs: { key: string; value: string }[]): IOtelSpan {
+  return { attributes: attrs } as unknown as IOtelSpan;
```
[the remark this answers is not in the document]
swetalin-10: Addressed — classifySpan() and isGenAITrace() now accept SpanAttrs ({ attributes: ReadonlyArray<IAttribute> }) instead of the full IOtelSpan. The test helper makeSpan() returns that narrower type directly, so no cast is needed.

packages/jaeger-ui/src/utils/genai/detect.test.ts
```diff
@@ -0,0 +1,76 @@
+// Copyright (c) 2026 The Jaeger Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+import { classifySpan, isGenAITrace } from './detect';
+import type { IOtelSpan } from '../../types/otel';
+
+function makeSpan(attrs: { key: string; value: string }[]): IOtelSpan {
+  const span = { attributes: attrs } satisfies IOtelSpan;
+  return span;
```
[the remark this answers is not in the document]
swetalin-10: Addressed — makeSpan() now returns { attributes: IAttribute[] } (not IOtelSpan) and there is no satisfies usage. The function signature change in classifySpan() means tests only need the attributes field.

packages/jaeger-ui/src/utils/genai/detect.ts
```diff
@@ -0,0 +1,41 @@
...
+  embeddings: 'LLM_CALL',
+  execute_tool: 'TOOL_CALL',
+  invoke_agent: 'AGENT',
+  create_agent: 'AGENT',
+  invoke_workflow: 'AGENT',
+  retrieval: 'RETRIEVAL',
+};
+
+export function isGenAISpan(span: SpanAttrs): boolean {
+  return span.attributes.some(a => a.key.startsWith('gen_ai.'));
```
anshgoyalevil: Lets also centralize this string instead of hardcoding it inline.

packages/jaeger-ui/src/utils/genai/detect.ts
```diff
@@ -0,0 +1,41 @@
...
+}
+
+export function classifySpan(span: SpanAttrs): GenAISpanKind {
+  let hasGenAI = false;
+  for (const attr of span.attributes) {
+    if (attr.key === 'gen_ai.operation.name' && typeof attr.value === 'string') {
+      return OPERATION_TO_KIND[attr.value] ?? 'UNKNOWN_GENAI';
+    }
+    if (!hasGenAI && attr.key.startsWith('gen_ai.')) {
```
anshgoyalevil: Same here

packages/jaeger-ui/src/utils/genai/detect.ts
```diff
@@ -0,0 +1,41 @@
...
+      hasGenAI = true;
+    }
+  }
+  return hasGenAI ? 'UNKNOWN_GENAI' : 'STANDARD';
+}
```
anshgoyalevil: This function seems hard to read. Can we simplify this?

packages/jaeger-ui/src/utils/genai/detect.ts
swetalin-10: Done — extracted into a `SpanAttributeNamespace` enum in `src/constants/span-attributes.ts` (same file as the span-icons PR). `isGenAISpan` now uses `SpanAttributeNamespace.GEN_AI` instead of the inline string.

packages/jaeger-ui/src/utils/genai/detect.ts
swetalin-10: Same fix applied here — `classifySpan` now uses `SpanAttributeNamespace.GEN_AI` for the prefix check. Also added `GEN_AI_OPERATION_NAME = 'gen_ai.operation.name'` as a named constant in the same file so that key is centralized too.

packages/jaeger-ui/src/utils/genai/detect.ts
swetalin-10: Simplified further — replaced the manual loop + `hasGenAI` flag with a `find()` for the operation name and a call to `isGenAISpan()` for the fallback. No mutable state, and the two-step logic is immediately readable:
1. If the span has a known `gen_ai.operation.name`, map it.
2. Otherwise fall back to `isGenAISpan()` to distinguish `UNKNOWN_GENAI` from `STANDARD`.

packages/jaeger-ui/src/constants/span-attributes.ts
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) 2026 The Jaeger Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+export enum SpanAttributeNamespace {
```
yurishkuro: why is this not a string const?

packages/jaeger-ui/src/utils/genai/detect.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export function classifySpan(span: SpanAttrs): GenAISpanKind {
+  const opAttr = span.attributes.find(({ key }) => key === GEN_AI_OPERATION_NAME);
+  if (opAttr && typeof opAttr.value === 'string') {
+    return OPERATION_TO_KIND[opAttr.value] ?? 'UNKNOWN_GENAI';
+  }
+  return isGenAISpan(span) ? 'UNKNOWN_GENAI' : 'STANDARD';
+}
+
+export function isGenAITrace(spans: ReadonlyArray<SpanAttrs>): boolean {
```
yurishkuro, line 42: all these methods do the same thing - loop over attributes. You already made a good change in otel.ts to compute this once and expose on the trace. You can extend this to spans as well and compute all of these things in a single pass over attributes when building OTEL wrapper.

packages/jaeger-ui/src/constants/span-attributes.ts
swetalin-10: Fair — a single-member enum added nothing over a plain const. Replaced `SpanAttributeNamespace` with `export const GEN_AI_NAMESPACE = 'gen_ai.'`.

packages/jaeger-ui/src/utils/genai/detect.ts
swetalin-10: Done — OtelSpanFacade now computes genAIKind once in its constructor via classifySpan, and exposes it through IOtelSpan, the same pattern as isGenAITrace on the trace. classifySpan itself is also now a genuine single pass over attributes (checks the operation-name key and the gen_ai.* prefix together in one loop, instead of a find() followed by a fallback scan). OtelTraceFacade.isGenAITrace now reads the cached per-span genAIKind values instead of re-scanning each span's attributes again at the trace level. Added test coverage for the new genAIKind classification on OtelSpanFacade directly.

packages/jaeger-ui/src/types/otel.ts
```diff
@@ -17,6 +17,8 @@ export enum StatusCode {
   ERROR = 'ERROR',
 }
 
+export type GenAISpanKind = 'LLM_CALL' | 'TOOL_CALL' | 'AGENT' | 'RETRIEVAL' | 'UNKNOWN_GENAI' | 'STANDARD';
```
yurishkuro: iiuc `STANDARD` actually means "not GenAI". I think this is not a good nomenclature - we should simply have this field in the span to be `?GenAISpanKind` and return undefined when it's not GenAI

yurishkuro (commented):
PR #4122 (merged 2026-06-30) already added packages/jaeger-ui/src/utils/genai/index.ts:
```
export function isGenAISpan(span: IOtelSpan): boolean {
  return span.attributes.some(attr => attr.key.startsWith('gen_ai.'));
}
```
It's wired into production in `SpanDetail/index.tsx (showGenAITab = isGenAISpan(span))`.

Before merge, this should be reconciled: fold `utils/genai/index.ts` into this module and repoint SpanDetail at it.

packages/jaeger-ui/src/types/otel.ts
swetalin-10: Fixed in 946f8461 — `GenAISpanKind` no longer includes `STANDARD`; `IOtelSpan.genAIKind` is now `GenAISpanKind | undefined`, and `classifySpan`/`isGenAISpan`/`isGenAITrace` return/check `undefined` instead of the sentinel. Updated all call sites (`OtelSpanFacade`, `OtelTraceFacade`) and test fixtures accordingly.

swetalin-10:
Fixed both remaining points in 946f8461:

**1. `STANDARD` sentinel** — `GenAISpanKind` now only lists actual GenAI kinds. `IOtelSpan.genAIKind` is `GenAISpanKind | undefined`, and `classifySpan`/`isGenAISpan`/`isGenAITrace` return/check `undefined` instead of `'STANDARD'`. Updated `OtelSpanFacade`, `OtelTraceFacade`, and the affected test fixtures.

**2. Reconciling `utils/genai/index.ts`** — folded the temporary stub into this module: `utils/genai/index.ts` now re-exports from `detect.ts` instead of duplicating `isGenAISpan`. Since `SpanDetail/index.tsx` already imports from the folder path (`../../../../utils/genai`), it now resolves to the canonical `detect.ts` logic with no change needed to `SpanDetail` itself. Removed the now-redundant `index.test.ts`; its one test case not already covered (keys that merely contain `gen_ai` mid-string, e.g. `custom.gen_ai.tag`) is now in `detect.test.ts`.

74 tests passing across the affected files, 729 passing across `utils/`, `model/`, and `SpanDetail/` more broadly. Clean `tsc-lint`/`oxlint`.

yurishkuro (approved):

PR #3862, opened 2026-05-11
feat(ui): render GenAI attributes with rich JSON tree

### LINKED ISSUE #3829: [Feature]: Better rendering of SpanDetail with GenAI attributes
## Problem

`SpanDetail/index.tsx` passes all span attributes to `AccordionAttributes`, which renders every attribute as a flat key-value row. For GenAI spans that follow OTel semantic conventions, `gen_ai.input.messages` and `gen_ai.output.messages` are serialized JSON arrays that can be multi-KB in size (e.g., a 20-turn conversation history). The current flat renderer truncates these values, making the primary diagnostic content of an AI span completely unreadable.

Affected attributes (defined in the [OTel GenAI semconv](https://opentelemetry.io/docs/specs/semconv/gen-ai/)):
- `gen_ai.input.messages` — prompt messages as a JSON array
- `gen_ai.output.messages` — completion messages as a JSON array
- `gen_ai.tool.call.arguments` — tool input arguments as a JSON object
- `gen_ai.tool.call.result` — tool output as a JSON object
- `gen_ai.retrieval.documents` — retrieved RAG documents as a JSON array

## Why This Is Worth a Standalone Fix

`react-json-view-lite` (v2.5.0) is **already listed in `package.json`** — no new dependency is needed. The fix is contained entirely within `SpanDetail/index.tsx`: detect attributes whose keys match the list above and render them through `react-json-view-lite` instead of the flat `AccordionAttributes` row.

This improvement benefits any user who already sends OTel-instrumented GenAI traces to Jaeger today, independently of the larger GenAI Observability work tracked in jaegertracing/jaeger#8401.

## Code Location

- Renderer: `packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/index.tsx`
- Dependency already available: `react-json-view-lite` ^2.5.0 in `packages/jaeger-ui/package.json`

## Proposed Fix

1. Define a set of attribute keys that should receive JSON rendering:
   `gen_ai.input.messages`, `gen_ai.output.messages`, `gen_ai.tool.call.arguments`,
   `gen_ai.tool.call.result`, `gen_ai.retrieval.documents`, `gen_ai.tool.definitions`

2. In `SpanDetail/index.tsx`, split `span.attributes` into `richAttrs` (keys in the set above) and `standardAttrs` (everything else).

3. Render `richAttrs` through `JsonView` from `react-json-view-lite` and `standardAttrs` through the existing `AccordionAttributes` component unchanged.

4. Wrap the JSON renderer in a collapsed accordion so it only parses the value on expand (lazy load — avoids parsing multi-KB strings at trace render time).

## Acceptance Criteria

- `gen_ai.tool.call.arguments` renders as an interactive JSON tree in SpanDetail
- `gen_ai.input.messages` renders as an interactive JSON tree in SpanDetail
- All non-GenAI attributes continue to render identically to today (no regression)
- No new runtime dependencies added

Related: jaegertracing/jaeger#8401

Harizz076:
- Fixes https://github.com/jaegertracing/jaeger-ui/issues/3829
- `SpanDetail` passes all span attributes to `AccordionAttributes`, which renders every attribute as a flat key-value row. For GenAI spans that follow OTel semantic conventions, attributes like `gen_ai.input.messages` and `gen_ai.output.messages` are serialized JSON arrays that can be multi-KB in size. The current flat renderer truncates these values, making the primary diagnostic content unreadable.

## Description of the changes

I have revamped the UI to natively support GenAI spans rather than rendering them as raw JSON trees. 

### `packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/index.tsx`
- **Tabbed Interface:** Introduced a two-tab layout ("GenAI" and "Default") for spans matching GenAI semantic conventions. The "Default" tab retains the original layout to prevent regressions.
- **Dedicated Prompt/Completion Blocks:** In the GenAI tab, we now parse and extract the input and output messages to display them as dedicated `Input` and `Output` blocks instead of raw flat attributes.
- **Copy Functionality:** Added a "Copy" button directly to the Input/Output blocks to easily grab prompt and completion payloads.
- **Performance Optimizations:** Added `destroyInactiveTabPane={true}` to the Ant Design `Tabs` component so heavy attribute lists don't mount when viewing the GenAI tab. Memoized the payload parsing (`React.useMemo`) to prevent eager re-parsing on every render.
- **Accessibility Improvements:** Made the custom accordion headers fully keyboard accessible (`role="button"`, `tabIndex={0}`, `Enter`/`Space` key handlers) and updated ARIA states to use `aria-expanded` instead of `aria-checked`.

### `packages/jaeger-ui/src/utils/genai/detect.ts`
- **Metadata Extraction:** Added robust helper functions (`getGenAiMetadata`, `isGenAiSpan`) to extract provider, model, token counts, and parse the prompt/completion messages from the span attributes and events.
- **Strict Typing:** Improved type safety for `GenAiAttributeValue` to properly handle objects and arrays, fixing TypeScript errors under strict mode.

- Manually tested locally in the UI by rendering a mock trace with deeply nested GenAI semantic convention payloads.
- Verified that the "GenAI" tab properly parses the messages array and renders the formatted text with a working copy button.
- Verified that switching tabs successfully lazy-loads content and destroys inactive DOM elements for performance.
- Checked non-GenAI tags and confirmed no regressions on the Default tab.
- Passed Vitest suite across all packages (and updated the `SpanDetail` test suite to explicitly test the new tab switching behavior).

## Checklist

*(See the comments below for a screen recording of the new UI in action!)*


yurishkuro:
What kind of research have you done about a good UX for GenAI span display?

Harizz076:
> What kind of research have you done about a good UX for GenAI span display?

While writing my LFX proposal for the GenAI observability work I looked into existing tools for UX reference.

Traceloop / OpenLLMetry (https://traceloop.com/docs/openllmetry/introduction) is the most technically relevant since it already emits standard OTLP spans with gen_ai.* attributes and points directly at Jaeger today. Their trace view does not use a flat Tags table at all, instead giving each span a semantic type badge (WORKFLOW, TASK, CHAT, QUERY, EMBEDDING) and dedicated Prompt / LLM Data tabs.

image


Langfuse shows the UX pattern most worth drawing from. Input and Output are first-class sections at the top of the detail panel, not buried inside an attributes table. Long completions show an inline expand link instead of being cut off, token counts are shown on each span, and there is a Formatted / JSON toggle. 

image


The immediate gap on the Jaeger side is that gen_ai.input.messages shows up as a truncated one-liner in the Tags table. This PR is a small first step fixing the rendering layer for those specific attributes. The deeper structural work (dedicated prompt section, token display) is a bigger lift for a future iteration. I am also adding a full tentative list of platforms I'll be going through:Arize Phoenix, Langfuse, LangSmith, Braintrust, Opik, Traceloop / OpenLLMetry. So incase you think this is not good enough, please wait till Im done writing my proposal and I can update this PR with more knowledge that I will accumulate :)

bhuvan-somisetty (commented):
Hey @Harizz076, nice work overall — the lazy parse, per-row accordion state, and raw string fallback all look correct. Two things worth fixing before this merges:

**1. Duplicated constant — please reuse `RICH_MEDIA_ATTRIBUTE_KEYS` from `utils/genai/detect.ts`**

`GENAI_RICH_ATTRIBUTE_KEYS` in `AccordionGenAIAttributes.tsx` is a duplicate of `RICH_MEDIA_ATTRIBUTE_KEYS` already exported from `packages/jaeger-ui/src/utils/genai/detect.ts`. Having two sources of truth for the same list will drift over time. Please import from there instead of redefining it here.

**2. `gen_ai.system_instructions` is missing**

`detect.ts`'s `RICH_MEDIA_ATTRIBUTE_KEYS` includes `gen_ai.system_instructions` but your set doesn't. That attribute carries the LLM system prompt and definitely warrants rich rendering — it should be included.

**3. Minor — `role="switch"` is semantically incorrect**

`role="switch"` is meant for binary on/off toggles. An expand/collapse accordion header should use a `<button>` element (or at minimum `role="button"`) so screen readers announce it correctly.

Harizz076:
@bhuvan-somisetty Thanks for your feedback ive made the changes which your comment pointed out. 

@yurishkuro please let me know what you think about this PR, it is just a small change...

Harizz076:
@yurishkuro this is a small fix however will help in the lfx '26 T2, can you please provide some feedback on this? thanks

yurishkuro:
@Harizz076 the screenshot in the description is confusing and not at all close to the dedicated tools. It is helpful to see before/after comparison of changes.

Overall this might be improving specific JSON attribute view, but it's not what's needed for GenAI spans, they require different dedicated layouts altogether.

Harizz076:
> @Harizz076 the screenshot in the description is confusing and not at all close to the dedicated tools. It is helpful to see before/after comparison of changes.
> 
> 
> 
> Overall this might be improving specific JSON attribute view, but it's not what's needed for GenAI spans, they require different dedicated layouts altogether.

image


Hi @yurishkuro, do you think this is good for the GenAI span view? I'll edit the PR description and screenshots with the same.

yurishkuro:
yes this is pretty good, but I would think we may want multiple horizontal tabs in the Details panel, one for the default / normal view, and another GenAI customized view

Harizz076:
> we may want multiple horizontal tabs in the Details panel, one for the default / normal view, and another GenAI customized view

Do you mean we would want to show GenAI spans in both ways? Like a normal span and in the customised view...?

Harizz076:
https://github.com/user-attachments/assets/520bf9f6-22ca-4cdc-8bc4-db22025405a3

i've completely refactored the genai span details based on the feedback.

here is a summary of all the changes pushed:

- ripped out the old accordion component and replaced it with a clean tabbed layout. there's now a dedicated "GenAI" tab and a "Default" tab.
- the default tab now acts as a true standard view and shows all the raw gen_ai.* tags without artificially filtering them out.
- added missing copy buttons to both the input and output boxes so prompts are easy to grab.
- addressed all the copilot feedback.

Harizz076:
@yurishkuro PTAL at the screen recording ive added, pr description has been updated too. I really hope it can be merged now.... Do let me know if something else needs to be added/changed

Harizz076:
sorry was fixing some things agent freaked out lol, will open a cleaner PR

Issue #3815, opened 2026-05-06
feat(VirtualizedTraceView): add logical view filter to hide infrastructure spans in GenAI traces


bhuvan-somisetty:
## Problem

`generateRowStatesFromTrace` in `VirtualizedTraceView.tsx` has no filter path for span kind. Every span in a trace is always rendered as a row, regardless of whether it carries GenAI semantic convention attributes or is a plain infrastructure span (HTTP, database, gRPC).

The current signature (reconstructed from component behaviour):

```typescript
function generateRowStatesFromTrace(
  spans: ReadonlyArray<IOtelSpan>,
  detailStates: Map<string, TDetailState>,
  childrenHiddenIDs: Set<string>,
  detailPanelMode: SpanDetailPanelMode,
  prunedServices: Set<string>
): RowState[]
```

`prunedServices` filters by service name only. There is no parameter to filter by GenAI span kind.

For a GenAI trace with 50 spans, 40 of them might be infrastructure spans (HTTP calls to the LLM API, DB lookups, gRPC hops). These obscure the 10 reasoning spans that actually describe what the agent decided to do. A developer debugging the agent's reasoning has to manually scan through all 50 rows to find the relevant ones.

Relates to jaegertracing/jaeger#8401 (Deliverable 4 - Simplified Logical View).

---

### 1. Add `logicalViewEnabled` to trace timeline state

Add a `logicalViewEnabled: boolean` field to `src/types/TTraceTimeline.ts` and wire it into the Zustand layout preferences store so the toggle persists across span selections.

### 2. Extend `generateRowStatesFromTrace`

Add a `logicalViewEnabled` parameter. When `true`, skip rows where `classifySpan(span) === 'STANDARD'` unless the span has at least one descendant that is non-`STANDARD`. This preserves the tree structure: a STANDARD span that wraps a GenAI child must stay visible so the child remains in context.

```typescript
// pseudocode for the filter
if (logicalViewEnabled && classifySpan(span) === 'STANDARD') {
  const hasGenAIDescendant = span.childSpans.some(
    c => classifySpan(c) !== 'STANDARD'
  );
  if (!hasGenAIDescendant) continue; // prune from row list
}
```

When `logicalViewEnabled` is `false`, the function must behave identically to the current implementation with zero overhead.

### 3. Add toggle to `TraceViewSettings.tsx`

[...]

PR #3864, opened 2026-05-11
feat(genai): pre-configure service filter with GenAI services on trace load

bhuvan-somisetty:
When a trace contains GenAI spans, the service filter now automatically pre-selects only the services that own at least one `gen_ai.*` span. Non-GenAI services are pruned by default, giving the user a focused view of the AI reasoning flow without requiring manual filter configuration.

The priority order is unchanged: a `svcFilter` URL param takes precedence, then localStorage defaults, then this GenAI auto-filter, and finally show-all. Root service protection is enforced at every level via the existing `sanitizePrunedServices` guard, so the root span's service is never hidden.

The implementation adds `getGenAIServiceNames(trace)` to `detect.ts`, which scans spans once and returns the set of service names with any GenAI attribute. `resolveInitialFilter` receives this set as an optional fifth parameter and uses it as the fallback tier. The hook computes the set via `useMemo` only when `trace.isGenAITrace` is true, so there is zero overhead for non-GenAI traces.

Depends on #3826 (GenAI span iconography). Relates to #3815.

Tests cover all new paths: GenAI filter applied, URL param overrides it, localStorage overrides it, all-services-are-GenAI (no prune), root protection, `skipDefaults` bypass, and non-GenAI trace skips the filter entirely.


Harizz076:
Hey buddy, I was kinda already working on this issue as you might already know :) . Im sure me and the maintainers are grateful that youre putting so many PRs. But, since youre already 5 PRs above your quota, can you please close this PR and I can put mine? Hope you dont mind, Thanks

bhuvan-somisetty:
Closing this PR to stay within the project's PR limit for new contributors (limit: 1 for first-time contributors, per the contributing guidelines).

I also realized that the GenAI observability space in jaeger-ui is currently being worked on as part of the active LFX Term 2 mentorship, so I don't want to create conflicts or duplicate effort there.

Will focus on getting my existing open PR (#3814) reviewed and merged first before opening new ones. Happy to revisit this idea in the future if it's still relevant. Thanks for your time.

PR #3872, opened 2026-05-12
feat(TracePage): Add GenAITimelineViewer view type with auto-activation

swetalin-10:

Part of #3820

`ETraceViewType` had no entry for GenAI traces. Traces containing `gen_ai.*` spans landed in the standard waterfall with no automatic routing to a dedicated view.

## Description of the changes

- Adds `GenAITimelineViewer` to `ETraceViewType`. When a trace contains any span with a `gen_ai.*` attribute, the view auto-activates via `useEffect`. When navigating away to a non-GenAI trace, it auto-deactivates back to the standard timeline.
- The GenAI View entry is always visible in the dropdown menu, allowing users to manually view any trace through the GenAI lens and toggle back.
- For now `GenAITimelineViewer` renders the same `TraceTimelineViewer` component — behavioral differences will be introduced in later PRs.
- Introduces `viewTypeShowsMinimap()` and `viewTypeIsNavigable()` helpers in `types.ts` so that adding future timeline-based views only requires updating one canonical location.
- Replaces per-view tracking functions (`trackGanttView`, `trackGraphView`, etc.) with a single `trackViewChange(viewType)` that uses the enum value directly. This covers all views uniformly, including Flamegraph and GenAI which previously had no tracking.
- Removes unused `viewOptions` prop that was declared but never passed.

- Tests cover: auto-activation on GenAI traces, no auto-activation on plain traces, auto-deactivation when navigating to a non-GenAI trace, GenAI View always shown in dropdown, and unified tracking for all view types.
- `npm run lint`, `npm test`, `npm run build` pass.

## Checklist

- [x] **Heavy**: AI generated most or all of the code changes


packages/jaeger-ui/src/components/TracePage/index.tsx
```diff
@@ -371,6 +372,10 @@ export function TracePageImpl(props: TProps) {
     return <ErrorMessage className="ub-m3" error={trace.error || 'Unknown error'} />;
   }
 
+  const traceIsGenAI = data
+    .asOtelTrace()
+    .spans.some(s => s.attributes.some(a => a.key.startsWith('gen_ai.')));
```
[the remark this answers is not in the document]
swetalin-10: Fixed — wrapped in `useMemo` keyed on `data` so the scan only runs when the trace changes, not on every re-render.

packages/jaeger-ui/src/components/TracePage/index.tsx
```diff
@@ -479,6 +485,20 @@ export function TracePageImpl(props: TProps) {
...
+        scrollToFirstVisibleSpan={sm.scrollToFirstVisibleSpan}
+        findMatchesIDs={spanFindMatches}
+        trace={data.asOtelTrace()}
+        criticalPath={criticalPath}
+        updateNextViewRangeTime={updateNextViewRangeTime}
+        updateViewRangeTime={updateViewRangeTime}
+        viewRange={viewRange}
+        useOtelTerms={useOtelTerms}
+      />
+    );
```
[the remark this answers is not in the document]
swetalin-10: Consolidated — both `TraceTimelineViewer` and `GenAITimelineViewer` view types now share a single render branch that outputs `<TraceTimelineViewer>`. No duplication.

packages/jaeger-ui/src/components/TracePage/index.tsx
```diff
@@ -193,6 +193,7 @@ export function TracePageImpl(props: TProps) {
   const [slimView, setSlimView] = useState(() => Boolean(embedded?.timeline?.collapseTitle));
   const [viewType, setViewType] = useState<ETraceViewType>(ETraceViewType.TraceTimelineViewer);
   const [viewRange, setViewRange] = useState<IViewRange>({ time: { current: [0, 1] } });
+  const [genAiBannerDismissed, setGenAiBannerDismissed] = useState(false);
```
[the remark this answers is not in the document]
swetalin-10: Fixed — initialised from `sessionStorage.getItem('jaeger.genai-banner-dismissed') === 'true'` and on dismiss calls `sessionStorage.setItem('jaeger.genai-banner-dismissed', 'true')`. Persists across full-page reloads and all tabs.

jkowall:
Thanks @swetalin-10 for the contribution. This is a useful step toward giving GenAI traces a dedicated entry point.

I spotted a few things that look worth addressing before this is ready:

1. The GenAI banner and dropdown entry are gated on `backendCapabilities.aiAssistant` (`TracePage/index.tsx`). That capability means the in-app AI assistant sidecar is available, and it defaults to false. The GenAI trace view is client-side trace visualization, so deployments without the assistant enabled would still never see the new banner or manual "GenAI View" option even when the trace has `gen_ai.*` spans.

2. `sessionStorage.getItem` / `setItem` are called directly from `TracePage`. Those APIs can throw in some privacy or embedded contexts; nearby tracing code already wraps sessionStorage access for that reason. It would be safer to treat persistence as best-effort and fall back to in-memory dismissal state.

3. Clicking "Switch to GenAI View" hides the banner only because the current view is GenAI. If the user later switches to another view in the same session, the banner can appear again. Since the PR describes this as "shows once per session," the action button should probably mark it dismissed too.

I did a static review of the PR head and did not run the full local npm suite.

swetalin-10:
Thanks for the detailed review! All three points are addressed in the latest push @jkowall 

**1. Wrong capability gate**
The banner and "GenAI View" dropdown entry are now driven by `traceIsGenAI` (whether the trace has `gen_ai.*` spans) rather than `backendCapabilities.aiAssistant`. The AI sidecar flag only belongs on the in-app assistant features, not on client-side trace visualization.

**2. sessionStorage safety**
Extracted a `dismissGenAiBanner` callback that wraps `setItem` in a try/catch and falls back to in-memory-only dismissal when storage is unavailable — matching the pattern in `src/tracing/page-attribution.ts`. The `useState` initializer for the dismissed flag also wraps `getItem` in a try/catch and returns `false` on failure.

**3. Switch button should mark dismissed**
"Switch to GenAI View" now calls `dismissGenAiBanner()` before switching the view, so the banner does not reappear if the user later switches to a different view in the same session.

yurishkuro:
@swetalin-10 I am not sold on this approach. How exactly do you define what "GenAI viewer" is? Currently all viewers we use have completely different structure and data layout. The GenAI view, on the other hand, is just a customization of the main Timeline Viewer with potentially the following properties, which activate automatically upon detecting GenAI trace:
  * the header might display some additional GenAI stats like total token usage
  * the Gantt chart should be hidden by default
  * the span details should default to sidebar instead of inline (which should actually be automatic if there is no gantt chart).
  * the Span Details panel shows additional GenAI tab for GenAI only attributes and some high level span summary (again, token counts if present, etc), while the standard span details display goes to Default tab in the panel.
  
This is not the only possible solution to the problem, and maybe it's exactly what you mean by adding what looks like a high-level view switch but the implementation would still be based on the priumary TimelineViewer with custom configuration?

swetalin-10:
Thanks for the clarification @yurishkuro . Yes , the current implementation already renders the same TraceTimelineViewer component for both view types; GenAITimelineViewer is just a configuration state, not a structurally different viewer. I agree it should not sit in the same dropdown as TraceGraph and Flamegraph.

I'll rework the PR so the 4 customizations (header token stats, hide Gantt, sidebar panel mode, GenAI tab) activate automatically on GenAI trace detection - removing the opt-in banner and the GenAITimelineViewer
enum value.

One question: should there be a way for users to manually toggle the Gantt chart back on once it auto-hides, or should auto-detection be the only control?

yurishkuro:
The ability to toggle back from GenAI view to normal view is the main advantage of your approach. Since we're reusing the view implementation I think it's ok to keep this as a new option in the main View dropdown.

packages/jaeger-ui/src/components/TracePage/TracePageHeader/AltViewOptions.tsx
```diff
@@ -25,6 +25,7 @@ type Props = {
   traceID: string;
   disableJsonView: boolean;
   viewType: ETraceViewType;
+  showGenAIView: boolean;
```
yurishkuro: This is starting to turn into spaghetti settings where you already have `disableJsonView` and `showGenAIView`. Can we come up with a more generic properties mechanism to control the eligibility of dropdown selections?

swetalin-10:
Got it. Yes, GenAITimelineViewer reuses the same TraceTimelineViewer component underneath, it's just configuration on the same view, not a structurally different viewer. I'll keep it in the dropdown as the toggle-back mechanism, auto-activate on detection, and remove the opt-in banner.

packages/jaeger-ui/src/components/TracePage/TracePageHeader/AltViewOptions.tsx
swetalin-10: Agreed - adding a boolean per view doesn't scale. I'll replace disableJsonView and showGenAIView with a single EligibilityContext object passed as one prop, and attach an optional isEligible predicate to each MENU_ITEMS entry. Future conditional views just add an isEligible to their entry without touching Props. Will push the refactor along with the auto-activation changes.

packages/jaeger-ui/src/components/TracePage/TracePageHeader/AltViewOptions.tsx
```diff
@@ -20,42 +20,40 @@ import prefixUrl from '../../../utils/prefix-url';
...
+  isGenAITrace: boolean;
+  jsonEnabled: boolean;
+};
+
 type Props = {
   onTraceViewChange: (viewType: ETraceViewType) => void;
   traceID: string;
-  disableJsonView: boolean;
   viewType: ETraceViewType;
+  eligibility: EligibilityContext;
```
yurishkuro: nesting it does not make it generic, I meant some kind of map keyed by ETraceViewType

packages/jaeger-ui/src/components/TracePage/TracePageHeader/AltViewOptions.tsx
```diff
@@ -20,42 +20,40 @@ import prefixUrl from '../../../utils/prefix-url';
 import { ETraceViewType } from '../types';
 import { getTargetBlankOrTop } from '../../../utils/config/get-target';
 
+export type EligibilityContext = {
+  isGenAITrace: boolean;
```
yurishkuro: how about we don't suppress this option at all? Let people select GenAI if they want, it could actually be useful if for some reason our auto-detection misfires.

packages/jaeger-ui/src/components/TracePage/TracePageHeader/AltViewOptions.tsx
swetalin-10: Good point. Replaced EligibilityContext with viewOptions?: Partial<Record<ETraceViewType, boolean>> - a map keyed by view type where false hides that entry. Adding new conditional views now only requires passing a key in the map, with no change to the Props type.

packages/jaeger-ui/src/components/TracePage/TracePageHeader/AltViewOptions.tsx
swetalin-10: Done removed the isEligible predicate from GenAI View so it always appears in the dropdown. Auto-detection still switches to it automatically, but users can also select it manually at any time.

yurishkuro:
@swetalin-10 let me take over this PR to unblock you, because you're running into a series of legacy tech debt issues here.

yurishkuro (approved):

Issue #3265, opened 2025-12-30
[Feature]: Upgrade to OpenTelemetry terminology


yurishkuro:
## Requirement

Since the input tracing data is predominantly OpenTelemetry these days, the Jaeger UI should be using OTEL terminology like attributes, events, links, resource, etc., instead of legacy names like "tags", "logs", "process".

## Problem

This is a very large refactoring and needs to be done incrementally.

## Proposal

[ADR 0002: Making Jaeger UI OpenTelemetry-Native](https://github.com/jaegertracing/jaeger-ui/blob/main/docs/adr/0002-otlp-api-v3-migration.md)

## Help Wanted status

This is a tracking issue for a large project. PRs should only be submitted for individual checklist items below. Read the ADR document for details.

---

### Phase 1: OTEL Facade Layer ✅

- [x] Create OTEL facade types (`IOtelSpan`, `IOtelTrace`, `IResource`, `IAttribute`, `IEvent`, `ILink`, `IStatus`, `SpanKind`, `StatusCode`)
- [x] Implement `OtelSpanFacade` and `OtelTraceFacade` classes mapping legacy → OTEL properties
- [x] Create facade selectors and React hooks (`useOtelTrace`, `useOtelSpan`)
- [x] Test all property mappings and performance benchmarks
- [x] Terminology toggle feature flag (`useOpenTelemetryTerms` in config)

### Phase 2: Component Migration ✅

**Pilot**
- [x] `SpanDetail/KeyValuesTable` — Tags → Attributes
- [x] `SpanDetail/AccordianLogs` — Logs → Events

**Core Display**
- [x] `VirtualizedTraceView` — main trace view
- [x] `SpanBarRow` — span bars + tooltip terminology
- [x] `SpanDetailRow` — span details
- [x] `TraceTimelineViewer/utils` — `isErrorSpan`, `isKindClient`, etc.

**Statistics & Analysis**
- [x] `TraceStatistics/tableValues` — tag iteration + terminology
- [x] `TraceStatistics/PopupSql` — tag extraction
- [x] `TraceStatistics/index` — statistics display + terminology

**Search & Results**
- [x] `SearchResults/ResultItem` — service name display
- [x] `SearchTracePage` — trace list
- [x] `SearchForm` — "Operation" → "Span Name", "Tags" → "Attributes"

**Supporting Components**
- [x] `TracePageHeader`
- [x] `TraceFlamegraph`
[...]

PR #3890, opened 2026-05-14
feat(api/v3): Expose Zod schemas for OTLP trace/span types

gkhulbe4:
- Addresses one of the unchecked items under [#3265 Phase 3.5–3.7](https://github.com/jaegertracing/jaeger-ui/issues/3265): *"Complete Zod schema coverage for full trace/span wire types."*
- `packages/jaeger-ui/src/api/v3/generated-client.ts` already contains Zod schemas for the full OTLP trace/span surface (`TracesData`, `ResourceSpans`, `ScopeSpans`, `Span`, `Span_Event`, `Span_Link`, `Resource`, `InstrumentationScope`, `KeyValue`, `AnyValue`, `ArrayValue`, `KeyValueList`, `Status`), but `schemas.ts` only re-exports the services + operations response schemas. Consumers that need to validate trace responses (e.g. the React-Query `getTrace` hook tracked in #3835 / Milestone 3.2) currently have no clean import path. The README in `src/api/v3/` says *"schemas.ts simply re-exports the generated schemas"* — this PR makes that true.
- While writing the new tests I also found that the four tagged-union-like OTLP types (`AnyValue`, `ArrayValue`, `KeyValue`, `KeyValueList`) had their `.partial()` calls stripped by the blanket regex in `scripts/postprocess-schemas.cjs`. Their TypeScript types are correctly emitted as `Partial<{...}>`, but the matching Zod schemas demanded every variant field at once — a shape no real OTLP attribute value satisfies (an `AnyValue` carries only one of `stringValue`/`boolValue`/`intValue`/…). Fixed in this PR. The broader strict-vs-wire-format mismatch on the non-union top-level schemas is intentionally out of scope and filed as #3890 for follow-up.

## Description of the changes
- Addresses one of the unchecked items under [#3265 Phase 3.5–3.7](https://github.com/jaegertracing/jaeger-ui/issues/3265): *"Complete Zod schema coverage for full trace/span wire types."*
- `packages/jaeger-ui/src/api/v3/generated-client.ts` already contains Zod schemas for the full OTLP trace/span surface (`TracesData`, `ResourceSpans`, `ScopeSpans`, `Span`, `Span_Event`, `Span_Link`, `Resource`, `InstrumentationScope`, `KeyValue`, `AnyValue`, `ArrayValue`, `KeyValueList`, `Status`), but `schemas.ts` only re-exports the services + operations response schemas. Consumers that need to validate trace responses (e.g. the React-Query `getTrace` hook tracked in #3835 / Milestone 3.2) currently have no clean import path. The README in `src/api/v3/` says *"schemas.ts simply re-exports the generated schemas"* — this PR makes that true.
- While writing the new tests I also found that the four tagged-union-like OTLP types (`AnyValue`, `ArrayValue`, `KeyValue`, `KeyValueList`) had their `.partial()` calls stripped by the blanket regex in `scripts/postprocess-schemas.cjs`. Their TypeScript types are correctly emitted as `Partial<{...}>`, but the matching Zod schemas demanded every variant field at once — a shape no real OTLP attribute value satisfies (an `AnyValue` carries only one of `stringValue`/`boolValue`/`intValue`/…). Fixed in this PR. The broader strict-vs-wire-format mismatch on the non-union top-level schemas is intentionally out of scope and filed as #3890 for follow-up.

- `npm test -w packages/jaeger-ui -- src/api/v3/schemas.test.ts` → 59/59 pass (was 27 before; +32 new).
- Full task-completion suite per `CLAUDE.md`: `npm run fmt && npm run lint && npm test && npm run build` all clean. Total test count 2865 → 2897.
- Verified the postprocess regex by re-running it mentally against the current generator output: the strip removes 4 `.partial()` calls; the union-type loop restores them on `AnyValue`, `ArrayValue`, `KeyValue`, `KeyValueList`. The "Could not restore" warning is emitted only if the generator emits a shape the regex can't match.
 

## Checklist


scripts/postprocess-schemas.cjs
```diff
@@ -36,13 +36,29 @@ const copyrightHeader = `// Copyright (c) 2026 The Jaeger Authors.
...
 }
 
 // 2. Remove .partial() calls
 const beforeCountPartial = (content.match(/\.partial\(\)/g) || []).length;
 content = content.replace(/\.partial\(\)\s*/g, '');
-const afterCountPartial = (content.match(/\.partial\(\)/g) || []).length;
+const afterStripCountPartial = (content.match(/\.partial\(\)/g) || []).length;
+const removedCountPartial = beforeCountPartial - afterStripCountPartial;
+const UNION_TYPES = ['AnyValue', 'ArrayValue', 'KeyValueList'];
+let restoredCountPartial = 0;
```
anshgoyalevil, line 48: This is a great defensive check to ensure that schema regeneration doesn't silently break the partial restoration logic.

scripts/postprocess-schemas.cjs
```diff
@@ -63,20 +79,48 @@ content = content.replace(
...
+const exportLines = [
+  'export { GetServicesResponse as ServicesResponseSchema };',
+  'export { GetOperationsResponse as OperationsResponseSchema };',
+  'export { Operation as OperationSchema };',
+  'export { TracesData as TracesDataSchema };',
+  'export { ResourceSpans as ResourceSpansSchema };',
+  'export { ScopeSpans as ScopeSpansSchema };',
+  'export { Span as SpanSchema };',
+  'export { Span_Event as SpanEventSchema };',
+  'export { Span_Link as SpanLinkSchema };',
```
anshgoyalevil, line 95: The use of sentinel comments for the convenience exports block is a nice touch for idempotency. It makes the script much more robust for repeated runs.

packages/jaeger-ui/src/api/v3/schemas.ts
```diff
@@ -5,11 +5,31 @@
  * Zod schemas for Jaeger v3 API responses
  *
  * These are imported from generated-client.ts which is auto-generated from OpenAPI spec
- * and post-processed to remove .partial() for strict validation.
+ * and post-processed for forward-compatible validation: required-by-Proto3 fields are
+ * enforced as required, the tagged-union types (AnyValue, ArrayValue, KeyValueList)
+ * keep .partial() so any one variant satisfies them, and every object uses .passthrough()
+ * so unknown future fields are preserved rather than rejected.
  */
```
anshgoyalevil, line 12: The update to the documentation clearly explains the validation strategy (passthrough + partial for unions), which is very helpful for future contributors.

anshgoyalevil (commented):
Thanks for the PR. Please look into copilot review and some suggestions below

DsThakurRawat:
took this for a run too, since it's the schema half of the same milestone #4129 is the parser half of, and the deliverable text in #4278 asks for the two to be connected. 234 commits behind now.

the PR's approach is: regenerate `generated-client.ts` from the OpenAPI spec, strip the blanket `.partial()` the codegen applies (proto3/OpenAPI optionality mismatch makes it too permissive), restore it only on the three attribute-value union types (they're a proto3 oneof, so a real message only ever sets one field), then inject named export aliases so consumers can `import { AnyValueSchema } from './schemas'` instead of reaching into the bundled `schemas.X` object.

that last mechanism doesn't work anymore. the codegen's naming convention changed since this branched: `AnyValue` is now `opentelemetry_proto_common_v1_AnyValue`, `Span` is `opentelemetry_proto_trace_v1_Span`, and so on, all package-qualified. postprocess-schemas.cjs's restore-partial regex looks for the bare name and silently no-ops (`console.warn`, not a thrown error), and the convenience-export block literally does `export { AnyValue as AnyValueSchema }` referencing an identifier that no longer exists, a hard `tsc` error the moment you regenerate against the current spec:

```
error TS2304: Cannot find name 'AnyValue'.
error TS2304: Cannot find name 'ArrayValue'.
... (16 of these)
```

worth noting separately: main already solved half of this problem, independently, while this sat open. `schemas.ts` on main now exports `ServicesResponseSchema`/`OperationsResponseSchema`/`OperationSchema`/`TraceSummariesResponseSchema` by destructuring them off the `schemas` bundle object by their qualified name (`const { jaeger_api_v3_GetServicesResponse } = schemas`), not by injecting bare-name exports into the generated file. that pattern doesn't break the next time the codegen renames something, since `schemas.<qualified-name>` is the one thing that stays stable across a naming change.

so the fix i'd suggest: drop the convenience-export injection from postprocess-schemas.cjs entirely, and add the 13 non-recursive trace/span exports (`TracesDataSchema`, `ResourceSpansSchema`, `SpanSchema`, `ResourceSchema`, etc.) to `schemas.ts` the same way main already does the other four. keep AnyValue/ArrayValue/KeyValueList/KeyValue's `.partial()` restoration in postprocess-schemas.cjs specifically, because those three are mutually recursive (`ArrayValue.values` is `AnyValue[]`, `KeyValueList.values` is `KeyValue[]`, `KeyValue.value` is `AnyValue`) and re-deriving one of them in schemas.ts with `.unwrap().partial()` looks like it works in isolation but leaves the nested fields pointing at the original, fully-required version, i checked. just match the restore regex by suffix (`\w*AnyValue`) instead of the exact bare name, so the next codegen rename doesn't silently break it again the same way.

proved the fix rather than asserting it: regenerated fresh against the live spec, confirmed `AnyValueSchema.safeParse({ stringValue: 'hello' })` passes, confirmed the recursive case (`{ arrayValue: { values: [{ stringValue: 'a' }] } }`) passes too, `tsc --noEmit` clean, and the `api/v3` suite is 106/106 green.

one honesty note: ran the FULL test suite too, and 15 unrelated files fail with `localStorage.clear is not a function` (search/store/GenAI-tab tests). checked before blaming this PR, that's already broken on a clean `upstream/main` checkout with none of my changes applied, so it's environment noise on this machine, not something this PR causes.

happy to push the resolved branch, or this is the exact recipe if you'd rather do it yourself.

sksingh2005:
@gkhulbe4 Hey I was going through this and just wanted to confirm I’m reading this right.

After a fresh `pnpm run generate:api-types` the codegen now emits `opentelemetry_proto_common_v1_AnyValue`, so I’m seeing:

- `postproces-schemas.cjs` restore for `AnyValue/ArrayValue/KeyValueList` no longer matches , it just warns and leaves the unions fully-required
- the `export { AnyValue as AnyValueSchema }` block then fails `tsc` with `Cannot find name 'AnyValue'`

I think it will make sense to:

1. move the 13 trace/span exports into `schemas.ts` via `schemas.<qualified-name>` like `main` already does for `ServicesResponseSchema` (instead of injecting them into `generated-client.ts`), and
2. keep the `.partial()` restore only for the recursive `AnyValue/ArrayValue/KeyValueList/KeyValue` in `postprocess`, but match by suffix (`\w*AnyValue`) so the qualified name still matches?

I tried that locally and it passes. Let me know if that sounds right  happy to help with fix if you’d like.

yurishkuro:
Thanks for contributing, this is being re-implemented in #4438 and follow-ups.

PR #3899, opened 2026-05-15
feat(span): lazy-load large attributes to prevent UI stalls

### LINKED ISSUE #3893: [Feature]: Lazy-load large GenAI attribute values to prevent trace-view stall on multi-KB prompts and completions
## The Problem — A Story Before the Code

I loaded a GenAI trace from an agent that had a 30-message conversation history. I clicked the `chat gpt-4o` span expecting to see the prompt that caused the wrong answer. The browser tab froze. The spinner ran for about two seconds. When the UI came back, what I saw was a single truncated string cell with `[{"role":"user","content":"You are a helpful assistant...` cut off at some arbitrary width. I had to copy it out, open a new tab, paste it into a JSON formatter, and read it there.

That was the entire debugging session. The core workflow — *why did my AI agent say that?* — is currently a four-tab, copy-paste, external-tool dance. It should be one click in Jaeger.

The fix for that readability problem is being tracked in #3808 and #3829 (rich-media rendering). But there is a **prerequisite architectural problem that neither issue addresses**: the attribute values are being parsed and rendered *eagerly* — before the developer even asks for them. This issue tracks that.

---

## Root Cause

`SpanDetail/index.tsx` passes the **entire** `span.attributes` array directly into `AccordionAttributes`:

```typescript
// SpanDetail/index.tsx
<AccordionAttributes
  label={useOtelTerms ? 'Attributes' : 'Tags'}
  attributes={span.attributes}   // <-- all attributes, all at once
  // ... toggle handlers
/>
```

`AccordionAttributes` renders each attribute as a table row. There is no check on value size, no deferral, no virtualization. When `gen_ai.input.messages` carries a 200-message conversation history, the browser must:

1. Allocate a string of 50–500 KB
2. Measure the table cell
3. Paint the truncated text

All of this happens **synchronously on the React render cycle** triggered by the span selection click.

---

## Why This Is Separate from #3808 / #3829

The rich-media rendering issues describe *what* to render (Markdown, JSON tree, images). This issue describes *when* to render it.

Even after #3808 ships a `GenAIAttributeRenderer` that renders `gen_ai.input.messages` as a beautiful Markdown panel — if it parses the 500 KB JSON string at span-selection time, the UI still freezes. The rendering quality improves. The stall does not.

The fix must happen at the **attribute-loading layer**, not the rendering layer.

---

## Concrete Attribute Sizes

Using real GenAI workloads and the canonical fixture from #3839:

| Attribute | Typical payload | Current behaviour |
|---|---|---|
| `gen_ai.input.messages` | 2 KB – 500 KB (conversation history) | Inline string cell, eager |
| `gen_ai.output.messages` | 1 KB – 200 KB (completion + reasoning) | Inline string cell, eager |
| `gen_ai.retrieval.documents` | 10 KB – 2 MB (RAG document chunks) | Inline string cell, eager |
| `gen_ai.tool.call.arguments` | 0.5 KB – 50 KB (structured args) | Inline string cell, eager |
| `gen_ai.system_instructions` | 1 KB – 100 KB (system prompt) | Inline string cell, eager |
| `http.url`, `db.statement` | < 512 bytes (typical) | Inline string cell — fine |

The bottom row shows why the current design works for non-GenAI traces: standard attributes are small. The problem is GenAI-specific.

---

### Step 1 — Size-gate the attribute split in `SpanDetail`

Before passing attributes to `AccordionAttributes`, split them into two buckets:

```typescript
const LARGE_ATTR_THRESHOLD_BYTES = 10_240; // 10 KB

const largeAttrs = span.attributes.filter(
  a => typeof a.value === 'string' && a.value.length > LARGE_ATTR_THRESHOLD_BYTES
);
const standardAttrs = span.attributes.filter(
  a => !(typeof a.value === 'string' && a.value.length > LARGE_ATTR_THRESHOLD_BYTES)
);
```

`standardAttrs` flows into `AccordionAttributes` unchanged — zero regression for non-GenAI spans.

### Step 2 — Render large attributes behind a lazy accordion

Large attributes get a separate `<details>`/accordion section that starts collapsed. The attribute value is **not parsed or rendered** until the section is expanded:

```typescript
// Pseudocode — collapsed by default, value parsed only on expand
<LazyAttributeSection
  attribute={attr}
  renderContent={() => <GenAIAttributeRenderer attribute={attr} />}
/>
```

`LazyAttributeSection` keeps a local `expanded: boolean` state. When `false`, it renders only the attribute key and a byte-size badge (e.g., `gen_ai.input.messages  •  48 KB  ▶ expand`). When `true`, it calls `renderContent()`.

### Step 3 — Move JSON parsing off the main thread for values > 100 KB

For attributes larger than 100 KB, `JSON.parse` itself can block for 10–50 ms. Offload to a Web Worker or `queueMicrotask` so the accordion can open and show a skeleton before the tree renders:

```
expand click
    │
    ├─► render skeleton immediately (0 ms)
    └─► postMessage to worker: parse JSON
             │
             └─► onmessage: swap skeleton for react-json-view-lite tree
```

A simpler short-term option is `requestIdleCallback` for the parse step — good enough for < 500 KB.

---

## Rendering Flow Comparison

**Today (eager):**
```
click span → React renders SpanDetail → AccordionAttributes receives all attrs
           → browser measures + paints 500KB string cell
           → UI unresponsive for 1–3 seconds
           → developer sees truncated string
```

**Proposed (lazy):**
```
click span → React renders SpanDetail → AccordionAttributes receives small attrs only
           → UI responds in < 50ms
           → large attrs render as collapsed keys with size badge

developer expands section → LazyAttributeSection calls renderContent()
           → JSON parsed (async for > 100KB)
           → react-json-view-lite tree renders
           → developer reads the prompt
```

---

## Acceptance Criteria

- [ ] Selecting a GenAI span with `gen_ai.input.messages` > 10 KB does not block the main thread for more than 50 ms (measurable with Chrome DevTools Performance panel).
- [ ] Large attribute keys are visible in the collapsed state with a byte-size indicator.
- [ ] Expanding the section shows the attribute value via `GenAIAttributeRenderer` (from #3808).
- [ ] Non-GenAI spans and spans with only small attributes show no change in behaviour.
- [ ] The threshold (default 10 KB) is a named constant, not a magic number, and is documented.
- [ ] Unit test: `SpanDetail` renders a skeleton and not the full value on initial render when attribute exceeds threshold.
- [ ] Unit test: After simulated expand, full `GenAIAttributeRenderer` output is present in the DOM.

---

## Related Issues

- #8401 — parent issue (mentions lazy loading explicitly in comments)
- #3808 — rich-media rendering (consumer of the lazy-load mechanism)
- #3829 — better SpanDetail for GenAI (related rendering concern)
- #3839 — canonical GenAI fixture (needed for testing this)

---

## Notes

This is a prerequisite for a correct implementation of #3808. Shipping rich-media rendering without lazy loading means the Markdown/JSON renderers will parse large payloads at span-selection time and the stall moves from `AccordionAttributes` into `GenAIAttributeRenderer`. The fix is the same either way — defer until the developer asks for it.

The `VirtualizedTraceView` already uses `@tanstack/react-virtual` to avoid rendering off-screen span rows. This issue applies the same principle one level down: don't render off-screen (or un-expanded) attribute values.

Harizz076:

- Resolves #3893
- Resolves review feedback on UX and correctness concerns regarding splitting attributes.

https://github.com/user-attachments/assets/d84aa2e4-f852-4efc-961e-ec308c7205ad

## Description of the changes

- **Inline Deferred Rendering (`LargeValueCell`)**: Instead of splitting large attributes into a separate section, large string values (>10,000 chars) are now replaced by a small inline "click to expand" placeholder button directly within the standard `AttributesTable`.
- **Preserved UX & Ordering**: By handling lazy expansion at the cell level, we preserve the original attribute ordering, the correct count in the "Attributes (N)" header, and the inline collapsed preview strings.
- **Performance Optimization**: The `LargeValueCell` uses local component state to defer the expensive `DOM` rendering of massive string payloads until the user explicitly clicks the placeholder, preventing the UI thread from blocking when selecting a span with large (e.g. GenAI) payloads.
- **Cleaned up Dead Code**: Removed the outdated  approach.

- Tested with a large trace containing GenAI payloads in both light and dark modes to verify the expansion button styling and unblocked main thread behavior.
- Ran `npm run lint` and `npm test` locally.

## Checklist


yurishkuro:
Thanks for the PR — the motivation is solid (large LLM payloads blocking the main thread is a real problem). I have some UX and correctness concerns before this can merge.

### UX concern: splitting attributes into two separate sections

The current approach puts small attributes in the existing `AccordionAttributes` section and large attributes in separate floating accordion rows below it. This breaks attribute ordering (large attrs always appear last), makes the count in the `Attributes (N)` header wrong, hides large attributes from the collapsed inline preview, and exposes an implementation detail (string byte length) as a visible structural UI distinction.

A more coherent approach: keep all attributes in the single `AccordionAttributes` list, but render large values lazily *within their row* — e.g., a collapsed "48.8 KB — click to expand" placeholder inline in the attribute table. This preserves order, grouping, and count while still deferring the expensive render.

### Correctness issues

**1. `LazyAttributeSection.tsx` — `onToggle` returns a cleanup function from a click handler; the timeout leaks**

`onToggle` conditionally returns `() => clearTimeout(timeout)` — a pattern valid only inside `useEffect`. React silently discards return values from event handlers. If the user clicks expand then clicks collapse within 50 ms, `isDeferredLoaded` is set to `true` by the orphaned timeout even though the section is collapsed, bypassing the deferred-load guard on the next expand.

Fix: move the timeout into a `useEffect` that depends on `isExpanded && isVeryLarge && !isDeferredLoaded`.

**2. `SpanDetail/index.tsx` — `linksGetter` is not passed to `LazyAttributeSection`**

`AccordionAttributes` receives `linksGetter={linksGetter}`, which `AttributesTable` uses to render clickable hyperlinks on attribute rows. `LazyAttributeSection` only receives `attribute={attr}`. Any large attribute that a plugin decorates with a link silently loses its hyperlink after the split.

**3. `SpanDetail/index.tsx` — large attributes ignore `isAttributesOpen`; sidepanel mode won't auto-expand them**

`DetailState.forDetailPanelMode('sidepanel')` sets `isAttributesOpen = true` so all attribute sections start expanded in the side panel. `LazyAttributeSection` manages its own independent `useState(false)` and is not wired to this. Large attributes will start collapsed in sidepanel mode, silently hiding data.

**4. `SpanDetail/index.tsx` — `AttributesSummary` inline preview is blind to large attributes**

When `AccordionAttributes` is collapsed it shows an inline key=value preview. After the split only `standardAttrs` are passed, so a large `error.message` or `exception.stacktrace` is entirely invisible in collapsed state — no indication it exists until the user scrolls down to the separate rows.

**5. `GenAIAttributeRenderer.tsx` — duplicates `tryParseJson` + `JsonView` rendering from `AttributesTable.tsx`**

`AttributesTable.tsx` already has an identical `tryParseJson` (same regex `/^(\[|\{)/`), imports the same `JsonView`/`allExpanded`/`collapseAllNested`/`defaultStyles`, and renders a `JsonView` with the same custom class-name style map. Please extract the shared logic rather than duplicating it.

**6. `LazyAttributeSection.tsx` — `DEFER_RENDER_THRESHOLD_BYTES` counts UTF-16 code units, not bytes**

JavaScript `.length` is UTF-16 code units. A string of emoji or CJK characters at 5 120 chars is ~10–20 KB of actual content but `.length` reports 5 120 — well under the 100 KB threshold. The constant name implies bytes; the measurement isn't. Worth documenting or renaming.

**7. `SpanDetail/index.tsx` — React key `${attr.key}-${attr.value.length}` can collide**

If two large attributes share the same key name (permitted by the OTel spec) and happen to have equal value lengths, they produce identical React keys. React would silently reuse one component instance, leaking `isExpanded`/`isDeferredLoaded` state between rows.

**8. Naming: `GenAIAttributeRenderer` is size-based, not GenAI-specific**

The dispatch in `index.tsx` is purely `attr.value.length >= 10240` — it applies to any large attribute (`db.statement`, `http.response.body`, etc.). The component name implies a scoped concern that doesn't match the actual selection logic.

Harizz076:
@yurishkuro : Thank you for the detailed review, ive completely changed the approach and implemented the changes base don your advice. Please check the screencast and let me know if anything looks off.

Harizz076:
@yurishkuro : Can I get a review on this PR? Lazy-Loading large attributes will be a good addition to Jaeger-ui, especially since we have the Gen-AI tab now. Please run the workflows tests too whenever you find time.

CC: @jkowall

Harizz076:
@yurishkuro Can you please run Co-pilot's review on this and let me know if the screencast in the PR description looks good.

Harizz076:
@yurishkuro : Dropping a reminder here for your review and merge if possible. Please check the screencast for the changes.

PR #3915, opened 2026-05-17
fix(critical-path): surface computation errors instead of silently returning empty

### LINKED ISSUE #3907: [Bug]: Critical path silently returns empty when computation fails, giving engineers false confidence that their trace has no bottleneck
## What's happening

When you open a trace and enable the **critical path** overlay, there are specific edge cases (clock skew, sanitized overflow children, cyclic span references in unusual instrumentation) where the computation throws internally. The error is caught, logged only to the browser console, and the function returns an empty array - as if the trace simply has no critical path.

The trace view renders with no critical path highlight and no warning of any kind. To a developer staring at the timeline, it looks like all spans are truly parallel with no bottleneck. That's almost never true, and acting on that conclusion causes engineers to optimize the wrong things.

---

## The code

`packages/jaeger-ui/src/components/TracePage/CriticalPath/index.ts` lines 79–87:

```ts
for (const rootSpan of trace.rootSpans) {
  try {
    const sanitizedSpanMap = sanitizeOverFlowingChildren(createCPSpanMap(rootSpan));
    criticalPath = computeCriticalPath(sanitizedSpanMap, rootSpan.spanID, criticalPath);
  } catch (error) {
    console.log('error while computing critical path for a trace', error);
    // ↑ error disappears here. criticalPath stays empty. caller never knows.
  }
}
return criticalPath;
```

The caller in `TracePage/index.tsx` line 433 does:

```tsx
const criticalPath = criticalPathEnabled ? memoizedTraceCriticalPath(data.asOtelTrace()) : [];
```

There is no way for it to distinguish "computation succeeded, this trace has no critical path" from "computation failed silently". Both look identical: `criticalPath = []`.

---

## What the user sees vs what's actually happening

```
What the user sees in the trace view:
┌─────────────────────────────────────────────────────────────────┐
│    frontend        ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━  820ms │
│    auth-service  ━━━━━━━━━━━━━━━━━━━━━  412ms               │
│    db-query      ━━━━━━━━━━━━━━━━━━━━━━━━━━━━  680ms        │
│    cache-lookup  ━━━━  80ms                                  │
│                                                                 │
│  [Critical path: none highlighted]                              │
│  ← user concludes: "all spans run in parallel, no bottleneck"  │
└─────────────────────────────────────────────────────────────────┘

What's actually happening in the background:
  computeCriticalPath threw a RangeError (clock skew on auth-service span)
  console.log("error while computing critical path...") → DevTools only
  returned [] → caller received [] → UI showed nothing → user saw nothing
```

```
What the user SHOULD see:
┌─────────────────────────────────────────────────────────────────┐
│    Critical path could not be computed for this trace.          │
│    This usually means one or more spans have clock skew or      │
│    timestamps outside their parent's bounds.                    │
│    Check browser console for details.                   [Dismiss]│
├─────────────────────────────────────────────────────────────────┤
│    frontend        ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━  820ms │
│    auth-service  ━━━━━━━━━━━━━━━━━━━━━  412ms               │
│    ...                                                          │
└─────────────────────────────────────────────────────────────────┘
```

---

## When does the computation actually fail?

The throw can come from three places:

- `createCPSpanMap` - while building the span map for a root span's subtree
- `sanitizeOverFlowingChildren` - while filtering spans whose end times overflow their parent, which can happen with **NTP clock drift** between instrumented services
- `computeCriticalPath` - if the algorithm reaches an unexpected span state in the recursive walk

Real-world traces with any of the following will hit this:
- Services on hosts with >1ms NTP drift
- Instrumentation that records `startTime` after `endTime` (a known race in some SDKs on process exit)
- Async frameworks that produce overlapping sibling spans

These are not exotic. In any production microservices environment you will see clock skew. The critical path computation failing there and giving no feedback is a real day-to-day pain point.

---

## What a fix looks like

`criticalPathForTrace` should return a discriminated result instead of a bare array:

```ts
type CriticalPathResult =
  | { ok: true; sections: CriticalPathSection[] }
  | { ok: false; sections: CriticalPathSection[]; error: unknown };
```

Or simpler — keep the return type the same but expose a separate `criticalPathError` signal:

```ts
// CriticalPath/index.ts
export function criticalPathForTrace(
  trace: IOtelTrace
): { sections: CriticalPathSection[]; failed: boolean } {
  let sections: CriticalPathSection[] = [];
  let failed = false;
  for (const rootSpan of trace.rootSpans) {
    try {
      const sanitizedSpanMap = sanitizeOverFlowingChildren(createCPSpanMap(rootSpan));
      sections = computeCriticalPath(sanitizedSpanMap, rootSpan.spanID, sections);
    } catch (error) {
      console.error('Critical path computation failed for root span', rootSpan.spanID, error);
      failed = true;
    }
  }
  return { sections, failed };
}
```

Then in `TracePage/index.tsx`:

```tsx
const { sections: criticalPath, failed: criticalPathFailed } =
  criticalPathEnabled ? memoizedTraceCriticalPath(data.asOtelTrace()) : { sections: [], failed: false };

// pass criticalPathFailed down to show a warning banner
```

---

## Acceptance criteria

- [ ] When `criticalPathForTrace` throws on any root span, the error is still surfaced in the console (but as `console.error`, not `console.log`)
- [ ] The TracePage shows a visible, dismissable warning when critical path computation failed
- [ ] The warning message is actionable: tells the user what likely caused it and that the overlay data is incomplete
- [ ] When computation fails **partially** (multi-root trace, some roots succeed), the warning still appears and the partial result is still shown
- [ ] Unit test: mock `sanitizeOverFlowingChildren` to throw, verify `failed: true` is returned and `sections` contains the result for non-failing roots

---

## Why this matters

The critical path feature exists specifically to help engineers answer "what is actually on the critical path of this request?" That's one of the most important questions in production performance debugging. Silently returning an empty result without feedback corrupts the answer and erodes trust in the tool - because the engineer has no way to know whether their trace genuinely has no sequential bottleneck or whether the tool quietly gave up.

A visible warning takes thirty minutes to implement and completely restores that trust.

swetalin-10:
## Summary

Fixes #3907.

When `criticalPathForTrace` threw on a root span (e.g. clock skew, overflowing timestamps, cyclic references), the error was silently swallowed — `console.log`ged and an empty array returned, indistinguishable from a trace that genuinely has no critical path. Engineers staring at the timeline had no way to know the overlay quietly failed.

- **`CriticalPath/index.ts`**: `criticalPathForTrace` now returns `{ sections: CriticalPathSection[]; failed: boolean }` instead of a bare array. `console.log` upgraded to `console.error`. Partial results from non-failing root spans are preserved in multi-root traces.
- **`TracePage/index.tsx`**: Destructures the new result. When `failed` is `true`, renders a dismissable `antd` `Alert` banner above the header explaining the likely cause and that the overlay may be incomplete.
- **`CriticalPath/index.test.js`**: All existing happy-path tests updated to destructure `{ sections, failed }`. Two new tests cover the error path: one verifies `failed: true` with partial sections when only one root span throws, and one verifies `failed: true` with empty sections when all root spans throw.

## Acceptance criteria checklist

- [x] Errors are surfaced via `console.error`, not `console.log`
- [x] TracePage shows a visible, dismissable warning when critical path computation failed
- [x] Warning message is actionable (explains likely cause, notes overlay may be incomplete)
- [x] Partial results from succeeding root spans are still shown when one root fails
- [x] Unit test: mock `sanitizeOverFlowingChildren` to throw, verify `failed: true` and partial sections

## Test plan

- [ ] Run `yarn test packages/jaeger-ui/src/components/TracePage/CriticalPath` — all tests green
- [ ] Open a trace in the UI with `criticalPathEnabled: true` — normal overlay still renders, no banner
- [ ] Temporarily throw in `sanitizeOverFlowingChildren` — banner appears, can be dismissed, console shows error

### DIFF
--- a/packages/jaeger-ui/src/api/v3/client.test.ts
+++ b/packages/jaeger-ui/src/api/v3/client.test.ts
@@ -3,6 +3,11 @@
 
 import { JaegerClient, jaegerClient } from './client';
 import { ZodError } from 'zod';
+import { parseOtlpTrace } from './parser';
+
+vi.mock('./parser', () => ({
+  parseOtlpTrace: vi.fn(),
+}));
 
 describe('JaegerClient', () => {
   let client: JaegerClient;
@@ -256,6 +261,61 @@ describe('JaegerClient', () => {
     });
   });
 
+  describe('getTrace', () => {
+    const mockTrace = { traceID: 'trace-abc', spans: [] } as any;
+
+    it('successfully fetches and parses a trace', async () => {
+      const mockData = { resourceSpans: [] };
+      mockFetch.mockResolvedValue({ ok: true, json: async () => mockData });
+      (parseOtlpTrace as ReturnType<typeof vi.fn>).mockReturnValue(mockTrace);
+
+      const promise = client.getTrace('trace-abc');
+      vi.runAllTimers();
+      const result = await promise;
+
+      expect(result).toBe(mockTrace);
+      expect(mockFetch).toHaveBeenCalledWith(
+        '/api/v3/traces/trace-abc',
+        expect.objectContaining({ signal: expect.any(AbortSignal) })
+      );
+      expect(parseOtlpTrace).toHaveBeenCalledWith(mockData);
+    });
+
+    it('URL-encodes the traceId', async () => {
+      mockFetch.mockResolvedValue({ ok: true, json: async () => ({}) });
+      (parseOtlpTrace as ReturnType<typeof vi.fn>).mockReturnValue(mockTrace);
+
+      const promise = client.getTrace('id with spaces');
+      vi.runAllTimers();
+      await promise;
+
+      expect(mockFetch).toHaveBeenCalledWith(
+        '/api/v3/traces/id%20with%20spaces',
+        expect.anything()
+      );
+    });
+
+    it('throws with traceId in message when response is not OK', async () => {
+      mockFetch.mockResolvedValue({ ok: false, status: 404, statusText: 'Not Found' });
+
+      const promise = client.getTrace('missing-trace');
+      vi.runAllTimers();
+
+      await expect(promise).rejects.toThrow('Failed to fetch trace "missing-trace": 404 Not Found');
+    });
+
+    it('throws on 500 errors', async () => {
+      mockFetch.mockResolvedValue({ ok: false, status: 500, statusText: 'Internal Server Error' });
+
+      const promise = client.getTrace('some-trace');
+      vi.runAllTimers();
+
+      await expect(promise).rejects.toThrow(
+        'Failed to fetch trace "some-trace": 500 Internal Server Error'
+      );
+    });
+  });
+
   describe('singleton instance', () => {
     it('exports a singleton jaegerClient instance', () => {
       expect(jaegerClient).toBeInstanceOf(JaegerClient);
@@ -264,6 +324,7 @@ describe('JaegerClient', () => {
     it('singleton instance has all expected methods', () => {
       expect(jaegerClient.fetchServices).toBeInstanceOf(Function);
       expect(jaegerClient.fetchSpanNames).toBeInstanceOf(Function);
+      expect(jaegerClient.getTrace).toBeInstanceOf(Function);
     });
   });
 });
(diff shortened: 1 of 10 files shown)


swetalin-10:
Closing in favour of a cleaner PR rebased directly onto main (single commit, no unrelated history).

PR #3916, opened 2026-05-17
fix(critical-path): surface computation errors instead of silently returning empty

### LINKED ISSUE #3907: [Bug]: Critical path silently returns empty when computation fails, giving engineers false confidence that their trace has no bottleneck

swetalin-10:

Fixes #3907

When `criticalPathForTrace` throws — clock skew, overflowing timestamps, cyclic refs — it logs a message and returns an empty array. From the UI side that looks identical to a trace that simply has no critical path. No banner, no indication anything went wrong.

## Description of the changes

`criticalPathForTrace` now returns `{ sections, failed }` instead of a plain array. If a root span throws, the error goes to `console.error` instead of `console.log`, and `failed` is set to `true`. Partial results from other root spans that didn't throw are still included.

`TracePage` destructures the new shape. When `failed` is `true` it shows a dismissable `antd` Alert above the timeline explaining what likely went wrong and that the overlay might be incomplete.

> **Note:** The error banner can be reproduced by temporarily throwing inside `sanitizeOverFlowingChildren` and opening a trace with critical path enabled.

Updated all existing happy-path tests in `CriticalPath/index.test.js` to use the new `{ sections, failed }` shape. Added two error-path tests: one where only one root span throws (partial sections, failed true) and one where all root spans throw (empty sections, failed true).

Ran `npm test` on the `CriticalPath` directory — all pass.

## Checklist


packages/jaeger-ui/src/components/TracePage/index.tsx
```diff
@@ -511,6 +516,16 @@ export function TracePageImpl(props: TProps) {
       {archiveEnabled && (
         <ArchiveNotifier acknowledge={acknowledgeArchive} archivedState={archiveTraceState} />
       )}
+      {cpResult.failed && !criticalPathErrorDismissed && (
```
anshgoyalevil, line 520: `criticalPathErrorDismissed` is component-level state that is never reset when the user navigates to a different trace. `TracePageImpl` is not remounted per trace id — the effect at line 298 (`if (prevIdRef.current !== id)`) is exactly where trace-scoped state gets reset (`updateViewRangeTime`, `clearSearch`). Because the dismiss flag isn't reset there, once a user dismisses this banner for one failing trace, it stays dismissed for every subsequent trace in the same session (until a full page reload), so a genuinely failed critical-path computation on a later trace will show no warning.

Consider resetting `setCriticalPathErrorDismissed(false)` inside that id-change effect so the dismiss is scoped per trace.

swetalin-10: Good catch - you're right, this state was falling through the cracks of the id-change effect. Fixed in b0d6259d: added `setCriticalPathErrorDismissed(false)` right alongside `updateViewRangeTime(0, 1)` and `clearSearch()` in that same effect, so it's scoped per trace like the rest.

Added a regression test for exactly the scenario you described - dismiss the banner on one failing trace, navigate to a different failing trace, banner should reappear. Verified it actually catches the bug: reverted the fix locally, confirmed the new test fails without it, then restored the fix - full suite is back to green (70/70), `tsc` and `oxlint` clean.

anshgoyalevil (approved):

packages/jaeger-ui/src/components/TracePage/CriticalPath/index.ts
```diff
@@ -73,19 +73,26 @@ const computeCriticalPath = (
...
     try {
       // Create a map of CPSpan objects for this root span's tree,
       // filtering out non-blocking subtrees during creation.
       const sanitizedSpanMap = sanitizeOverFlowingChildren(createCPSpanMap(rootSpan));
-      criticalPath = computeCriticalPath(sanitizedSpanMap, rootSpan.spanID, criticalPath);
+      sections = computeCriticalPath(sanitizedSpanMap, rootSpan.spanID, sections);
     } catch (error) {
-      console.log('error while computing critical path for a trace', error);
+      console.error('Critical path computation failed for root span', rootSpan.spanID, error);
+      failed = true;
```
yurishkuro: it doesn't make sense to me that the error message, which probably contains a description of the actual issue, is swallowed here and later an alert with "maybe this, maybe that" is shown. Why not capture the actual error and show it?

swetalin-10: Fair point, and fixed - `criticalPathForTrace` now captures `error.message` per failing root span into a returned `errors: string[]` instead of discarding it after the `console.error`. The alert in `TracePage` renders that real text (`cpResult.errors.join('; ')`) instead of the old canned copy.

Added a couple of tests to lock in the behavior: one that asserts the captured string actually contains the thrown error's message (not a mock), and one confirming the banner renders it. Full suite green, `tsc` and `oxlint` clean.

packages/jaeger-ui/src/components/TracePage/CriticalPath/index.test.js
```diff
@@ -45,51 +45,70 @@ describe('criticalPathForTrace immutability', () => {
...
     test2.trace.spans.forEach((span, index) => {
-      expect(span.links).toBe(originalLinks[index].array); // Same reference
+      expect(span.links).toBe(originalLinks[index].array);
       expect(span.links.length).toBe(originalLinks[index].length);
     });
   });
 
   it('should not modify non-blocking spans parent childSpans', () => {
-    // Store original childSpans of the parent in test5
-    // test5 has a producer/consumer pair where the consumer is non-blocking
```
yurishkuro, line 83: unnecessary removal of comments in this file. This one in particular seems relevant

swetalin-10: Fixed in bf50a14e — restored. Checked the rest of what got removed in this file against main: everything else was a pure restatement of the line right below it (`// Run the critical path algorithm` above `TraceCriticalPath(...)`, `// Same reference` next to `.toBe()`, etc.), so left those out. This one was different - it explains *why* test5 specifically is used here (producer/consumer pair, consumer is non-blocking per `isBlockingSpan` in cpspan.ts), which isn't derivable from the test code itself.

yurishkuro (approved):

PR #3924, opened 2026-05-18
fix(transform-trace-data): Avoid stack overflow on deep parent-child …

gkhulbe4:
- Fixes ##3923
- `transformTraceData` in `packages/jaeger-ui/src/model/transform-trace-data.ts` traversed the span tree via recursive `processSpan`. Each parent-child step consumed a JS call-stack frame, and once tree depth exceeded the engine's limit (~10k under Vitest, similar in browser renderer processes), the call threw `RangeError: Maximum call stack size exceeded`. The error propagated out of the reducer's `fetchTraceDone` and the trace failed to render — no fallback. Triggers in production for deep traces from recursive instrumented functions, retry loops, long microservice call sequences, and instrumented batch jobs / workflow engines.

## Description of the changes
- **`packages/jaeger-ui/src/model/transform-trace-data.ts`** — replace the recursive `processSpan` with an explicit-stack iterative depth-first traversal. Root spans are pushed in reverse so the smallest `startTime` is popped first; each parent's children are pushed in reverse after the parent is recorded, preserving pre-order DFS. All observable semantics (`span.depth`, `hasChildren`, `relativeStartTime`, `svcCounts`, `references.span`, `subsidiarilyReferencedBy`, child sort order, flat-spans ordering) are preserved bit-for-bit — the existing 17 tests in `transform-trace-data.test.js` continue to pass unchanged.
- **`packages/jaeger-ui/src/model/transform-trace-data.test.js`** — one new regression test: transforms a 12000-deep parent-child chain and asserts no throw, correct `depth` on the deepest span, and correct DFS ordering of the flat `spans` array. Fails against pre-fix source with `RangeError: Maximum call stack size exceeded`. 

- `npm test -w packages/jaeger-ui -- src/model/transform-trace-data.test.js` → 18/18 pass with the fix.
- Verified the new test fails against pre-fix source via `git stash` of the source change — confirms the regression test genuinely exercises the bug.
- Full task-completion suite per `CLAUDE.md`: `npm run fmt && npm test && npm run build` clean. `npm run lint` fails on an unrelated pre-existing stale `yaml` override in root `package.json` (also fails against `main` without my changes).

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/model/transform-trace-data.test.js
+++ b/packages/jaeger-ui/src/model/transform-trace-data.test.js
@@ -451,5 +451,35 @@ describe('transformTraceData()', () => {
       const ids = result.spans.map(s => s.spanID);
       expect(ids).toEqual(['root', 'child1', 'grandChild1', 'child2']);
     });
+
+    it('processes a deep parent-child chain without exhausting the JS call stack', () => {
+      const depth = 12000;
+      const spans = [];
+      for (let i = 0; i < depth; i++) {
+        spans.push({
+          traceID: 'deep',
+          spanID: `s${i}`,
+          operationName: `op${i}`,
+          startTime: i + 1,
+          duration: 1,
+          processID: 'p',
+          references: i === 0 ? [] : [{ refType: 'CHILD_OF', spanID: `s${i - 1}`, traceID: 'deep' }],
+          tags: [],
+          logs: [],
+        });
+      }
+      const trace = {
+        traceID: 'deep',
+        spans,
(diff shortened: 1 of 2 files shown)


parshipcy:
please do not raise a PR without the actual issue being approved by a maintainer, as mentioned in the [contributing guidelines](https://github.com/jaegertracing/jaeger/blob/main/CONTRIBUTING_GUIDELINES.md#making-a-change).

yurishkuro:
feels like a theoretical / made up issue - which traces have 10k depth?

the implementation will be refactored anyway due to adr/0002-otlp-api-v3-migration.md

PR #3928, opened 2026-05-19
Fix timeline side panel header resizer

### LINKED ISSUE #3925: [Bug]: Column width adjustment on Timeline view is inconsistent
### What happened?

The column dividers on the trace timeline view behave differently. The divider to the right of the Service Name column works correctly - it shows the moving boundary at both the table and the header and the cursor shows draggable state also in both table and header.

Image

But the divider to the left of the Span Details column only works in the table, not in the header.

Image

Also, the z-axis is off for the divider line

Image

### Steps to reproduce

mouse over the column borders in both examples and try to drag.

### Expected behavior

they should behave identically

### Relevant log output

```shell

```

### Screenshot

_No response_

### Additional context

_No response_

### Jaeger backend version

_No response_

### SDK

_No response_

### Pipeline

_No response_

### Stogage backend

_No response_

### Operating system

_No response_

### Deployment model

_No response_

### Deployment configs

```shell

```

MohitGoyal09:
  - Resolves #3925

  ## Description of the changes
  - Move the timeline side panel resize handle into the fixed header row so the divider can be dragged from the header as well as the body area.
  - Share the side panel resize bounds from the timeline viewer with the header row.
  - Update timeline/header tests to cover the side panel header resizer and callback wiring.

  - `npm test -w packages/jaeger-ui -- src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow/TimelineHeaderRow.test.jsx src/components/TracePage/
  TraceTimelineViewer/index.test.jsx`
  - `npm test`
  - `npm run fmt`
  - `npm run lint`
  - `npm run build`
  - Browser verification with mocked trace data: confirmed the side panel drag handle starts at the fixed header top and aligns with the side panel
  boundary.

  ## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow/TimelineHeaderRow.test.jsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow/TimelineHeaderRow.test.jsx
@@ -6,6 +6,11 @@ import { fireEvent, render, screen } from '@testing-library/react';
 import '@testing-library/jest-dom';
 
 import TimelineHeaderRow from './TimelineHeaderRow';
+import { SIDE_PANEL_WIDTH_MAX, SIDE_PANEL_WIDTH_MIN } from '../store.constants';
+
+const { RESIZER_DRAG_DELTA } = vi.hoisted(() => ({
+  RESIZER_DRAG_DELTA: 0.05,
+}));
 
 vi.mock('../TimelineRow', () => {
   const TimelineRowMock = ({ children, className }) => (
@@ -23,7 +28,7 @@ vi.mock('../TimelineRow', () => {
 
 vi.mock('../../../common/VerticalResizer', () => ({
   default: ({ position, min, max, onChange }) => {
-    const nextPosition = position - 0.05;
+    const nextPosition = position - RESIZER_DRAG_DELTA;
     return (
       <button
         type="button"
@@ -142,15 +147,17 @@ describe('<TimelineHeaderRow>', () => {
   });
 
   describe('side panel visible', () => {
-    const sidePanelWidth = 0.3;
+    const sidePanelWidth = 1 - SIDE_PANEL_WIDTH_MAX;
+    const sidePanelResizerMin = 1 - SIDE_PANEL_WIDTH_MAX;
+    const sidePanelResizerMax = 1 - SIDE_PANEL_WIDTH_MIN;
     const sidePanelProps = {
       ...props,
       sidePanelVisible: true,
       sidePanelWidth,
       sidePanelLabel: 'Span Details',
       resizerMax: 1 - sidePanelWidth,
-      sidePanelResizerMin: 0.35,
-      sidePanelResizerMax: 0.8,
+      sidePanelResizerMin,
+      sidePanelResizerMax,
     };
 
     it('renders the side panel header cell', () => {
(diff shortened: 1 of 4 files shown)


packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow/TimelineHeaderRow.test.jsx
```diff
@@ -149,8 +161,29 @@ describe('<TimelineHeaderRow>', () => {
...
+      expect(nameColumnResizer).toHaveAttribute('data-max', String(1 - sidePanelWidth));
+    });
+
+    it('renders a side panel resizer aligned with the side panel boundary', () => {
+      render(<TimelineHeaderRow {...sidePanelProps} />);
+      const [, sidePanelResizer] = screen.getAllByTestId('vertical-resizer');
+
+      expect(sidePanelResizer).toHaveAttribute('data-position', String(1 - sidePanelWidth));
+      expect(sidePanelResizer).toHaveAttribute('data-min', '0.35');
+      expect(sidePanelResizer).toHaveAttribute('data-max', '0.8');
```
anshgoyalevil: Suggestion: Using the imported constants (like `SIDE_PANEL_WIDTH_MIN`) here instead of hardcoded strings like `'0.35'` would make the tests more resilient to changes in layout defaults.

anshgoyalevil (commented):
Please address some minor suggestions

anshgoyalevil:
Please take a look at Copilot comments, and also, please post the before/after screenshot/video.

MohitGoyal09:
Yes sure . i will look into this and fix this

yurishkuro:
commits not signed. please resolve (with a comment) review comments that are already addressed or stale.

yurishkuro (commented):
Looked through the diff — overall this is the right direction. The body-side resizer was scrolling out of view in long traces, and moving it under the sticky header lets the existing `height: 100vh-minus-nav` dragger keep the boundary grabbable from anywhere on screen. Bounds calculation is also better than the old one — using `headerNameWidth = nameColumnWidth * mainFraction` (header-coordinate) rather than the raw `spanNameColumnWidth` (body-coordinate) is the correct fraction for the normalized [0, 1] resizer position. The `clamp01` helper and the explicit `Math.min(..., sidePanelResizerMax)` to keep `min ≤ max` are nice defensive touches.

A few things worth tightening before merge:

**PR description doesn't match the code.** It says "so the divider can be dragged from the header as well as the body area," but the body `VerticalResizer` is deleted. Dragging from the body area still works only because `VerticalResizer.css` gives the dragger `100vh`-minus-nav height while the resizer itself sits under the sticky header. Please rephrase so the description reads true to the diff (e.g. "moved into the sticky header; the existing full-height dragger keeps body-area grabbing working").

**`vi.mock('./duck', ...)` is added in `index.test.jsx`.** I can't tell from this diff alone why it became necessary — `index.tsx` doesn't import the duck in the visible change. If the production component dispatches Redux actions alongside the Zustand setter (per the "Zustand + Redux" test name), this mock was presumably needed all along, in which case its arrival here looks like unrelated drift. Could you say a sentence in the PR body about what prompted the new mock?

**Import-path inconsistency for `SIDE_PANEL_WIDTH_MIN/MAX`.** The new `TimelineHeaderRow.test.jsx` imports them from `'../store.constants'`, while `index.test.jsx` mocks them on `'./store'`. If both modules export the constants (one re-exporting the other), it's harmless but inconsistent. Could you confirm the canonical location and import from one place?

**Optional props in `TimelineHeaderRow` are over-cautious.** `onSidePanelWidthChange?`, `sidePanelResizerMin?`, `sidePanelResizerMax?` all optional, with a runtime `showSidePanelResizer` boolean that checks each via `typeof === 'number'`. `TimelineHeaderRow` has one caller, which always passes all three. Required props would simplify the type and drop the branching. Stylistic, not a bug.

**Regression-test name could be more informative.** `"does not render a body VerticalResizer between main and side panel when timeline bars are visible"` is purely a regression guard; a one-line comment ("moved into the header row in #3928") or a tweak to `"now lives in TimelineHeaderRow instead of the body"` would help future readers.

**Checklist nit.** "I have signed all commits" is unchecked in the PR template. Project requires DCO sign-off — please confirm with `git log --show-signature` (or `git interpret-trailers`) and add a sign-off if missing.

Nothing blocking, all minor. Once the description matches the diff and the `./duck` mock is explained, this should be ready to go.

PR #3929, opened 2026-05-19
fix(api): accept otlp proto3 default omissions

### LINKED ISSUE #3891: [Bug]: Strict OTLP schemas (Span/ResourceSpans/InstrumentationScope) reject real wire-format fixtures due to omitted zero-valued required fields
### What happened?

After `scripts/postprocess-schemas.cjs` strips `.partial()` calls from the generated Zod schemas in `packages/jaeger-ui/src/api/v3/generated-client.ts`, the top-level OTLP types (`Span`, `ResourceSpans`, `ScopeSpans`, `InstrumentationScope`, `Resource`, `Span_Event`, `Span_Link`, `Status`) treat *every* field as required. Real OTLP wire format omits scalar fields whose value is the proto3 default (empty string, 0, false, empty array), so these schemas reject conformant backend responses.

Reproducing against the existing fixture:

```ts
import { schemas } from './generated-client';
import fixture from '../../utils/fixtures/otlp2jaeger-in.json';
schemas.TracesData.parse(fixture); // throws ZodError with 40+ issues
```

A non-exhaustive list of failure paths on otlp2jaeger-in.json:

resourceSpans[0].resource.droppedAttributesCount — missing
resourceSpans[0].scopeSpans[0].scope.{version, attributes, droppedAttributesCount} — missing
resourceSpans[0].scopeSpans[0].spans[0].{traceState, flags, events, droppedEventsCount, links, droppedLinksCount, droppedAttributesCount} — missing
resourceSpans[0].scopeSpans[0].spans[0].status.{code, message, details} — status: {} on the wire, but Status requires all three

### Steps to reproduce

The bug is reproducible against an existing fixture in the repo, no backend required:

1. Add a one-shot test at `packages/jaeger-ui/src/api/v3/_probe.test.ts`:

   ```ts
   import { schemas } from './generated-client';
   import fixture from '../../utils/fixtures/otlp2jaeger-in.json';

   describe('OTLP wire-format probe', () => {
     it('TracesData schema accepts a real Jaeger fixture', () => {
       const result = schemas.TracesData.safeParse(fixture);
       expect(result.success).toBe(true);
     });
   });

2. Run npm test -w packages/jaeger-ui -- src/api/v3/_probe.test.ts.

3. Observe ~40 Invalid input: expected ... received undefined issues across the paths listed in "What happened?".


### Expected behavior

Schemas should accept proto3-conformant wire data: fields with default scalar values (and empty repeated/message fields) may be absent. Only fields that are semantically required for the UI to function correctly — traceId, spanId, name, kind, startTimeUnixNano, endTimeUnixNano, status — should be enforced as required.

### Relevant log output

```shell
N/A
```

### Screenshot

N/A

### Additional context

This is the broader instance of the bug fixed for the four union types in #3890. The fix there was small because the four union types had a clean signal (Partial<{...}> in their TS types), so the postprocess script could be made surgical. The top-level types don't have that signal: their generated TS shape declares every field, even though the wire-format reality is more nuanced. Possible directions:

Per-field truly-required allowlist in postprocess: for each top-level schema, name the fields that are semantically required; everything else gets .optional() regardless of the OpenAPI spec.
Fix the OpenAPI spec in jaegertracing/jaeger-idl to mark proto3-default-omittable fields as nullable: true / not in required: — but that's a cross-repo change.
Drop strict mode: accept the original .partial() output and rely on consumers + TypeScript types to enforce required fields.
(1) is least invasive and keeps strict validation where it matters. Needs maintainer input on the per-schema required-field lists.

### Jaeger backend version

N/A

### SDK

N/A

### Pipeline

N/A

### Stogage backend

N/A

### Operating system

N/A

### Deployment model

N/A

### Deployment configs

```shell
N/A
```

MohitGoyal09:
  - Resolves #3891

  ## Description of the changes
  - Restore proto3 JSON optionality for OTLP schemas generated in `generated-client.ts`.
  - Keep semantically required span fields enforced while allowing default-valued scalar/repeated/message fields to be omitted from wire-format responses.
  - Update `scripts/postprocess-schemas.cjs` so regenerated API schemas preserve this per-field optionality.
  - Add schema tests covering the existing OTLP fixture and preserving rejection for spans missing required identity fields.

  - `make lint`
  - `NODE_OPTIONS=--localstorage-file=/private/tmp/jaeger-ui-node-localstorage-full-3891 make test`
  - `NODE_OPTIONS=--localstorage-file=/private/tmp/jaeger-ui-node-localstorage npm test -w packages/jaeger-ui -- src/api/v3/schemas.test.ts`

  ## Checklist

### DIFF
--- a/packages/jaeger-ui/src/api/v3/generated-client.ts
+++ b/packages/jaeger-ui/src/api/v3/generated-client.ts
@@ -31,83 +31,93 @@ const Operation = z.object({ name: z.string(), spanKind: z.string() }).passthrou
 const GetOperationsResponse = z.object({ operations: z.array(Operation) }).passthrough();
 const GoogleProtobufAny = z.object({ '@type': z.string() }).passthrough();
 const Status = z
-  .object({ code: z.number().int(), message: z.string(), details: z.array(GoogleProtobufAny) })
+  .object({
+    code: z.number().int().optional(),
+    message: z.string().optional(),
+    details: z.array(GoogleProtobufAny).optional(),
+  })
   .passthrough();
 const GetServicesResponse = z.object({ services: z.array(z.string()) }).passthrough();
-const ArrayValue: z.ZodType<ArrayValue> = z.lazy(() => z.object({ values: z.array(AnyValue) }).passthrough());
+const ArrayValue: z.ZodType<ArrayValue> = z.lazy(() =>
+  z.object({ values: z.array(AnyValue).optional() }).passthrough()
+);
 const KeyValueList: z.ZodType<KeyValueList> = z.lazy(() =>
-  z.object({ values: z.array(KeyValue) }).passthrough()
+  z.object({ values: z.array(KeyValue).optional() }).passthrough()
 );
 const AnyValue: z.ZodType<AnyValue> = z.lazy(() =>
   z
     .object({
-      stringValue: z.string(),
-      boolValue: z.boolean(),
-      intValue: z.string(),
-      doubleValue: z.number(),
-      arrayValue: ArrayValue,
-      kvlistValue: KeyValueList,
-      bytesValue: z.string(),
+      stringValue: z.string().optional(),
+      boolValue: z.boolean().optional(),
+      intValue: z.string().optional(),
+      doubleValue: z.number().optional(),
+      arrayValue: ArrayValue.optional(),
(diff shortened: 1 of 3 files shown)


yurishkuro:
this is not an issue today as we do not use schemas to validate OTLP

PR #4000, opened 2026-05-31
refactor(ids): normalize trace/span IDs at API boundary

### LINKED ISSUE #3963: Normalize trace/span IDs at API boundary, not scattered across UI
## Problem

Trace and span IDs from the Jaeger backend have historically been returned both with and without leading zeros (e.g. `0000abcd...` vs `abcd...`). As a result, the UI defensively strips leading zeros in multiple unrelated places:

- `src/hooks/useTraceLoading.ts` — `getCachedTrace` does a two-lookup fallback with `id.replace(/^0*/, '')`
- `src/utils/filter-spans.ts` (lines 61, 71) — strips zeros on both sides of span ID comparisons
- `src/components/SearchTracePage/SearchResults/index.tsx` (line 217) — strips zeros when looking up span links by trace ID

This is the wrong layer for normalization. Trace/span IDs should be treated as opaque strings everywhere in the UI. The stripping logic should happen once, at the API boundary — in `transformTraceData` or the v3 client — so that all downstream code receives a canonical form and never needs to know about the inconsistency.

## Proposal

* Identify the exact point where trace/span IDs enter the application (v1 `transformTraceData`, v3 client response parsing) and normalize there (e.g. strip leading zeros, or enforce a fixed-width padded form — whichever matches what the backend guarantees).
* Publish a proposal to change, with inventory.
* Then remove all the scattered `replace(/^0*/, '')` calls from the UI.

swetalin-10:

Closes #3963

Trace and span IDs from the Jaeger backend have historically been returned both with and without leading zeros (e.g. `0000abcd…` vs `abcd…`), so the same trace/span could be looked up under two different string keys. This PR adds a single `normalizeId` helper and calls it at the two places IDs actually enter the app (`transformTraceData`, the v1/legacy API boundary, and the v3 client), so every ID has one canonical form from there on.

**Canonical form: strip, not pad.** `normalizeId(id) = id.replace(/^0+/, '').toLowerCase() || '0'`. IDs are treated as opaque hex blobs — this only removes the padding-vs-not-padding ambiguity and case ambiguity, it does not reinterpret the ID as a number.

## Description of the changes

**`model/transform-trace-data.ts`** — `normalizeId(id)` strips leading zeros and lowercases (the `|| '0'` guard avoids returning an empty string for an all-zero input). `transformTraceData` now applies it to:
- `traceID`
- `span.spanID`, in the first pass, before the dedup map (`spanIdCounts`) and `spanMap` are seeded
- `ref.spanID` for every reference on every span, in the same pass, so the second-pass `spanMap.get(ref.spanID)` parent/child lookups always hit

**`api/v3/client.ts`** — `fetchTraceSummaries` now normalizes `s.traceId` before it becomes `TraceSummary.traceID`, so search results and navigation see the canonical form too.

## Test changes

- `transform-trace-data.test.js` — `normalizeId()` unit tests now cover strip+lowercase (previously only lowercase); the integration test that asserted IDs pass through **unchanged** is replaced with one that builds a padded/uppercase ID and asserts it comes back stripped and lowercased.
- `CriticalPath/testCases/test1,2,4-9.js` + `findLastFinishingChildSpan.test.js` + `sanitizeOverFlowingChildren.test.js` — span IDs lowercased consistently on both the input and expected sides, since `transformTraceData` now actually lowercases them.
- `CriticalPath/testCases/test3.js` — its span ID fixture has a genuine leading `00...`; the expected `criticalPathSections[0].spanID` is updated to the stripped form.
- `api/v3/client.test.ts` — added a test that feeds a padded/uppercase `traceId` through `fetchTraceSummaries` and asserts the stripped/lowercased result, so the v3 boundary is actually exercised.

## What changed since the last review

- `normalizeId` previously had zero callers anywhere in source — it was defined and unit-tested but never wired into `transformTraceData` or the v3 client, so no normalization happened at either API boundary and `Closes #3963` wasn't accurate. It's now called at both sites described above.
- Dropped the unexplained `testTimeout: 30000` bump in `vitest.config.ts` — unrelated to this change and unjustified.
- Reverted the `SearchResults/index.test.jsx` edit that rewrote object keys to identical literal strings — `SearchResults` isn't touched by this PR's production code, so that edit was a no-op.
- Rebased onto latest `main`; `reducers/trace.ts`/`trace.test.js` (the Redux trace slice) was deleted upstream in the meantime, so this PR's edit to `trace.test.js` was dropped along with the file.

- `npm test` — full suite passes
- `npx tsc -p packages/jaeger-ui/tsconfig.json` — no new type errors in any file this PR touches

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/api/v3/client.test.ts
+++ b/packages/jaeger-ui/src/api/v3/client.test.ts
@@ -319,6 +319,23 @@ describe('JaegerClient', () => {
       expect(summary.services).toEqual(mockApiSummary.services);
     });
 
+    it('normalizes traceId: strips leading zeros and lowercases', async () => {
+      const paddedSummary = {
+        ...mockApiSummary,
+        traceId: '0000BBBBCCCCDDDD0000111122223333',
+      };
+      mockFetch.mockResolvedValue({
+        ok: true,
+        json: async () => ({ summaries: [paddedSummary] }),
+      });
+
+      const promise = client.fetchTraceSummaries(query);
+      vi.runAllTimers();
+      const [summary] = await promise;
+
+      expect(summary.traceID).toBe('bbbbccccdddd0000111122223333');
+    });
+
     it('falls back to defaults for all optional fields when absent', async () => {
       // Only traceId is required; everything else including timestamps is optional.
       const minimal = { traceId: 'aaaabbbbccccdddd0000111122223333' };
--- a/packages/jaeger-ui/src/api/v3/client.ts
+++ b/packages/jaeger-ui/src/api/v3/client.ts
@@ -10,6 +10,7 @@
 
 import prefixUrl from '../../utils/prefix-url';
 import { ServicesResponseSchema, OperationsResponseSchema, TraceSummariesResponseSchema } from './schemas';
+import { normalizeId } from '../../model/transform-trace-data';
 import type { SearchQuery } from '../../types/search';
 import type { TraceSummary, ServiceSummary } from '../../types/trace-summary';
 import type { Microseconds } from '../../types/units';
@@ -104,7 +105,7 @@ export class JaegerClient {
         errorSpanCount: svc.errorSpanCount,
       }));
       return {
-        traceID: s.traceId,
+        traceID: s.traceId ? normalizeId(s.traceId) : '',
         traceName:
           rootServiceName && rootOperationName
             ? `${rootServiceName}: ${rootOperationName}`
--- a/packages/jaeger-ui/src/components/TracePage/CriticalPath/testCases/test1.js
+++ b/packages/jaeger-ui/src/components/TracePage/CriticalPath/testCases/test1.js
@@ -22,12 +22,12 @@ const testTrace = {
   traceID: 'test1-trace',
   spans: [
     {
-      spanID: 'span-E',
+      spanID: 'span-e',
       operationName: 'operation E',
       references: [
         {
           refType: 'CHILD_OF',
-          spanID: 'span-C',
+          spanID: 'span-c',
         },
       ],
       startTime: 50,
@@ -35,7 +35,7 @@ const testTrace = {
       processID: 'p1',
     },
     {
-      spanID: 'span-C',
+      spanID: 'span-c',
       operationName: 'operation C',
       references: [],
       startTime: 1,
@@ -43,12 +43,12 @@ const testTrace = {
       processID: 'p1',
     },
     {
-      spanID: 'span-D',
+      spanID: 'span-d',
       operationName: 'operation D',
       references: [
         {
           refType: 'CHILD_OF',
-          spanID: 'span-C',
+          spanID: 'span-c',
         },
       ],
       startTime: 20,
@@ -67,27 +67,27 @@ const transformedTrace = transformTraceData(testTrace);
 
 const criticalPathSections = [
   {
-    spanID: 'span-C',
+    spanID: 'span-c',
     sectionStart: 60,
     sectionEnd: 101,
   },
   {
-    spanID: 'span-E',
+    spanID: 'span-e',
     sectionStart: 50,
     sectionEnd: 60,
   },
   {
-    spanID: 'span-C',
+    spanID: 'span-c',
     sectionStart: 40,
     sectionEnd: 50,
   },
   {
-    spanID: 'span-D',
+    spanID: 'span-d',
     sectionStart: 20,
     sectionEnd: 40,
   },
   {
-    spanID: 'span-C',
+    spanID: 'span-c',
     sectionStart: 1,
     sectionEnd: 20,
   },
--- a/packages/jaeger-ui/src/components/TracePage/CriticalPath/testCases/test2.js
+++ b/packages/jaeger-ui/src/components/TracePage/CriticalPath/testCases/test2.js
@@ -24,7 +24,7 @@ const happyTrace = {
   traceID: 'trace-123',
   spans: [
     {
-      spanID: 'span-X',
+      spanID: 'span-x',
       operationName: 'op1',
       startTime: 1,
       duration: 100,
@@ -34,14 +34,14 @@ const happyTrace = {
       logs: [],
     },
     {
-      spanID: 'span-A',
+      spanID: 'span-a',
       operationName: 'op2',
       startTime: 10,
       duration: 40,
       references: [
         {
           refType: 'CHILD_OF',
-          spanID: 'span-X',
+          spanID: 'span-x',
         },
       ],
       processID: 'p1',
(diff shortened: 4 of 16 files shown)


yurishkuro:
> Converts a trace or span ID to its canonical form: lower-case, **no leading zeros**.

this is the opposite of what we need. The days of 64bit trace IDs are long gone, with OTEL data the IDs are 128bit which are supposed to be treated as opaque blobs, not as "numbers". So the hex encoding of these blobs preserves all the bits, including any leading zeros.

The UI needs to minimize the corrective actions, not add more of them. The only corrective action that would make sense is when we take trace/span IDs from user input (textbox or URL) and they are shorter than expected then we can try to pad them with left zeros to reach the required lenght.

swetalin-10:
You're right, I had the wrong mental model - normalizing by stripping zeros treats the ID as a number rather than an opaque bit string, which is exactly the wrong thing to do for OTEL IDs.

Updated the PR:

- `normalizeId` now only lowercases. Leading zeros from the backend are left intact since they're part of the actual 128-bit representation.
- For user input (filter textbox, `spanLinks` config keys): instead of stripping, short inputs are left-padded to match the expected length - `filter.padStart(span.spanID.length, '0')` - so a user typing an abbreviated ID still finds the right span.
- Restored `006c3cf93508f205` in test3 (that leading zero was genuine data, not something to clean up).

The tests that were validating strip behaviour are replaced with ones that confirm padding works correctly.

yurishkuro:
why do we need to lowercase?

swetalin-10:
Good point - we don't need to. Backend IDs are opaque blobs and their encoding should come through unchanged. Lowercasing was a holdover from before this PR that I carried forward without questioning it.

Removed it from both `transformTraceData` and the v3 client. The only places that still call `toLowerCase()` are the user-input boundaries (URL, filter textbox, `spanLinks` config keys), where we also pad to 32 chars - that's intentional so case-insensitive user input resolves to the right trace.

yurishkuro:
Thanks for tackling #3963 — but I think this PR needs significant rework before it does what it claims. I reviewed the diff against the head branch and the linked issue, and the implementation and the description have diverged sharply.

**The core change is a no-op.** `normalizeId` is exported and unit-tested, but `grep -rn normalizeId` over all non-test source on this branch returns only the definition — **zero callers**. Inside `transformTraceData`, `traceID` and `span.spanID` are still returned verbatim, and the v3 client maps `traceID: s.traceId ?? ''` rather than through `normalizeId`. So no normalization happens at any API boundary, and the PR's own new test *"passes traceID and spanIDs through unchanged from the backend"* asserts exactly that (`spanID === '00' + rootSpanID`). As written, `Closes #3963` isn't accurate — the issue's core ask is unaddressed.

**Even if it were wired in, it wouldn't fix the issue.** `normalizeId` only does `id.toLowerCase()`. #3963 is specifically about leading zeros (`0000abcd` vs `abcd`); lowercasing alone leaves those as distinct keys. The description's "strip vs pad" discussion refers to `replace(/^0+/, '') || '0'` logic that isn't in the code.

**The new comment in `useTraceLoading.ts` is misleading.** It states *"IDs are normalized once at the API boundary (transform-trace-data.ts)…"*, but that invariant is never established. This will lead future maintainers to trust a canonical form that doesn't exist.

**The PR needs a rebase to drop the dead-code / no-op churn.** Please rebase onto latest `main` and remove:
- the unused `normalizeId` helper (or actually call it), and
- the behavior-neutral test churn — the 8 `CriticalPath/testCases/*` files plus the two CriticalPath util tests lowercase span IDs on *both* the input and the expected sides (no-ops since nothing normalizes), and the `SearchResults`/`transform-trace-data` test edits rewrite keys to identical values. These inflate a 17-file diff whose real production change is ~3 lines and make it hard to see what actually changed.

**Also unexplained:** the `testTimeout: 30000` bump in `vitest.config.ts` is unrelated to ID normalization and not mentioned in the description — this usually masks a slow test rather than fixing it. Please justify or drop it.

**Direction from here** — pick one:
1. Implement the issue for real: add a strip-and-lowercase normalizer (`id.replace(/^0+/, '') || '0'`, lowercased) and call it in both `transformTraceData` (traceID, span.spanID, ref.spanID) **and** the v3 client, then keep only tests that actually exercise `0000abcd → abcd`; or
2. Close this and open a focused PR once the approach is settled.

Happy to re-review once it's rebased. Could you also update the description to match the final implementation, since the current inventory (7 call sites, strip semantics) describes changes that aren't in the diff?

swetalin-10:
@yurishkuro Thanks for the detailed review — I've reworked this per your "implement it for real" option. Summary of what changed:

- `normalizeId` now does `id.replace(/^0+/, '').toLowerCase() || '0'` (strip + lowercase, not lowercase-only), and is actually called at both API boundaries:
  - `transformTraceData`: `traceID`, `span.spanID`, and every `ref.spanID` (in the first pass, before `spanMap` is seeded, so the parent/child lookups still hit)
  - `api/v3/client.ts`: the `traceId` → `TraceSummary.traceID` mapping
- Replaced the test that asserted IDs pass through unchanged with ones that actually exercise `0000ABCD` → `abcd` in both `transform-trace-data.test.js` and `client.test.ts`.
- Dropped the unused-helper/no-op churn: reverted `SearchResults/index.test.jsx`'s key rewrite (that file isn't touched by this PR's production code), and dropped the unexplained `testTimeout: 30000` bump in `vitest.config.ts`.
- Kept the `CriticalPath` test-fixture lowercasing — it's no longer a no-op now that normalization is real (removing it would break those tests against the actual fix).
- Rebased onto latest `main`. In the process, `main` had deleted the dead `reducers/trace.ts`/`trace.test.js` Redux slice (#3999) — dropped my now-moot edit to that file along with it.
- Updated the PR description to match this implementation.

Full suite (`npm test`) and `tsc` are clean, and CI is green (DCO, build-and-check, lint-build, unit-tests, CodeQL). Ready for re-review whenever you have a chance.

yurishkuro:
Thanks for the thorough work here, @swetalin-10 — the inventory of scattered `replace(/^0+/, '')` calls and the v1-vs-v3 boundary analysis in your comment on #3963 were genuinely useful, and your open question ("strip or pad?") turned out to be the crux.

After thinking it through, we went a different direction, now merged in #4118: **treat trace/span IDs as fully opaque and normalize nothing.** The reasoning is written up in [my comment on #3963](https://github.com/jaegertracing/jaeger-ui/issues/3963#issuecomment-4970714251), but the short version:

- The two original motivations for normalization are obsolete on the output path — the infinite-loop bug (#128) was a caching-architecture issue that React Query + an authoritative post-load URL redirect now handle, and the backend no longer strips leading zeros.
- The answer to your open question turned out to be **neither strip nor pad**. Both are hex-only, format-assuming transformations. The moment an ID arrives base64-encoded (OTLP), a character-level `0`-strip or `0`-pad corrupts it — base64 leading *bytes* are not `'0'` *characters*. So the only representation-safe choice is to never mutate the string.
- #4118 removed all the sites you inventoried and replaced them with a single reconciliation point: after a trace loads, if the backend's canonical ID differs from the URL, redirect and seed the cache under both keys.

Because this PR normalizes at the boundary (the opposite of what shipped), I'm closing it as superseded rather than asking you to rework it. Really appreciate the analysis — it directly shaped the final approach. Hope to see you on other issues.

PR #4006, opened 2026-06-01
fix(search): Hide unsupported trace summary columns

### LINKED ISSUE #4005: [trace summaries] Suppress columns not supported by storage
In one of the implementations of 3rd party gRPC storage it returns the trace summaries like these:
```json
        {
            "traceId": "d557570a0bc414eb4c7bdeb39a6df1ff",
            "rootServiceName": "some-service",
            "rootOperationName": "some-endpoint",
            "minStartTimeUnixNano": "1780336447581003008",
            "maxEndTimeUnixNano": "1780336447744034048",
            "spanCount": 55,
            "services": []
        }
```
Namely, it never returns `services` and never returns `errorCount` or other counts, only `spanCount`. The table display could adapt to this by hiding the columns from the table altogether instead of showing `-` for services and `0` for error (the latter is completely wrong and misleading).

Ref: docs/adr/010-trace-summary-api.md in the backend repo

StantonMatt:
## Summary

- Track whether v3 trace summaries include service summaries and supplemental error/count data.
- Hide Services and Errors columns when a partial storage backend does not provide those fields.
- Render unsupported row-level error counts as unknown when mixed with supported summaries.

Fixes #4005

## Verification

- PATH=/opt/homebrew/opt/node@24/bin:$PATH npm run fmt
- PATH=/opt/homebrew/opt/node@24/bin:$PATH npm test -w packages/jaeger-ui -- src/api/v3/client.test.ts src/components/SearchTracePage/SearchResults/TraceTable.test.tsx
- PATH=/opt/homebrew/opt/node@24/bin:$PATH npm run lint
- PATH=/opt/homebrew/opt/node@24/bin:$PATH npm test
- PATH=/opt/homebrew/opt/node@24/bin:$PATH npm run build
- review-fix-loop clean

### DIFF
--- a/packages/jaeger-ui/src/api/v3/client.ts
+++ b/packages/jaeger-ui/src/api/v3/client.ts
@@ -103,6 +103,7 @@ export class JaegerClient {
         spanCount: svc.spanCount ?? 0,
         errorSpanCount: svc.errorSpanCount ?? 0,
       }));
+      const hasServiceSummaries = services.length > 0;
       return {
         traceID: s.traceId,
         traceName:
@@ -117,6 +118,9 @@ export class JaegerClient {
         errorSpanCount: s.errorSpanCount ?? 0,
         orphanSpanCount: s.orphanSpanCount ?? 0,
         services,
+        serviceSummariesSupported: hasServiceSummaries,
+        errorSpanCountSupported:
+          s.errorSpanCount !== undefined || s.orphanSpanCount !== undefined || hasServiceSummaries,
       };
     });
   }
--- a/packages/jaeger-ui/src/api/v3/client.test.ts
+++ b/packages/jaeger-ui/src/api/v3/client.test.ts
@@ -317,6 +317,8 @@ describe('JaegerClient', () => {
       expect(summary.errorSpanCount).toBe(1);
       expect(summary.orphanSpanCount).toBe(0);
       expect(summary.services).toEqual(mockApiSummary.services);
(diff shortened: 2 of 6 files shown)


packages/jaeger-ui/src/api/v3/client.ts
```diff
@@ -117,6 +118,9 @@ export class JaegerClient {
         errorSpanCount: s.errorSpanCount ?? 0,
         orphanSpanCount: s.orphanSpanCount ?? 0,
         services,
+        serviceSummariesSupported: hasServiceSummaries,
```
yurishkuro, line 121: this doesn't below in the client, it's a view concern and it can compute these flags across all rows.

packages/jaeger-ui/src/api/v3/client.ts
```diff
@@ -117,6 +118,9 @@ export class JaegerClient {
         errorSpanCount: s.errorSpanCount ?? 0,
         orphanSpanCount: s.orphanSpanCount ?? 0,
```
yurishkuro, line 119: these need to be relaxed to `number | undefined` in the type.

StantonMatt:
Thanks for the quick follow-up here. Since #4007 has now merged and fixes #4005 on `main`, I’m closing this PR as superseded rather than rebasing an alternate implementation over the same surface. I verified the conflict was from #4007, and the issue is now addressed upstream.

PR #4046, opened 2026-06-08
fix: Fix sidepanel resize when timeline is hidden

sksingh2005:
## Which problem is this PR solving?
- Resolves #4045 

## Description of the changes
- Updated the render condition for `VerticalResizer` in `TimelineHeaderRow.tsx` to ensure it remains visible and active when either the timeline bars OR the side panel are visible.
- Adjusted the `resizerMax` calculation logic in `TraceTimelineViewer/index.tsx` to properly account for the available layout width (`1 - SIDE_PANEL_WIDTH_MIN`) when the timeline is disabled but the side panel is open, preventing negative or invalid resize boundaries.

## How was this change tested?

https://github.com/user-attachments/assets/2f0f20b4-8486-44ec-b71a-0d0ef6f6362e

- Ran the existing automated test suite (`npm run fmt`, `npm run lint`, `npm test`, `npm run build`) to ensure no regressions.

## Checklist
- [x] I have read https://github.com/jaegertracing/jaeger/blob/main/CONTRIBUTING_GUIDELINES.md
- [x] I have signed all commits
- [ ] I have added unit tests for the new functionality
- [x] I have run lint and test steps successfully: `make lint test`

## AI Usage in this PR (choose one)
See [AI Usage Policy](https://github.com/jaegertracing/jaeger/blob/main/CONTRIBUTING_GUIDELINES.md#ai-usage-policy).
- [ ] **None**: No AI tools were used in creating this PR
- [x] **Light**: AI provided minor assistance (formatting, simple suggestions)
- [ ] **Moderate**: AI helped with code generation or debugging specific parts
- [ ] **Heavy**: AI generated most or all of the code changes

PR #4047, opened 2026-06-08
fix: keep span detail side panel resizable when timeline bars are hidden

### LINKED ISSUE #4045: [Bug]: Span Detail sidepanel is not resizable when timeline is hidden
### What happened?

Per title

### Steps to reproduce

Per title

### Expected behavior

Side panel should remain resizable

### Screenshot

No resize icon

Image

94xhn:

- Resolves #4045

When the timeline bars are hidden while the span detail side panel is active, the side panel cannot be resized — no resize handle is shown.

Both vertical resizers were gated behind `timelineBarsVisible`:
- the name-column resizer in `TimelineHeaderRow`, and
- the side-panel resizer in `TraceTimelineViewer`.

With the bars hidden, the timeline column collapses and the side panel absorbs it, leaving a single boundary between the Service/Operation name column and the side panel — but neither resizer is rendered, so that boundary has no draggable handle.

## Description of the changes

- In `TraceTimelineViewer`, always render a `VerticalResizer` while the side panel is active:
  - **Timeline bars visible** (unchanged): the resizer sits at `1 - sidePanelWidth` and drives `sidePanelWidth`.
  - **Timeline bars hidden** (new): the side panel width is derived from `spanNameColumnWidth` (`effectiveSidePanelWidth = 1 - spanNameColumnWidth`), so the resizer is placed at the name-column / side-panel boundary (`spanNameColumnWidth`) and resizes the side panel by adjusting the name column width, clamped between `SPAN_NAME_COLUMN_WIDTH_MIN` and `min(SPAN_NAME_COLUMN_WIDTH_MAX, 1 - SIDE_PANEL_WIDTH_MIN)`.

- Updated `TraceTimelineViewer/index.test.jsx`:
  - the layout-combination matrix and the former "does not render a VerticalResizer when timeline bars are hidden" case now expect the resizer to be present whenever the side panel is active;
  - added a case asserting that, with the bars hidden, the resizer's `onChange` drives `setSpanNameColumnWidth` and leaves `setSidePanelWidth` untouched.
- `vitest run` on `TraceTimelineViewer/index.test.jsx` → 16 passing.
- `tsc` typecheck and `oxlint` pass; formatted with `vp fmt`.

## Checklist

## AI Usage in this PR

- [x] **Moderate**: AI helped with code generation and debugging specific parts


94xhn:
Thanks for the review! Both points addressed in 0c71f260:

- **Side panel max width**: clamped the hidden-bars resizer `min` to `Math.max(SPAN_NAME_COLUMN_WIDTH_MIN, 1 - SIDE_PANEL_WIDTH_MAX)`, so the derived side panel width (`1 - spanNameColumnWidth`) can no longer exceed `SIDE_PANEL_WIDTH_MAX`. Combined with the existing `max`, the name column is now bounded so the panel stays within `[SIDE_PANEL_WIDTH_MIN, SIDE_PANEL_WIDTH_MAX]`.
- **Test contract**: added `expect(props.setSidePanelWidth).not.toHaveBeenCalled()` to the hidden-bars resize test so both the Zustand store and the Redux-tracking prop are covered.

parshipcy (commented):
Can you please add a screencast in the PR description?

parshipcy (changes requested):
`setSpanNameColumnWidth` (and related clamping in `store.layout.ts` / `duck.ts`) using the formula: `1 - sidePanelWidth - MIN_TIMELINE_COLUMN_WIDTH` even when bars are hidden, so i think dragging can feel stuck sometime.

parshipcy:
@yurishkuro i think this PR supersedes #4046 once the store fixes are in... same bug fix, but resize handling stays in the body in this PR but for #4046 the same is in the header. What do you think?

94xhn:
@yurishkuro @parshipcy Just pushed the store-level fix (c587c725): `setSpanNameColumnWidth` and `setColumnWidth` now branch on `timelineBarsVisible`. With the bars hidden the side panel absorbs the timeline column, so the name column drags across the full range instead of sticking at ~0.575 (the old clamp still reserved `MIN_TIMELINE_COLUMN_WIDTH`). Added regression tests for the bars-hidden case in store.test.ts and duck.test.js.

So #4047 is now self-contained. The resize handling here lives in the timeline body (index.tsx), next to the columns it controls, rather than in the header. Happy to align with whatever you and the maintainers consider the cleaner home, and to reconcile anything useful from #4046.

I'll add a screencast of the bars-hidden resize to the PR description shortly.

yurishkuro:
4046 was much further along, I merged it.

Issue #3926, opened 2026-05-19
Tracking: ADR 0004 Migrate from Redux to Zustand + TanStack React Query


parshipcy:
## Summary

Track the **incremental migration** away from classic **Redux** (`redux`, `redux-actions`, `connect`) toward the architecture described in [ADR 0004: State Management Strategy](https://github.com/jaegertracing/jaeger-ui/blob/main/docs/adr/0004-state-management-strategy.md) and [ADR 0008: Target state management architecture](https://github.com/jaegertracing/jaeger-ui/blob/main/docs/adr/0008-target-state-management-architecture.md).

- [x] #3704 -> ADR 0008

### Phase 0: Foundations
- [x] #3662 -> Shared TanStack Query client, Zustand + class bridge, Type alignment

### Phase 1: Zustand for client UI state (Redux still holds server data)
- [x] #3702  -> Trace compare (traceDiff)
- [x] #3715  -> Archive notifier (archive)
- [x] #3719 -> layout prefs (trace timeline)
- [x] #3721 -> collapse/expand and detail state (trace timeline)
- [x] #3750 -> DDG view modifiers
- [x] #3761 -> Embedded / Page / TopNav flags
- [x] #3781 -> `config` Redux slice

### Phase 2: TanStack Query for server state
- [x] #3835 -> Single trace load
- [x] #3943 -> Search + file upload
- [x] #3966 -> Services and operations discovery
- [x] #3991 -> Dependencies page

parshipcy:
@yurishkuro can you please reopen this?

parshipcy:
remaining redux work is also an LFX mentorship term 3 deliverable under #4278, and that [RFC 0004](https://github.com/jaegertracing/jaeger-ui/blob/main/docs/rfc/0004-state-management-strategy.md) is the checklist.

PR #4048, opened 2026-06-08
feat(monitor): Migrate metrics fetching from Redux to React Query

parshipcy:

- Resolves part of #3926

## Description of the changes

- Replaces the Redux metrics reducer and action creators with TanStack Query hooks as part of RFC 0004 Phase 2f.
- Adds `useServiceMetricsQuery` and `useOperationMetricsQuery` beside `Monitor/ServicesView`.
- Preserves Monitor service, span-kind, and timeframe URL synchronization while loading metrics through the new hooks.
- Removes the metrics reducer from the root reducer and `ReduxState`.
- Keeps partial metrics results when individual requests fail and associates failures with explicit metric labels.
- Updates the Monitor tests and RFC 0004 migration status.

- Jaeger UI test suite: 3,739 tests passed.
- Plexus test suite: 115 tests passed.
- Type checking, formatting, linting, and production builds passed.
- Manual navigation to `/monitor`.

## Checklist

- [ ] I have read https://github.com/jaegertracing/jaeger/blob/main/CONTRIBUTING_GUIDELINES.md


packages/jaeger-ui/src/components/Monitor/ServicesView/useMetricsQuery.ts
```diff
@@ -0,0 +1,320 @@
...
+      const payload = (await Promise.allSettled([
+        JaegerAPI.fetchMetrics('latencies', [serviceName], { ...params, quantile: 0.5 }),
+        JaegerAPI.fetchMetrics('latencies', [serviceName], { ...params, quantile: 0.75 }),
+        JaegerAPI.fetchMetrics('latencies', [serviceName], { ...params }),
+        JaegerAPI.fetchMetrics('calls', [serviceName], { ...params }),
+        JaegerAPI.fetchMetrics('errors', [serviceName], { ...params }),
+      ])) as FetchedAllServiceMetricsResponse;
+      return transformServiceMetrics(payload);
+    },
+    enabled: Boolean(serviceName && params),
```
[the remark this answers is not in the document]
parshipcy: no runtime impact

packages/jaeger-ui/src/hooks/useMetricsQuery.ts
```diff
@@ -0,0 +1,320 @@
...
+ * (`Promise.allSettled` across 5 `GET /api/metrics/*` calls).
+ */
+export function useServiceMetricsQuery(
+  serviceName: string | undefined,
+  params: MetricsAPIQueryParams | undefined
+): UseQueryResult<ServiceMetricsResult, ApiError> {
+  return useQuery({
+    queryKey: serviceMetricsQueryKey(serviceName, params!),
+    queryFn: async () => {
+      if (!serviceName || !params) throw new Error('useServiceMetricsQuery: service and params required');
```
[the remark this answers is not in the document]
parshipcy: i think not necessary

packages/jaeger-ui/src/hooks/useMetricsQuery.ts
```diff
@@ -0,0 +1,320 @@
...
+          metricPoints: metrics.metrics[0].metricPoints.map((p: MetricPointObject) => {
+            let y: number | null;
+            try {
+              y = parseFloat(p.gaugeValue.doubleValue.toFixed(2));
+              max = y > max ? y : max;
+            } catch {
+              y = null;
+            }
+            return { x: new Date(p.timestamp).getTime(), y };
+          }),
```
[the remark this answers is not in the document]
parshipcy: fixed

packages/jaeger-ui/src/components/Monitor/ServicesView/index.tsx
```diff
@@ -205,28 +204,34 @@ export function MonitorATMServicesViewImpl(props: TProps) {
     }
   }, [servicesLoading, updateDimensions]);
 
-  // componentDidUpdate equivalent
+  //fresh fetch whenever the service list loads for the first time
```
yurishkuro: This pattern seems overly complicated to me. If we're using TanStack Query for loading both services and their metrics, shouldn't this be just two useXyz statements?

```
services = useServices()
currentService = ...
metrics = useServiceMetricsQuery(currentService)
```

Why do we need to re-fetch if `services` array changes if the selected service did not change?

parshipcy: valid point

packages/jaeger-ui/src/components/Monitor/ServicesView/useMetricsQuery.ts
```diff
@@ -0,0 +1,334 @@
...
+      const payload = (await Promise.allSettled([
+        JaegerAPI.fetchMetrics('latencies', [serviceName!], { ...params!, quantile: 0.5, endTs }),
+        JaegerAPI.fetchMetrics('latencies', [serviceName!], { ...params!, quantile: 0.75, endTs }),
+        JaegerAPI.fetchMetrics('latencies', [serviceName!], { ...params!, endTs }),
+        JaegerAPI.fetchMetrics('calls', [serviceName!], { ...params!, endTs }),
+        JaegerAPI.fetchMetrics('errors', [serviceName!], { ...params!, endTs }),
+      ])) as FetchedAllServiceMetricsResponse;
+      return transformServiceMetrics(payload);
+    },
+    enabled: Boolean(serviceName && params),
```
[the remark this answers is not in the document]
parshipcy: not required

packages/jaeger-ui/src/components/Monitor/ServicesView/index.tsx
```diff
@@ -117,6 +96,48 @@ export function MonitorATMServicesViewImpl(props: TProps) {
     store.getNumber('lastAtmSearchTimeframe', ONE_HOUR_MS)
   );
 
+  const currentService = selectedService || services[0];
+
+  // endTs is intentionally absent — it is resolved to Date.now() inside each
```
yurishkuro: I think we should have a Refresh button that would reset endTs and still pass it to queries as part of the key. endTs should also be refreshed by service & spanKind dropdown selections. Otherwise not only is the query behavior inconsistent with how Search screen works but also there's no way to update the timeseries except full page reload.

packages/jaeger-ui/src/components/Monitor/ServicesView/useMetricsQuery.ts
```diff
@@ -0,0 +1,321 @@
+// Copyright (c) 2026 The Jaeger Authors.
```
yurishkuro, line 1: why is this file in hooks/? It's not shareable with other components.

packages/jaeger-ui/src/components/Monitor/ServicesView/useMetricsQuery.ts
```diff
@@ -0,0 +1,321 @@
...
+
+function operationMetricsQueryKey(
+  serviceName: string | undefined,
+  params: MetricsHookParams
+): readonly ['operationMetrics', string | undefined, MetricsHookParams] {
+  return ['operationMetrics', serviceName, params] as const;
+}
+
+// Transformation helpers (extracted from src/reducers/metrics.ts)
+export type ServiceMetricsResult = {
```
yurishkuro, line 20: please define all types at the top of the file, before functions

packages/jaeger-ui/src/components/Monitor/ServicesView/useMetricsQuery.ts
```diff
@@ -0,0 +1,321 @@
...
+          }
+          serviceMetrics[metrics.name]!.push(metric);
+        } else {
+          serviceMetrics[metrics.name] = metric;
+        }
+      }
+    } else {
+      const reason = (promiseResult as PromiseRejectedResult).reason;
+      switch (i) {
+        case 0:
```
yurishkuro, line 108: is this how the original code was? Looks very fragile, don't we have some labels in the results to key off, instead of array positions?

packages/jaeger-ui/src/components/Monitor/ServicesView/index.tsx
parshipcy: done. can you please check this: 117e27e3ff50492349d6d5ebc8db4016f9933f9f ?

packages/jaeger-ui/src/components/Monitor/ServicesView/useMetricsQuery.ts
parshipcy: FIxed in 3547270d6768f7d23165286411e10483049d36ca

packages/jaeger-ui/src/components/Monitor/ServicesView/useMetricsQuery.ts
parshipcy: yes, this is same as the original code but I also feel that it's a bit fragile

packages/jaeger-ui/src/hooks/useMetricsQuery.ts
```diff
@@ -0,0 +1,321 @@
...
+      ? serviceMetricsQueryKey(serviceName, params)
+      : (['serviceMetrics', 'disabled'] as const),
+    queryFn: async () => {
+      const endTs = Date.now();
+      const payload = (await Promise.allSettled([
+        JaegerAPI.fetchMetrics('latencies', [serviceName!], { ...params!, quantile: 0.5, endTs }),
+        JaegerAPI.fetchMetrics('latencies', [serviceName!], { ...params!, quantile: 0.75, endTs }),
+        JaegerAPI.fetchMetrics('latencies', [serviceName!], { ...params!, endTs }),
+        JaegerAPI.fetchMetrics('calls', [serviceName!], { ...params!, endTs }),
+        JaegerAPI.fetchMetrics('errors', [serviceName!], { ...params!, endTs }),
```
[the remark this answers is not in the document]
parshipcy: @yurishkuro should we fix it?

packages/jaeger-ui/src/components/Monitor/ServicesView/index.tsx
```diff
@@ -205,28 +202,44 @@ export function MonitorATMServicesViewImpl(props: TProps) {
...
     return <LoadingIndicator vcentered centered />;
   }
 
   return (
     <>
-      {_isEmpty(metrics && metrics.serviceMetrics && metrics.serviceMetrics.service_latencies) && (
+      {_isEmpty(serviceMetrics && serviceMetrics.service_latencies) && (
         <Alert
           message={
             <>
```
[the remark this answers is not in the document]
parshipcy: yes I think this is valid, @yurishkuro what do u think?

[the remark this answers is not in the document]
yurishkuro: yes - why should we render anything (aside from spinner) if the data is still loading?

parshipcy:
@yurishkuro ptal when you have time

DsThakurRawat:
took this for a run since it's been open since June and two independent URL-sync PRs (#4119, #4182) landed on main while it sat here. main is 177 commits ahead now.

merged main into the PR head locally. four files actually conflict, and none of them are mechanical, they're all the same root cause: this PR and the URL-sync work both touch `ServicesView` in overlapping ways because neither knew about the other.

`reducers/index.ts` is trivial, main independently dropped the `trace` reducer (a different migration) while this PR drops `metrics`. composes cleanly, both go.

the real one is `index.tsx`. this PR's `MonitorATMServicesViewImpl` takes zero props and gets everything from the new hooks. main's current version takes `props: TProps` with `search`/`navigate` from `withRouteProps`, because #4119/#4182 added URL-param syncing after this PR branched. resolving it naively either way breaks something: keep this PR's signature and the URL sync has nothing to read `search` from; keep main's `TProps` and you're back to the Redux dispatch props this PR is trying to delete. the fix is `MonitorATMServicesViewImpl(props: TOwnProps)`, i.e. keep just the two fields URL-sync needs and drop the Redux half, then thread `syncFiltersToUrl` calls into this PR's `setEndTime`-based handlers instead of the old `fetchAllServiceMetrics` dispatch calls they came attached to.

doing that surfaced two real bugs, not merge artifacts, both reproduced with the actual test suite:

`currentService` was `selectedService || services[0]`, no sanitization, while the render path (`getSelectedService()`) and every handler already call `resolveService()` to fall back when the selected value isn't in the services list. so an invalid or attacker-crafted URL param (`?service=evil%26foo=bar`) would render the sane fallback service name in the UI while firing the actual metrics query against the raw unsanitized string. `useServiceMetricsQuery`/`useOperationMetricsQuery` calls confirmed this with the wrong argument before the fix. one-line fix, `currentService = resolveService(selectedService, services)`.

second one's subtler. this PR's persistence effect writes `lastAtmSearchService` unconditionally whenever `currentService` changes. the URL-sync tests require that a URL-seeded filter never gets written back as the new stored default (`does not persist URL-sourced filters to localStorage`), and the old Redux-era `fetchMetrics` callback this PR deletes used to gate every write on `!urlOwned.current.<field>` for exactly that reason. dropping that callback as dead code (correctly, nothing else called it) also silently dropped the guard. replaced the one unconditional effect with one that persists service/spanKind/timeframe each gated on their own `urlOwned` flag.

adapted the test file the same way: kept this PR's own `on page load` describe as-is, ported all 15 tests from `URL query params` and `URL write-back` onto the hooks API (swap `mockFetchAllServiceMetrics` assertions for `useServiceMetricsQuery` ones, drop the Redux prop injection), and dropped `on page switch` plus the `mapStateToProps()`/`mapDispatchToProps()` describes, since those functions don't exist anymore and the one thing `on page switch` checked is already covered by this PR's own `on page load` test.

verified rather than assumed: `tsc --noEmit` clean, the full Monitor suite (10 files, 150 tests) green, the targeted ServicesView suite 40/40 (it was 5 failing on the two bugs above before those fixes landed).

one more thing this surfaced: ADR-0004's phase table doesn't exist anymore, #4275 split it into RFC 0004 back in July while this sat open, so the ADR update in this PR's own diff needs to go into `docs/rfc/0004-state-management-strategy.md` §2f instead. minor, but the reviewer bot won't catch a stale target file.

happy to push the resolved branch somewhere if that saves you the merge, or just leave this as the exact recipe. either way, worth landing, the hooks migration itself is solid.

yurishkuro (approved):

yurishkuro:
E2E smoke testing passed at `67861720` against the real Jaeger SPM development stack (Jaeger v2.21.0, Prometheus, and microsim).

The following workflows passed:

- The Monitor page rendered live service-level latency, error-rate, and request-rate graphs.
- The operation metrics table rendered data returned by Jaeger.
- Changing the service updated the URL and displayed the selected service metrics.
- Filtering operations reduced the table to matching rows.
- Changing the timeframe updated the URL, graph range, and operation data.
- Changing the span kind updated the URL; a span kind without metrics displayed the no-data state, and returning to `server` restored the data.
- Refresh displayed its loading state, fetched a newer metrics window, updated the rendered values, and cleared the loading state.

The backend returned nine services. For the `redis` service over a two-hour window, the three grouped metrics endpoints returned populated latency, call-rate, and error-rate series.

No PR-specific failures were found. Two pre-existing observations appeared during the run: Ant Design reports that `Alert.message` is deprecated, and the no-data instructions link resolves as a relative URL instead of the configured external URL.

Issue #444, opened 2019-08-27
Revisit policy for squashing commits


yurishkuro:
We've done an experiment with not squashing commits, imo it was a complete failure. Here's a snippet of git log since the last release
```
1969210 - Merge pull request #414 from tiffon/layered-digraph-schema (6 weeks ago) <Everett>
72b4e7a - Merge branch 'master' into layered-digraph-schema (6 weeks ago) <Everett>
0b43edf - Merge branch 'master' into ddg-hops-selector-component (6 weeks ago) <Joe Farro>
c2ba69c - Update DeepDependencies/Header className and test (6 weeks ago) <Everett Ross>
9d912fa - Move chevron out of button (6 weeks ago) <Everett Ross>
9b816a5 - Iterating on codecov config (#418) (6 weeks ago) <Joe Farro>
b4897af - Merge branch 'master' into layered-digraph-schema (6 weeks ago) <Joe Farro>
6b631d3 - Fix typo in codecov config (6 weeks ago) <Joe Farro>
2f0c3f3 - Merge branch 'master' into rm-patch-check-from-master (6 weeks ago) <Joe Farro>
efe8c50 - Centralize clickHandler, restrict passed props (6 weeks ago) <Everett Ross>
d1b21af - Fix typo in codecov config (#417) (6 weeks ago) <Joe Farro>
2245d4d - Merge branch 'master' into layered-digraph-schema (6 weeks ago) <Joe Farro>
7a74235 - Merge branch 'master' into rm-patch-check-from-master (6 weeks ago) <Joe Farro>
4a9a6ed - Merge branch 'master' into rm-patch-check-from-master (6 weeks ago) <Joe Farro>
b12a0c8 - Fix typo in codecov config (6 weeks ago) <Joe Farro>
351218f - Merge branch 'master' into layered-digraph-htmlgroup-measurableNodes (6 weeks ago) <Joe Farro>
611b20d - Allow MeasurableNodesLayer to be standalone (6 weeks ago) <Joe Farro>
861c016 - Merge pull request #415 from tiffon/rm-patch-check-from-master (6 weeks ago) <Yuri Shkuro>
1ce258a - HtmlLayersGroup MeasurableNodesLayer in plexus (6 weeks ago) <Joe Farro>
baf176c - Upgrade TypeScript to latest (6 weeks ago) <Joe Farro>
4a775bb - Remove patch check from master branch (6 weeks ago) <Joe Farro>
[...]

PR #4079, opened 2026-06-15
fix(plexus): Use currentColor for DDG edge weight labels in dark mode

### LINKED ISSUE #4070: [Bug/UI]: System Architecture edge weights are unreadable in Dark Mode (Contrast Issue)
### What happened?

While exploring the **System Architecture** graph (Deep Dependency Graph) in Dark Mode, I noticed a severe contrast issue. The text showing the relationship weights on the edges (the numbers between the nodes) is rendered in solid black. 

Against the dark blue/gray canvas background used in Dark Mode, this black text is almost entirely invisible, making it very hard to see the edge metrics.

### Steps to reproduce

1. Generate some traces (e.g., using HotROD).
2. Switch Jaeger UI to **Dark Mode** using the header toggle.
3. Navigate to the **System Architecture** tab.
4. Look at the numbers on the arrows connecting the services.

### Expected behavior

The edge text color should adapt to the theme. In dark mode, it should use a lighter color (like `var(--text-secondary)`) so it has sufficient contrast against the dark canvas, similar to how node labels are handled.

### Relevant log output

```shell

```

### Screenshot

Image

### Additional context

_No response_

### Jaeger backend version

_No response_

### SDK

_No response_

### Pipeline

_No response_

### Stogage backend

_No response_

### Operating system

_No response_

### Deployment model

_No response_

### Deployment configs

```shell

```

shubhtrek:
Resolves: #4070

## Context & Motivation

I was testing the System Architecture graph in Dark Mode and noticed the numbers
on the arrows connecting services were completely invisible. After digging into the
code, I found that the edge weight labels in `SvgEdge.tsx` had a hardcoded
`fill="#000"` — solid black text on a dark background. Similarly, `SvgEdgesLayer.tsx`
had `stroke: '#000'` hardcoded as the default edge stroke, with a pre-existing TODO
comment acknowledging it needed to be fixed.

Yurishkuro confirmed this is a bug — fallout from the dark theme migration.

## Screenshots

**Before (broken — labels invisible in dark mode):**
image

**After (fixed — labels readable in dark mode):**
Screenshot 2026-06-15 092044

## Implementation Details

The fix replaces hardcoded `#000` with `currentColor`, which is the SVG standard
way to inherit color from the CSS `color` property. Since Ant Design's theme engine
already sets `color` correctly for both light and dark modes, this just works.

- **`SvgEdge.tsx`** — changed `fill="#000"` to `fill="currentColor"` on the label text
- **`SvgEdgesLayer.tsx`** — changed `stroke: '#000'` to `stroke: 'currentColor'` and
  cleaned up the stale TODO comment
- **`SvgEdgesLayer.test.jsx`** — updated the existing assertion to match the new value

I went with `currentColor` instead of `var(--text-secondary)` because CSS variables
don't work as inline SVG attributes — they only work in CSS property context.
`currentColor` is the right tool here and is already used elsewhere for icon colors.

## Impact

- Edge weight labels are now readable in dark mode
- Light mode looks exactly the same — the DDG's CSS (`stroke: #444` on `.Ddg--Edges`)
  still overrides the stroke for arrow lines
- Removed a hardcoded magic value and a stale TODO from the plexus library

## Testing

- ✅ `npm run tsc-lint` — no type errors
- ✅ `npm test` — all 2,980 tests pass
- ✅ `npm run fmt` — no formatting changes needed
- ✅ `npm run build` — builds successfully

### DIFF
--- a/packages/plexus/src/Digraph/SvgEdge.tsx
+++ b/packages/plexus/src/Digraph/SvgEdge.tsx
@@ -76,7 +76,7 @@ function SvgEdge<U = {}>(props: TProps<U>) {
       />
 
       {label && (
-        <text x={labelX} y={labelY} fill="#000" fontSize="1rem" fontWeight="bold">
+        <text x={labelX} y={labelY} fill="currentColor" fontSize="1rem" fontWeight="bold">
           {label}
         </text>
       )}
--- a/packages/plexus/src/Digraph/SvgEdgesLayer.test.jsx
+++ b/packages/plexus/src/Digraph/SvgEdgesLayer.test.jsx
@@ -103,9 +103,9 @@ describe('SvgEdgesLayer', () => {
       expect(SvgLayer.lastProps.classNamePart).toBe('SvgEdgesLayer');
     });
 
-    it('passes extraWrapper with black stroke to SvgLayer', () => {
+    it('passes extraWrapper with currentColor stroke to SvgLayer', () => {
       render(<SvgEdgesLayer {...defaultProps} />);
-      expect(SvgLayer.lastProps.extraWrapper).toEqual({ stroke: '#000' });
+      expect(SvgLayer.lastProps.extraWrapper).toEqual({ stroke: 'currentColor' });
     });
 
     it('passes getClassName to SvgLayer', () => {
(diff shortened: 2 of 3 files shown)


parshipcy (approved):
LGTM

@yurishkuro 👀

yurishkuro (approved):

PR #4095, opened 2026-06-15
fix(ci): Use full npm install and bump engines.npm to 11.13.0

yurishkuro:
## Summary
- Fixes the Dependabot lockfile fixup workflow still producing diffs with missing `libc` fields
- Root cause: `--package-lock-only` doesn't fully resolve platform bindings, so npm skips writing `libc` metadata
- Fix: use `npm install --ignore-scripts` (full resolution) instead
- Also bumps `engines.npm` from `>=11.6.0` to `>=11.13.0` — the CI was installing 11.6.0 which predates consistent `libc` field writing. The lockfile on main was generated with 11.13.0

## Test plan
- [ ] After merge, re-run the workflow on PR #4059 (or wait for next Dependabot PR)
- [ ] Verify the lockfile diff no longer removes `libc` fields

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/dependabot-lockfile-fix.yml
+++ b/.github/workflows/dependabot-lockfile-fix.yml
@@ -30,7 +30,7 @@ jobs:
           npm install -g npm@$NPM_VERSION
 
       - name: Regenerate lockfile
-        run: npm install --package-lock-only --ignore-scripts
+        run: npm install --ignore-scripts
 
       - name: Commit if changed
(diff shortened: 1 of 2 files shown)


yurishkuro:
Superseded by #4096 which uses corepack for a proper sealed toolchain instead of post-hoc lockfile regeneration.

Issue #4061, opened 2026-06-12
[Bug]: AccordionEvents leaks setTimeout and requestAnimationFrame callbacks on unmount


Abhay-sonkar:
### What happened?

In AccordionEvents.tsx, the notifyListReflow callback schedules two setTimeout calls and one requestAnimationFrame call to emit reflow events on the window object, but it never captures the returned timeout IDs or the rAF handle.

If the component unmounts before these callbacks fire (for example, the user rapidly toggles the accordion and then navigates away  within 50ms), the emit function still executes after unmount and dispatches a "jaeger:list-resize" or "jaeger:detail-measure" event on the window object. This is a memory/event leak — the component has no way to cancel pending work on cleanup.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/AccordionEvents.tsx
Lines: 69-73

### Steps to reproduce

1. Open a trace that has events/logs in the SpanDetail panel
2. Expand the AccordionEvents section (click to open it)
3. Immediately navigate away from the trace page (or unmount the component) before the 50ms timeout fires
4. The emit() function still executes after unmount, dispatching "jaeger:list-resize" / "jaeger:detail-measure" events on window from a component that no longer exists

### Expected behavior

When AccordionEvents unmounts, any pending setTimeout calls and the requestAnimationFrame callback scheduled by notifyListReflow should
be cancelled via clearTimeout / cancelAnimationFrame, so no events  are dispatched on the window object after the component is gone.

### Relevant log output

```shell

```

### Additional context

A fix has been prepared:
- Store the two setTimeout IDs in a useRef array
- Store the requestAnimationFrame ID in a useRef
- Add a cleanup useEffect that calls clearTimeout on each stored  timer ID and cancelAnimationFrame on the stored rAF handle when the component unmounts

A PR implementing this fix will be linked once the issue is opened.

### Deployment configs

```shell

```

PR #4104, opened 2026-06-16
Cancel pending event reflow timers

starkmarkus:
Addresses #4061 by clearing pending timeout and animation frame callbacks when AccordionEvents unmounts, and adds a regression test that verifies the callbacks do not fire after unmount.


starkmarkus:
Thanks for the review. I will add a browser guard around the cleanup path and switch the timer test to safer spies/cleanup so it does not leak globals on failure.

starkmarkus:
Implemented the requested review fixes: `window` is now guarded in the reflow cleanup/scheduling path, and the event reflow tests now use `jest.spyOn(...)` with `finally` cleanup.

Issue #4099, opened 2026-06-16
[Bug]: TraceGraph mode buttons (S / T / ST) are missing aria-labels for screen readers


shubhtrek:
### What happened?

As a user relying on a screen reader I want the TraceGraph mode toggle buttons (S, T, ST) to have proper descriptive labels so that I can actually understand what they do instead of just hearing random single letters.

Right now, if you use the TraceGraph sidebar, the buttons just say "S", "T", and "ST". For normal users, there's a mouse hover tooltip, but for keyboard/screen reader users, the tool just reads "S, button". It's super confusing. Also, we can't tell which mode is currently active because it's missing the aria-pressed attribute.

### Steps to reproduce

1. Run Jaeger UI locally and open any trace.
2. Go to the "Trace Graph" tab.
3. Press Tab on your keyboard to reach the right-side sidebar buttons (S, T, ST).
4. Check with a screen reader (or DevTools accessibility tree) — notice it just says the letter without any proper description.

### Expected behavior

The buttons should have proper aria-label tags. For example:

S button should say "Color by service"
T button should say "Color by total time"
ST button should say "Color by self time"
Also, the currently active button should have aria-pressed={true} so users know which mode is on.

### Relevant log output

```shell

```

### Screenshot

<img width="1145" height="932" alt="Image" src="https://github.com/user-attachments/assets/fe3e0a22-66ab-41fc-80ad-5953780f3c7c" />

### Additional context

This violates basic accessibility rules (WCAG SC 4.1.2 and SC 1.3.1). I noticed this while fixing similar issues around the app. File: packages/jaeger-ui/src/components/TracePage/TraceGraph/TraceGraph.tsx around line 210. (Note: This is related to my other a11y PRs like #4065)

### Deployment configs

```shell

```

shubhtrek:
|| Under Construction ||
I have officially taken ownership of this issue and am currently writing the code for it

PR #4105, opened 2026-06-16
Add trace graph accessibility labels

starkmarkus:
Addresses #4099 by adding accessible names to the trace graph mode buttons and updating the test to assert the new accessible labels.


starkmarkus:
Thanks for the review. I will update the buttons to expose the active state with `aria-pressed`, make the tooltip text match the self-time wording, and relax the test assertions so they are less brittle.

starkmarkus:
Implemented the review fixes here: the mode buttons now expose `aria-pressed`, the tooltip/title text is `Self time`, and the tests use regex matchers instead of brittle exact-label checks.

Issue #3996, opened 2026-05-30
Speed up the Vitest unit-test suite


yurishkuro:
## Context

Wall-clock for the full `packages/jaeger-ui` Vitest run is ~33 s, but CPU time spent in tests is ~87 s. Of 244 test files, the top 20 (8%) account for **63%** of the test time. The `verbose` reporter shows the cumulative cost is dominated by infrastructure rather than test logic:

```
Duration  33.68s (transform 9.09s, setup 15.99s, import 104.50s,
                  tests 96.88s, environment 126.87s)
```

### Slowest test files (file-level duration)

| File | Time |
|---|---|
| `TracePage/TraceStatistics/index.test.jsx` | 7.2 s |
| `Monitor/ServicesView/operationDetailsTable/index.test.jsx` | 5.3 s |
| `App/TopNav.test.jsx` | 4.0 s |
| `DependencyGraph/DAGOptions.test.jsx` | 3.1 s |
| `App/index.test.jsx` | 3.0 s |
| `SearchTracePage/SearchForm.test.jsx` | 3.0 s |
| `SearchTracePage/index.test.jsx` | 2.9 s |
| `TracePage/TraceSpanView/index.test.jsx` | 2.9 s |
| `TracePage/TracePageHeader/TraceViewSettings.test.jsx` | 2.7 s |
| `Monitor/ServicesView/index.test.jsx` | 2.7 s |
| `App/ThemeProvider.test.tsx` | 2.7 s |

### Slowest individual cases

| Test | Time |
|---|---|
| `<MonitorATMPage> does not explode and renders initial elements` | 2149 ms |
| `<TraceSpanView> does not explode` | 1588 ms |
| `<TopNav> renders the \"Jaeger\" link` | 1274 ms |
| `<TraceTagOverview> check handler` | 1258 ms |
| `<OperationTableDetails> sort row` | 1215 ms |

The 1.2–2.1 s per-test costs are most plausibly attributable to full AntD-Table mounts (Table v6 runs several layout passes via `ResizeObserver`) plus per-`it` fresh `render()` of large component trees — not `waitFor` polling.

[...]

PR #4132, opened 2026-06-22
perf(tests): extract searchInTable as pure function

Anujkumar9081:
Part of #3996

Extracts `searchInTable` as an exported pure function and updates 3 tests 
to call it directly instead of via componentRef, eliminating 3 unnecessary 
component renders.

Before: TraceStatistics/index.test.jsx ~5804ms
After:  TraceStatistics/index.test.jsx ~5695ms


yurishkuro:
the performance gain does not warrant the size/complexity of the change.

PR #4136, opened 2026-06-23
fix(bug): clear trace ID input field on navigation/reload (#4135)

### LINKED ISSUE #4135: [Bug]: Find by Trace ID field does not clear on reload
Historically when you paste a trace ID into the Find by Trace ID box:

Image

The page would reload with the new trace and the box will be cleared. Now not only is it not cleared, but it stays this way even when user navigates away from trace page and comes back to it (even for another trace).

Most likely this regression was introduced when AI Assistant logic was added, which is co-located in the same text box, and for the assistant it is reasonable to preserve the textbox content across pages. But when the trace ID is recognized and loaded the box should be cleared.

shubhtrek:
Resolves: #4135

Context & Motivation
I noticed that when looking up a trace by pasting its ID into the "Lookup by Trace ID..." search box, the UI correctly navigates to the trace page but the input box does not clear. Since the top navigation menu persists across page loads, the ID remains stuck in the search box even if you click around the app or try to look up another trace. 

Historically, this box cleared automatically after a search. The issue started when the new combined AI Assistant textarea logic was added. While it makes sense for the assistant to retain context, the trace ID lookup should be cleared once the navigation starts.

Video:
https://drive.google.com/file/d/12TZER0_Ilp2w-0nHOPWL5GdUw3Jzp4Ii/view?usp=sharing

Implementation Details
The problem was that the `TraceLookupSearchInput` was rendered as an uncontrolled Ant Design `<Input />` component. Because the navigation is handled via client-side routing (`useNavigate`) and the header doesn't remount, the input element kept its DOM value.

- Converted `TraceLookupSearchInput` in `JaegerAskSearchInput.tsx` to a controlled component using a React state variable (`value`).
- Updated the submit handler to reset the state (`setValue('')`) after a successful navigation.
- Added assertions (`expect().toHaveValue('')`) to the unit tests in `JaegerAskSearchInput.test.jsx` to verify that both the default lookup input and the assistant textarea are cleared on form submit.

Impact
- Looking up a trace ID now clears the text field immediately upon navigation.
- The input is clean and ready for the next query without requiring the user to manually clear it.

Testing
✅ npm test — all unit tests pass (including the updated `JaegerAskSearchInput.test.jsx` test cases)
✅ npm run lint — no lint or type-check errors
✅ npm run build — builds successfully

### DIFF
--- a/packages/jaeger-ui/src/components/App/JaegerAskSearchInput.test.jsx
+++ b/packages/jaeger-ui/src/components/App/JaegerAskSearchInput.test.jsx
@@ -50,7 +50,7 @@ describe('<JaegerAskSearchInput /> trace lookup (assistant disabled)', () => {
     expect(screen.getByPlaceholderText('Lookup by Trace ID...')).toBeInTheDocument();
   });
 
-  it('navigates to trace page on form submit', () => {
+  it('navigates to trace page on form submit and clears the input', () => {
     render(
       <MemoryRouter>
         <JaegerAskSearchInput />
@@ -62,6 +62,7 @@ describe('<JaegerAskSearchInput /> trace lookup (assistant disabled)', () => {
 
     expect(mockNavigate).toHaveBeenCalledTimes(1);
     expect(mockNavigate).toHaveBeenCalledWith(`/trace/${traceId}`);
+    expect(screen.getByTestId('idInput')).toHaveValue('');
   });
 
   it('does not navigate when input is empty', () => {
@@ -123,6 +124,7 @@ describe('<JaegerAskSearchInput /> assistant mode', () => {
     fireEvent.keyDown(textarea, { key: 'Enter', shiftKey: false });
 
     expect(mockNavigate).not.toHaveBeenCalled();
+    expect(textarea).toHaveValue('');
   });
 
   it('still navigates for 16–32 hex trace ids (preserves original form)', () => {
@@ -140,6 +142,7 @@ describe('<JaegerAskSearchInput /> assistant mode', () => {
     fireEvent.keyDown(textarea, { key: 'Enter', shiftKey: false });
 
     expect(mockNavigate).toHaveBeenCalledWith(`/trace/${traceId}`);
+    expect(textarea).toHaveValue('');
   });
 
   it('navigates for base64-encoded trace ids', () => {
@@ -157,6 +160,7 @@ describe('<JaegerAskSearchInput /> assistant mode', () => {
     fireEvent.keyDown(textarea, { key: 'Enter', shiftKey: false });
 
     expect(mockNavigate).toHaveBeenCalledWith('/trace/%2FlhpXXcq1Bdw%2B4twt863jg%3D%3D');
+    expect(textarea).toHaveValue('');
   });
 
   it('Shift+Enter does not submit', () => {
--- a/packages/jaeger-ui/src/components/App/JaegerAskSearchInput.tsx
+++ b/packages/jaeger-ui/src/components/App/JaegerAskSearchInput.tsx
@@ -18,13 +18,17 @@ const ASK_JAEGER_PLACEHOLDER = 'Ask Jaeger or lookup trace';
 
 function TraceLookupSearchInput() {
   const navigate = useNavigate();
+  const [value, setValue] = React.useState('');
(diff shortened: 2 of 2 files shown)


yurishkuro:
Thanks for the fix! This is a valid bug, but there's a simpler solution that avoids converting to a controlled component. Since the input is already uncontrolled and you have access to the form, you can just call `form.reset()`:

```tsx
function TraceLookupSearchInput() {
  const navigate = useNavigate();
  const goToTrace = React.useCallback(
    (event: React.FormEvent<HTMLFormElement>) => {
      event.preventDefault();
      const form = event.currentTarget;
      const value = (form.elements.namedItem('idInput') as HTMLInputElement)?.value;
      if (value) {
        navigate(getUrl(value));
        form.reset();
      }
    },
    [navigate]
  );
  // ... rest unchanged
}
```

This is a one-line addition (`form.reset()`) with no need for `useState`, no need to wire up `value`/`onChange`, and no dependency array issue (the current PR puts `value` in the `useCallback` deps, which re-creates the handler on every keystroke — harmless here but unnecessary).

yurishkuro:
and please update PR to up to date main

shubhtrek:
@yurishkuro Please review again!

packages/jaeger-ui/src/components/App/JaegerAskSearchInput.test.jsx
```diff
@@ -50,7 +50,7 @@ describe('<JaegerAskSearchInput /> trace lookup (assistant disabled)', () => {
     expect(screen.getByPlaceholderText('Lookup by Trace ID...')).toBeInTheDocument();
   });
 
-  it('navigates to trace page on form submit', () => {
+  it('navigates to trace page on form submit and clears the input', () => {
```
yurishkuro, line 53: your original version had a check that the input was cleared, now it's removed, why?

shubhtrek: I accidentally dropped those assertions while refactoring the input back to an uncontrolled component with form.reset(). I've just added them back.

packages/jaeger-ui/src/components/App/JaegerAskSearchInput.tsx
```diff
@@ -25,6 +25,7 @@ function TraceLookupSearchInput() {
       const value = (form.elements.namedItem('idInput') as HTMLInputElement)?.value;
       if (value) {
         navigate(getUrl(value));
+        form.reset();
```
parshipcy: I think `form.reset()` is not working here as ant design's `<Input>` manage its own internal state, so the HTML's `form.reset()` doesn't reset what the user sees. I think the consistent solution will be to use a controlled value + `setValue('')` after navigation (same pattern as `JaegerAskAssistantSearchInput`) and like what witten in the PR description.

shubhtrek: One sec! I check manually and tell you

shubhtrek: I Tryed It's works but please ensure again

parshipcy: yes, but the previous fix where you added just `form.reset()`, that was not working

yurishkuro (approved):

yurishkuro:
@parshipcy are you saying this fix is wrong? I didn't test it live.

shubhtrek:
@yurishkuro the @parshipcy said the past code is not good but this one is good I'v passed test I taught

parshipcy:
> @parshipcy are you saying this fix is wrong? I didn't test it live.

the previous fix where just the `form.reset()` was added as the fix was not working, reason: https://github.com/jaegertracing/jaeger-ui/pull/4136#discussion_r3457002306. The current fix is working fine but one small nit: the `valueRef` pattern is not really needed for a submit handler that only runs on submit, we could simplify to:
```
const goToTrace = React.useCallback(
  (event: React.FormEvent<HTMLFormElement>) => {
    event.preventDefault();
    if (value) {
      navigate(getUrl(value));
      setValue('');
    }
  },
  [navigate, value]
);
```
same behavior, a bit easier to read, not a blocker though.

PR #4143, opened 2026-06-24
refactor(TracePage): Decouple search logic

### LINKED ISSUE #4134: refactor(TracePage): Decouple view-specific search logic from TracePage
## Problem

`TracePage` currently owns the search/find logic for all view types. It computes `traceDagEV` (the graph DAG), calls `filterSpans` for timeline views, and calls `getUiFindVertexKeys` for the graph view — all in the parent component. This means:

1. The parent must hold internal data structures (`traceDagEV`) that belong to child views.
2. Adding a new view that supports search requires modifying `TracePage` to add another branch in the search computation block.
3. The parent has intimate knowledge of *how* each view searches, not just *that* it supports search.

Related: `hideMap` and `navigable` were recently addressed with helper functions (`viewTypeShowsMinimap`, `viewTypeIsNavigable`), but search is the bigger coupling issue.

## Proposed Solution

Use a callback prop (push model) so each view owns its own search logic and reports results upward. The data flow is fully declarative/reactive:

1. The parent passes `uiFind` (the search query string) down as a prop — it already does this today.
2. Each view recomputes matches internally whenever `uiFind` changes (via `useMemo`).
3. Each view pushes results back to the parent via an `onSearchResults` callback.

No imperative "please search now" signal is needed — just data flowing down and results flowing up.

```tsx
function TraceGraph({ trace, uiFind, onSearchResults }) {
  const dagEV = useMemo(() => calculateTraceDagEV(trace), [trace]);
  const matches = useMemo(() => getUiFindVertexKeys(uiFind, dagEV.vertices), [uiFind, dagEV]);

  useEffect(() => { onSearchResults(matches); }, [matches, onSearchResults]);

  return <Graph ev={dagEV} highlights={matches} />;
}
```

The parent becomes:

```tsx
// Parent provides query and receives results
<TraceGraph trace={trace} uiFind={uiFind} onSearchResults={setFindMatches} />
<TraceTimelineViewer trace={trace} uiFind={uiFind} onSearchResults={setFindMatches} />
```

Each view internally computes matches (using whatever data structures it needs) and pushes `{ count: number, matches: Set<string> }` to the parent via `onSearchResults`. The parent only displays the count in the header and passes matches back for highlighting.

This eliminates:
- `traceDagEV` computation in `TracePage`
- `graphFindMatches` / `spanFindMatches` branching in `TracePage`
- The need to modify `TracePage` when adding searchable views

## Scope

- Move `calculateTraceDagEV` call into `TraceGraph`
- Move `filterSpans` / `filterPrunedSpanIDs` calls into `TraceTimelineViewer`
- Define a common `onSearchResults` callback prop signature
- Update `TracePage` to receive and display results without knowing how they were computed
- Ensure `TraceFlamegraph`, `TraceSpanView`, etc. can opt in to the same pattern if they support search

## Context

This pattern is sometimes called "inversion of control on search" — the parent knows *that* views are searchable, not *how* they search. It follows the same principle as the `viewTypeShowsMinimap()` helpers: view-specific behavior should be declared/owned by the view, not encoded in the parent.

vanshika2720:
Move search computation ownership from `TracePage` into `TraceGraph` and `TraceTimelineViewer`. Each view now computes its own search matches and reports the result count via an `onSearchResults` callback.

`TraceGraph` now owns DAG creation and graph-specific search logic, while `TraceTimelineViewer` owns span filtering and pruned-span search logic. `TracePage` is reduced to coordinating the search query and displaying the result count.

`TraceStatistics` search behavior is intentionally preserved unchanged to avoid behavioral regressions and keep the scope of this refactor focused.

Closes #4134

* Resolves #4134
* Removes view-specific search logic from `TracePage`
* Reduces coupling between `TracePage` and individual trace views
* Makes it easier for future views to implement search without requiring changes to `TracePage`

## Description of the changes

* Added a shared `onSearchResults` callback contract for reporting search result counts
* Moved `calculateTraceDagEV()` and `getUiFindVertexKeys()` into `TraceGraph`
* Moved `filterSpans()` and `filterPrunedSpanIDs()` into `TraceTimelineViewer`
* Updated `TracePage` to manage only the search count state and callback handling
* Removed graph-specific search branching from `TracePage`
* Reset search count when switching between trace views
* Removed unused props and imports introduced by the previous architecture
* Preserved existing search highlighting and navigation behavior
* Left `TraceStatistics` search implementation unchanged to avoid regressions

* Ran formatting and lint checks successfully
* Ran TypeScript validation successfully
* Verified production build succeeds
* Verified graph search highlighting and result counts
* Verified timeline search highlighting, result counts, and next/previous navigation
* Verified search counts update correctly when switching views
* All 3205 unit tests passed

## Checklist

* [ ] I have added unit tests for the new functionality

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceGraph/TraceGraph.test.jsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceGraph/TraceGraph.test.jsx
@@ -7,7 +7,6 @@ import { render, screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { LayoutManager } from '@jaegertracing/plexus';
 import transformTraceData from '../../../model/transform-trace-data';
-import calculateTraceDagEV from './calculateTraceDagEV';
 import TraceGraph, { setOnEdgePath } from './TraceGraph';
 import { MODE_SERVICE, MODE_TIME, MODE_SELFTIME } from './OpNode';
 import testTrace from './testTrace.json';
@@ -66,7 +65,6 @@ vi.mock('@jaegertracing/plexus', () => {
 });
 
 const transformedTrace = transformTraceData(testTrace);
-const ev = calculateTraceDagEV(transformedTrace.asOtelTrace());
 
 describe('<TraceGraph>', () => {
   let props;
--- a/packages/jaeger-ui/src/components/TracePage/TraceGraph/TraceGraph.test.jsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceGraph/TraceGraph.test.jsx
@@ -74,7 +74,7 @@ describe('<TraceGraph>', () => {
   beforeEach(() => {
     props = {
       headerHeight: 60,
-      ev,
+      trace: transformedTrace.asOtelTrace(),
     };
   });
 
@@ -181,12 +181,17 @@ describe('<TraceGraph>', () => {
   it('handles uiFind mode correctly', () => {
     const propsWithUiFind = {
       ...props,
-      uiFind: 'test-service',
-      uiFindVertexKeys: new Set(['key1', 'key2']),
+      uiFind: 'service1',
     };
     render(<TraceGraph {...propsWithUiFind} />);
     const wrapper = screen.getByTestId('mock-digraph').parentElement;
     expect(wrapper).toHaveClass('is-uiFind-mode');
+  });
+
+  it('fires onSearchResults callback when trace and uiFind are provided', () => {
+    const onSearchResults = jest.fn();
+    render(<TraceGraph {...props} uiFind="service1" onSearchResults={onSearchResults} />);
+    expect(onSearchResults).toHaveBeenCalledWith(expect.objectContaining({ count: expect.any(Number) }));
   });
 
   it('initializes with correct default mode', () => {
--- a/packages/jaeger-ui/src/components/TracePage/TraceGraph/TraceGraph.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceGraph/TraceGraph.tsx
@@ -16,21 +16,25 @@ import {
   MODE_SELFTIME,
   getHelpTable,
 } from './OpNode';
-import { TEv, TSumSpan } from './types';
+import { TSumSpan } from './types';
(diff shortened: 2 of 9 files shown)


anshgoyalevil (commented):
Thanks for taking this on — the direction here (push model with `onSearchResults`) is exactly what the issue asked for! While testing it live I ran into a behavior regression when switching between views with an active search query, and wanted to flag it with a concrete repro so it's easy to fix.

**Repro:** Open a trace, type a term into the Find box while on *Trace Timeline* (the header correctly shows the match count, e.g. "1"). Then switch to *Trace Graph* using the view selector. The header count drops to **0** and stays there — even though the matching span/node is clearly highlighted. Switching back to *Trace Timeline* reproduces the same thing. It only recovers if you retype the search text.

**Root cause:** in `TracePage/index.tsx` there's:
```tsx
useEffect(() => {
  setFindCount(0);
}, [viewType]);
```
This resets the count whenever `viewType` changes, but `TraceGraph`/`TraceTimelineViewer` also report their real count from their own effect when they mount/update. React fires effects child-before-parent within a commit, so the child's freshly-reported count is always overwritten by this parent effect's `setFindCount(0)` right after. (This can even happen on first render if `uiFind` is already set via the URL.) Since nothing else re-triggers a report until the search text itself changes, the header stays stuck at 0.

For reference, `TraceStatistics`/`TraceSpansView` don't hit this because their count is computed inside `TracePage` itself in a single effect gated on `[spanFindMatches, viewType]` — there's no separate effect in another component racing with it.

**Suggested fix:** rather than resetting the count via a competing effect, tag the reported result with the view type it was computed for (e.g. store `{ viewType, count }` from `handleSearchResults` and derive `findCount = findResult.viewType === viewType ? findResult.count : 0`). That avoids the cross-component effect ordering race entirely, since there's nothing that needs to "win" a race — the stale result just doesn't match the current view type.

Happy to see this land — nice job scoping the rest of the refactor!

vanshika2720:
@anshgoyalevil Thanks for the detailed review! I've pushed the requested changes and added a regression test for the reported issue. Would appreciate another review when you have time.

vanshika2720:
@yurishkuro Please review this.

Issue #4017, opened 2026-06-05
[Feature]: Allow enhancing span names with "summary fields" overlays


JnRouvignac:
### Requirement

As a person troubleshooting traces with hundreds of thousands of spans, I find it hard to spot patterns when span names follow OpenTelemetry best practices of low-cardinality.
I want span names in the UI to be overlaid with "summary fields"
so that I can investigate large traces AND follow OpenTelemetry best practices for span naming

### Problem: High Cardinality in span names vs. UI Observability

When investigating large traces containing hundreds of spans with the same name, it is incredibly difficult to distinguish between them or spot patterns.
Because OpenTelemetry best practices require low-cardinality span names (e.g., `get_user` instead of `get_user/314159`), critical identifying data (like `customerId` or `orderId`) must be set as span attributes.
This creates friction when investigating large traces:

* When there are many spans in a trace, it can be hard to get the information you are looking for out of span attributes.
* Users must click on dozens of individual spans, expand the attributes panel, and hunt for the specific ID to find the one they need.

Using high-cardinality span names (e.g., appending `customerId` or `orderId`) would fix the UI issue, but it violates OpenTelemetry standards and duplicates data already stored in attributes.

References:

* OpenTelemetry best practices
    * HTTP span names: https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name
    * General span names: https://github.com/open-telemetry/opentelemetry-specification/blob/v1.56.0/specification/trace/api.md#span

### Proposal: "Summary Fields"

A feature similar to Google Cloud Logs Explorer's "Summary Fields" may solve the problem.

This feature would allow users to dynamically select a span attribute (like `customerId`) and append or overlay that attribute value next to the span names across the entire trace view.
This way users can quickly scan for patterns or specific IDs across hundreds of spans at a glance, without bloating index data or breaking OpenTelemetry low-cardinality best practices.

References:
* Google logs explorer "summary fields":
    * doc: https://docs.cloud.google.com/logging/docs/view/logs-explorer-interface#add_summary_fields
    * demo: https://storage.googleapis.com/gweb-cloudblog-publish/original_images/11.logs_explorer_simple_query_mode_hide_default_summary_fields.gif

udita-0707:
Hi! @JnRouvignac I'd like to work on this issue.

[...]

PR #4149, opened 2026-06-27
feat(trace): Add HTTP status summary chips

udita-0707:
Introduce HTTP status span pills on trace timeline span rows, gated behind `traceTimeline.spanPillsEnabled` (`default true`). Part 1 of the incremental summary-fields plan for #4017.

- First incremental step for #4017 — users can see HTTP status codes on span rows without opening span details
- Supersedes #4063
- Reshapes the former “data layer only” PR into a small, user-visible improvement (per maintainer feedback)

## Description of the changes

- Add `spanPills.ts` with `ISpanPill`, `getSpanPillsForSpan()`, and private `HTTP` status attribute handling (PR1: HTTP status only - no generic lookup yet)
- Add `useSpanPillsEnabled()` in the same module for config gating (enabled unless explicitly set to `false`)
- Compute pills per visible span in `SpanBarRow` (not up front for the whole trace); error severity (`isError`) is set where pills are built
- Pass only a boolean `spanPillsEnabled` through `TraceTimelineViewer` → `VirtualizedTraceView` → `SpanBarRow`
- Add `traceTimeline.spanPillsEnabled` config flag (default true; set to `false` to disable)

Architecture: Pill sources, labels, and error styling live in `spanPills.ts`. Components stay generic. `VirtualizedTraceView` does not know about pill contents.

1. Manual: with default config, pills appear on spans with`http.status_code` / `http.response.status_code`, red styling for 5xx, no chips when disabled or attribute missing

- `npm test -w packages/jaeger-ui -- src/components/TracePage/TraceTimelineViewer/spanPills.test.ts src/components/TracePage/TraceTimelineViewer/useSpanPills.test.ts src/components/TracePage/TraceTimelineViewer/SpanBarRow.test.jsx src/components/TracePage/TraceTimelineViewer/index.test.jsx src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.test.jsx`
- `npm run lint`
- `npm run build`

## Checklist


packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/summaryFieldsUtils.ts
```diff
@@ -0,0 +1,110 @@
...
+    if (hasValue) {
+      result.set(span.spanID, values);
+    }
+  }
+
+  return result;
+}
+
+export const buildAvailableFields = memoizeOne(buildAvailableFieldsImpl);
+export const buildSummaryLookup = memoizeOne(buildSummaryLookupImpl);
```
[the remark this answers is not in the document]
udita-0707: Good catch, `memoize-one` uses reference equality, so a new `selectedFields` array with the same contents would bypass the cache. Added a custom equality fn to compare array contents (trace still compared by reference). Test added in 5f983ceb.

parshipcy (dismissed):
I don't see any blockers, lgtm👍

udita-0707:
@yurishkuro PTAL Thanks!

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/summaryFieldsUtils.ts
```diff
@@ -0,0 +1,130 @@
...
+  const result = new Map<string, Record<string, string>>();
+  if (selectedFields.length === 0) {
+    return result;
+  }
+
+  const fieldSet = new Set(selectedFields);
+
+  for (const span of trace.spans) {
+    const values: Record<string, string> = {};
+    let hasValue = false;
```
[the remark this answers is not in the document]
udita-0707: Valid point, trace attribute keys are untrusted. Switched `values` to `Object.create(null)` so keys like `__proto__` can't mutate the prototype chain. Added a test.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/store.layout.ts
```diff
@@ -31,18 +34,54 @@ type TraceTimelineLayoutPrefsStore = {
...
+    const trimmed = key.trim();
+    if (!trimmed || seen.has(trimmed)) continue;
+    seen.add(trimmed);
+    normalized.push(trimmed);
+    if (normalized.length >= MAX_SUMMARY_FIELDS) break;
+  }
+  return normalized;
+}
+
+function parseStoredSummaryFields(): string[] {
```
yurishkuro: the functionality seems to be to "load" fields, not to "parse" them (parse is an implementation detail of loading)

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/store.layout.ts
```diff
@@ -31,18 +34,54 @@ type TraceTimelineLayoutPrefsStore = {
...
+  selectedSummaryFields: string[];
   setSpanNameColumnWidth: (width: number) => void;
   setSidePanelWidth: (width: number) => void;
   // Updates layout fields only; use `setDetailPanelMode` from `./store` to also sync detail panel state.
   applyDetailPanelModeToLayout: (mode: SpanDetailPanelMode) => void;
   setTimelineBarsVisible: (visible: boolean) => void;
+  setSelectedSummaryFields: (fields: string[]) => void;
 };
 
+function normalizeSummaryFields(fields: unknown[]): string[] {
```
yurishkuro: comment needed - what does "normalize" mean?

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/summaryFieldsUtils.ts
```diff
@@ -0,0 +1,130 @@
...
+
+export const MAX_SUMMARY_FIELDS = 3;
+
+export type AvailableField = {
+  key: string;
+  coverage: number;
+  total: number;
+};
+
+function safeStringify(value: unknown): string {
```
yurishkuro: why is the input `unknown` when it's called with known 3 types?

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/summaryFieldsUtils.ts
```diff
@@ -0,0 +1,130 @@
...
+  }
+  for (let i = 0; i < next.length; i += 1) {
+    if (next[i] !== prev[i]) {
+      return false;
+    }
+  }
+  return true;
+}
+
+export const buildAvailableFields = memoizeOne(buildAvailableFieldsImpl);
```
yurishkuro: memoization can be done by the caller. And the modern patterns typically do not require the use of memoizeOne in the callers (maybe useMemo or such).

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/summaryFieldsUtils.ts
```diff
@@ -0,0 +1,130 @@
+// Copyright (c) 2026 The Jaeger Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+import memoizeOne from 'memoize-one';
+
+import { AttributeValue, IOtelTrace } from '../../../types/otel';
+
+export const MAX_SUMMARY_FIELDS = 3;
```
yurishkuro: What's the objective of this limit? 3 seems like a very arbitrary value - if it's performance related then 3 or 5 probably wouldn't make much difference (20 would). Since we are using virtualized view for the timeline, computing N pills per span is usually not a big deal. At most the Settings panel could show a warning saying "selecting too many fields may degrade UI performance".

yurishkuro (commented):
I think [I suggested a slightly suboptimal breakdown](https://github.com/jaegertracing/jaeger-ui/pull/4063#pullrequestreview-4580892951) of the original PR. The most useful way to structure PRs is by delivering incremental value with small changes. E.g.
  * PR1 - implement minimal stubs of summaryFieldUtils, e.g. by checking for HTTP status attribute only, and the display wiring that actually shows it in the timeline. And a config feature that enables the whole thing. (which is two lines of code). This immediately gives you a working version of the final capability, just limited to a single field which is not configurable.
  * PR2 - extend with roughly what you have here (again maybe hardcode the fields selection to a few more attributes) - the focus is on wiring the real trace introspection while it's actually in use already.
  * PR3 - add the settings implementation with persistence.
  * PR4 - make config option on by default (allows users to disable if there are performance concerns in production).

udita-0707:
> I think [I suggested a slightly suboptimal breakdown](https://github.com/jaegertracing/jaeger-ui/pull/4063#pullrequestreview-4580892951) of the original PR. The most useful way to structure PRs is by delivering incremental value with small changes. E.g.
> 
> * PR1 - implement minimal stubs of summaryFieldUtils, e.g. by checking for HTTP status attribute only, and the display wiring that actually shows it in the timeline. And a config feature that enables the whole thing. (which is two lines of code). This immediately gives you a working version of the final capability, just limited to a single field which is not configurable.
> * PR2 - extend with roughly what you have here (again maybe hardcode the fields selection to a few more attributes) - the focus is on wiring the real trace introspection while it's actually in use already.
> * PR3 - add the settings implementation with persistence.
> * PR4 - make config option on by default (allows users to disable if there are performance concerns in production).

@yurishkuro Following your latest feedback, here's a concrete plan for #4017, please confirm before I reshape #4149.

**PR1** - Minimal slice: hardcoded HTTP status chip + span wiring + `summaryFieldsEnabled` (default false). No settings, no persistence, no full trace scan.

**PR2** - Generalized `buildSummaryLookup` + small hardcoded field set; still no user configuration.

**PR3** - `buildAvailableFields` + trace-aware filtering (introspection while feature is already in use).

**PR4** - Settings picker + localStorage persistence (incl. your naming/comments: load vs parse, document sanitize, drop memoizeOne, revisit max-fields).

**PR5** - Flip config default to true (opt-out).

**PR6-7** (later) - URL persistence and "Add to summary fields" in span detail.

I'll close/reshape #4149 accordingly. Does this match what you had in mind, or would you merge PR2+PR3?

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/store.layout.ts
udita-0707: Agreed - the public intent here is loading persisted field keys from localStorage, not parsing. I'll rename this to `loadStoredSummaryFields` and keep JSON parsing as an internal detail of that function.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/store.layout.ts
udita-0707: I'll add a comment explaining what this does: trim whitespace, drop non-strings/empty keys, dedupe while preserving order, and cap the list length. If you prefer a more explicit name like `sanitizeSummaryFieldKeys`, I'm happy to rename it as well.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/summaryFieldsUtils.ts
udita-0707: This helper is only reached for array/object branches after `formatAttributeValue` has already narrowed the type. I'll either tighten the parameter type to match those cases or inline `JSON.stringify` at the call sites and remove the helper.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/summaryFieldsUtils.ts
udita-0707: Makes sense. I'll remove `memoize-one` from the utils and export plain functions instead.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/summaryFieldsUtils.ts
udita-0707: The limit was meant as a UX guard (avoid cluttering span rows), not a performance optimization.

I'll remove the hard cap from the utils/store layer. When the settings UI lands, I'll add a soft warning in the panel (e.g. "selecting many fields may affect readability/performance") rather than silently truncating at an arbitrary number.

yurishkuro:
@udita-0707 looks fine, as long as each PR lands some small user facing improvement.

udita-0707:
@yurishkuro This PR was reshaped from the original “data layer + localStorage, no UI” approach into PR1 of the 7-PR incremental plan: HTTP status chips + feature gate only. Generic lookup, settings UI, and persistence will land in follow-up PRs stacked on this one.

yurishkuro:
The pill machinery is split between specific and generic at the wrong altitude. `VirtualizedTraceView` imports `HTTP_STATUS_SUMMARY_FIELDS` and `SpanBarRow` imports `isHttpStatusCode5xx`, so two component files now carry HTTP-specific knowledge. Since more pill types are coming, I'd like the component layer to stay generic and have **all knowledge of which attributes become pills, and which pills are errors, live in one module.** The rendering side is already generic — it maps over a collection and draws a `<Tag>` per entry — so this is mostly removing names and imports, not adding code.

### Let each pill carry its own presentation

Compute the error flag where the pill is built, so components don't have to know what 5xx means:

```ts
// spanPills.ts  (renamed from summaryFieldsUtils.ts)

// PR1: HTTP status is the only pill source. Private — callers never see these keys.
const HTTP_STATUS_ATTR_KEYS = ['http.status_code', 'http.response.status_code'] as const;

export type ISpanPill = { label: string; value: string; isError?: boolean };

function httpStatusPill(span: IOtelSpan): ISpanPill | undefined {
  for (const key of HTTP_STATUS_ATTR_KEYS) {
    const attr = span.attributes.find(a => a.key === key);
    const value = attr?.value == null ? '' : String(attr.value);
    if (value) {
      const code = Number(value.trim());
      return { label: 'http.status_code', value, isError: code >= 500 && code < 600 };
    }
  }
  return undefined;
}

// Pure and unit-testable. Owns *which* attributes become pills; callers do not.
export function buildSpanPills(trace: IOtelTrace): Map<string, ISpanPill[]> {
  const result = new Map<string, ISpanPill[]>();
  for (const span of trace.spans) {
    const pill = httpStatusPill(span);
    if (pill) result.set(span.spanID, [pill]);
  }
  return result;
}
```

### Move the config gating into a hook

The `summaryFieldsEnabled` check and memoization currently sit inline in `TraceTimelineViewerImpl`. That belongs in a hook, which also gives the module a better name than "utils." Keep the pure builder for testing and wrap it:

```ts
export function useSpanPills(trace: IOtelTrace): Map<string, ISpanPill[]> {
  const enabled = useConfig().traceTimeline?.spanPillsEnabled === true;
  return useMemo(() => (enabled ? buildSpanPills(trace) : new Map()), [enabled, trace]);
}
```

### Resulting component code

`TraceTimelineViewerImpl` — no `useConfig`, no `useMemo`, no enabled flag:

```tsx
const spanPills = useSpanPills(trace);
// ...
<VirtualizedTraceView ... spanPills={spanPills} />
```

`VirtualizedTraceView` — no `HTTP_STATUS_SUMMARY_FIELDS`, no `selectedFields`:

```tsx
const pills = spanPills.get(spanID);
// ...
<SpanBarRow ... pills={pills} />
```

`SpanBarRow` — imports nothing pill-specific; the pill tells it how to render:

```tsx
{pills?.map(pill => (
  <Tag
    key={pill.label}
    aria-label={`${pill.label}: ${pill.value}`}
    className={cx('SpanBarRow--pill', { 'is-error': pill.isError })}
  >
    {pill.value}
  </Tag>
))}
```

### Renames

| Current | Proposed | Why |
| --- | --- | --- |
| `summaryFieldsUtils.ts` | `spanPills.ts` | name the domain, not "utils" |
| `buildHttpStatusSummaryLookup` | `buildSpanPills` + `useSpanPills` | generic seam; HTTP stays internal |
| `HTTP_STATUS_SUMMARY_FIELDS` / `HTTP_STATUS_SUMMARY_FIELD` (exported) | `HTTP_STATUS_ATTR_KEYS` (private) | components shouldn't import specific-pill knowledge |
| `isHttpStatusCode5xx` (exported, used by `SpanBarRow`) | inlined into the builder as `pill.isError` | severity decided where the pill is built |
| `summaryValues` prop | `pills` | it's a label→value/severity list, not "values" |
| `selectedFields` prop | *removed* | redundant — the pill list already encodes which pills exist |
| `Map<string, Record<string, string>>` | `Map<string, ISpanPill[]>` | pill carries its own label + severity |
| `summaryLookup` | `spanPills` | consistent vocabulary |
| `.SpanBarRow--summaryChip` | `.SpanBarRow--pill` | consistent |
| `traceTimeline.summaryFieldsEnabled` (config + `Config` type + `default-config.ts`) | `traceTimeline.spanPillsEnabled` | consistent vocabulary across config, code, and CSS |

Two things fall out of this, both good: the `Object.create(null)` guard and its `__proto__` test can go, since pill labels are now module-controlled constants rather than arbitrary attribute keys, so there's no pollution vector left; and the `key !== …` guard inside the old `isHttpStatusCode5xx` disappears with the inlining.

Please rename the config key `traceTimeline.summaryFieldsEnabled` → `traceTimeline.spanPillsEnabled` as part of this too — update the `Config` type in `types/config.ts`, the default in `default-config.ts`, and the read site in the `useSpanPills` hook. After this, no `summary*` naming should remain anywhere in the feature: module, functions, hook, props, types, CSS class, and config key all consistently on `spanPills`.

udita-0707:
@yurishkuro Makes sense. I'll refactor so components stay generic: `spanPills.ts` owns pill building + error severity, `useSpanPills` owns config gating, and rename `summaryFieldsEnabled` → `spanPillsEnabled` throughout. Will push an update shortly.

udita-0707:
@yurishkuro refactoring has been done. PTAL!

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/useSpanPills.ts
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) 2026 The Jaeger Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+import { useMemo } from 'react';
+
+import { useConfig } from '../../../hooks/useConfig';
+import { IOtelTrace } from '../../../types/otel';
+import { buildSpanPills, ISpanPill } from './spanPills';
+
+export function useSpanPills(trace: IOtelTrace): Map<string, ISpanPill[]> {
```
yurishkuro: I don't think it's necessary to separate into a useSpanPills file, there's more boilerplate here than business logic.

Just add it to spanPills.ts

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
```diff
@@ -370,6 +373,7 @@ export const VirtualizedTraceViewImpl = React.memo(function VirtualizedTraceView
         prunedServices.size > 0 &&
         span.childSpans.some(child => prunedServices.has(child.resource.serviceName));
       const criticalPathSections = criticalPathContext.sectionsFor(span, isCollapsed, hasPrunedChildren);
+      const pills = spanPills.get(spanID);
```
yurishkuro: Why does virtualized view need to know about span pills? Aren't they purely a SpanBarRow concept? Because of the current design you are forced to compute pills for all spans in a trace upfront, instead of as needed when the span row comes into view.

packages/jaeger-ui/src/constants/default-config.ts
```diff
@@ -121,6 +121,7 @@ const defaultConfig: Config = {
   traceTimeline: {
     enableSidePanel: true,
     defaultDetailPanelMode: 'inline',
+    spanPillsEnabled: false,
```
yurishkuro: You can make then on by default, would be nice feature. The flag is just to disable it in case of issues.

udita-0707: Done

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/useSpanPills.ts
udita-0707: Added to spanPills.ts

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
udita-0707: Removed pill map from `VirtualizedTraceView`; `SpanBarRow` computes pills per span via `getSpanPillsForSpan`

packages/jaeger-ui/src/types/config.ts
```diff
@@ -238,6 +238,10 @@ export type Config = {
     // 'sidepanel' makes the side panel the default experience for new users.
     // Default: 'inline'.
     defaultDetailPanelMode?: SpanDetailPanelMode;
+
+    // spanPillsEnabled controls span pill overlays in the trace timeline.
+    // Default: true. Set to false to disable if needed.
+    spanPillsEnabled?: boolean;
```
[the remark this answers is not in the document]
udita-0707: updated the PR descrption

packages/jaeger-ui/src/constants/default-config.ts
```diff
@@ -121,6 +121,7 @@ const defaultConfig: Config = {
   traceTimeline: {
     enableSidePanel: true,
     defaultDetailPanelMode: 'inline',
+    spanPillsEnabled: true,
   },
```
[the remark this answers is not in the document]
udita-0707: updated the PR description

packages/jaeger-ui/src/types/config.ts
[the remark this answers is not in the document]
udita-0707: updated the PR description

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/spanPills.ts
```diff
@@ -0,0 +1,37 @@
...
+/** Builds pills for a single span. Owns which attributes become pills; callers do not. */
+export function getSpanPillsForSpan(span: IOtelSpan): ISpanPill[] {
+  const pill = httpStatusPill(span);
+  return pill ? [pill] : [];
+}
+
+/** Enabled unless explicitly disabled via config (default on). */
+export function useSpanPillsEnabled(): boolean {
+  return useConfig().traceTimeline?.spanPillsEnabled !== false;
+}
```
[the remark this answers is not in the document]
udita-0707: updated the PR description

packages/jaeger-ui/src/constants/default-config.ts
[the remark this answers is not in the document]
udita-0707: updated the PR description

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/spanPills.test.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  describe('useSpanPillsEnabled', () => {
+    beforeEach(() => {
+      mockUseConfig.mockReturnValue({ traceTimeline: {} });
+    });
+
+    it('defaults to enabled when config omits the flag', () => {
+      const { result } = renderHook(() => useSpanPillsEnabled());
+      expect(result.current).toBe(true);
+    });
```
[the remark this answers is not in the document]
udita-0707: The PR description has been updated

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/spanPills.ts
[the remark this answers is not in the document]
udita-0707: The PR description has been updated.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/spanPills.test.ts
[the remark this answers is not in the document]
udita-0707: The PR description has been updated.

udita-0707:
@yurishkuro PTAL

yurishkuro (approved):

Issue #4148, opened 2026-06-27
[Bug]: "View traces" on System Architecture DAG fails with 400 Bad Request


kams19-ops:
### What happened?

On the System Architecture page's DAG view, right-clicking (or clicking) a service 
node and selecting "View traces" opens a new tab to the Search page, but the search 
fails with a 400 error.

This happens because `handleViewTraces` in `DAG.tsx` builds the search URL with only 
the `service` param.

No `lookback`, `start`, or `end` is included. `SearchTracePage` has existing logic 
(in `searchQueryFromUrl`) to derive `start`/`end` from a `lookback` param when only 
`lookback` is present - but since this URL has no `lookback` either, that derivation 
never runs, `start`/`end` stay empty, and the backend API call fails with a 400.

### Steps to reproduce

1. Run a local Jaeger backend with some multi-service trace data (e.g. via the HotROD demo).
2. Go to the **System Architecture** tab.
3. Click on any non-focal service node (e.g. `customer`) to open its context menu.
4. Click **"View traces"**.
5. Observe: a new tab opens to `/search?service=customer` and shows:
   `Failed to fetch trace summaries: 400 Bad Request`

### Expected behavior

Clicking "View traces" on a node should open the Search page with a valid time range 
(e.g. a default lookback) included, so the search succeeds instead of erroring out.

### Relevant log output

```shell

```

### Screenshot

<img width="1662" height="824" alt="Image" src="https://github.com/user-attachments/assets/dc0ee28d-1949-4675-9720-9466f1dd4e1d" />

<img width="1663" height="803" alt="Image" src="https://github.com/user-attachments/assets/bf2ba7da-f5f1-4296-98fc-02b4f8f4b60d" />

### Deployment configs

```shell

```

kams19-ops:
Hi @jpkrohling @kinghuang @yurishkuro i found a bug where clicking "View traces" on a node in the System Architecture DAG view fails with a 400 error, due to a missing lookback param in the generated URL. I've already got a fix ready and tested locally. Would appreciate it if someone could take a quick look and confirm whether this is worth a PR, or if I'm missing some context on why it's built this way!

parshipcy:
Yes, this looks like a valid bug.

kveni12:
[...]

PR #4151, opened 2026-06-27
fix(DependencyGraph): include lookback in View traces URL to prevent 400 error.

kams19-ops:

- #4148 

`handleViewTraces` in the System Architecture DAG view (`DAG.tsx`) builds the 
"View traces" URL with only the `service` param:

```js
window.open(getSearchUrl({ service: hoveredNode?.key }), '_blank');
```

Since no `lookback`, `start`, or `end` is included, `SearchTracePage`'s existing 
logic to derive `start`/`end` from `lookback` (added for legacy/HotROD-style links) 
never runs, and the search request fails with a 400 Bad Request.

## Description of the changes
-  Adds a default `lookback: '1h'` to the generated URL, so the existing 
  start/end-derivation logic in `searchQueryFromUrl` kicks in correctly.
- Adds a guard so `handleViewTraces` does nothing if `hoveredNode` is null 
  (previously it would still call `window.open` with an invalid URL).
- Updates the two existing `handleViewTraces` tests in `DAG.test.jsx` to match 
  the new expected behavior.

- -Ran the app locally with the Jaeger backend + HotROD demo data.
- Reproduced the 400 error before the fix (screenshots in linked issue).
- Applied the fix, confirmed the generated URL now includes `lookback=1h`, and 
  the Search page successfully loads results instead of erroring.
- Ran the full `DAG.test.jsx` suite - all 30 tests pass, including the two 
  updated `handleViewTraces` tests.

## Checklist
- [-] I have read https://github.com/jaegertracing/jaeger/blob/main/CONTRIBUTING_GUIDELINES.md
- [-] I have signed all commits

- [-] **Moderate**: AI helped with code generation or debugging specific parts

https://github.com/user-attachments/assets/4e4648d5-cb62-4198-9b20-30e45e5e67c3

### DIFF
--- a/packages/jaeger-ui/src/components/DependencyGraph/DAG.tsx
+++ b/packages/jaeger-ui/src/components/DependencyGraph/DAG.tsx
@@ -12,6 +12,7 @@ import { getUrl as getSearchUrl } from '../SearchTracePage/url';
 
 import './dag.css';
 import { DAG_MAX_NUM_SERVICES } from '../../constants';
+import { DEFAULT_LOOKBACK } from '../../constants/search-form';
 
 type TProps = {
   data: {
@@ -59,7 +60,7 @@ export const renderNode = (
 
 export const handleViewTraces = (hoveredNode: TVertex | null) => {
   if (!hoveredNode?.key) return;
-  window.open(getSearchUrl({ service: hoveredNode.key, lookback: '1h' }), '_blank');
+  window.open(getSearchUrl({ service: hoveredNode.key, lookback: DEFAULT_LOOKBACK }), '_blank');
 };
 
 export const createHandleNodeClick =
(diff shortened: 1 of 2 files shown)


parshipcy (commented):
Please sign your commits

kams19-ops:
> Please sign your commits

Done.

packages/jaeger-ui/src/components/DependencyGraph/DAG.tsx
```diff
@@ -58,7 +58,8 @@ export const renderNode = (
 };
 
 export const handleViewTraces = (hoveredNode: TVertex | null) => {
-  window.open(getSearchUrl({ service: hoveredNode?.key }), '_blank');
+  if (!hoveredNode?.key) return;
+  window.open(getSearchUrl({ service: hoveredNode.key, lookback: '1h' }), '_blank');
```
parshipcy: It would be better to use `DEFAULT_LOOKBACK` from `constants/search-form` instead of hardcoding 1h, so it stays in sync with search defaults.

kams19-ops: Yes sir, i have done the changes.

parshipcy (approved):
Thanks! LGTM👍

PR #4153, opened 2026-06-27
refactor(TracePage): decouple TraceGraph search logic from TracePage

### LINKED ISSUE #4134: refactor(TracePage): Decouple view-specific search logic from TracePage

akshatsinghai6682-sketch:
- Resolves #4134

## Description of the changes
- Moved `calculateTraceDagEV` call from `TracePage` into `TraceGraph` component
- `TraceGraph` now computes its own DAG and search matches internally using `useMemo`
- Added `onSearchResults` callback prop to `TraceGraph` so it reports match results back to `TracePage` via `useEffect`
- Removed `traceDagEV`, `graphFindMatches` computation from `TracePage`
- `TracePage` now receives search results from child views instead of computing them itself
- `findCount` in header is now derived per view type: from `onSearchResults` for Graph view, from `spanFindMatches` for Timeline and other views

- Manually tested Timeline view search — match count and highlighting work correctly
- Manually tested Trace Graph view search — match count updates correctly via `onSearchResults` callback
- TypeScript lint passes with 0 errors (`npm run tsc-lint`)

## Checklist


akshatsinghai6682-sketch:
Hi @yurishkuro, I've addressed the Copilot review comments:

1. Added `useEffect` to clear stale `findMatches` when switching views or `uiFind` changes
2. Updated `TraceGraph.test.jsx` to use new `trace` and `onSearchResults` props
3. Updated `index.test.jsx` to reflect new callback-based search flow

Ready for review!

dolliecoder:
hey I took a look locally and there are two blocking issues:
compilation error: so a recent commit accidentally deleted export const OLDEST_FIRST = 'OLDEST_FIRST'; from src/model/order-by.ts. This breaks the typescript build (tsc-lint fails with error TS2305: module '"./order-by"' has no exported member 'OLDEST_FIRST').
broken test: The AI autofix in TraceId.test.jsx replaced rerender() with a second renderComponent() call, which incorrectly dumps a second component into the DOM. The test now falsely passes by verifying the stale, short component instead of the new, full-length one

akshatsinghai6682-sketch:
Hi @dolliecoder, thank you for the review!

1.  Restored `OLDEST_FIRST` export in `order-by.ts` — fixed in the latest commit.

2. Regarding `TraceId.test.jsx` — this file was not modified in this PR. Running `git diff` confirms this PR only touches:
   - `TraceGraph.tsx`
   - `TraceGraph.test.jsx`
   - `TracePage/index.tsx`
   - `TracePage/index.test.jsx`

Please let me know if there's anything else to address!

jkowall:
Thanks for iterating on this. I am going to leave the PR open for now, but I expect we will close this one because the churn has made it hard to review confidently.

The main issues are:

1. This does not fully resolve #4134. The issue is about moving view-specific search ownership out of `TracePage`, but this PR only moves the `TraceGraph` path. `TracePage` still computes span and timeline matches with `filterSpansMemo` and `filterPrunedSpanIDs`, so the parent remains coupled to child-view search internals.

2. The current TracePage test does not prove the new callback flow works. It now asserts that `resultCount` defaults to `0` when switching to graph view, but it does not simulate `TraceGraph` calling `onSearchResults` or verify that the header count updates from reported graph matches.

3. The PR history has become too noisy for the final small diff: unrelated start-time sort work, accidental removals/restores, repeated AI autofix follow-ups, and multiple reversions. Even though current CI is green, the review signal is now dominated by churn rather than the intended refactor.

If you want to continue with this work, the best path is a fresh PR from current `main` with a clean, focused implementation of #4134, including the non-graph search paths and tests that exercise the callback contract end to end.

akshatsinghai6682-sketch:
Ok @jkowall Thank You

PR #4158, opened 2026-06-28
refactor(tracepage): generalize scroll-page for element containers

dolliecoder:
- Addresses the maintainer `TODO` in `scroll-page.ts` to decouple the smooth-scroll utility from `window`, allowing it to target arbitrary DOM Elements (such as side panels).

## Description of the changes
- Added DOM abstraction helpers (`getScrollTop`, `setScrollTop`) to safely bridge the API gap between `window.scrollY` and `element.scrollTop`, ensuring horizontal scroll is preserved in both cases.
- Added an optional `container` parameter to `scrollBy` and `scrollTo`, defaulting to `window` to ensure **zero existing callers** are broken.
- Kept the `lastTween` singleton intact to avoid scope creep, as Jaeger UI currently has no execution paths where multiple elements scroll simultaneously.

- Manually verified that all 10 existing `window` tests pass mathematically.
- Added 2 new `describe` blocks in `scroll-page.test.js` using a mock DOM element to verify that `scrollTop` is accurately read and mutated.

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/scroll-page.test.js
+++ b/packages/jaeger-ui/src/components/TracePage/scroll-page.test.js
@@ -154,7 +154,7 @@ describe('scroll-by', () => {
         scrollTop: 50,
         scrollLeft: 0,
       };
-      Tween.mockReset();
+      Tween.mockClear();
       cancel();
     });
 
@@ -199,6 +199,28 @@ describe('scroll-by', () => {
 
       expect(tweenOptions.from).toBe(50);
       expect(tweenOptions.to).toBe(60);
+    });
+
+    it('does not clear the active tween when an older cross-container tween completes', () => {
+      window.scrollY = 100;
+
+      // Start tween A (window).
+      scrollBy(100);
+
+      // Start tween B (element) while A is still active.
+      scrollBy(10, false, container);
+
+      expect(tweenInstances.length).toBe(2);
+      const tweenA = tweenInstances[0];
+      const tweenB = tweenInstances[1];
+
+      // A finishes first — invoke its onUpdate with done: true.
+      tweenA.onUpdate({ value: 200, done: true });
+
+      // cancel() should still cancel B since B is the active tween.
+      cancel();
+      expect(tweenB.cancel.mock.calls.length).toBe(1);
+      expect(tweenA.cancel.mock.calls.length).toBe(0);
     });
   });
 });
--- a/packages/jaeger-ui/src/components/TracePage/scroll-page.test.js
+++ b/packages/jaeger-ui/src/components/TracePage/scroll-page.test.js
@@ -47,7 +47,9 @@ describe('scroll-by', () => {
         scrollBy(yDelta, true);
         expect(Tween.mock.calls.length).toBe(1);
         scrollBy(yDelta, false);
-        expect(Tween.mock.calls[0]).toEqual(Tween.mock.calls[1]);
+        expect(Tween.mock.calls[0][0].to).toBe(Tween.mock.calls[1][0].to);
+        expect(Tween.mock.calls[0][0].duration).toBe(Tween.mock.calls[1][0].duration);
+        expect(Tween.mock.calls[0][0].from).toBe(Tween.mock.calls[1][0].from);
       });
 
       it('is additive when an in-progress scroll is the same direction', () => {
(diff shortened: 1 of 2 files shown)


packages/jaeger-ui/src/components/TracePage/scroll-page.ts
```diff
@@ -10,15 +10,30 @@ let lastTween: Tween | void;
...
+function _onTweenUpdate(container: Element | Window, { done, value }: { done: boolean; value: number }) {
+  setScrollTop(container, value);
   if (done) {
     lastTween = undefined;
   }
 }
 
-export function scrollBy(yDelta: number, appendToLast = false) {
-  const { scrollY } = window;
+export function scrollBy(yDelta: number, appendToLast = false, container: Element | Window = window) {
```
jkowall, line 45: This adds the `container` parameter, but the append logic below still treats `lastTween` as globally valid. That means an in-progress window scroll can be appended to a later element scroll, or vice versa, because `targetFrom = lastTween.to` is reused without knowing which scroll surface produced it.

A minimal fix would be to track the active container alongside the tween, for example `let lastTweenContainer: Element | Window | void;`, set it whenever a new `Tween` is created, and only enter the append path when `lastTweenContainer === container`. Clear both `lastTween` and `lastTweenContainer` in `cancel()` and when `_onTweenUpdate(... done: true)` fires. If the container differs, this should behave like a fresh scroll from `getScrollTop(container)`.

packages/jaeger-ui/src/components/TracePage/scroll-page.test.js
```diff
@@ -144,4 +146,42 @@ describe('scroll-by', () => {
       expect(twCancel.mock.calls.length).toBe(0);
     });
   });
+
+  describe('element scrolling', () => {
```
jkowall, line 150: Please add a regression test that covers append behavior across different containers. The current element tests prove basic `scrollTop` reads/writes, but they do not exercise the singleton `lastTween` state that `appendToLast` uses.

One useful case is: set `window.scrollY = 100`, call `scrollBy(100)` to create a window tween, then use a mock element with `scrollTop = 50` and call `scrollBy(10, true, container)`. The second tween should have `from: 50` and `to: 60`; today it reuses the prior window target and jumps to `210`.

jkowall (changes requested):
Thanks for generalizing this utility. I found one correctness issue that should be fixed before merging: the existing singleton `lastTween` now needs to be scoped to the scroll container, otherwise append behavior can mix window and element scroll targets.

I verified the existing focused test passes, but a temporary cross-container probe reproduced the issue: after `scrollBy(100)` on `window`, calling `scrollBy(10, true, container)` where `container.scrollTop` is `50` produced a target of `210` instead of the expected `60`.

dolliecoder:
@jkowall I updated the code so that appending only happens within the same scroll container by keeping track of which container the tween belongs to, and I added the regression test you suggested to cover this case and I also confirmed that the updated scroll-page tests all pass

packages/jaeger-ui/src/components/TracePage/scroll-page.test.js
jkowall: Thanks for adding this. The regression test still needs one adjustment: `Tween.mockReset()` in the nested `beforeEach` wipes the constructor implementation installed by the outer `beforeEach`, so the first `scrollBy(100)` records constructor args but returns `undefined` and never leaves a truthy `lastTween`. The second `scrollBy(10, true, container)` therefore skips the `appendToLast && lastTween` path entirely, which means this test would pass even if the `lastTweenContainer === container` guard were removed.

Please use `mockClear()` here, or re-install the mock implementation after the reset, so the test actually exercises the singleton `lastTween` path that the earlier bug lived in.

packages/jaeger-ui/src/components/TracePage/scroll-page.ts
dolliecoder: Implemented this approach. The tween now records its associated scroll container, and the append path only executes when lastTweenContainer === container and if the container differs scrolling starts as a fresh tween

packages/jaeger-ui/src/components/TracePage/scroll-page.test.js
dolliecoder: Added the requested regression test covering append behavior across different scroll containers. The test creates a window tween first and verifies that an element scroll starts from the element's current scroll position instead of reusing the previous tween state and updated the regression test so it exercises the intended cross-container append scenario and verifies the constructor arguments for the second tween and the updated scroll-page tests pass with this change

packages/jaeger-ui/src/components/TracePage/scroll-page.ts
```diff
@@ -6,39 +6,69 @@ import Tween from './Tween';
...
+  } else {
+    (container as Element).scrollTop = value;
+  }
+}
+
+function _onTweenUpdate(container: Element | Window, { done, value }: { done: boolean; value: number }) {
+  setScrollTop(container, value);
   if (done) {
     lastTween = undefined;
+    lastTweenContainer = undefined;
```
jkowall, line 41: This completion callback can belong to an older tween while a newer tween is active. For example, start window tween A, then start element tween B; both remain scheduled. If A reports `done` first, this unconditionally clears `lastTween` and `lastTweenContainer`, even though those globals now refer to B. B keeps animating, but `cancel()` becomes a no-op and a later `scrollBy(..., true, containerB)` cannot append to B's target.

I reproduced this with a temporary unit test: after starting A and B, invoking A's `onUpdate({ done: true, ... })` and then calling `cancel()` expected B's `cancel` mock to be called once, but it was called zero times.

Please associate each callback with its owning tween (or a generation token) and only clear the singleton state when the completing tween is still the active one. A regression test should cover an older cross-container tween completing after a newer one starts.

dolliecoder: done fixed with a generation token: onTweenUpdate only clears globals when gen === tweenGeneration and the regression test covers that cross-container case.

PR #4163, opened 2026-06-28
fix: Increase critical path bar contrast in dark mode

### LINKED ISSUE #4162: [Bug]: Critical path segments are invisible in dark mode
## Steps to reproduce

1. Enable dark mode.
2. Open a trace and turn on critical path (the button in the timeline header).
3. The critical path bar overlaid on spans is invisible.

## What's happening

`vars.css` defines the critical path tokens in both the `:root` block and the `[data-theme='dark']` block with the exact same values:

```css
--critical-path-color: #000;
--critical-path-outline: #fff;
```

So in dark mode the bar is still black, but the surface background is `#162338` (dark navy), making it invisible.

The token is used in `SpanBar.tsx:58`:
```tsx
style={{ background: 'var(--critical-path-color)' }}
```
and in `SpanBar.css:43-44` for the outline box-shadow.

## Expected

The critical path bar should be visible in dark mode. Swapping the token values in the dark block would fix it:

```css
[data-theme='dark'] {
  --critical-path-color: #fff;
  --critical-path-outline: #000;
}
```

## Files

- `src/components/common/vars.css:155-156, 263-264`
- `src/components/TracePage/TraceTimelineViewer/SpanBar.tsx:58`
- `src/components/TracePage/TraceTimelineViewer/SpanBar.css:43-44`

bhavyamsharmaa:
Closes #4162

The critical path bar was invisible in dark mode because --critical-path-color and --critical-path-outline had the same values in both the :root block and the [data-theme='dark'] block in vars.css, so dark mode still rendered a black bar on a dark navy surface.

Root cause: lines 263-264 of vars.css duplicated the light-mode values instead of inverting them.

Fix: swap the two token values in the [data-theme='dark'] block so the bar is white with a black outline in dark mode.

The three screenshots below show the same trace. The critical path bars are the horizontal markers running through the span bars (thick black in light mode, invisible/white in dark mode).

light mode — critical path bars clearly visible as black stripes through the spans (this is the intended behaviour):

light mode

dark mode before fix — same bars are invisible (black on dark navy background):

dark mode before

dark mode after fix — bars visible again as white stripes:
dark mode after

### DIFF
--- a/packages/jaeger-ui/src/components/common/vars.css
+++ b/packages/jaeger-ui/src/components/common/vars.css
@@ -260,8 +260,8 @@ SPDX-License-Identifier: Apache-2.0
   --trace-emphasis-highlight: #3d3520;
 
   /* Critical Path */
-  --critical-path-color: #000;
-  --critical-path-outline: #fff;
+  --critical-path-color: #fff;
+  --critical-path-outline: #000;
 
   /* ============================================
     TRACE TIMELINE TOKENS - Dark Mode


bhavyamsharmaa:
@yurishkuro Could you please add the changelog label? The check-label CI check is failing due to the missing label — I believe changelog:bugfix-or-minor-feature would be the right one for this fix.

yurishkuro:
I can clearly see critical path in dark mode, how is it "invisible"?

bhavyamsharmaa:
@yurishkuro fair point — "invisible" was an overstatement, sorry for the confusion.

The actual problem is contrast. Before this fix, `--critical-path-color` in the dark-mode CSS block was `#000` — the same value as light mode, where black on a white/light background is fine. In dark mode the timeline sits on a dark navy background, so a black-filled bar blends right in. The only thing technically remaining was the 0.5 px white `box-shadow` outline — barely a hairline at that thickness.

The fix swaps the two tokens in the `[data-theme='dark']` block so the fill becomes `#fff` and the outline becomes `#000`, which is the correct inversion for a dark surface.

**Before** (`--critical-path-color: #000`): fill is invisible against the dark bg; only the sub-pixel outline survives.
**After** (`--critical-path-color: #fff`): white bar is clearly visible across each span.

yurishkuro (approved):

PR #4164, opened 2026-06-28
fix(assistant): Render tool-call parts in chat thread

SoumyaRaikwar:
Part of #3932 
- Tool-call message parts from the AG-UI runtime are silently dropped in the Ask Jaeger chat panel. When the AI agent calls an MCP tool (e.g. `get_services`, `read_skill`), the user sees nothing while the tool runs  and no confirmation it completed, only the final text reply appears.

## Description of the changes
- Add `JaegerToolCallIndicator` component that renders tool-call parts  inline in the chat thread:
  - **While running** (`status.type === 'running'` or `requires-action`): shows *"Calling tool_name…"* in italic
  - **When complete**: shows **"Called tool_name"** as a collapsible `<details>` block with Input/Output sections (collapsed by default)
  - **On failure** (`status.type === 'incomplete'` or `isError`): shows **"Failed tool_name"** with an error-colored border
- Reuse `react-json-view-lite` (`JsonView`) from `AttributesTable` for JSON results; plain text results fall back to a `<pre>` block
- Extend the `MessagePrimitive.Parts` render function in  `JaegerThreadMessageBody` to handle `part.type === 'tool-call'` alongside the existing `'text'` branch
- Use `part.status` for lifecycle state instead of inferring from `result === undefined`
- Conditionally render the Output section only when `result` is defined

## Screenshots
Screenshot from 2026-06-29 02-39-31

## Screencast
[Screencast from 2026-06-29 02-38-12.webm](https://github.com/user-attachments/assets/93b769e5-586f-4a56-975f-5bc1bb20a75a)

- Added 4 new test cases to `JaegerAssistantPanel.test.jsx`:
  - Running indicator renders "Calling tool_name…" when result is undefined
  - Completed indicator renders "Called tool_name" with result text
  - Long results are truncated at 300 chars with ellipsis
  - Text parts still render correctly alongside tool-call parts (regression)
- Verified end-to-end with Jaeger backend + Gemini AI sidecar: tool calls for `get_services` and `read_skill` render correctly in the chat panel
- All 3217 tests pass, lint and build clean

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/App/JaegerAssistantPanel.tsx
+++ b/packages/jaeger-ui/src/components/App/JaegerAssistantPanel.tsx
@@ -31,8 +31,6 @@ function JaegerAssistantBootstrap() {
   return null;
 }
 
-const TOOL_RESULT_MAX_LEN = 300;
-
 function formatToolResult(result: unknown): string {
   if (typeof result === 'string') return result;
   try {
@@ -43,25 +41,39 @@ function formatToolResult(result: unknown): string {
   }
 }
 
-function truncateToolResult(result: unknown): string {
-  const text = formatToolResult(result);
-  if (text.length <= TOOL_RESULT_MAX_LEN) return text;
-  return `${text.slice(0, TOOL_RESULT_MAX_LEN)}…`;
-}
-
-function JaegerToolCallIndicator({ toolName, result }: { toolName: string; result?: unknown }) {
+function JaegerToolCallIndicator({
+  toolName,
+  argsText,
+  result,
+}: {
+  toolName: string;
+  argsText?: string;
+  result?: unknown;
+}) {
   const isRunning = result === undefined;
-  return (
-    <div className="JaegerAssistantPanel-toolCall">
-      {isRunning ? (
+  if (isRunning) {
+    return (
+      <div className="JaegerAssistantPanel-toolCall">
         <span className="JaegerAssistantPanel-toolCallStatus">Calling {toolName}…</span>
-      ) : (
-        <>
-          <span className="JaegerAssistantPanel-toolCallName">Called {toolName}</span>
-          <span className="JaegerAssistantPanel-toolCallResult">{truncateToolResult(result)}</span>
-        </>
-      )}
-    </div>
+      </div>
+    );
+  }
+  return (
+    <details className="JaegerAssistantPanel-toolCall">
+      <summary className="JaegerAssistantPanel-toolCallName">Called {toolName}</summary>
+      <div className="JaegerAssistantPanel-toolCallBody">
+        {argsText && (
+          <div className="JaegerAssistantPanel-toolCallSection">
+            <span className="JaegerAssistantPanel-toolCallLabel">Input</span>
+            <pre className="JaegerAssistantPanel-toolCallPre">{argsText}</pre>
+          </div>
+        )}
+        <div className="JaegerAssistantPanel-toolCallSection">
+          <span className="JaegerAssistantPanel-toolCallLabel">Output</span>
+          <pre className="JaegerAssistantPanel-toolCallPre">{formatToolResult(result)}</pre>
+        </div>
+      </div>
+    </details>
   );
 }
 
@@ -82,7 +94,13 @@ function JaegerThreadMessageBody({ variant }: { variant: 'user' | 'assistant' })
             );
           }
           if (part.type === 'tool-call') {
-            return <JaegerToolCallIndicator toolName={part.toolName} result={part.result} />;
+            return (
+              <JaegerToolCallIndicator
+                toolName={part.toolName}
+                argsText={part.argsText}
+                result={part.result}
+              />
+            );
           }
           return null;
         }}
--- a/packages/jaeger-ui/src/components/App/JaegerAssistantPanel.tsx
+++ b/packages/jaeger-ui/src/components/App/JaegerAssistantPanel.tsx
@@ -45,12 +45,14 @@ function JaegerToolCallIndicator({
   toolName,
   argsText,
   result,
+  status,
 }: {
   toolName: string;
   argsText?: string;
   result?: unknown;
+  status?: { type: string };
 }) {
-  const isRunning = result === undefined;
+  const isRunning = status?.type === 'running' || status?.type === 'requires-action';
   if (isRunning) {
     return (
       <div className="JaegerAssistantPanel-toolCall">
@@ -99,6 +101,7 @@ function JaegerThreadMessageBody({ variant }: { variant: 'user' | 'assistant' })
                 toolName={part.toolName}
                 argsText={part.argsText}
                 result={part.result}
+                status={part.status}
               />
             );
           }
--- a/packages/jaeger-ui/src/components/App/JaegerAssistantPanel.tsx
+++ b/packages/jaeger-ui/src/components/App/JaegerAssistantPanel.tsx
@@ -36,7 +36,8 @@ const TOOL_RESULT_MAX_LEN = 300;
 function formatToolResult(result: unknown): string {
   if (typeof result === 'string') return result;
   try {
-    return JSON.stringify(result);
+    const json = JSON.stringify(result);
+    return json ?? String(result);
   } catch {
     return String(result);
   }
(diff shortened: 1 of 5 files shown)


SoumyaRaikwar:
@jkowall  tool calls currently show raw JSON. So updated the Ask Jaeger panel to show tool calls as expandable blocks (description in header, raw output inside). i think its much more cleaner now 

[Screencast from 2026-06-29 02-38-12.webm](https://github.com/user-attachments/assets/344f1511-9a0a-44f2-ad03-e8b5d75389a0)

packages/jaeger-ui/src/components/App/JaegerAssistantPanel.tsx
```diff
@@ -31,21 +31,79 @@ function JaegerAssistantBootstrap() {
...
+function JaegerToolCallIndicator({
+  toolName,
+  argsText,
+  result,
+}: {
+  toolName: string;
+  argsText?: string;
+  result?: unknown;
+}) {
+  const isRunning = result === undefined;
```
jkowall: This should key off the part lifecycle status from `MessagePrimitive.Parts`, not the presence of `result`. A tool call can complete without returning a result payload, in which case assistant-ui marks the part complete while `result` remains `undefined`; this code would keep rendering `Calling tool…` forever. Please pass `part.status` into `JaegerToolCallIndicator` and treat only `running` or `requires-action` as the in-progress states. A regression test with `{ type: 'tool-call', status: { type: 'complete' }, result: undefined }` would cover the edge case.

jkowall (changes requested):
Thanks for tightening up the tool-call display. One lifecycle edge case needs fixing before merge: the renderer should use assistant-ui's `part.status` instead of inferring running state from `result === undefined`.

packages/jaeger-ui/src/components/App/JaegerAssistantPanel.tsx
SoumyaRaikwar: Updated the indicator to key off `part.status.type` instead of `result === undefined`:  only `running` and `requires-action` now show the "Calling…" state. Added a regression test with `{ type: 'tool-call', status: { type: 'complete' }, result: undefined }` to cover the edge case you described.

Also switched the completed state to a collapsible `<details>` element with Input/Output sections while at it.

packages/jaeger-ui/src/components/App/JaegerAssistantPanel.tsx
```diff
@@ -31,21 +31,82 @@ function JaegerAssistantBootstrap() {
...
 
+function formatToolResult(result: unknown): string {
+  if (typeof result === 'string') return result;
+  try {
+    const json = JSON.stringify(result);
+    return json ?? String(result);
+  } catch {
+    return String(result);
+  }
+}
```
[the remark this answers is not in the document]
SoumyaRaikwar: Skipping this suggestion:  `JSON.stringify` never returns `null` for non-null inputs (it either returns a string or throws), so the `json ?? String(result)` fallback is dead code. Also, we intentionally removed truncation since the result is now inside a collapsible `<details>` element where the full content is appropriate. The current implementation is simpler and handles the same cases.

packages/jaeger-ui/src/components/App/JaegerAssistantPanel.tsx
```diff
@@ -31,21 +31,84 @@ function JaegerAssistantBootstrap() {
...
-              component="div"
-              className="JaegerAssistantPanel-messageText"
-              smooth={false}
-            />
-          ) : null
-        }
+        {({ part }) => {
+          if (part.type === 'text') {
+            return (
+              <MessagePartPrimitive.Text
```
yurishkuro, line 129: How come this component receives no input from the actual agent response? What exactly does it display?

packages/jaeger-ui/src/components/App/JaegerAssistantPanel.tsx
```diff
@@ -31,21 +31,84 @@ function JaegerAssistantBootstrap() {
...
+  argsText,
+  result,
+  status,
+}: {
+  toolName: string;
+  argsText?: string;
+  result?: unknown;
+  status?: { type: string };
+}) {
+  const isRunning = status?.type === 'running' || status?.type === 'requires-action';
```
yurishkuro, line 83: what about other states like incomplete or error, `isError`?
```
const isDone = status?.type === 'complete' || status?.type === 'incomplete';
const isFailed = status?.type === 'incomplete' || part.isError;
```

packages/jaeger-ui/src/components/App/JaegerAssistantPanel.tsx
```diff
@@ -31,21 +31,84 @@ function JaegerAssistantBootstrap() {
   return null;
 }
 
+function formatToolResult(result: unknown): string {
+  if (typeof result === 'string') return result;
+  try {
+    const json = JSON.stringify(result);
```
yurishkuro: We do have a widget that can display JSON nicely, it's used in the Span Details for JSON encoded attributes. Can we reuse it here?

yurishkuro (commented):
this is very nice!

packages/jaeger-ui/src/components/App/JaegerAssistantPanel.tsx
SoumyaRaikwar: `MessagePartPrimitive.Text ` is a context-driven component from `@assistant-ui/react `.  it subscribes to the runtime store and re-renders as streaming text deltas arrive via SSE. The `MessagePrimitive.Parts` wrapper iterates over message parts and sets up context for each one; the primitive reads the text content from that context, not via explicit props. We could render part.text directly (like we do with `part.toolName` for tool calls), but the primitive handles live streaming updates without us managing store subscriptions manually.

packages/jaeger-ui/src/components/App/JaegerAssistantPanel.tsx
SoumyaRaikwar: Updated  the indicator now tracks incomplete and isError states. Failed tool calls render "Failed <tool_name>" with an error-colored border. Added tests for both isError: true and status: { type: 'incomplete' } cases.

packages/jaeger-ui/src/components/App/JaegerAssistantPanel.tsx
SoumyaRaikwar: done

SoumyaRaikwar:
[Screencast from 2026-06-30 00-27-03.webm](https://github.com/user-attachments/assets/0015ca1e-a94f-4368-b92b-3af938177caa)

updated

packages/jaeger-ui/src/components/App/JaegerAssistantPanel.tsx
```diff
@@ -31,21 +32,136 @@ function JaegerAssistantBootstrap() {
...
+          </div>
+        )}
+        {result !== undefined && (
+          <div className="JaegerAssistantPanel-toolCallSection">
+            <span className="JaegerAssistantPanel-toolCallLabel">Output</span>
+            <ToolCallResultValue result={result} />
+          </div>
+        )}
+      </div>
+    </details>
```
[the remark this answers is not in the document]
SoumyaRaikwar: updated the pr description

packages/jaeger-ui/src/components/App/JaegerAssistantPanel.tsx
```diff
@@ -31,21 +32,136 @@ function JaegerAssistantBootstrap() {
...
+function formatToolResult(result: unknown): string {
+  if (typeof result === 'string') return result;
+  try {
+    return JSON.stringify(result);
+  } catch {
+    return String(result);
+  }
+}
+
+const jsonViewStyles = {
```
yurishkuro: The style bridging is already in AttributesTable.tsx (lines 82–97) — the exact same 12-property style object inlined into <JsonView>. The PR copies it verbatim into jsonViewStyles instead of extracting it into a shared constant (e.g. in a utils/jsonView.ts or in AttributesTable.tsx itself as an export). This is straightforward duplication that should be factored out.

SoumyaRaikwar: Extracted the json-markup style object into a shared constant at `src/utils/jsonViewStyles.ts`. Both `AttributesTable` and `JaegerAssistantPanel` now import from there:  no more duplication.

yurishkuro (approved):

PR #4170, opened 2026-06-30
fix(filter-spans): use JSON.stringify for object/array attribute values in span search

### LINKED ISSUE #4165: [Bug]: Span search silently fails for object/array attribute values
## Describe the bug

Searching for spans in the trace timeline filter box silently returns no matches when the search term appears inside an object or array attribute value.

The root cause is in `packages/jaeger-ui/src/utils/filter-spans.ts` at line 42:

```typescript
const valueString = String(value);
```

`String(value)` on a JavaScript object always returns `"[object Object]"`, and on an array returns a comma-joined string (e.g. `"a,b"`) that loses keys and nested structure. So any span attribute whose value is a JSON object can never be found by value search, regardless of what text the user types.

## Steps to reproduce

1. Open a trace containing spans with structured attribute values — for example:
   - A GenAI trace where `gen_ai.input.messages` is an array of `{role, content}` objects
   - Any HTTP span where response headers are stored as an object
   - Any span with a nested-JSON attribute value
2. In the trace timeline, type a word that visually appears inside one of those values in the attributes table (e.g. `"user"`, `"gpt-4o"`, `"application/json"`).
3. No spans are highlighted.

## Expected behavior

Spans whose attribute values **contain** the search term anywhere — including inside nested objects or arrays — should be highlighted as matches.

## Actual behavior

```js
String({ role: "user", content: "hello world" })
// → "[object Object]"
// search for "user" or "hello" → no match
```

The span is never highlighted even though the text is clearly visible in the attributes table.

## Root cause

`filter-spans.ts` line 42 uses `String(value)` for all attribute value types. For objects and arrays it should use `JSON.stringify(value)` so the full structure is searchable.

The fix is a one-liner:

```typescript
// before
const valueString = String(value);

// after
const valueString = typeof value === 'object' && value !== null ? JSON.stringify(value) : String(value);
```

Note: `AttributesTable.tsx` already has exactly this logic in its local `getFilterValue()` helper (line 108). The global span filter utility has the same gap but the fix was never applied there.

## Impact

- Affects the span-level filter in the trace timeline (the text input that highlights matching spans)
- Affects all traces with OpenTelemetry structured attributes — increasingly common with GenAI, HTTP header capture, and rich instrumentation libraries
- Failure is silent: the UI shows zero results with no error, so users assume the span does not contain the data they searched for

## Environment

- File: `packages/jaeger-ui/src/utils/filter-spans.ts`
- Affects: main branch
- Triggered by: any span attribute whose value is an object, array, or Uint8Array

MOHITPARMAR007:
## Description

Fixes span search silently failing when the search term appears inside an object or array attribute value.

`String(value)` on a JavaScript object always returns `"[object Object]"`, and on an array returns a comma-joined string that loses keys and nested structure. So any span attribute whose value is a JSON object (e.g. `gen_ai.input.messages`, HTTP headers) could never be found by value search.

## Fix

Use `JSON.stringify(value)` for object/array attribute values so the full structure is searchable — matching the existing logic already used in `AttributesTable.tsx`'s `getFilterValue()` helper.

## Testing

Added 2 new test cases in `filter-spans.test.js` covering search inside object and array attribute values. All 22 tests pass.

Fixes #4165

### DIFF
--- a/packages/jaeger-ui/src/utils/filter-spans.test.js
+++ b/packages/jaeger-ui/src/utils/filter-spans.test.js
@@ -235,4 +235,42 @@ describe('filterSpans', () => {
   it('should return an empty set if no spans match the filter', () => {
     expect(filterSpans('-processTagKey1', spans)).toEqual(new Set());
   });
+  it('should return spans whose tags have an object value containing the filter text', () => {
+    const spanWithObjectTag = {
+      spanID: 'span-id-object',
+      operationName: 'operationNameObject',
+      process: {
+        serviceName: 'serviceNameObject',
+        tags: [],
+      },
+      tags: [
+        {
+          key: 'gen_ai.input.messages',
+          value: { role: 'user', content: 'hello world' },
+        },
+      ],
+      logs: [],
+    };
+    expect(filterSpans('hello', [spanWithObjectTag])).toEqual(new Set(['span-id-object']));
+    expect(filterSpans('user', [spanWithObjectTag])).toEqual(new Set(['span-id-object']));
+  });
+
+  it('should return spans whose tags have an array value containing the filter text', () => {
+    const spanWithArrayTag = {
+      spanID: 'span-id-array',
+      operationName: 'operationNameArray',
+      process: {
+        serviceName: 'serviceNameArray',
+        tags: [],
+      },
+      tags: [
+        {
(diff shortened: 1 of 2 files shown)


yurishkuro:
solved in #4171

Issue #4147, opened 2026-06-26
[Feature]: Override Jaeger UI timeout


carlos4ndre:
### Requirement

Ability to override the default timeout of 10s for requests sent to the Jaeger backend.

### Problem

Ideally requests sent to Jaeger Backend should be fast and never take more than 10s which is the [default timeout](https://github.com/jaegertracing/jaeger-ui/blob/782de5ffc5c6bbf71eca3e0baf81b6bf010e86a5/packages/jaeger-ui/src/api/v3/client.ts#L131), but for more heavy queries those might take a longer time to process, otherwise they will fail.

### Proposal

Provide an ability to override the Jaeger UI timeout potentially via the `ui-config.json` configuration file

<img width="1726" height="701" alt="Image" src="https://github.com/user-attachments/assets/d2bbafe6-5944-4fe5-a071-001b9992f211" />

parshipcy:
@yurishkuro @jkowall PTAL

shubhtrek:
Hi, I looked into this and the approach of adding requestTimeoutMs to ui-config.json seems like a clean and backward-compatible solution. The pattern already exists in the codebase (similar to traceGraph.layoutManagerMemory).

The change would be minimal — a new optional field in config.ts:

```ts
api?: {
  // Overrides the default 10s request timeout (in milliseconds).
  requestTimeoutMs?: number;
};
```
And then reading it in client.ts instead of the hardcoded value:

```ts
// before
private async fetchWithTimeout(url: string, timeout = 10000)

// after
private async fetchWithTimeout(url: string, timeout = getConfigValue('api.requestTimeoutMs') ?? 10000)

```

yurishkuro:
as a feature it makes sense

PR #4172, opened 2026-06-30
Add configurable API request timeout

dolliecoder:
-  Implements the configurable backend request timeout proposed in #4147 by allowing the default 10-second timeout to be overridden through the UI configuration

## Description of the changes
- Add a new optional `api.requestTimeoutMs` configuration field.
- Use the configured timeout in the v3 API client instead of the hardcoded 10-second default.
- Preserve the existing 10-second timeout when no configuration is provided.
- Add a unit test verifying that the configured timeout is respected.

- Added a unit test covering the configurable timeout.
- Verified that the existing default timeout behavior remains unchanged
- ran npm test -- src/api/v3/client.test.ts
- All 32 tests passed.

## Checklist
- [z] I have added unit tests for the new functionality

### DIFF
--- a/packages/jaeger-ui/src/api/v3/client.ts
+++ b/packages/jaeger-ui/src/api/v3/client.ts
@@ -9,6 +9,7 @@
  */
 
 import prefixUrl from '../../utils/prefix-url';
+import getConfig from '../../utils/config/get-config';
 import { ServicesResponseSchema, OperationsResponseSchema, TraceSummariesResponseSchema } from './schemas';
 import type { SearchQuery } from '../../types/search';
 import type { TraceSummary, ServiceSummary } from '../../types/trace-summary';
@@ -124,11 +125,11 @@ export class JaegerClient {
   /**
    * Fetch with timeout support using AbortController.
    * @param url - The URL to fetch
-   * @param timeout - Timeout in milliseconds (default: 10 seconds)
+   * @param timeout - Timeout in milliseconds (defaults to the configured request timeout)
    * @returns Promise<Response>
    * @throws Error if request times out or network error occurs
    */
-  private async fetchWithTimeout(url: string, timeout = 10000): Promise<Response> {
+  private async fetchWithTimeout(url: string, timeout = getConfig().api.requestTimeoutMs): Promise<Response> {
     const controller = new AbortController();
     const timeoutId = setTimeout(() => controller.abort(), timeout);
 
--- a/packages/jaeger-ui/src/constants/default-config.ts
+++ b/packages/jaeger-ui/src/constants/default-config.ts
@@ -66,6 +66,10 @@ const defaultConfig: Config = {
     },
     maxLimit: 1500,
   },
+  api: {
+    requestTimeoutMs: 10000,
+  },
+
   traceIdDisplayLength: 7,
   backendCapabilities: {
     archiveStorage: false,
--- a/packages/jaeger-ui/src/api/jaeger.test.js
+++ b/packages/jaeger-ui/src/api/jaeger.test.js
@@ -26,6 +26,7 @@ import JaegerAPI, {
 
 const defaultOptions = {
   credentials: 'same-origin',
+  signal: expect.any(AbortSignal),
 };
 
 describe('archiveTrace', () => {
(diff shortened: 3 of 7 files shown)


packages/jaeger-ui/src/constants/default-config.ts
```diff
@@ -66,6 +66,10 @@ const defaultConfig: Config = {
     },
     maxLimit: 1500,
   },
+  api: {
+    requestTimeoutMs: undefined,
```
yurishkuro: why isn't it 10000 here?

packages/jaeger-ui/src/api/v3/client.ts
```diff
@@ -128,7 +129,10 @@ export class JaegerClient {
    * @returns Promise<Response>
    * @throws Error if request times out or network error occurs
    */
-  private async fetchWithTimeout(url: string, timeout = 10000): Promise<Response> {
+  private async fetchWithTimeout(
+    url: string,
+    timeout = getConfig().api?.requestTimeoutMs ?? 10000
+  ): Promise<Response> {
```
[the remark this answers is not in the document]
yurishkuro: +1

packages/jaeger-ui/src/api/v3/client.ts
yurishkuro: default is not 10 seconds, it's whatever config says OR 10 seconds.

dolliecoder:
@yurishkuro I updated the default config to 10000, simplified the client to use the configured value directly andd updated the documentation comment and verified the tests still pass

packages/jaeger-ui/src/api/v3/client.ts
```diff
@@ -124,11 +125,14 @@ export class JaegerClient {
...
    * @param url - The URL to fetch
-   * @param timeout - Timeout in milliseconds (default: 10 seconds)
+   * @param timeout - Timeout in milliseconds (defaults to the configured request timeout)
    * @returns Promise<Response>
    * @throws Error if request times out or network error occurs
    */
-  private async fetchWithTimeout(url: string, timeout = 10000): Promise<Response> {
+  private async fetchWithTimeout(
+    url: string,
+    timeout = getConfig().api?.requestTimeoutMs
```
yurishkuro: this could theoretically return undefined (because of `?.`)

Does the section in the config actually need to be optional if we have the default value that will always be set from default-config.ts?

dolliecoder: I'll make the api section non-optional since it's always populated by default config.ts

dolliecoder:
@yurishkuro I've addressed the review comments and verified the tests pass could you please take another look when you have a chance?

anshgoyalevil (commented):
Nice implementation for the `api/v3/client.ts` path (and the mockdefault/undefined issue is already being discussed above). One gap I don't see raised yet: `api.requestTimeoutMs` is only consumed by `packages/jaeger-ui/src/api/v3/client.ts`. The v3 client is currently only used by `useTraceDiscovery.ts` (one hook). The primary API client, `packages/jaeger-ui/src/api/jaeger.ts` (used via `actions/jaeger-api.ts` for trace fetching, search, dependency graph, service metrics, and trace archiving — i.e. essentially every other network call in the app), calls `isomorphic-fetch` directly with no `AbortController`/timeout at all, configured or otherwise. So as written, this PR doesn't actually give users control over “the default 10-second request timeout” for the vast majority of API requests described in #4147 — only for the small v3 surface. Worth confirming whether `jaeger.ts` needs the same timeout wiring (or clarifying scope in the PR description) before merge.

dolliecoder:
@anshgoyalevil thanks just pushed the final cosmetic tweak to the array length

Dugera27:
Hi @dolliecoder thank you for working to get this feature. Teams I work with are interested in it. I was wondering is there much effort to resolve the conflicts to get this merged?

dolliecoder:
Hi @Dugera27 Thanks for reaching out. I've just rebased the branch against main and resolved all the merge conflicts. It should be good to go now! Let me know if there's anything else needed.

Dugera27:
@anshgoyalevil does this look good now on your end?

PR #4177, opened 2026-07-01
perf(tests): reduce redundant renders in TopNav.test.jsx

### LINKED ISSUE #3996: Speed up the Vitest unit-test suite

Shruti2110-coder:
Closes #3996 (partial — TopNav.test.jsx from item 5)

## What
Moves from `beforeEach(render)` to `beforeAll(render)` in two
describe blocks whose tests are all read-only assertions on the same
component tree. Uses `@testing-library/react/pure` to disable RTL's
automatic `afterEach(cleanup)` and calls `cleanup()` manually in
`afterAll` instead.

Tests that mutate state (`builds the Compare link`, `standalone renders`)
keep isolated renders with `afterEach(cleanup)`.

## Measurements
| | Duration |
|---|---|
| Before | 747 ms |
| After | 306 ms |
| Delta | **-59%** |

Run with:
`npx vitest run --reporter=json src/components/App/TopNav.test.jsx`

### DIFF
--- a/packages/jaeger-ui/src/components/App/TopNav.test.jsx
+++ b/packages/jaeger-ui/src/components/App/TopNav.test.jsx
@@ -2,12 +2,13 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import React from 'react';
-import { render, screen } from '@testing-library/react';
+import { render, screen, cleanup } from '@testing-library/react/pure';
 import { BrowserRouter } from 'react-router-dom';
 import '@testing-library/jest-dom';
 
 import { TopNavImpl as TopNav } from './TopNav';
 import { useTraceDiffStore } from '../../stores/trace-diff-store';
+
 
 vi.mock('antd', async () => {
   const actual = await vi.importActual('antd');
@@ -113,15 +114,14 @@ describe('<TopNav>', () => {
     pathname: '/search',
   };
 
-  beforeEach(() => {
-    useTraceDiffStore.setState({ a: null, b: null, cohort: [] });
-    useConfigMock.mockReset();
-    useConfigMock.mockReturnValue(defaultProps.config);
-  });
+  
 
   describe('renders the default menu options', () => {
     let component;
-    beforeEach(() => {
+    beforeAll(() => {
+      useTraceDiffStore.setState({ a: null, b: null, cohort: [] });
+    useConfigMock.mockReset();
+    useConfigMock.mockReturnValue(defaultProps.config);
       component = render(
         <BrowserRouter>
           <TopNav {...defaultProps} />
@@ -129,8 +129,9 @@ describe('<TopNav>', () => {
       );
     });
 
-    afterEach(() => {
+    afterAll(() => {
       component.unmount();
+      cleanup();
     });
 
     it('renders the "Jaeger" link', () => {
@@ -166,7 +167,10 @@ describe('<TopNav>', () => {
 
   describe('renders the custom menu', () => {
     let component;
-    beforeEach(() => {
+    beforeAll(() => {
+    useTraceDiffStore.setState({ a: null, b: null, cohort: [] });
+    useConfigMock.mockReset();
+    useConfigMock.mockReturnValue(defaultProps.config);
       component = render(
         <BrowserRouter>
           <TopNav {...defaultProps} />
@@ -174,8 +178,9 @@ describe('<TopNav>', () => {
       );
     });
 
-    afterEach(() => {
+    afterAll(() => {
       component.unmount();
+      cleanup();
     });
 
     it('renders the top-level item', () => {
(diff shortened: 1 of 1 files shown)


packages/jaeger-ui/src/components/App/TopNav.test.jsx
```diff
@@ -219,52 +224,61 @@ describe('<TopNav>', () => {
...
-    render(
-      <BrowserRouter>
-        <TopNav {...defaultProps} />
-      </BrowserRouter>
-    );
-
-    const navMenu = screen.getAllByTestId('mock-menu')[0];
-    expect(navMenu).toHaveAttribute('data-selectedkeys', JSON.stringify(['/search']));
+  describe('standalone renders', () => {
+beforeEach(() => {
```
anshgoyalevil, line 228: This `beforeEach` (and the `describe('standalone renders'` block around it) has inconsistent indentation and trailing whitespace, unlike the rest of the file which is consistently 2/4-space indented. Since `lint` includes an `fmt-lint` (format) check, this block will likely fail `make fmt --check` on CI — worth running the formatter again before merge.

PR #4181, opened 2026-07-01
feat(assistant): Add markdown rendering and panel resize

SoumyaRaikwar:
- Assistant replies in the Ask Jaeger panel render as plain text, discarding the markdown structure (lists, tables, inline/fenced code, blockquotes) that LLM output typically contains
- The panel is fixed at 26rem wide, which is cramped for responses containing tables or code blocks

## Description of the changes
- Render assistant text parts with `markdown-to-jsx` (9.8.2, ~15 kB gzip), with raw-HTML parsing disabled as defense-in-depth and streaming-optimized parsing enabled; user-typed messages keep the plain-text path
- Extract the rendering config into a shared `src/utils/markdownOptions.ts` so the upcoming GenAI span detail tab can reuse identical markdown rendering; element styles use design tokens from `vars.css` so both themes work
- Add drag-to-resize on the panel's left edge (320–800 px, default 416 px), with document-listener cleanup on unmount and a `max-width: 100%` cap so the panel never overflows narrow viewports

## Screenshots and screencast
Screenshot from 2026-07-02 03-07-54

### panel resizing
[Screencast from 2026-07-02 03-08-12.webm](https://github.com/user-attachments/assets/a382be10-8867-44b6-a225-a10def20b12d)

- 13 new unit tests (markdown vs. plain-text path selection per variant, drag widen/narrow, min/max clamping, drag stop on mouseup); the panel suite is 30 tests total with 98.7% statement / 100% function coverage on the changed component
- Full suite passes: 3,412 tests across both packages, plus `npm run lint`, `npm run tsc-lint`, and a production build
- Manual E2E against a live Jaeger backend + Gemini AI sidecar over AG-UI: verified streamed replies render tables, lists, bold, and code correctly alongside tool-call indicators, and the panel resizes smoothly

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/App/JaegerAssistantPanel.tsx
+++ b/packages/jaeger-ui/src/components/App/JaegerAssistantPanel.tsx
@@ -240,6 +240,11 @@ export function JaegerAssistantDock() {
 
   const onDragStart = React.useCallback(
     (e: React.MouseEvent) => {
+      if (e.button !== 0) return;
+      // A mouseup released outside the window can be missed; tear down any
+      // in-flight drag before starting a new one so listeners never stack.
+      dragCleanupRef.current?.();
+      e.preventDefault();
       const startX = e.clientX;
       const startWidth = width;
       document.body.style.cursor = 'ew-resize';
--- a/packages/jaeger-ui/src/components/App/JaegerAssistantPanel.tsx
+++ b/packages/jaeger-ui/src/components/App/JaegerAssistantPanel.tsx
@@ -12,10 +12,12 @@ import {
 } from '@assistant-ui/react';
 import { IoClose } from 'react-icons/io5';
 import { JsonView, collapseAllNested } from 'react-json-view-lite';
+import Markdown from 'markdown-to-jsx';
 
 import { useJaegerAssistant, useJaegerAssistantOptional } from './JaegerAssistantContext';
 import { useJaegerAssistantConfigured } from '../../hooks/useJaegerAssistant';
 import jsonViewStyles from '../../utils/jsonViewStyles';
+import { sharedMarkdownOptions } from '../../utils/markdownOptions';
 
 import './JaegerAssistantPanel.css';
 
@@ -117,6 +119,17 @@ function JaegerToolCallIndicator({
   );
 }
 
+function AssistantMarkdownText({ text }: { text: string }) {
+  return (
+    <Markdown
+      className="JaegerAssistantPanel-messageText JaegerAssistantPanel-md"
+      options={{ ...sharedMarkdownOptions, optimizeForStreaming: true }}
+    >
+      {text}
+    </Markdown>
+  );
+}
+
 function JaegerThreadMessageBody({ variant }: { variant: 'user' | 'assistant' }) {
   return (
     <MessagePrimitive.Root
@@ -125,6 +138,9 @@ function JaegerThreadMessageBody({ variant }: { variant: 'user' | 'assistant' })
       <MessagePrimitive.Parts>
         {({ part }) => {
           if (part.type === 'text') {
+            if (variant === 'assistant' && part.text) {
+              return <AssistantMarkdownText text={part.text} />;
+            }
             return (
               <MessagePartPrimitive.Text
                 component="div"
@@ -202,6 +218,10 @@ function JaegerAssistantThreadView({ focusComposer }: { focusComposer: boolean }
   );
 }
 
+const PANEL_MIN_WIDTH = 320;
+const PANEL_MAX_WIDTH = 800;
+const PANEL_DEFAULT_WIDTH = 416; // 26rem at 16px base
+
 /**
  * Right dock for Ask Jaeger.
  *
@@ -213,6 +233,40 @@ function JaegerAssistantThreadView({ focusComposer }: { focusComposer: boolean }
 export function JaegerAssistantDock() {
   const assistant = useJaegerAssistantOptional();
   const assistantConfigured = useJaegerAssistantConfigured();
+  const [width, setWidth] = React.useState(PANEL_DEFAULT_WIDTH);
+  // Holds the teardown for an in-flight drag so unmounting mid-drag does not
+  // leave document listeners attached or body cursor/user-select overrides stuck.
+  const dragCleanupRef = React.useRef<(() => void) | null>(null);
+
+  const onDragStart = React.useCallback(
+    (e: React.MouseEvent) => {
+      const startX = e.clientX;
+      const startWidth = width;
+      document.body.style.cursor = 'ew-resize';
+      document.body.style.userSelect = 'none';
+
+      const onMouseMove = (ev: MouseEvent) => {
+        const delta = startX - ev.clientX;
+        setWidth(Math.min(PANEL_MAX_WIDTH, Math.max(PANEL_MIN_WIDTH, startWidth + delta)));
+      };
+
+      const stopDragging = () => {
+        document.body.style.cursor = '';
+        document.body.style.userSelect = '';
+        document.removeEventListener('mousemove', onMouseMove);
+        document.removeEventListener('mouseup', stopDragging);
+        dragCleanupRef.current = null;
+      };
+
+      document.addEventListener('mousemove', onMouseMove);
+      document.addEventListener('mouseup', stopDragging);
+      dragCleanupRef.current = stopDragging;
+    },
+    [width]
+  );
+
+  React.useEffect(() => () => dragCleanupRef.current?.(), []);
+
   if (!assistant || !assistantConfigured) {
     return null;
   }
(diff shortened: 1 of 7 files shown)


packages/jaeger-ui/src/components/App/JaegerAssistantPanel.tsx
```diff
@@ -213,6 +233,45 @@ function JaegerAssistantThreadView({ focusComposer }: { focusComposer: boolean }
...
+        document.body.style.cursor = '';
+        document.body.style.userSelect = '';
+        document.removeEventListener('mousemove', onMouseMove);
+        document.removeEventListener('mouseup', stopDragging);
+        dragCleanupRef.current = null;
+      };
+
+      document.addEventListener('mousemove', onMouseMove);
+      document.addEventListener('mouseup', stopDragging);
+      dragCleanupRef.current = stopDragging;
```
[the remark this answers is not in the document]
SoumyaRaikwar: Applied the window blur teardown. Skipped document mouseleave: it fires whenever the pointer exits the viewport, which happens routinely mid-drag when overshooting the panel edge — cancelling there would end legitimate drags

packages/jaeger-ui/src/components/App/JaegerAssistantPanel.tsx
```diff
@@ -12,10 +12,12 @@ import {
 } from '@assistant-ui/react';
 import { IoClose } from 'react-icons/io5';
 import { JsonView, collapseAllNested } from 'react-json-view-lite';
+import Markdown from 'markdown-to-jsx';
```
jkowall: Non-blocking nit: since this is React-only usage, consider importing from `markdown-to-jsx/react` instead of the root `markdown-to-jsx` entry. The package keeps the root import for backward compatibility, but its 9.x docs recommend `/react` for React code, so switching now avoids depending on a legacy entry point without changing the API.

```suggestion
import Markdown from 'markdown-to-jsx/react';
```

SoumyaRaikwar: Done

yurishkuro (commented):
Three findings from a closer read, ordered by severity:

**1. Confirmed bug — empty-string `part.text` misroutes the first streaming token**

```tsx
if (variant === 'assistant' && part.text) {   // '' is falsy
```

`@assistant-ui/react-ag-ui`'s run-aggregator emits the first snapshot with `{ type: "text", text: "" }` — the text part is marked touched before any delta arrives. `part.text === ""` is falsy, so the condition falls through to `<MessagePartPrimitive.Text>` (the plain-text/user path) for every assistant message at stream start. The guard should be `part.type === 'text'` alone (already checked one level up); `AssistantMarkdownText` handles an empty string fine.

**2. Confirmed perf bug — `options` spread fires `compiler()` on every streamed token**

```tsx
options={{ ...sharedMarkdownOptions, optimizeForStreaming: true }}
```

`markdown-to-jsx` feeds `options` into a `useMemo([options])` → `mergedOptions`, which gates a second `useMemo([content, mergedOptions])` that runs `compiler()`. A new object identity on every render makes both memos miss. During a 500-token stream this fires `compiler()` 500 extra times — ironically undermining the `optimizeForStreaming` flag, which only sanitises trailing partial `<tag` fragments in input rather than reducing call frequency. Fix: hoist to a module-level constant in `markdownOptions.ts`:

```ts
export const streamingMarkdownOptions: MarkdownToJSX.Options = {
  ...sharedMarkdownOptions,
  optimizeForStreaming: true,
};
```

**3. Maintenance trap — `[width]` dependency and `startWidth` are load-bearing but non-obvious**

`onDragStart` captures `startWidth = width` from the closure, so `[width]` in the dep array is necessary — without it, every drag would start from `PANEL_DEFAULT_WIDTH`. But the dep also causes `onDragStart` to be recreated on every `setWidth` (60× per second during a drag). These two concerns are silently coupled: removing `[width]` to fix the churn would break drag start position. A future reader will be tempted to remove it. Consider a mirrored ref:

```tsx
const widthRef = React.useRef(PANEL_DEFAULT_WIDTH);
// when setting width: widthRef.current = w; setWidth(w);
// in onDragStart: const startWidth = widthRef.current;
// dep array: []
```

packages/jaeger-ui/src/components/App/JaegerAssistantPanel.tsx
```diff
@@ -224,8 +287,9 @@ export function JaegerAssistantDock() {
       className="JaegerAssistantPanel"
       aria-label="Ask Jaeger assistant"
       aria-hidden={!panelOpen}
-      style={{ display: panelOpen ? undefined : 'none' }}
+      style={{ display: panelOpen ? undefined : 'none', width, maxWidth: '100%' }}
     >
+      <div className="JaegerAssistantPanel-resizeHandle" onMouseDown={onDragStart} />
```
jkowall: Can we make the resize affordance keyboard and screen-reader operable before merging? Right now this is a mouse-only `div`, so the new panel resize control has no role, accessible name, focus target, keyboard path, or min/max/current value state.

A good fit here is a focusable `role="separator"` with `aria-orientation="vertical"`, `aria-valuemin`, `aria-valuemax`, `aria-valuenow`, and ArrowLeft/ArrowRight handlers that reuse the same width clamp logic as dragging. That keeps the visual behavior while making the control available to non-mouse users.

jkowall (commented):
One accessibility issue from the resize changes.

SoumyaRaikwar:
> Three findings from a closer read, ordered by severity:

FIxed all three as suggested, Thanks for the close read.

packages/jaeger-ui/src/components/App/JaegerAssistantPanel.tsx
SoumyaRaikwar: Added keyboard and screen-reader support as suggested

packages/jaeger-ui/src/utils/markdownOptions.ts
```diff
@@ -0,0 +1,104 @@
...
+      component: SafeImage,
+    },
+  },
+};
+
+// A stable object identity is required: markdown-to-jsx memoizes its parser
+// on the options reference, so spreading a new options object on every
+// render (e.g. { ...sharedMarkdownOptions, optimizeForStreaming: true })
+// defeats that memoization and re-runs the compiler on every streamed token.
+export const streamingMarkdownOptions: MarkdownToJSX.Options = {
```
yurishkuro: const declarations should be at the top of the file

packages/jaeger-ui/src/utils/markdownOptions.ts
```diff
@@ -0,0 +1,104 @@
...
+  ...sharedMarkdownOptions,
```
yurishkuro: why does sharedMarkdownOptions need to be exported or even defined? It looks like this is the only place using it, so why not just have a single streamingMarkdownOptions declaration?

for future span detail panel rendering, `optimizeForStreaming` is still useful because span attributes could be truncated at ingestion and the markdown renderer should be tolerant of unclosed tags and such.

packages/jaeger-ui/src/utils/markdownOptions.test.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+import { sharedMarkdownOptions, streamingMarkdownOptions } from './markdownOptions';
+
+describe('streamingMarkdownOptions', () => {
+  it('is the same reference across repeated reads', () => {
+    // markdown-to-jsx memoizes its parser on the options object identity. A
+    // fresh { ...sharedMarkdownOptions, optimizeForStreaming: true } literal
+    // built inline on every render would defeat that memoization on every
+    // streamed token; a module-level constant guarantees a stable reference.
+    expect(streamingMarkdownOptions).toBe(streamingMarkdownOptions);
```
yurishkuro: this condition is always true regardless of the reference. This test is not needed.

packages/jaeger-ui/src/utils/markdownOptions.ts
SoumyaRaikwar: done

packages/jaeger-ui/src/utils/markdownOptions.ts
SoumyaRaikwar: merged into one, ` optimizeForStreaming: true` is now just part of sharedMarkdownOptions,  good point that truncated/unclosed-tag tolerance is useful for the span detail tab's ingested content too, not just live streaming, so there was never a real reason to split them. ` JaegerAssistantPanel.tsx` now imports sharedMarkdownOptions directly with no spread.

packages/jaeger-ui/src/utils/markdownOptions.test.tsx
SoumyaRaikwar: removed

yurishkuro (approved):

PR #4191, opened 2026-07-05
feat(genai-tab): Add GenAI span detail content

swetalin-10:
- Part of jaegertracing/jaeger#8401 (GenAI observability): the "Rich-Media Side Panel" deliverable requires high-fidelity rendering of GenAI span data (token usage, prompts/completions, tool calls), but the GenAI tab added in #4122 is currently just a placeholder ("GenAI details coming soon"). This PR fills in that content.

## Description of the changes
- `GenAITab/genAiData.ts` (new): extracts provider/model, token usage (input/output/cache read/cache write/reasoning), input/output conversation, tool calls, and any other `gen_ai.*` attributes from a span's raw attributes.
- Conversation messages follow the current OTel GenAI semantic-conventions shape (`gen_ai.input.messages`/`output.messages` as `{ role, parts: [...] }`, not a flat `content` field). Each part renders per its `type`:
  - `text`/`reasoning`/`compaction` → content shown directly
  - `tool_call`/`server_tool_call` → shown as `→ name(args)`
  - `tool_call_response`/`server_tool_call_response` → shown as `← response`
  - `blob`/`file`/`uri` (media) → shown as a `[modality attached]` placeholder rather than silently dropped
  - Unrecognized/future part types fall back to a JSON dump so nothing disappears
- Still supports the deprecated flat `gen_ai.prompt`/`gen_ai.completion` shape (`{ role, content }`) for older instrumentation, matching the OTel deprecated-registry definition for those attributes.
- `GenAITab/index.tsx`: renders the above as a provider/model row, a token-usage row, a `Conversation` section (system instructions + input/output messages), a `Tool Call` section (arguments/result as a JSON tree, reusing the existing `jsonViewStyles`/`react-json-view-lite` pattern from `AttributesTable`), and an `Other GenAI Attributes` accordion so no `gen_ai.*` attribute is ever silently hidden.
- `SpanDetail/index.tsx`: the GenAI tab now defaults active (`defaultActiveKey="genai"`) when the selected span is a GenAI span, so users land directly on the enhanced view instead of the generic attributes tab. Non-GenAI spans are unaffected — only the single "Details" tab, unchanged.
- Markdown rendering is intentionally out of scope for this PR: prompts/completions render as plain preformatted text, so what's displayed always matches exactly what the model transmitted, without the risk of misinterpreting arbitrary model output as markdown or needing HTML sanitization this PR doesn't add. Rich formatting can be layered on top of this as a follow-up.
- `isGenAISpan` is currently sourced from a temporary `utils/genai/index.ts` helper. #3856 introduces a more complete `classifySpan`/`detect.ts` utility; once that merges, this import should be swapped to use it (tracked as a follow-up, not a blocker here since the two utilities are behaviorally equivalent for gating this tab).

The exact visual layout for token/message display is a first iteration and open to feedback — happy to adjust based on review.

## Screenshots
Captured from a running local build (`npm start`) against a sample trace with an `invoke_agent` span and a child `execute_tool` span, uploaded through Jaeger UI's built-in file loader — not mockups.

**Trace timeline** — both spans show the GenAI icon:
Trace timeline showing two spans with GenAI icons

**`invoke_agent` span — GenAI tab**: provider/model, token usage, and full conversation. The embedded `tool_call` part inside the assistant message renders as `→ get_weather({"city":"Paris"})` — the `parts[]` parsing fix in action, not a placeholder:
GenAI tab showing provider, model, token usage, and conversation

**`execute_tool` span — Tool Call section**: the separate top-level `gen_ai.tool.call.*` attributes render as an interactive JSON tree for both arguments and result:
GenAI tab showing tool call arguments and result as JSON trees

- Unit tests in `genAiData.test.ts` covering the `parts[]` schema (text, tool_call, tool_call_response, server_tool_call_response, media placeholder, unknown part fallback), the deprecated flat-content fallback, token usage (including numeric-string and cache/reasoning variants), and role validation.
- Unit tests in `index.test.tsx` covering provider/model, token usage, conversation rendering, tool calls (including unparseable JSON arguments), the empty state, and the "Other GenAI Attributes" accordion's collapsed/expanded states.
- Full `GenAITab`/`SpanDetail` suite passes: 137 tests across 13 files.
- `npm run tsc-lint` and `npm run oxlint` clean on all changed files.

## Checklist


packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/genAiData.ts
```diff
@@ -0,0 +1,257 @@
...
+    }
+  }
+  if (!Array.isArray(parsed)) return [];
+  return parsed.map((entry): GenAiMessage => {
+    if (typeof entry !== 'object' || entry === null) {
+      return { role: undefined, content: String(entry) };
+    }
+    const rec = entry as Record<string, unknown>;
+    const role = typeof rec.role === 'string' ? (rec.role as GenAiRole) : undefined;
+    if (Array.isArray(rec.parts)) {
```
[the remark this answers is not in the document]
swetalin-10: Fixed in 5b4a3f6a — added `asRole()` which validates against the actual OTel Role enum (system/user/assistant/tool) before assigning, falling back to `undefined` for any other string. Added a test asserting an unrecognized role like "developer" is dropped to `undefined`.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/genAiData.ts
```diff
@@ -0,0 +1,257 @@
...
+  switch (rec.type) {
+    case 'text':
+    case 'reasoning':
+    case 'compaction':
+      return typeof rec.content === 'string' ? rec.content : stringifyValue(rec.content ?? rec);
+    case 'tool_call':
+    case 'server_tool_call': {
+      const name = typeof rec.name === 'string' ? rec.name : 'unknown_tool';
+      const args = rec.type === 'tool_call' ? rec.arguments : rec.server_tool_call;
+      return `→ ${name}(${JSON.stringify(args)})`;
```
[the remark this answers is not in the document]
swetalin-10: Fixed in 5b4a3f6a — missing arguments/server_tool_call now default to `{}` instead of `undefined`, so this renders as `→ name({})` rather than `→ name(undefined)`. Added a test covering the no-arguments case.

yurishkuro:
Nice work — the extraction is well-factored and the test coverage is genuinely thorough. Two things I'd want addressed before this goes in, plus a couple of minor notes.

### 1. String-encoded tool arguments get double-encoded in `renderPart`

```ts
case 'tool_call':
case 'server_tool_call': {
  const name = typeof rec.name === 'string' ? rec.name : 'unknown_tool';
  const args = (rec.type === 'tool_call' ? rec.arguments : rec.server_tool_call) ?? {};
  return `→ ${name}(${JSON.stringify(args)})`;
}
```

Instrumentation commonly emits tool-call `arguments` as an **already-JSON-encoded string**. When it does, `JSON.stringify` escapes it a second time and the user sees:

```
→ get_weather("{\"city\":\"Paris\"}")
```

Note the top-level tool-call path (`JsonBlock`) already handles this exact case by attempting `JSON.parse` on strings first — so the same data shape is rendered two different ways depending on where it appears. Please normalize string args through parse-then-stringify here too (fall back to the raw string if it isn't JSON). The current tests only cover object-valued args, so this path is unguarded — worth a test with a string-encoded `arguments`.

### 2. `server_tool_call` reads its args from a different field than `tool_call`

In the same block, `tool_call` reads `rec.arguments` but `server_tool_call` reads `rec.server_tool_call`. That asymmetry looks suspicious — can you confirm the field name against the OTel GenAI parts schema? If a `server_tool_call` part actually carries its args under `arguments` (or something else), this renders `→ name()` for every server tool call. There's no test for `server_tool_call` args (only `server_tool_call_response` is covered), so a wrong field name would pass CI silently.

### Nits (non-blocking)

- The final empty-state guard re-derives several already-computed booleans (`!showTokens && !showConversation && !toolCall && !provider && !model && otherAttributes.length === 0`). Collapse to a single `hasContent` value.
- `tool_call_response` string responses render quoted (`← "rainy, 57°F"`), slightly at odds with the plain-text philosophy stated in the description — minor/subjective.

swetalin-10:
Thanks for the thorough review! Fixed in 3334a56f:

**1. Double-encoding** — confirmed and fixed. Added `stringifyToolValue()` which parses JSON-encoded strings before re-stringifying (falling back to the raw string if it isn't JSON), mirroring what `JsonBlock` already does for the top-level `gen_ai.tool.call.*` attributes. Applied to `tool_call`/`server_tool_call` arguments and `tool_call_response`/`server_tool_call_response` values. Added tests for both the string-encoded-object case and the plain-non-JSON-string case. As a side effect this also un-quotes plain-string `tool_call_response` values, so the second nit is resolved too.

**2. `server_tool_call` field name** — checked directly against the OTel GenAI parts JSON schema (`ServerToolCallPart` in `gen-ai-input-messages.json`): `ToolCallRequestPart` carries its payload under `arguments`, but `ServerToolCallPart` carries it under a same-named `server_tool_call` field (a polymorphic, provider-specific object) — required field, confirmed via `"required": ["type", "name", "server_tool_call"]`. So the asymmetry is correct per spec, not a bug. Added a code comment citing this and a test exercising `server_tool_call` args (previously only `server_tool_call_response` had coverage).

Also collapsed the empty-state guard into a single `hasContent` value per the other nit.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.tsx
```diff
@@ -1,12 +1,184 @@
...
+  // string, same as gen_ai.input.messages - try to parse strings so they get
+  // the interactive tree view too, not just a raw text dump.
+  let parsed = value;
+  if (typeof value === 'string') {
+    try {
+      parsed = JSON.parse(value);
+    } catch {
+      parsed = value;
+    }
+  }
```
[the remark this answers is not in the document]
swetalin-10: Fixed in 0b555c22 — added `tryParseJson()` using the same regex guard as `AttributesTable.tryParseJson` (only attempts `JSON.parse` when the string starts with `{`/`[`), and switched both `JsonBlock` here and `stringifyToolValue` (added for the double-encoding fix) to use it instead of a blind parse-then-catch. Added a test for the one new branch this introduces.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/genAiData.ts
```diff
@@ -0,0 +1,297 @@
...
+    findAttribute(attributes, 'gen_ai.provider.name') ?? findAttribute(attributes, 'gen_ai.system')
+  );
+  const model = asString(
+    findAttribute(attributes, 'gen_ai.response.model') ?? findAttribute(attributes, 'gen_ai.request.model')
+  );
+
+  const inputMessagesRaw =
+    findAttribute(attributes, 'gen_ai.input.messages') ?? findAttribute(attributes, 'gen_ai.prompt');
+  const outputMessagesRaw =
+    findAttribute(attributes, 'gen_ai.output.messages') ?? findAttribute(attributes, 'gen_ai.completion');
```
yurishkuro: please document the motivation for the fallbacks

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/genAiData.ts
```diff
@@ -0,0 +1,297 @@
...
+}
+
+/**
+ * Extracts GenAI span data from OTel attributes.
+ * Checks current attribute names first, falling back to deprecated ones
+ * still commonly emitted by existing instrumentation:
+ * - gen_ai.provider.name (current) vs gen_ai.system (deprecated)
+ * - gen_ai.input.messages/output.messages (current) vs gen_ai.prompt/completion (deprecated)
+ */
+export function extractGenAiData(attributes: ReadonlyArray<IAttribute>): GenAiData {
```
yurishkuro: the attributes array is being linearly scanned N times in this function. Create a map once and do by-key lookups.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/genAiData.ts
```diff
@@ -0,0 +1,297 @@
...
+    cacheReadInputTokens: asNumber(findAttribute(attributes, 'gen_ai.usage.cache_read.input_tokens')),
+    cacheCreationInputTokens: asNumber(findAttribute(attributes, 'gen_ai.usage.cache_creation.input_tokens')),
+    reasoningOutputTokens: asNumber(findAttribute(attributes, 'gen_ai.usage.reasoning.output_tokens')),
+  };
+
+  const toolName = asString(findAttribute(attributes, 'gen_ai.tool.name'));
+  const toolArguments = findAttribute(attributes, 'gen_ai.tool.call.arguments');
+  const toolResult = findAttribute(attributes, 'gen_ai.tool.call.result');
+  const toolId = asString(findAttribute(attributes, 'gen_ai.tool.call.id'));
+  const hasToolCall = toolName != null || toolArguments != null || toolResult != null;
```
yurishkuro: is `toolId` intentionally excluded from this check?

yurishkuro (commented):
# GenAI span-detail rendering: architecture options

Context: review of [jaegertracing/jaeger-ui#4191](https://github.com/jaegertracing/jaeger-ui/pull/4191), which adds real content to `GenAITab` (provider/model, token usage, conversation, tool calls, "other attributes"). The PR's `genAiData.ts` reads ~15 individual `gen_ai.*` attributes into named variables via repeated `attributes.find()` calls, and separately maintains a `HANDLED_KEYS` allowlist so anything not extracted falls into a generic "Other GenAI Attributes" section without duplication.

## Principle: no data hiding

`GenAITab` is a debugging/observability surface. Its job is to make GenAI span data legible, not to decide on the user's behalf what's worth showing. So whenever this design has to choose between "guess the right value and discard the rest" and "show everything, even imperfectly," it chooses the latter. A field the code doesn't specifically recognize - an attribute key, a legacy value that disagrees with its current replacement, a token-usage field added after this code was written, a section type from a future registry entry - must still surface *somewhere*, even if only as a raw key/value pair, rather than being silently dropped. An ugly-but-honest fallback beats a clean UI that quietly threw data away. This principle is why every fallback mechanism in this design (`get()` claiming keys instead of `HANDLED_KEYS`, short-circuiting current/legacy reads, `Object.keys(usage)` instead of a fixed field list, the generic `default` case in the section dispatcher) is built the same way: something unrecognized always has a place to land.

## Problems with the current (PR #4191) approach

- **`HANDLED_KEYS` drift.** Every new extracted field needs a *second*, unenforced edit to `HANDLED_KEYS`. Miss it and the field renders twice — once via its named field, once again in "Other GenAI Attributes."
- **Concrete bug from the drift risk, and a violation of no-data-hiding:** `hasToolCall` only checks `toolName`/`toolArguments`/`toolResult`, not `toolId` (`genAiData.ts:272`). A span carrying only `gen_ai.tool.call.id` gets no tool-call section — and because that key is in `HANDLED_KEYS`, it's also excluded from "other," so it disappears entirely, with no path in the UI where a user could ever see it.
- **Repeated O(n) scans.** `findAttribute` does a fresh linear `attributes.find()` per call; `extractGenAiData` calls it ~15 times, then a 16th full pass for `otherAttributes`. Not currently a performance problem at realistic per-span attribute counts, but it's a lot of redundant work for a design that's already fragile for other reasons.
- **Same duplication, moved to the render layer.** `GenAITab/index.tsx` has five near-identical JSX blocks, one per token-usage field, each gated on `usage.<field> != null` and differing only in which field it reads and its label text. `GenAiTokenUsage` (the type) and this JSX (the render) each separately enumerate the same five fields, so it's easy to add a sixth usage field to the type and forget the matching JSX block — the field would be extracted but silently never shown. See "Rendering-side redundancy the registry output also resolves" below.

## Proposed design: single-pass registry of renderers

Build one `Map` from the span's attributes. Walk a small ordered list of `build` functions; each one reads whichever keys it cares about through a shared `get()` accessor, and `get()` itself removes a key from the map the moment it's read. Ownership of a key is therefore defined by nothing but "some `build` function called `get()` on it" — there is no second, separately declared key list to keep in sync with the read logic, which was the exact shape of the `HANDLED_KEYS` bug. Whatever's left in the map after every entry has run is the generic/"other" bucket automatically. Render order is just `REGISTRY`'s declaration order, plus `other` appended last — no separate `rank` field is needed: the loop walks the registry array itself (developer-controlled, fixed order), not the span's `attributes` array (whose order is arbitrary and instrumentation-dependent), so the output order is already deterministic before anything gets built. A `rank` number would only earn its keep if sections were instead being discovered by scanning attributes in *their* order and dispatching per-attribute — that's not what this loop does.

An important authoring rule falls out of this, in the opposite direction from what might seem natural: for a genuine current/deprecated key pair, use a short-circuiting `get(current) ?? get(legacy)` (or `||` for plain strings), not two unconditional reads. If both keys are present and *disagree* — a real instrumentation bug, not just an expected legacy-to-current transition — the unused legacy value must stay unclaimed rather than being silently discarded, per the no-data-hiding principle above. Leaving it unclaimed means it falls through to "Other GenAI Attributes," where the conflict is still visible; claiming and discarding it would make a real divergence invisible. This only applies when the pair is *actually* deprecated-vs-current per spec — `gen_ai.request.model`/`gen_ai.response.model` looks like the same shape but isn't: per the OTel semconv registry, both are current, and they record different things (requested model vs. model that actually responded), so they're handled by showing both rather than falling back between them (see the `meta` entry below).

This is a design sketch, not the final code:

```typescript
// Unchanged from the current PR's genAiData.ts - shown here only because
// GenAiSection below references them.
type GenAiMessage = { role?: 'system' | 'user' | 'assistant' | 'tool'; content: string };
type GenAiToolCall = { id?: string; name?: string; arguments?: unknown; result?: unknown };
type GenAiTokenUsage = {
  inputTokens?: number;
  outputTokens?: number;
  cacheReadInputTokens?: number;
  cacheCreationInputTokens?: number;
  reasoningOutputTokens?: number;
};

// Every variant separates its discriminant (`type`) from its payload
// (`data`) so a generic fallback renderer can walk `Object.entries(data)`
// for a section type it doesn't have a specific case for - see the
// `default:` arm in the dispatcher below.
type GenAiSection =
  | { type: 'meta'; data: { provider?: string; model?: string } }
  | { type: 'tokens'; data: GenAiTokenUsage }
  | { type: 'conversation'; data: { systemInstructions?: string; inputMessages: GenAiMessage[]; outputMessages: GenAiMessage[] } }
  | { type: 'toolCall'; data: GenAiToolCall }
  | { type: 'other'; data: { attributes: IAttribute[] } };

// get() claims a key as a side effect of reading it (removes it from the
// shared pool). A build function's "owned keys" are exactly the keys it
// happens to call get() on - nothing is declared twice.
type GetAttr = (key: string) => AttributeValue | undefined;
type SectionBuilder = (get: GetAttr) => GenAiSection | undefined;

// Array order IS render order - no separate rank field to keep in sync
// with position.
const REGISTRY: SectionBuilder[] = [
  // meta
  get => {
    // `gen_ai.provider.name`/`gen_ai.system` is a genuine current/deprecated
    // pair - `||` short-circuits so `gen_ai.system` is only read (and
    // claimed) when `gen_ai.provider.name` is absent/empty, leaving a
    // disagreeing legacy value unclaimed so it surfaces in "Other GenAI
    // Attributes" instead of being silently discarded.
    const provider = asString(get('gen_ai.provider.name')) || asString(get('gen_ai.system'));

    // `gen_ai.request.model`/`gen_ai.response.model` are NOT a
    // current/deprecated pair per the OTel spec - both are current, and
    // they record different things: which model was requested vs. which
    // model actually served the response (a provider substituting a
    // pinned version, e.g. request "gpt-4" / response "gpt-4-0613", is
    // the expected case, not a conflict). Show both rather than
    // collapsing one into a fallback for the other.
    const requestModel = asString(get('gen_ai.request.model'));
    const responseModel = asString(get('gen_ai.response.model'));
    const model =
      responseModel && requestModel && responseModel !== requestModel
        ? `${responseModel} (requested: ${requestModel})`
        : responseModel || requestModel;
    return provider || model ? { type: 'meta', data: { provider, model } } : undefined;
  },
  // tokens
  get => {
    const usage: GenAiTokenUsage = {
      inputTokens: asNumber(get('gen_ai.usage.input_tokens')),
      outputTokens: asNumber(get('gen_ai.usage.output_tokens')),
      cacheReadInputTokens: asNumber(get('gen_ai.usage.cache_read.input_tokens')),
      cacheCreationInputTokens: asNumber(get('gen_ai.usage.cache_creation.input_tokens')),
      reasoningOutputTokens: asNumber(get('gen_ai.usage.reasoning.output_tokens')),
    };
    return hasAnyTokenUsage(usage) ? { type: 'tokens', data: usage } : undefined;
  },
  // conversation
  get => {
    // Same short-circuit-on-purpose rule as the meta builder above: only
    // fall back to `gen_ai.prompt`/`gen_ai.completion` when the current
    // key is absent, so a legacy value that disagrees with the current one
    // is left unclaimed and surfaces in "Other GenAI Attributes" instead
    // of being silently dropped.
    const inputMessages = parseMessages(get('gen_ai.input.messages') ?? get('gen_ai.prompt'));
    const outputMessages = parseMessages(get('gen_ai.output.messages') ?? get('gen_ai.completion'));
    const systemInstructions = parseSystemInstructions(get('gen_ai.system_instructions'));
    return inputMessages.length || outputMessages.length || systemInstructions
      ? { type: 'conversation', data: { inputMessages, outputMessages, systemInstructions } }
      : undefined;
  },
  // toolCall
  get => {
    const id = asString(get('gen_ai.tool.call.id'));
    const name = asString(get('gen_ai.tool.name'));
    const args = get('gen_ai.tool.call.arguments');
    const result = get('gen_ai.tool.call.result');
    return id || name || args !== undefined || result !== undefined
      ? { type: 'toolCall', data: { id, name, arguments: args, result } }
      : undefined;
  },
];

export function extractGenAiSections(attributes: ReadonlyArray<IAttribute>): GenAiSection[] {
  const remaining = new Map(attributes.map(a => [a.key, a.value]));
  const get: GetAttr = key => {
    const value = remaining.get(key);
    remaining.delete(key);
    return value;
  };

  const sections = REGISTRY.map(build => build(get)).filter((s): s is GenAiSection => s !== undefined);

  const other = [...remaining].filter(([k]) => k.startsWith('gen_ai.')).map(([key, value]) => ({ key, value }));
  if (other.length) sections.push({ type: 'other', data: { attributes: other } });

  return sections;
}
```

`index.tsx` becomes a single `sections.map(...)` with a `switch (section.type)`, one case per shape, `other` feeding the accordion. Every builder runs unconditionally (there are only 4-5 of them, so no need for a presence pre-check before calling it) — each one decides for itself, from what `get()` returns, whether it has anything to show.

This fixes the `hasToolCall`/toolId bug directly: the tool-call section's firing condition now includes `id`, because `id` is simply one of the values `build` reads, with no separate list to have gotten out of sync. It also fixes the redundancy in the first version of this sketch, where a `keys: [...]` array on each entry duplicated the exact same literal strings passed to `get(...)` inside `build` — same drift risk as `HANDLED_KEYS`, just re-introduced at per-entry scope. It does **not** by itself fix the `parseSystemInstructions` non-array bug found in review — that lives inside an individual `build` function and is orthogonal to the loop structure.

## Rendering-side redundancy the registry output also resolves

First, an example of the `sections.map(...)`/`switch (section.type)` dispatcher mentioned above, actually consuming `extractGenAiSections()`'s full output - this is the part of `GenAITab` that has exactly one case per section *type* (meta, tokens, conversation, toolCall, other), so there's no repetition to fix among these named cases; each is structurally different from the others. But the switch as written has no `default`, so a sixth section type added to `REGISTRY` in the future without a matching `case` here would silently render nothing - the same silent-drop bug recurring a third time, now at the dispatch layer instead of the extraction or per-field-render layer. Unlike `gen_ai.*` attribute keys (an open set defined by external instrumentation, which is why `HANDLED_KEYS`/`get()` needed the "unclaimed things still show up somewhere" treatment), the set of `GenAiSection` types is closed and authored in this same module - so the primary defense should be a compile-time nudge (e.g. a lint rule that flags a non-exhaustive `switch` over a union) to add a real `case` for a new type. But that nudge can be bypassed (a loosened type, a disabled rule, a rushed review), so the `default` arm below is a second, runtime-only line of defense: it can't know how to render a type it's never heard of *well*, but because every variant now shares the same `{type, data}` shape, it can always render *something* - a generic key/value dump of `data`'s own fields, the same `Object.entries` trick `TokensRow` uses below for individual token fields, just one level up. Each value gets rendered through `JsonBlock`, the same component the current PR already uses for the `toolCall` section's arguments/result (and modeled on `AttributesTable`'s pattern): it tries to parse a string as JSON and tree-views it if that succeeds, tree-views a value that's already a nested object/array (`AttributeValue` in this codebase, `packages/jaeger-ui/src/types/otel.ts:20-25`, is recursive - `{ [key: string]: AttributeValue }` and `Array<AttributeValue>` are both legal attribute values here, not just JSON-encoded strings), and falls back to plain text otherwise - `JSON.stringify`-ing a value inline would both dump unformatted JSON as a wall of text and reinvent a worse version of a component that already exists in this file:

This is a design sketch, not the final code:

```tsx
function UnknownSection({ type, data }: { type: string; data: Record<string, unknown> }) {
  return (
    <div className="GenAITab--section">
      <h3 className="GenAITab--sectionTitle">{type}</h3>
      {Object.entries(data).map(([key, value]) => (
        <div key={key} className="GenAITab--toolSubsection">
          <span className="GenAITab--toolLabel">{key}</span>
          <JsonBlock value={value} />
        </div>
      ))}
    </div>
  );
}

export default function GenAITab({ span }: Props): React.ReactElement {
  const sections = useMemo(() => extractGenAiSections(span.attributes), [span.attributes]);

  if (sections.length === 0) {
    return <div className="GenAITab--empty">No GenAI-specific attributes found on this span.</div>;
  }

  return (
    <div className="GenAITab">
      {sections.map(section => {
        switch (section.type) {
          case 'meta':
            return <MetaRow key="meta" {...section.data} />;
          case 'tokens':
            return <TokensRow key="tokens" usage={section.data} />;
          case 'conversation':
            return <ConversationSection key="conversation" {...section.data} />;
          case 'toolCall':
            return <ToolCallSection key="toolCall" toolCall={section.data} />;
          case 'other':
            return (
              <AccordionAttributes
                key="other"
                label="Other GenAI Attributes"
                data={section.data.attributes}
                linksGetter={null}
              />
            );
          default:
            return <UnknownSection key={section.type} type={section.type} data={section.data} />;
        }
      })}
    </div>
  );
}
```

The `default` arm is dead code today - every variant `extractGenAiSections` can currently produce has an explicit `case` above it - and that's the point: it costs nothing while the switch stays exhaustive, and it only ever executes if that invariant is broken. Per the no-data-hiding principle above, an ugly-but-honest key/value dump in that case beats rendering nothing.

Whether a case spreads `section.data` (`meta`, `conversation`) or passes it as a single named prop (`tokens`, `toolCall`, `other`) isn't arbitrary, though it's easy to misread as inconsistent: it depends on whether the target component's props line up 1:1 with `data`'s own fields, or whether the component needs the whole object as one unit. `MetaRow`'s props are exactly `provider`/`model` - `data`'s two fields - so `{...section.data}` is just `provider={...} model={...}` written once. `TokensRow` takes a single `usage` prop and runs `Object.keys(usage)` over it internally (the fix from the tokens-redundancy section above); spreading `section.data` there would be structurally wrong, since none of `data`'s own keys is named `usage` - it needs the whole object passed through as one value precisely so it can iterate it generically instead of naming its fields. `ToolCallSection` and `AccordionAttributes` are the same: the whole `toolCall`/`attributes` value is one cohesive thing to those components, not a set of independent props.

The duplication PR #4191's review flagged is *inside* the `'tokens'` case specifically (`TokensRow` below), not across these cases — five near-identical JSX blocks, one per token-usage field, differing only in which `usage.*` property they read and their label text:

This is an abbreviated illustration of the current PR's code, not a verbatim quote — the surrounding `<div>` is the same one the JSX below replaces the *inside* of, so the two snippets can be compared directly:

```tsx
<div className="GenAITab--tokens">
  {usage.inputTokens != null && (
    <span className="GenAITab--tokenItem">
      <span className="GenAITab--tokenLabel">Input</span> {formatTokenCount(usage.inputTokens)}
    </span>
  )}
  {usage.outputTokens != null && ( /* same shape, different field/label */ )}
  {usage.cacheReadInputTokens != null && ( /* same shape, different field/label */ )}
  {usage.cacheCreationInputTokens != null && ( /* same shape, different field/label */ )}
  {usage.reasoningOutputTokens != null && ( /* same shape, different field/label */ )}
</div>
```

That's the exact same failure mode as `HANDLED_KEYS`, just moved to the render layer instead of the extraction layer: `GenAiTokenUsage` (the type) and this JSX (the render) each separately enumerate the same five fields, so adding a sixth usage field to the type is easy to do while forgetting to copy-paste a sixth JSX block — the new field would be extracted but silently never shown.

An earlier version of this fix replaced the five blocks with a `.map()` over a fixed `{key, label}` array — but that array is itself a third enumeration of the same five field names, so a sixth `GenAiTokenUsage` field still needs a matching entry added by hand, or it's dropped exactly as before, just with a smaller footprint. The actual fix is to loop over `usage`'s own keys instead of a separately declared list, and keep a labels map purely for cosmetic display - a key with no entry in it still renders, under its raw key name, rather than disappearing:

This is a design sketch, not the final code:

```tsx
// Cosmetic only - a key missing from here still renders, just under its
// raw field name instead of a friendly label.
const TOKEN_LABELS: Partial<Record<keyof GenAiTokenUsage, string>> = {
  inputTokens: 'Input',
  outputTokens: 'Output',
  cacheReadInputTokens: 'Cached (read)',
  cacheCreationInputTokens: 'Cached (write)',
  reasoningOutputTokens: 'Reasoning',
};

// same <div>, same className - only what's inside it changed:
function TokensRow({ usage }: { usage: GenAiTokenUsage }) {
  return (
    <div className="GenAITab--tokens">
      {(Object.keys(usage) as Array<keyof GenAiTokenUsage>).map(key => {
        const value = usage[key];
        if (value == null) return null;
        return (
          <span key={key} className="GenAITab--tokenItem">
            <span className="GenAITab--tokenLabel">{TOKEN_LABELS[key] ?? key}</span> {formatTokenCount(value)}
          </span>
        );
      })}
    </div>
  );
}
```

This is the `TokensRow` the dispatcher above calls for `case 'tokens'`. A sixth `GenAiTokenUsage` field renders automatically the moment it's added to the object built in the `tokens` builder (`Object.keys(usage)` picks it up for free, since that object always assigns all its fields, even to `undefined`) — `TOKEN_LABELS` is optional polish, not a second required list. This is the no-data-hiding principle again, one level down: an unmapped thing still surfaces (here, under a raw key instead of a friendly label; earlier, in "Other GenAI Attributes" instead of a named field) rather than silently vanishing.

## Survey of comparable vendor UIs

Six GenAI/agent-observability UIs were checked for how they extract and render span-level GenAI attributes, to see whether any already use a registry-of-renderers pattern like the one proposed above.

- **agent-prism** (`packages/data/src/open-telemetry/adapter.ts`) — named lookups. Each field (`getSpanCost`, `getSpanTokensCount`, `getSpanInputOutput`) is its own function doing its own `attributes.find()` calls. No rank/priority. Tool calls and multi-part messages aren't handled at all — constants exist (`gen_ai.tool.name`, `llm.input_messages`) but are unconsumed; they'd fall into the raw JSON-or-plain generic attribute renderer (`DetailsViewAttributesTab.tsx:37-83`).

- **Arize Phoenix** (`app/src/pages/trace/SpanDetails.tsx`) — named lookups, dispatched by span kind. Top-level `switch (spanKind)` → `LLMSpanInfo`/`RetrieverSpanInfo`/etc., each with ~7 near-identical "pull key, validate, fallback" `useMemo` blocks (`SpanDetails.tsx:581-677`). Token counts are precomputed server-side, not parsed from raw attributes client-side. Un-mapped fields fall into an always-present raw-JSON "Attributes" tab (`AttributesJSONBlock`) — a separate tab, not inlined. Tool calls (`LLMMessage`, `SpanDetails.tsx:1297-1486`) and multi-part content (text/image, `MessageContentsList`) are both handled, with an explicit deprecated `function_call` → `tool_calls` fallback.

- **Langfuse** (`packages/shared/src/server/otel/OtelIngestionProcessor.ts`) — named lookups, but with real ranked-fallback lists at ingestion. `extractModelName` walks an *ordered* array of candidate keys in a loop (closest thing to "one key wins, ranked"). `extractUsageDetails`/`extractModelParameters` are large if/else chains keyed on `instrumentationScopeName`, each hardcoding its own field names — i.e. the "registry" is per-vendor-SDK, not per-attribute. Leftover attributes become generic `metadata` (`IOPreviewPretty.tsx`, always rendered last — fixed order, not rank-driven). Tool *definitions* vs tool *call invocations* are extracted and rendered separately (`useChatMLParser.ts`, `ToolCallInvocationsView.tsx`).

- **OpenLIT** (`src/client/src/constants/traces.ts`, `helpers/client/trace.ts`) — hybrid: a table exists but isn't used for rendering. `TraceMapping` is a genuine `{key: {label, path, paths, prefix, ...}}` table walked once by `normalizeTrace()` — the closest prior art to the proposed design. But the **rendering** layer (`trace-detail-page.tsx`, `chat-view.tsx`) doesn't walk that table; it pulls named fields individually with raw string-literal attribute keys. An `SPAN_ATTR_INFO_PILL_KEYS` ordered array and `KNOWN_SPAN_ATTR_KEYS` set look purpose-built for exactly a "walk registry, render ordered, treat rest as unknown" pattern — but grep confirms neither is actually consumed anywhere; they're dead exports. Unmapped attributes go through an undeduped raw-JSON grid (no exclusion of already-shown keys).

- **OpenSearch agent-health** (`services/traces/utils.ts`, `messageExtraction.ts`) — named lookups, forked per span-type/vendor. `getKeyAttributes` branches on span name/operation (if/else: eval / bedrock-llm / tool / default), each branch its own hardcoded fallback chain. `messageExtraction.ts` forks again into `extractClaudeCodeMessages` vs `extractGenericMessages`, each re-deriving prompt/completion/tool fields independently. `EXPECTED_ATTRIBUTES: Record<SpanCategory, string[]>` exists but is only used for compliance warnings, not rendering. Unmapped attributes fall through to a flat key/value table (`SimpleSpanAttributesTable.tsx`) — nothing dropped, but adding a new `gen_ai.*` key means touching several independent fallback chains across files.

- **OpenSearch Dashboards** (`src/plugins/agent_traces/public/application/pages/traces/hooks/span_transforms.ts`) — named lookups. `traceHitToAgentSpan` is one field per line (`genAiSystem: attrs?.gen_ai?.system || ''`, etc.) building a fixed struct. No tool-call handling at all (no `gen_ai.tool.*` reads anywhere in the plugin). Unmapped fields are only reachable via a raw "Raw Span" JSON accordion — no structured "other" section. Notably, the *codebase* (not this plugin) has real registry-sorted-by-rank precedent elsewhere: `DocViewsRegistry.getDocViewsSorted()` (`src/plugins/discover/public/application/doc_views/doc_views_registry.ts:33-54`) filters by a `shouldShow(hit)` predicate and sorts by numeric `order` — structurally identical to the proposed `keys`/`rank`/`build` registry, just applied to document-view plugins rather than GenAI attributes. `field_formats_registry.ts` is a second precedent (`Map<id, FieldFormat>` with a generic lowest-priority default formatter).

### Takeaway

None of the six surveyed projects render GenAI span attributes via a single-pass, order-preserving registry of builders. All six independently reinvented some flavor of "named variable per field, repeated per field," and every one of them shows the same symptom this review flagged in PR #4191: scattered fallback logic that has to be updated in multiple places when a new attribute is added, with generic/"other" attribute handling bolted on separately rather than falling out of the same mechanism. OpenLIT's `TraceMapping`/`SPAN_ATTR_INFO_PILL_KEYS` and OpenSearch Dashboards' `DocViewsRegistry` are the two pieces of real prior art for a "walk a table once, generic lowest-priority fallback for anything unclaimed" shape close to what's proposed above (OpenSearch Dashboards' version does use an explicit numeric `order`, since `DocViewsRegistry` dispatches over document *hits* in arbitrary order rather than a fixed builder list) — but in both cases, the table exists and the actual rendering code doesn't use it. That's a second, independent argument for building the registry so it verifiably drives the render, rather than living beside it as another artifact that can drift out of sync.

## Span events: a source PR #4191 doesn't read at all

This is just for the record and for future improvements.

PR #4191's `extractGenAiData(span.attributes)` reads **only span attributes**. It never looks at `span.events`. That's a real gap, not a hypothetical one — OTel's GenAI semantic conventions historically modeled message content as span *events*, and a meaningful amount of real instrumentation still emits that way:

- The original OTel GenAI semconv defined `gen_ai.system.message`, `gen_ai.user.message`, `gen_ai.assistant.message`, `gen_ai.tool.message` (input) and `gen_ai.choice` (output) as span events, each carrying the message content in the event's own attributes — before the newer `gen_ai.input.messages`/`gen_ai.output.messages` span-attribute convention that this PR targets existed.
- A separate legacy convention (Semantic Kernel) emits a single `gen_ai.content.prompt` event and a single `gen_ai.content.completion` event instead.
- This isn't dead: **OpenLIT** actively reads both conventions off `span.events` in its Strands instrumentation (`sdk/python/src/openlit/instrumentation/strands/utils.py:139-206`, mirrored in the TypeScript SDK), and several of its own instrumentations (Cohere, Multion, Together, AI21, Mistral, PremAI, Julep) still *write* `gen_ai.content.prompt`/`gen_ai.content.completion` via `span.add_event(...)` as a legacy path alongside attributes.
- **Langfuse**'s ingestion (`OtelIngestionProcessor.ts:1679-1773`) treats events as a real, load-bearing input: after attribute-based framework-specific extraction fails, it falls back to scanning `span.events` for exactly the event names above, reduces each event's `attributes[]` into a message object, and merges the result into the same `{input, output}` shape produced from attributes.
- **OpenSearch agent-health** goes further: `SpanDetailsPanel.tsx:33-34` and `messageExtraction.ts` treat `span.events` (`llm.request`/`llm.response`, or Claude-Code-specific `user_prompt`/`tool_decision`/`tool_result` events) as the *primary* source, with attribute-based extraction as the documented fallback "when no events present." Some fields (`llm.system_prompt`, `bedrock.tool_count`) exist **only** on events — there's no attribute equivalent anywhere in that codebase.
- By contrast, **Phoenix**, **agent-prism**, and **OpenSearch Dashboards'** `agent_traces` plugin all ignore events for GenAI purposes too (Phoenix special-cases only `exception` events; OpenSearch Dashboards' PPL transformer even fetches `events` off the index response and then never reads the field it fetched — a plumbed-but-dead path, the same shape of bug as `HANDLED_KEYS` drift, just at the events layer instead of the attributes layer).

**jaeger-ui already has the events plumbing PR #4191 would need.** `IOtelSpan.events: IEvent[]` (`packages/jaeger-ui/src/types/otel.ts:80`) is populated for every span — `OtelSpanFacade` derives it from the legacy `Span.logs` field (`packages/jaeger-ui/src/model/OtelSpanFacade.ts:56-60`), and `SpanDetail/index.tsx:114-129` already renders it via `AccordionEvents`. `genAiData.ts` simply never looks at it. Practical consequence: a span produced by any instrumentation using the events-based GenAI convention (OpenLIT/Strands today; anything else on the older semconv) would show `GenAITab`'s empty state — "No GenAI-specific attributes found on this span" — despite the span carrying real prompt/completion/message content, visible only if the user separately opens the generic Events accordion and reads it as raw JSON.

**Implication for the registry design above:** `get()` should not be assumed to resolve only against span attributes. Because ownership in this design is implicit — derived entirely from which keys a `build` function calls `get()` on, not from a declared list — folding events in only means changing what `get()` reads from; no individual registry entry needs to change. Two ways to do it, following the two real precedents found above:

1. **Langfuse's way — merge into one lookup surface.** Before building `remaining`, also scan `span.events` for names in a known GenAI-event set, flatten each matching event's `attributes[]` into the same `remaining` map (namespaced or not, depending on whether collisions with span-level attributes are possible), then run the same single pass. Simplest change to the design above — `get`/`remaining` stay a single map, registry entries don't need to know whether a value came from an attribute or an event.
2. **agent-health's way — event source ranked ahead of attribute source.** Keep two maps (attributes, events-flattened-by-name) and have `get(key)` check the event map first, falling back to attributes — matching that a span emitting both is more likely to have put authoritative content in the event (the newer/older convention distinction), not the reverse.

Given jaeger-ui's spans always carry an `events` array already normalized by `OtelSpanFacade` regardless of whether the underlying data was OTel-native or converted from legacy Jaeger logs, option 1 is the better fit here: fold matching `gen_ai.*` events into the same map the attribute registry entries already read from, so there's exactly one code path per field rather than two parallel ones to keep in sync.

swetalin-10:
Reworked the extraction per your architecture review — pushed in 3322434a.

`extractGenAiData()` + `HANDLED_KEYS` + `findAttribute()` are gone, replaced with `extractGenAiSections()`: one `Map` built once from the span's attributes, a `REGISTRY` of 4 builder functions (meta/tokens/conversation/toolCall) that each read whichever keys they need through a shared `get()` that removes a key the moment it's read, and whatever's left unclaimed afterward becomes the `other` section automatically. No second list to keep in sync with what's actually extracted.

This fixes the `toolId` bug directly — it's now just one of the keys the `toolCall` builder reads, so a span with only `gen_ai.tool.call.id` set produces a `toolCall` section instead of disappearing. Added a test for exactly that case.

Also fixed, per the review:
- `gen_ai.request.model`/`gen_ai.response.model` now show both when they differ (`"gpt-4-0613 (requested: gpt-4)"`) instead of falling back between them, since both are current per spec and record different things
- `parseSystemInstructions` now handles a single already-parsed instruction object that isn't array-wrapped, instead of silently returning `undefined` for it (this was the bug mentioned in passing near the end of the design-sketch section, not a separately filed comment — found it on a full re-read)
- Documented the fallback rationale inline at each `get(current) ?? get(legacy)` site
- `index.tsx` now dispatches on `section.type` via `switch`, with a `default` case that generically dumps an unrecognized section's fields via `JsonBlock` instead of rendering nothing — added a test that mocks the extractor to force that path, since it's unreachable through normal use today
- `TokensRow` now loops over `Object.keys(usage)` instead of 5 hardcoded blocks

58/58 tests passing on the GenAITab suite, 134/134 across the wider `SpanDetail` tree. Clean `tsc-lint`/`oxlint`.

Span events are still out of scope here per your note — tracked as a follow-up, not attempted in this PR.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/genAiData.ts
swetalin-10: Documented at each fallback site in the new registry builders — e.g. `gen_ai.provider.name`/`gen_ai.system` is a genuine current/deprecated pair (comment explains the short-circuit), while `gen_ai.request.model`/`gen_ai.response.model` is explicitly NOT one (comment explains why both are shown instead). Pushed in 3322434a.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/genAiData.ts
swetalin-10: Fixed — replaced the ~15 findAttribute() linear scans with a single Map built once, read via a shared get() accessor. Part of the registry rework in 3322434a.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/genAiData.ts
swetalin-10: Not intentional — real bug, thanks for catching it. Fixed in 3322434a: toolId is now just one of the keys the toolCall builder reads via get(), so a span with only gen_ai.tool.call.id set produces a toolCall section instead of disappearing. Added a dedicated test for this exact case.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/genAiData.ts
```diff
@@ -0,0 +1,316 @@
...
+  if (typeof value === 'string') {
+    try {
+      parsed = JSON.parse(value);
+    } catch {
+      // Not JSON - treat the whole string as a single message with no role.
+      return [{ role: undefined, content: value }];
+    }
+  }
+  if (!Array.isArray(parsed)) return [];
+  return parsed.map((entry): GenAiMessage => {
```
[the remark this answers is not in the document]
swetalin-10: Fixed in a631461e — wrapped a non-array parsed value in a one-element array before mapping, same fix already applied to parseSystemInstructions for the same class of bug. Added a test for a single already-parsed message object.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/genAiData.ts
```diff
@@ -0,0 +1,316 @@
...
+
+  const sections = REGISTRY.map(build => build(get)).filter((s): s is GenAiSection => s !== undefined);
+
+  const other = [...remaining]
+    .filter(([k]) => k.startsWith('gen_ai.'))
+    .map(([key, value]) => ({ key, value }));
+  if (other.length) sections.push({ type: 'other', data: { attributes: other } });
+
+  return sections;
+}
```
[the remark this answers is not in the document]
swetalin-10: Good catch — real bug. Fixed in a631461e: claimed attributes are now tracked by index (a Set<number>) rather than a Map keyed by attribute name, so a repeated gen_ai.* key's earlier occurrence is no longer silently overwritten before any builder runs. Added a test with a duplicate key confirming the first occurrence is claimed and the second still surfaces under "Other GenAI Attributes".

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/genAiData.test.ts
```diff
@@ -0,0 +1,523 @@
...
+function section<T extends keyof SectionDataMap>(
+  sections: GenAiSection[],
+  type: T
+): SectionDataMap[T] | undefined {
+  return sections.find(s => s.type === type)?.data as SectionDataMap[T] | undefined;
+}
+
+describe('extractGenAiSections', () => {
+  describe('meta section (provider/model)', () => {
+    it('prefers current provider/model attribute names over deprecated ones when they agree', () => {
```
yurishkuro:
```suggestion
    it('prefers current provider attribute names over deprecated one when they disagree', () => {
```
there is no test for model here

swetalin-10:
One more fix, caught on my own re-check (1c65601a): the previous commit's fix for the duplicate-key bug used \`attributes.findIndex()\` inside \`get()\`, which re-scans the whole array on every call — that's exactly the "linearly scanned N times" problem from the original review, reintroduced while fixing the duplicate-key issue. Rebuilt it as a \`Map<key, index[]>\` built once up front, so \`get()\` is back to O(1) per call while still correctly preserving repeated keys. Existing duplicate-key test still passes unchanged, confirming behavior didn't shift.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.tsx
```diff
@@ -1,12 +1,222 @@
...
+        />
+      ))}
+    </div>
+  );
+}
+
+function ToolCallSection({
+  toolCall,
+}: {
+  toolCall: { id?: string; name?: string; arguments?: unknown; result?: unknown };
```
yurishkuro: * id is never used
* it's better to `import type GenAiToolCall`

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.tsx
```diff
@@ -1,12 +1,222 @@
...
+          key={`output-${i}`}
+          message={{ role: message.role || 'assistant', content: message.content }}
+          index={inputMessages.length + i}
+        />
+      ))}
+    </div>
+  );
+}
+
+function ToolCallSection({
```
yurishkuro: `function ToolCallSection({ id, name, arguments: args, result }: GenAiToolCall)`

this would've caught unused ID in the linter.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.tsx
```diff
@@ -1,12 +1,222 @@
...
+  toolCall,
+}: {
+  toolCall: { id?: string; name?: string; arguments?: unknown; result?: unknown };
+}) {
+  return (
+    <div className="GenAITab--section">
+      <h3 className="GenAITab--sectionTitle">Tool Call{toolCall.name ? `: ${toolCall.name}` : ''}</h3>
```
yurishkuro: can this not be simplified to `toolCall.name || ''`?

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/genAiData.test.ts
swetalin-10: Applied, thanks — updated in ae795058. You're right that the title over-claimed model coverage that wasn't actually there.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.tsx
swetalin-10: Exported GenAiToolCall and switched ToolCallSection to your suggested destructured signature in ae795058 — that's exactly what surfaced id was unused. Now actually rendering it as an ID row above Arguments/Result.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.tsx
swetalin-10: Applied verbatim in ae795058.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.tsx
swetalin-10: Not applied literally as name || '' — that would drop the ': ' separator whenever a name is present ("Tool Callget_weather" instead of "Tool Call: get_weather"). Used name && `: ${name}` instead in ae795058, which renders identically to the original ternary in both branches but drops the redundant explicit ': '' alternative. Let me know if you had something else in mind.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/genAiData.ts
```diff
@@ -0,0 +1,336 @@
...
+      return value;
+    }
+  }
+  const parts = Array.isArray(parsed) ? parsed : [parsed];
+  const text = parts
+    .map(part =>
+      typeof part === 'object' && part !== null ? (part as Record<string, unknown>).content : part
+    )
+    .filter(content => content != null)
+    .map(content => (typeof content === 'string' ? content : JSON.stringify(content)))
```
yurishkuro:
```suggestion
    .map(stringifyValue)
```
to protect against malformed data

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/genAiData.ts
```diff
@@ -0,0 +1,336 @@
...
+ * so the output isn't double-encoded (`"{\"city\":\"Paris\"}"` instead of
+ * `{"city":"Paris"}`); non-JSON strings are kept as-is. Mirrors JsonBlock's
+ * handling of the same shape for the top-level gen_ai.tool.call.* attributes.
+ */
+function stringifyToolValue(value: unknown): string {
+  if (typeof value === 'string') {
+    const parsed = tryParseJson(value);
+    return parsed === value ? value : JSON.stringify(parsed);
+  }
+  return JSON.stringify(value ?? {});
```
yurishkuro: why not use stringifyValue here for safety?

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/genAiData.ts
swetalin-10: Applied verbatim in 4bcdfa04. Had to move stringifyValue's definition above parseSystemInstructions to clear a resulting use-before-define lint warning.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/genAiData.ts
swetalin-10: Good point on the safety, but not routed through stringifyValue directly here — it pretty-prints with 2-space indentation, which would turn the single-line `→ name(args)` tool-call format into multi-line output (e.g. the existing test expecting `→ get_weather({"city":"Paris"})` compact). Added the same try/catch safety (falls back to String(value) on a throw) without the formatting change, in 4bcdfa04. Added a test with a genuinely circular argument object confirming the fallback engages instead of throwing. Let me know if you'd rather have the pretty-printed format here regardless.

yurishkuro (approved):

PR #4192, opened 2026-07-05
fix(trace): fix view selector arrow key navigation

### LINKED ISSUE #4178: [Bug]: Arrow keys in Trace view selector dropdown resize minimap instead of navigating options.
### What happened?

On the Trace page, when the view selector dropdown ("Trace Timeline",  "Trace Graph", etc.) is open, pressing Up/Down arrow keys does not navigate  between the dropdown options. Instead, the arrow key events fall through to the SpanGraph minimap component behind the dropdown, causing the visible  time window to resize unexpectedly. The only way to navigate between dropdown options via keyboard is to press Tab repeatedly through each item individually, which is non-standard and significantly slower than the expected arrow key navigation.

### Steps to reproduce

2. Press Tab until the "Trace Timeline" button (top right) receives focus.
3. Press Enter or Space to open the dropdown.
4. Press the Down arrow key.
5. Expected: focus moves to the next option (e.g. "Trace Graph").
   Actual: the minimap timeline scrubber resizes instead - a completely unrelated component is affected.
6. To navigate between options at all, the user must press Tab again to move through each option individually - there is no way to use arrow keys to navigate the open dropdown, forcing an inefficient and non-standard keyboard interaction.

### Expected behavior

Per WAI-ARIA authoring practices for dropdown menus:
- Down/Up arrow keys should navigate between options
- Enter should select the focused option
- Escape should close without selecting

### Relevant log output

```shell

```

### Screenshot

Image

## Root cause

`AltViewOptions.tsx` uses Ant Design's `Dropdown` + `Button` combination.
Unlike Ant Design's `Select` component, `Dropdown` does not:
1. Intercept arrow key events for internal menu navigation
2. Stop those events from propagating to components underneath

The `SpanGraph` minimap listens for arrow key events to control the visible time window - since `Dropdown` doesn't stop propagation, those events reach `SpanGraph` and resize the minimap instead of navigating the dropdown.

## Proposed solution

Replace `Dropdown` + `Button` with Ant Design's `Select` component, which natively supports full keyboard navigation (arrow keys, Enter, Escape) and correctly stops event propagation.

This is consistent with the rest of the codebase - the Search page's Service, Operation, and Lookback fields already use `Select` (via a `SearchableSelect` wrapper) and work correctly with keyboard navigation. `AltViewOptions` is the only place that uses `Dropdown` for this kind of view-switching UI.

### Jaeger backend version

_No response_

### SDK

_No response_

### Pipeline

_No response_

### Stogage backend

_No response_

### Operating system

_No response_

### Deployment model

_No response_

### Deployment configs

```shell

```

69iqqq:

* Fixes #4178

## Description of the changes

* Added a check in combokeyHandler (TracePage/index.tsx) to ignore arrow key events when the focus is inside an interactive overlay (dropdown, popover, or modal using .closest([role="menu"])). This prevents tab from changing the plane when the drop down menu is open
* Added an onOpenChange handler to theAltViewOptions dropdn to focus the first menu item when it opens. This lets users use the Up/Down arrows right away without needing to press Tab first.

* Manually tested that the view selector dropdn can be navigated using the Up/Down arrow keys without cahnging the pane and also not increasing or dcing the cnt
* Ran npm test, npm run lint, and npm run build successfully.

## Attachment

https://github.com/user-attachments/assets/07238d4f-a95d-45a9-87ca-0fbc9d0dc442

## Checklist

* [ ] I have added unit tests for the new functionality *(N/A, this fix mainly relies on DOM focus handling and event targets.)*
* [x] I have run lint and test successfully.

* [ ] **None**
* [x ] **Light** to understand the code base faster
* [ ] **Moderate**
* [ ] **Heavy**

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/index.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/index.tsx
@@ -134,6 +134,10 @@ export const shortcutConfig: { [name: string]: [number, number] } = {
 export function makeShortcutCallbacks(adjRange: (start: number, end: number) => void): ShortcutCallbacks {
   function getHandler([startChange, endChange]: [number, number]): CombokeysHandler {
     return function combokeyHandler(event: React.KeyboardEvent<HTMLElement>) {
+      const target = event.target as HTMLElement;
+      if (target?.closest?.('[role="menu"], .ant-dropdown, .ant-popover, .ant-modal')) {
+        return;
+      }
       event.preventDefault();
       adjRange(startChange, endChange);
     };
--- a/packages/jaeger-ui/src/components/TracePage/TracePageHeader/AltViewOptions.test.jsx
+++ b/packages/jaeger-ui/src/components/TracePage/TracePageHeader/AltViewOptions.test.jsx
@@ -13,8 +13,13 @@ vi.mock('antd', async () => {
   const originalModule = await vi.importActual('antd');
   return {
     ...originalModule,
-    Dropdown: ({ children, menu }) => (
-      <div data-testid="dropdown">
+    Dropdown: ({ children, menu, onOpenChange }) => (
+      <div
+        data-testid="dropdown"
+        onClick={() => {
+          if (onOpenChange) onOpenChange(true);
+        }}
+      >
         {children}
         <div data-testid="dropdown-menu">
           {menu.items.map(item => (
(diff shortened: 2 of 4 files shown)


parshipcy (commented):
Always add screenshots or screencasts in the PR description for ui changes

packages/jaeger-ui/src/components/TracePage/index.tsx
```diff
@@ -134,6 +134,10 @@ export const shortcutConfig: { [name: string]: [number, number] } = {
 export function makeShortcutCallbacks(adjRange: (start: number, end: number) => void): ShortcutCallbacks {
   function getHandler([startChange, endChange]: [number, number]): CombokeysHandler {
     return function combokeyHandler(event: React.KeyboardEvent<HTMLElement>) {
+      const target = event.target as HTMLElement;
+      if (target?.closest?.('[role="menu"], .ant-dropdown, .ant-popover, .ant-modal')) {
+        return;
+      }
```
parshipcy, line 140: please add a test for this guard

69iqqq:
> Always add screenshots or screencasts in the PR description for ui changes
my bad, I've now added screencast to the PR description

packages/jaeger-ui/src/components/TracePage/index.tsx
69iqqq: i've added the test. could you please check it ?

69iqqq: Added tests to cover all  changes introduced in this PR, once it passes all the test , i will mark
* [ ] I have added unit tests for the new functionality *(N/A, this fix mainly relies on DOM focus handling and event targets.)* in my pr

parshipcy:
@yurishkuro does this approach looks good to you, or should we use `select` instead?

PR #4194, opened 2026-07-07
fix(ListView): Remove array-length pre-allocation that created holes

dolliecoder:
## Which problem is this PR solving?
- part of #4189 

## Description of the changes
- Removed `items.length = end - start + 1` from `ListView.render()` which created N empty holes in the items array, followed by `items.push(...)` appending N real elements producing a `length 2N` array with N leading holes that broke React's virtualized DOM key mapping.
- Added a regression test asserting rendered item count equals the drawn range (`end - start + 1`) so this cannot silently regress.

## How was this change tested?
- Ran full TraceTimelineViewer test suite: 33 test files, 556 tests  all passing.
- Verified that the regression test catches the old bug: `items.length` now correctly equals `end - start + 1` (was 2N with N leading holes).
- `npm test`: all trace-timeline tests pass.
- `npm run tsc-lint`: no new type errors.
- `npm run fmt`: formatting compliant.

## Checklist
- [x] I have read https://github.com/jaegertracing/jaeger/blob/main/CONTRIBUTING_GUIDELINES.md
- [x] I have signed all commits
- [x] I have added unit tests for the new functionality
- [x] I have run lint and test steps successfully: `make lint test`

## AI Usage in this PR (choose one)
See [AI Usage Policy](https://github.com/jaegertracing/jaeger/blob/main/CONTRIBUTING_GUIDELINES.md#ai-usage-policy).
- [ ] **None**: No AI tools were used in creating this PR
- [x] **Light**: AI provided minor assistance (formatting, simple suggestions)
- [ ] **Moderate**: AI helped with code generation or debugging specific parts
- [ ] **Heavy**: AI generated most or all of the code changes

Issue #4193, opened 2026-07-05
[Bug]: perf(VirtualizedTraceView): mapSpanIndexToRowIndex is O(n) per keyboard navigation call


bhavyamsharmaa:
### What happened?

`mapSpanIndexToRowIndex` in `VirtualizedTraceView.tsx` searches through the entire row state array every time it's called.

```ts
const mapSpanIndexToRowIndex = useCallback(
  (index: number) => {
    const rows = getRowStates();
    const max = rows.length;
    for (let i = 0; i < max; i++) {
      const { spanIndex } = rows[i];
      if (spanIndex === index) return i;
    }
    throw new Error(`unable to find row for span index: ${index}`);
  },
  [getRowStates]
);
```

This function is used by `ScrollManager` when navigating between visible spans with keyboard shortcuts. Since it scans the entire row list every time, the lookup becomes more expensive as the trace grows. With large traces (around 80k spans or more), repeated keyboard navigation can cause noticeable UI lag.

`generateRowStatesFromTrace` is already memoized using `memoizeOne`, so the row states are reused. However, the reverse lookup from `spanIndex` to `rowIndex` still scans the array on every call.

### Steps to reproduce

1. Open a trace with a large number of spans (for example, 80k or more).
2. Use the keyboard shortcuts to move to the next or previous visible span repeatedly.
3. Notice the delay while navigating through the trace.

### Expected behavior

Looking up a `rowIndex` from a `spanIndex` should be a constant-time operation instead of scanning the entire row array.

One possible approach is:

- Build a `Map<spanIndex, rowIndex>` while generating the row states.
- Store that map together with the generated row state data.
- Replace the loop in `mapSpanIndexToRowIndex` with a direct map lookup.

This avoids repeatedly scanning the same data and should keep keyboard navigation responsive even for very large traces.

### Relevant files

- `packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx`
- `packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/generateRowStates.ts`

Related: #4189

Anas-wasti:
Hi @bhavyamsharmaa,

I came across this issue while looking into the performance problems with large traces. The repeated scan in `mapSpanIndexToRowIndex` makes sense as a potential bottleneck, especially during keyboard navigation on large traces.

[...]

Issue #4189, opened 2026-07-05
[Bug]: UI freezes and crashes when interacting with large traces (e.g. 80k spans)


dolliecoder:
### What happened?

As a Jaeger user troubleshooting complex requests, I want to be able to load and interact with large traces (e.g. 80k spans) in the Trace Timeline so that I can analyze them without the UI locking up or crashing.

Currently, when loading a massive trace and trying to interact with it (like collapsing or expanding the span tree), the UI completely locks up and crashes. This is caused by an $O(N)$ object allocation bottleneck in generateRowStates.ts. On every interaction, the application iterates through all 80k spans and instantiates a fully materialized RowState object for every single visible row. This allocates tens of thousands of objects on the heap, triggering massive Garbage Collection (GC) churn and blocking the main thread.

### Steps to reproduce

1. Load a large trace (e.g. ~80,000 spans) into the Jaeger UI.
2. Attempt to interact with the trace timeline (e.g., collapse a parent span tree).
3. Observe the UI completely freeze and eventually crash due to the main thread being blocked by GC and object allocation.

### Expected behavior

The UI should remain responsive when interacting with large traces. The timeline should smoothly collapse and expand without locking up the browser.

### Relevant log output

```shell

```

### Screenshot

https://github.com/user-attachments/assets/cbb871bb-6f7b-496b-9715-887a8b4dd46a

### Additional context

here is the proposed architectural approach we plan to take to fix this via a PR:

1. Modernize ListView.tsx: Upgrade it from an older Class Component to a React Functional Component using hooks (useRef, useEffect), keeping the props API identical.
2. Virtualize Row State Generation: Instead of allocating full RowState objects for all 80k rows in generateRowStates.ts, we'll do a fast pass to populate packed primitive arrays (number[] for span indices and row types). V8 highly optimizes arrays of primitives, so this avoids object allocation entirely.
3. Lazy Materialization: generateRowStates will return a lightweight view object. VirtualizedTraceView.tsx will only call .getRow(index) to construct the RowState on-the-fly for the ~50 rows that ListView actually renders.
This reduces the generateRowStates memory allocation from $O(N)$ objects down to $O(1)$ objects, completely eliminating the GC churn and UI freeze.

### Operating system

Tested on Windows, but impacts all OS browsers

### Deployment configs

```shell

```

dolliecoder:
I am currently working on this and will open a PR once the maintainers agree with the proposed approach

[...]

PR #4195, opened 2026-07-07
fix(trace): Replace recursive traversals with iterative to prevent stack overflow on deep traces

dolliecoder:
- part of resolving #4189 (together with #4193 and #4194)

## Description of the changes
- Recursive tree traversals in three functions overflow the JS call stack (~10k frame limit) when processing traces with deep nesting up to 72k levels observed in an 80k-span trace.
**Changes:**
1. **`processSpan` in `transform-trace-data.ts`**  Converted from recursion to iterative stack-based DFS. This is the most critical fix: without it, deeply nested traces crash before rendering.
2. **`createCPSpanMap` in `cpspan.ts`**  Converted to iterative with explicit stack and `blockingChildren[]` arrays, preserving bottom-up post-order traversal.
3. **`findAllDescendants` and `collectFromSubtree` in `criticalPath.ts`** Both converted from recursive to iterative stack-based DFS.
All iterative versions use a while-loop with explicit stack, pushing children in reverse to preserve left-to-right traversal order. The output is identical to the recursive versions. Time complexity remains O(n).

- **Unit tests**: All 593 tests in the affected test suites pass (17 transform-trace-data tests, 21 CriticalPath tests, 555 TraceTimelineViewer tests)
- **Full test suite**: 3344/3345 tests pass across the entire jaeger-ui package. The single failure is a pre-existing Monitor timeout test unrelated to this change
- **Manual verification**: Loaded and rendered an 80,000-span trace with 72,000 levels of nesting. The trace timeline displays correctly, critical path lines render, and there are no console errors
- **Correctness**: Each iterative function was compared against its recursive predecessor to verify the same tree traversal order is produced

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/CriticalPath/utils/cpspan.test.ts
+++ b/packages/jaeger-ui/src/components/TracePage/CriticalPath/utils/cpspan.test.ts
@@ -1,0 +1,24 @@
+// Copyright (c) 2026 The Jaeger Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+import { createCPSpanMap } from './cpspan';
+
+describe('createCPSpanMap', () => {
+  it('should not overflow stack on a chain of 15000 blocking spans', () => {
+    const spans: any[] = [];
+    for (let i = 0; i < 15000; i++) {
+      spans.push({
+        spanID: `s${i}`,
+        kind: 'INTERNAL',
+        startTime: 0,
+        endTime: 1,
+        duration: 1,
+        childSpans: [],
+        parentSpanID: i === 0 ? undefined : `s${i - 1}`,
+      });
+    }
+    for (let i = 0; i < 14999; i++) spans[i].childSpans = [spans[i + 1]];
+    for (let i = 1; i < 15000; i++) spans[i].parentSpan = spans[i - 1];
+    expect(() => createCPSpanMap(spans[0])).not.toThrow();
+  });
+});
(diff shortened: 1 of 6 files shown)


dolliecoder:
@yurishkuro @jkowall PTAL

jkowall (changes requested):
Thanks for addressing the recursive traversal paths. The implementation looks directionally correct, but I think this needs a focused regression test before merge.

Please add a test that builds a deeply nested parent-child chain large enough to fail with the previous recursive implementation, then verifies the iterative paths complete without a call stack overflow. At minimum, the test should cover `transformTraceData`; ideally it should also exercise the critical-path helper paths touched here (`createCPSpanMap` and the collapsed/pruned critical-path traversal). A smaller synthetic depth than the real 72k-span trace is fine if it reliably exceeds the JS call stack limit in the old implementation.

The reason I am asking for this is that the PR is specifically a stack-safety fix. Existing behavior tests can pass while a future refactor accidentally reintroduces recursion on these hot paths, and this is exactly the kind of large-trace failure Jaeger UI needs to guard explicitly.

dolliecoder:
@jkowall Done pushed commit e2b89e78 with 3 stack-safety regression tests covering transformTraceData , createCPSpanMap and collapsed critical-path traversal all passing with depths of 10k–15k spans.

jkowall:
Thanks, I re-ran the added tests against the pre-fix recursive implementation.

The new `transformTraceData` and `createCPSpanMap` tests are effective: both fail with `RangeError: Maximum call stack size exceeded` on the old code and pass on `e2b89e78`.

The collapsed critical-path test at 10,000 spans is not yet a regression test on Node 24. It still passes against the old recursive `findAllDescendants` implementation. Increasing the depth to 15,000 makes the old code fail with a stack overflow while the current iterative implementation passes.

Please raise that fixture to at least 15,000, preferably using a shared `depth` constant for all derived values and asserting that the returned sections are complete. I am leaving the requested-changes review in place until this test reliably detects the regression.

dolliecoder:
@jkowall raised depth to 15,000 with a shared (DEPTH) constant and completeness assertion all three tests now fail on the old recursive code and pass on fa2bc5e4

PR #4199, opened 2026-07-08
fix(FileLoader): stop reporting OTLP backend-transform failures as parse errors

### LINKED ISSUE #4197: [Bug]: Importing an OTLP/JSON trace file silently requires a live Jaeger backend, and failure is reported as a JSON parse error
## What's happening

The "Click or drag files to this area" JSON importer on the Search page (`FileLoader.tsx`) tells you it supports "JSON files containing one or more traces." For a legacy Jaeger trace JSON, that's true end-to-end, entirely in the browser - no network call involved.

For an OTLP/JSON trace file, it's a different story. `readJsonFile.ts` detects the `resourceSpans` shape and, instead of parsing it locally, ships the whole payload to a backend endpoint:

```ts
// packages/jaeger-ui/src/utils/readJsonFile.ts
if ('resourceSpans' in traceObj) {
  JaegerAPI.transformOTLP(traceObj)
    .then((result: string) => resolve(result))
    .catch((err: unknown) => {
      const cause = err instanceof Error ? `: ${err.message}` : '';
      reject(new Error(`Error converting OTLP trace to Jaeger${cause}`, { cause: err }));
    });
} else {
  resolve(traceObj);
}
```

```ts
// packages/jaeger-ui/src/api/jaeger.ts
transformOTLP(traces: any): Promise<any> {
  return getJSON(`${this.apiRoot}transform`, { method: 'POST', body: JSON.stringify(traces) });
},
```

There is no client-side OTLP → legacy-span conversion anywhere in this codebase - I checked (`grep -rn resourceSpans packages/jaeger-ui/src` only turns up this one call site and the generated v3 API client, which is unrelated). If `POST /api/transform` isn't reachable - no query service running, an older backend that predates this endpoint, a proxy that only forwards `/api/traces`, or you're just poking at a static build with no backend behind it at all - the OTLP path has nothing to fall back to. The legacy-JSON path never needed a backend in the first place, so this asymmetry is easy to miss until you hit it.

---

## What the user sees vs. what's actually happening

```
Drop a legacy Jaeger trace.json:

+------------------------------------------+
| my-trace.json                            |
|                                          |
| Click or drag files to this area.        |
| JSON files containing one or more traces |
| are supported.                           |
+------------------------------------------+
              |
              | 100% client-side (JSON.parse only)
              v
     Trace loads. No network request fired.


Drop an OTLP/JSON trace file (same UI, same drop zone):

+------------------------------------------+
| otel-file-exporter-output.json           |
|                                          |
| Click or drag files to this area.        |
| JSON files containing one or more traces |
| are supported.                           |
+------------------------------------------+
              |
              | POST /api/transform  (silently required)
              v
     backend unreachable
              |
              v
     Toast: "Failed to parse otel-file-exporter-output.json:
             Error converting OTLP trace to Jaeger: Failed to fetch"
```

The file's JSON was completely valid - it never failed to *parse* anything. The message says "Failed to parse" because `FileLoader.tsx`'s catch handler doesn't distinguish "your JSON is malformed" from "the network call your JSON quietly triggered didn't work":

```ts
// packages/jaeger-ui/src/components/SearchTracePage/FileLoader.tsx
.catch((err: unknown) => {
  message.error(
    `Failed to parse ${file.name}: ${err instanceof Error ? err.message : String(err)}`
  );
});
```

Nothing in the UI ever tells the user that dropping *this particular* file format needs a live backend while every other file it accepts doesn't.

---

## Why this is a real day-to-day problem, not an edge case

OTLP/JSON isn't some obscure format jaeger-ui has to accept begrudgingly - it's the standard output of the OTel Collector's File Exporter, and it's exactly what you'd have on hand in the situations where a lightweight, backend-free viewer matters most:

- **Air-gapped / restricted environments** - pulling a trace file out to inspect it on a laptop with no route to the cluster's query service.
- **Bug reports and support tickets** - a colleague or customer attaches a `trace.json` they exported; you drag it into your local jaeger-ui to look at it, with no backend pointed at *their* data.
- **CI artifacts** - a failing integration test dumps an OTLP trace as a build artifact; someone downloads it and drops it into jaeger-ui running locally, disconnected from that CI's tracing backend.
- **Demo / static-hosting deployments** - any jaeger-ui instance that isn't wired to a full query service (there's precedent for this exact scenario in #3541, "Improve dev mode by loading backend settings dynamically").

In every one of these, the person doing the debugging already has exactly the file they need, open in front of them - and the tool tells them it "failed to parse" it.

---

## What a fix looks like

A couple of options, not mutually exclusive:

1. **Give OTLP import a client-side path**, mirroring the direction the reverse conversion already went in `utils/span-to-otlp.ts` (span → OTLP). A small `otlp-to-legacy.ts` covering the common cases (attributes, events, links, status, span/trace IDs) would let `readJsonFile` parse OTLP fully offline, falling back to the backend only if it hits something it can't map.
2. **At minimum, stop mislabeling the failure.** `readJsonFile` already knows it's on the OTLP branch when the transform call fails - it should raise something like `"OTLP import requires a reachable Jaeger backend (POST /api/transform) - this file's JSON itself is valid"` instead of letting `FileLoader` wrap it in "Failed to parse."

## Acceptance criteria

- [ ] Dropping a valid OTLP/JSON file with no backend reachable produces an error that clearly says the backend transform failed - not a generic "failed to parse" message
- [ ] The error is specific enough that a user unfamiliar with the codebase understands *why* (backend-dependent conversion), not just *that* it failed
- [ ] If a client-side conversion path is added: legacy JSON and OTLP JSON both load fully offline; the backend call is only a fallback for shapes the client parser doesn't recognize
- [ ] Unit test in `readJsonFile.test.ts` / `FileLoader.test.tsx` covering "OTLP file, transform endpoint unreachable" and asserting the surfaced message is accurate

## Why this matters

The whole point of a drag-and-drop JSON importer is that it works without asking anything of your environment - that's the promise the UI text makes, and it's true for one of the two formats it claims to accept. For the other, the tool silently phones home, and when that call fails, it blames the file instead of the network. Someone hits this at exactly the moment they're already stuck (no backend, need to look at a trace *right now*), and the error message points them in the wrong direction to boot.

bhuvan-somisetty:
- Closes #4197

## Description of the changes
- Dropping an OTLP/JSON trace file into the Search page importer routes it through `JaegerAPI.transformOTLP` (`POST /api/transform`), since there's no client-side OTLP-to-legacy conversion in this codebase. When that call fails — no query service reachable, an older backend without the endpoint, a static build with nothing behind it — `readJsonFile` wrapped the failure in a generic `Error`, and `FileLoader`'s catch handler prefixed it with `"Failed to parse <file>"`. The file's JSON was completely valid the whole time; the failure happens afterward, in a network call the user never sees.
- Added an `OtlpTransformError` class exported from `readJsonFile.ts` so `FileLoader` can tell this failure apart from an actual parse error and report it accurately — the file name plus a message that says the OTLP conversion needs a reachable backend, instead of blaming the file's JSON.
- This is the minimal fix from the issue (making the error message accurate). Adding a full client-side OTLP → legacy conversion path so OTLP import works fully offline is a larger, separate change and is left as follow-up work per the issue's acceptance criteria.

- Updated the existing `readJsonFile.test.js` case that asserted the old message text, and added a new case asserting the rejection is an `OtlpTransformError` instance.
- Added a `FileLoader.test.jsx` case asserting the toast shown for an OTLP transform failure names the file and explains the backend requirement, instead of the old "Failed to parse" wording.
- `npx vitest run src/utils/readJsonFile.test.js src/components/SearchTracePage/FileLoader.test.jsx` — all 27 tests pass.
- `npx vitest run src/components/SearchTracePage/ src/utils/readJsonFile.test.js` — full directory, 293 tests pass.
- `tsc --noEmit` and `vp fmt --check` clean on all four changed files.

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/SearchTracePage/FileLoader.test.jsx
+++ b/packages/jaeger-ui/src/components/SearchTracePage/FileLoader.test.jsx
@@ -5,9 +5,13 @@ import React from 'react';
 import { render, screen, act } from '@testing-library/react';
 import '@testing-library/jest-dom';
 
-vi.mock('../../utils/readJsonFile', () => ({
-  default: jest.fn(),
-}));
+vi.mock('../../utils/readJsonFile', async () => {
+  const actual = await vi.importActual('../../utils/readJsonFile');
+  return {
+    ...actual,
+    default: jest.fn(),
+  };
+});
 
 vi.mock('../../model/transform-trace-data', () => ({
   default: jest.fn(),
@@ -48,7 +52,7 @@ vi.mock('antd', async () => {
 });
 
 import FileLoader, { loadTracesFromFile } from './FileLoader';
-import readJsonFile from '../../utils/readJsonFile';
+import readJsonFile, { OtlpTransformError } from '../../utils/readJsonFile';
 import transformTraceData from '../../model/transform-trace-data';
 import { traceToTraceSummary } from '../../model/trace-summary';
 import { populateTraceCache } from '../../hooks/useTraceLoading';
@@ -246,6 +250,25 @@ describe('<FileLoader />', () => {
     });
 
     expect(message.error).toHaveBeenCalledWith(expect.stringContaining('not valid JSON'));
+    expect(mockOnTracesLoaded).not.toHaveBeenCalled();
+  });
+
+  it('shows a backend-required message, not "Failed to parse", when OTLP transform fails', async () => {
+    const { message } = await import('antd');
+    readJsonFile.mockRejectedValue(
+      new OtlpTransformError(
+        'OTLP import requires a reachable Jaeger backend (POST /api/transform): Failed to fetch'
+      )
+    );
+
+    render(<FileLoader onTracesLoaded={mockOnTracesLoaded} />);
+    await act(async () => {
+      global.mockBeforeUpload(makeFile('otel-file-exporter-output.json'));
+    });
+
+    expect(message.error).toHaveBeenCalledWith(
+      'otel-file-exporter-output.json: OTLP import requires a reachable Jaeger backend (POST /api/transform): Failed to fetch'
+    );
     expect(mockOnTracesLoaded).not.toHaveBeenCalled();
   });
 
--- a/packages/jaeger-ui/src/components/SearchTracePage/FileLoader.tsx
+++ b/packages/jaeger-ui/src/components/SearchTracePage/FileLoader.tsx
@@ -5,7 +5,7 @@ import * as React from 'react';
 import { Upload, message } from 'antd';
 import { IoDocumentAttachOutline } from 'react-icons/io5';
 
-import readJsonFile from '../../utils/readJsonFile';
+import readJsonFile, { OtlpTransformError } from '../../utils/readJsonFile';
 import transformTraceData from '../../model/transform-trace-data';
 import { traceToTraceSummary } from '../../model/trace-summary';
 import { populateTraceCache } from '../../hooks/useTraceLoading';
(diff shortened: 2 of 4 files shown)


bhuvan-somisetty:
@yurishkuro and @jkowall Can u please review this pr.

swetalin-10 (approved):
Looks good to me from what I tested - checked this against a dead backend and it holds up in most of the scenarios I ran. Also tested it against #4337 (open for the same issue) - this one's right when the backend's down or missing the route, #4337's right when it's up but rejects the payload. Neither tells those two apart by status code — worth a follow-up here, or reconciling with #4337. Leaving the actual call to @yurishkuro.

PR #4213, opened 2026-07-15
feat(genai-tab): Add content-sensitive rendering toggle (plain/markdown/JSON) for message content

swetalin-10:

Part of jaegertracing/jaeger#8401 (GenAI observability). The GenAI tab (`SpanDetail/GenAITab`) renders `gen_ai.input.messages`/`gen_ai.output.messages` content as plain preformatted text, so model output containing Markdown (lists, code blocks, emphasis) displays as raw syntax instead of formatted text.

**Reworked per [review feedback](https://github.com/jaegertracing/jaeger-ui/pull/4213#discussion_r3591340586)**: this PR originally rendered all message content unconditionally as Markdown. That assumption doesn't hold — a GenAI text attribute has no declared format in the OTel GenAI semantic conventions (it can be prose, Markdown, JSON, YAML, or a template), and unconditional Markdown silently corrupts anything that isn't Markdown, including Jaeger's own flattened tool-call composite strings (`parseMessages`/`renderParts` joins prose with `→ name({...})` call markers, which Markdown then mangles).

## Description of the changes

- `GenAITab/index.tsx`: message content format is no longer assumed. Per message:
  1. **JSON detection first** — reuses the existing `tryParseJson` guard; content that parses as an object/array routes through the existing `JsonBlock` tree view (the same component already used for tool call arguments/results).
  2. **Default to plain text** otherwise — a `<pre>`, not Markdown.
  3. **Format dropdown** (Plain / Markdown / JSON) lets the user override the renderer per message.
  4. **Persisted per attribute name** — the chosen format is stored in `localStorage` keyed by `gen_ai.input.messages` / `gen_ai.output.messages` / `gen_ai.system_instructions`, so it's remembered and applies to every message from that attribute, not globally and not per-individual-message.
- The format-override state lives in `ConversationSection`, not in each `MessageBlock` instance, so changing the dropdown on one message immediately updates every other currently-rendered message sharing the same attribute — not just future mounts.
- A 150KB size gate skips Markdown parsing for oversized content, falling back to plain (avoids pathological reflow/parse cost on huge attributes).
- The opt-in Markdown path still reuses `sharedMarkdownOptions` (from #4181) unchanged — same safety properties (`disableParsingRawHTML`, link allowlist) as before.
- Each message's Format dropdown has a distinct accessible name (e.g. "Content format for message 2 (assistant)") instead of a shared generic label, so screen reader users can tell which message a control applies to in a multi-message conversation.

**Update per [latest review round](https://github.com/jaegertracing/jaeger-ui/pull/4213#pullrequestreview-4727375594):**

- Fixed a real rendering bug: a single-sentence message (no other Markdown formatting) compiles to one inline node, and `markdown-to-jsx`'s `forceWrapper` then wraps it in a bare `<span>` instead of a block element. Padding on an inline element only shows at the start/end of the whole run, not around each wrapped line, so the message showed an indented first line with no padding on the rest of the paragraph once it wrapped. Fixed by passing `forceBlock: true` alongside the existing shared Markdown options for this tab specifically, so every message always compiles to a `<p>` inside a real `<div>`, matching the box the CSS already assumes. Confirmed by rendering the actual `<Markdown>` output and inspecting the DOM: before the fix a one-sentence message produced `<span class="GenAITab--messageContent">…</span>`; after the fix it produces `<div class="GenAITab--messageContent"><p>…</p></div>`, the same shape multi-paragraph messages already used.
- The JSON option in the Format dropdown is now disabled (with a tooltip) on any message whose content doesn't actually parse as JSON, the same way the Markdown option is already disabled for oversized content. Since the format preference is stored per attribute name (not per message), a message can inherit a "JSON" preference from a sibling message under the same attribute even when its own content isn't JSON — that case now falls back to rendering plain text instead of silently showing "JSON" selected while nothing parses.
- `MessageFormat`, the storage-key prefix, and the size-limit constant moved out of `index.tsx` into a new `messageFormat.ts` (plain functions, not a hook), matching the existing `ThemeStorage.ts` / `ThemeProvider.tsx` pattern already used elsewhere in this codebase for localStorage-backed UI state.
- Removed the `screenshots/` directory from the branch; the PR description below no longer references committed screenshot files.

**Update per [follow-up discussion](https://github.com/jaegertracing/jaeger-ui/pull/4213#discussion_r3607409631):** attached the trace behind the earlier screenshots (trace ID `bde4caeb1c1a8d4c8b591f597d19d5af`, `jaeger: /api/ai/chat`, 13 spans) as a PR comment attachment — it also has a genuine JSON payload (`gen_ai.tool.definitions`, a JSON-encoded array of MCP tool schemas), which doubles as an answer to the "share a trace with JSON payload" ask.

That trace also gave a concrete example of a real usability problem in `AttributesTable.tsx` (the key:value table shared by the standard span "Details" tab, GenAITab's "Other GenAI Attributes", the span Events accordion, and the Trace Logs view): a JSON-tree value was squeezed into the narrow value column next to a fixed-width key column, forcing horizontal scrolling for anything but a small object. `AttributesTable.tsx`/`.css` now give a JSON-tree value its own full-width row instead — a key-only row followed by a full-width value row below it — while every other value shape (scalars, the existing header string-list case) is unchanged. Did not extend this to reusing the GenAITab format dropdown on Details-view values (flagged as an open follow-up question rather than building it unprompted, since it's a bigger UX call on shared, widely-used code).

## Test changes

- `index.test.tsx`: replaced the unconditional-Markdown tests with coverage for the new behavior — plain-text default, JSON auto-detection routing to the tree view, Markdown rendering once the dropdown is switched, format persistence per attribute name across remounts, live sync across sibling messages sharing an attribute when the format changes, isolation between different attributes, pretty-printed JSON with leading whitespace still routing to the tree view, the dropdown staying in sync (not showing "Markdown" selected) when content exceeds the size limit and falls back to plain, each message's format dropdown having a distinct accessible name, a single-sentence message rendering inside a real block element instead of a bare inline span, the JSON option being disabled on non-JSON content, and a message falling back to plain text (with the JSON option disabled) when it inherits a "JSON" attribute-level preference its own content doesn't match.
- `genAiData.test.ts`: added unit tests for `tryParseJson` directly — plain JSON, JSON with leading whitespace/newlines (both object and array), non-JSON strings, and malformed JSON-looking strings.
- `AttributesTable.test.jsx`: updated the row-count assertion for the new two-row JSON-tree layout, and added coverage for a JSON-tree value rendering on its own full-width row and a non-JSON-tree value staying unchanged on one row with its key.

- `npm test` — `GenAITab/index.test.tsx`: 26/26 passing, `GenAITab/genAiData.test.ts`: 55/55 passing, `AttributesTable.test.jsx`: 11/11 passing. Also ran every other consumer of `AttributesTable` (`AccordionAttributes.test.jsx`, `AccordionEvents.test.jsx`, `TraceLogsView/index.test.jsx`, `SpanDetail/index.test.jsx`, `markdownOptions.test.tsx`): all passing, no regressions. Full `TracePage` suite: 1175/1175 passing.
- `npx tsc -p packages/jaeger-ui/tsconfig.json` — no new type errors.
- `npm run lint` (oxlint + knip + license/copyright/tsx-naming checks) — clean on all changed files.
- Verified the Markdown wrapper fix directly against the reported trace by rendering the actual message content and inspecting the produced DOM before and after the fix.

## Checklist


packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.tsx
```diff
@@ -21,16 +23,16 @@ import './index.css';
...
-function MessageBlock({ message, index }: { message: GenAiMessage; index: number }) {
+function MessageBlock({ message }: { message: GenAiMessage }) {
   return (
     <div className={`GenAITab--message GenAITab--message-${message.role || 'unknown'}`}>
       <div className="GenAITab--messageHeader">
         <span className="GenAITab--messageRole">{message.role || 'message'}</span>
         <CopyIcon copyText={message.content} tooltipTitle="Copy message" buttonText="Copy" />
       </div>
-      <pre className="GenAITab--messageContent" data-testid={`genai-message-${index}`}>
+      <Markdown className="GenAITab--messageContent" options={sharedMarkdownOptions}>
```
yurishkuro: How do we know the content is a markdown? Do we select it purely based on the type of the message ("prompt", or "input", etc.)?

I think we need a more sensitive approach, such as:
  1. attempt to parse the string as JSON as the most detectable format
  1. if not then default to plain text
  1. support a Format dropdown to allow user to change the renderer
  1. store the selection (per attribute name) in local storage to remember user preference

swetalin-10: Fixed in 0aded5ae, following your 4 points exactly:

1. JSON parse attempted first (`tryParseJson` on the content string) - if it parses to an object/array, it routes to the existing `JsonBlock` tree view.
2. Otherwise defaults to plain `<pre>` text, not Markdown.
3. Added a per-message Format dropdown (Plain / Markdown / JSON) so the user can override the renderer.
4. The selection is persisted in localStorage keyed by attribute name (`gen_ai.input.messages` / `gen_ai.output.messages` / `gen_ai.system_instructions`), so it's remembered per attribute, not globally and not per-individual-message.

Also added a 150KB size gate that skips Markdown parsing even if selected (falls back to plain), and lifted the format-override state up to `ConversationSection` so switching the dropdown on one message immediately updates every other currently-rendered message from the same attribute, not just future mounts.

PR description updated to match. Tests rewritten (20/20 passing) to cover the new default-plain/JSON-detect/dropdown/persistence/live-sync/per-attribute-isolation behavior.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.tsx
```diff
@@ -58,6 +88,49 @@ function JsonBlock({ value }: { value: unknown }) {
...
+        <div className="GenAITab--messageHeaderActions">
+          <select
+            className="GenAITab--formatSelect"
+            aria-label="Content format"
+            value={format}
+            onChange={e => onFormatChange(e.target.value as MessageFormat)}
+          >
+            <option value="plain">Plain</option>
+            <option value="markdown">Markdown</option>
+            <option value="json">JSON</option>
```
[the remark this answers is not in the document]
swetalin-10: Good catch - fixed in 1f068271. The dropdown now displays `effectiveFormat` (what's actually rendered) instead of the raw stored preference, and the Markdown option is disabled with a tooltip once content exceeds the 150KB limit, so it can no longer show "Markdown" selected while the block silently renders as plain. Added a regression test covering this exact scenario.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.tsx
```diff
@@ -13,26 +14,55 @@ import {
...
-      <pre className="GenAITab--messageContent" data-testid={`genai-message-${index}`}>
-        {message.content}
-      </pre>
-    </div>
-  );
+// Message text has no declared format in the OTel GenAI conventions - it can be prose,
+// Markdown, JSON, code, or a template. Per review discussion, the format is not assumed:
+// content that parses as JSON defaults to the tree view, everything else defaults to
+// plain text, and the user can override per attribute via the dropdown below. The
+// override is remembered per attribute name (not globally, and not per-message), so
```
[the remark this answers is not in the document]
swetalin-10: Fixed - title updated to "feat(genai-tab): add a per-attribute format toggle (plain/markdown/JSON) for message content" and the description was already rewritten in 0aded5ae to describe the opt-in behavior (JSON auto-detect, plain default, per-attribute persistence) rather than unconditional markdown.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.tsx
```diff
@@ -13,26 +14,55 @@ import {
...
+} {
+  const [overrides, setOverrides] = useState<Partial<Record<string, MessageFormat>>>({});
+
+  const getFormat = (attributeKey: string, content: string): MessageFormat => {
+    const override = overrides[attributeKey];
+    if (override) return override;
+    const stored = storage.getString(MESSAGE_FORMAT_STORAGE_PREFIX + attributeKey);
+    if (stored === 'plain' || stored === 'markdown' || stored === 'json') return stored;
+    const parsed = tryParseJson(content);
+    return typeof parsed === 'object' && parsed !== null ? 'json' : 'plain';
```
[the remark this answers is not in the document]
swetalin-10: Good catch - fixed in 768a95da. `jsonObjectOrArrayStartRegex` now allows leading whitespace (`/^\s*[[{]/` instead of `/^(\[|\{)/`), so pretty-printed JSON like "\n{...}" is detected correctly - `JSON.parse` already tolerates the leading whitespace, so no other behavior changes. Added 5 unit tests on `tryParseJson` directly.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.tsx
```diff
@@ -58,6 +88,55 @@ function JsonBlock({ value }: { value: unknown }) {
...
+              Markdown{isOversized ? ' (too large)' : ''}
+            </option>
+            <option value="json">JSON</option>
+          </select>
+          <CopyIcon copyText={message.content} tooltipTitle="Copy message" buttonText="Copy" />
+        </div>
+      </div>
+      {effectiveFormat === 'json' ? (
+        <JsonBlock value={message.content} />
+      ) : effectiveFormat === 'markdown' ? (
```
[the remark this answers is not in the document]
swetalin-10: Same root cause as the sibling comment - fixed by the same change in 768a95da. Since `JsonBlock` calls the shared `tryParseJson` from `genAiData.ts`, fixing the regex there fixes both the auto-detection path and the explicit-JSON-selection path with one change. Added a UI-level test confirming pretty-printed JSON with a leading newline routes to the tree view and the dropdown shows "JSON".

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.tsx
```diff
@@ -58,6 +88,55 @@ function JsonBlock({ value }: { value: unknown }) {
...
+    <div className={`GenAITab--message GenAITab--message-${message.role || 'unknown'}`}>
+      <div className="GenAITab--messageHeader">
+        <span className="GenAITab--messageRole">{message.role || 'message'}</span>
+        <div className="GenAITab--messageHeaderActions">
+          <select
+            className="GenAITab--formatSelect"
+            aria-label="Content format"
+            value={effectiveFormat}
+            onChange={e => onFormatChange(e.target.value as MessageFormat)}
+          >
```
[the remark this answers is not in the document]
swetalin-10: Good catch - fixed in 15658d56. Each format dropdown's accessible name now includes the message's position and role, e.g. "Content format for message 2 (assistant)", instead of the shared generic "Content format". `ConversationSection` now builds one flattened, ordered list of messages (system, then input, then output) so the numbering is stable and sequential. Added a test asserting two messages get distinct accessible names.

yurishkuro (commented):
Please remove screenshots/ dir from the branch.

Can you share the trace you are using for testing?

yurishkuro:
image
image

The formatting of the Markdown render is poor: 
  1. having margin on the first word is not typical (see GitHub's rendering in the PR)
  2. there is no horizontal padding around the paragraph

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.tsx
```diff
@@ -13,26 +14,55 @@ import {
...
+
+// Above this length, Markdown parsing is skipped even if selected - avoids pathological
+// reflow/parse cost on huge attributes. Plain text and the JSON tree view have no such
+// cap since neither does Markdown's block-level reparsing.
+const MARKDOWN_SIZE_LIMIT = 150_000;
+
+// Lives in the parent (ConversationSection), not per-MessageBlock: a format choice for
+// an attribute name must apply to every currently-rendered message from that attribute
+// immediately, not just on that one message's own instance or a future remount.
+function useMessageFormatOverrides(): {
```
yurishkuro: we normally define these in separate files

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.tsx
```diff
@@ -58,6 +88,57 @@ function JsonBlock({ value }: { value: unknown }) {
...
+          >
+            <option value="plain">Plain</option>
+            <option
+              value="markdown"
+              disabled={isOversized}
+              title={isOversized ? 'Markdown is disabled for messages over 150KB' : undefined}
+            >
+              Markdown{isOversized ? ' (too large)' : ''}
+            </option>
+            <option value="json">JSON</option>
```
yurishkuro: should this be also disabled if the value is not a JSON?

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.tsx
```diff
@@ -58,6 +88,57 @@ function JsonBlock({ value }: { value: unknown }) {
...
+      <div className="GenAITab--messageHeader">
+        <span className="GenAITab--messageRole">{message.role || 'message'}</span>
+        <div className="GenAITab--messageHeaderActions">
+          <select
+            className="GenAITab--formatSelect"
+            aria-label={`Content format for message ${messageNumber} (${message.role || 'message'})`}
+            value={effectiveFormat}
+            onChange={e => onFormatChange(e.target.value as MessageFormat)}
+          >
+            <option value="plain">Plain</option>
```
yurishkuro, line 100: where do we stand on YAML rendered, is there a lightweight library for syntax highlighting?

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.tsx
```diff
@@ -58,6 +88,57 @@ function JsonBlock({ value }: { value: unknown }) {
...
+            >
+              Markdown{isOversized ? ' (too large)' : ''}
+            </option>
+            <option value="json">JSON</option>
+          </select>
+          <CopyIcon copyText={message.content} tooltipTitle="Copy message" buttonText="Copy" />
+        </div>
+      </div>
+      {effectiveFormat === 'json' ? (
+        <JsonBlock value={message.content} />
```
yurishkuro: Can you share a trace with JSON payload?

It would be good to improve JSON rendering in the standard Details view - the main issue there is that the attributes are rendered in a table with key:value columns, but the JSON content is better viewed on a separate line from the key (and I'd argue it could also use the same renderer dropdown that you have here, why should the two displays be different?

swetalin-10:
Thanks for the detailed pass — went through each point:

**Screenshots dir** — removed (the tracked `screenshots/before.png` / `after.png` and the PR description's `raw.githubusercontent.com` links to them are both gone now; committed in e77f2f3e).

**Markdown padding/margin bug** — found the actual cause instead of guessing. `markdown-to-jsx` only wraps its compiled output in a block element once it has more than one top-level node. A single-sentence message with no other formatting (like the SYSTEM prompt in the screenshots) compiles to exactly one inline node, and `forceWrapper` then wraps it in a bare `<span>` instead of a `<div>`. Padding on an inline element only shows at the very start/end of the whole run, not around each wrapped line — that's exactly the "indented first word, no padding on the rest" you saw. Confirmed by rendering the component directly and inspecting the DOM:

- Before: `<span class="GenAITab--messageContent">…</span>`
- After (`forceBlock: true` added for this tab): `<div class="GenAITab--messageContent"><p>…</p></div>` — same shape multi-paragraph messages already used correctly.

Fixed in 44b49e4f, scoped to GenAITab only (didn't touch `sharedMarkdownOptions` itself since the assistant chat panel puts its padding on an outer container, not the Markdown root, so it isn't affected by this — didn't want to change behavior there without its own testing pass).

**File separation** — moved `MessageFormat`, the storage prefix, the size-limit constant, and the read/write/detect logic out of `index.tsx` into a new `messageFormat.ts`, following the same pattern `ThemeStorage.ts`/`ThemeProvider.tsx` already use elsewhere for localStorage-backed UI state.

**JSON option disable** — good catch, yes it needed the same treatment as Markdown. The JSON option is now disabled (with a tooltip) whenever the message content doesn't actually parse as JSON. Since the preference is stored per attribute name, a message can inherit "JSON" from a sibling under the same attribute even when its own content isn't JSON — that case now falls back to plain text instead of showing "JSON" selected while nothing renders. Added a regression test for exactly that cross-message case.

**YAML rendering / lightweight library** — checked actual bundle sizes rather than guessing:

| Option | Real measured size (core + a YAML grammar/language, gzip) |
|---|---|
| Prism.js core + `prism-yaml` component | ~4KB (3.2KB + 0.9KB) |
| highlight.js core + `languages/yaml` | ~24KB |
| `refractor` (React/hast wrapper over Prism) | ~7-10KB+ (wrapper overhead on top of Prism, plus `hastscript`/`parse-entities` as transitive deps) |
| `lowlight` (hljs wrapper) default import | ~289KB (has a tree-shakeable core, but same wrapper-overhead conclusion as refractor for one language) |

Prism core + the YAML grammar only is the clear lightest real option. For detection, wouldn't reach for `js-yaml` (that's a full parser, ~30-40KB, overkill just to detect the format) — a regex heuristic similar to the existing `tryParseJson` guard is enough. For rendering, would tokenize with `Prism.tokenize()` and walk the tokens into React elements directly rather than `dangerouslySetInnerHTML`, keeping the same no-raw-HTML posture the Markdown path already has. Didn't implement this in this PR — it's a new format type (a 4th option) on top of what's already here, and feels like its own follow-up rather than folding into this one. Happy to open that as a separate PR if that sizing makes sense to you.

**Trace / JSON payload trace** — working on getting you a trace to attach; will follow up with that separately once I have it, rather than posting a screenshot-only answer here.

**Details-view suggestion** — agree the key:value table is a bad fit for JSON content specifically, and reusing the same format dropdown there instead of a second bespoke renderer makes sense. Would want to scope that as its own PR too given it touches the standard attributes table, not just GenAITab — let me know if you'd rather see a proposal for that before I start.

swetalin-10:
Follow-up on the two open items from my last comment:

**Trace with JSON payload** — attaching the actual trace behind the screenshots ([benchmark-trace-bde4caeb.json](https://github.com/user-attachments/files/30152266/benchmark-trace-bde4caeb.json), trace ID `bde4caeb1c1a8d4c8b591f597d19d5af`, `jaeger: /api/ai/chat`, 13 spans). It also happens to answer the JSON-payload ask directly: the `generate_content` spans carry a `gen_ai.tool.definitions` attribute — a JSON-encoded array of the MCP tool schemas available to the sidecar (`get_critical_path`, `get_service_dependencies`, `get_services`, etc., each with full parameter/response JSON Schema). One trace, both asks.

**Details-view JSON rendering** — went ahead and fixed the concrete part of this: `AttributesTable` now gives a JSON-tree value its own full-width row instead of squeezing it into the value column next to a fixed-width key column. Loading `gen_ai.tool.definitions` from the attached trace is a good way to see it — that attribute isn't claimed by any GenAITab section, so it falls through to "Other GenAI Attributes," which shares this exact table with the standard Details tab, so the fix applies in both places (and in the span Events accordion and Trace Logs view, since they all use the same component). Verified all four still pass their full suites (1175 tests) with no regressions.

Didn't extend this to reusing the Plain/Markdown/JSON dropdown on Details-view values, though — that dropdown makes sense for GenAITab because message content has genuinely ambiguous format. A Details-view value that's already confirmed to parse as JSON isn't ambiguous the same way, and wiring an interactive format-switcher onto every JSON-shaped attribute across spans/events/logs app-wide felt like a bigger call than I should make unilaterally on shared code based on one comment. Wanted to flag that split explicitly rather than silently deciding it either way — let me know if you'd still want that built out, happy to take it as a follow-up.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/genAiData.ts
```diff
@@ -96,12 +96,14 @@ function parseSystemInstructions(value: AttributeValue | undefined): string | un
...
 
-const jsonObjectOrArrayStartRegex = /^(\[|\{)/;
+const jsonObjectOrArrayStartRegex = /^\s*[[{]/;
 
 /**
  * Only attempts JSON.parse when the string looks like an object/array literal,
  * same guard as AttributesTable.tryParseJson - avoids relying on try/catch
  * exceptions for control flow on the common case of plain, non-JSON strings.
+ * Leading whitespace is tolerated so pretty-printed/indented JSON (e.g.
+ * "\n{...}") is still detected - JSON.parse itself already ignores it.
```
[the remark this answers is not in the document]
swetalin-10: Good catch — fixed in 1e873711. Reworded the docstring from "same guard as AttributesTable.tryParseJson" to "a similar guard to AttributesTable.tryParseJson", and added an explicit note that unlike AttributesTable's guard, this one tolerates leading whitespace. The two are genuinely different now (this one's regex is `/^\s*[[{]/`, AttributesTable's is `/^(\[|\{)/`) since this guard was made whitespace-tolerant earlier in the PR to detect pretty-printed JSON, so the comment needed to stop claiming equivalence.

yurishkuro:
Did a broader pass. Two things.

**1. The persisted format preference should be a zustand store, not a hand-rolled module + `useState`.**

`messageFormat.ts` together with the `overrides` state in `ConversationSection` reimplements what we already do for persisted UI prefs — see [`search-panel-store.ts`](../blob/main/packages/jaeger-ui/src/components/SearchTracePage/search-panel-store.ts) and [`store.layout.ts`](../blob/main/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/store.layout.ts): a localStorage-backed `create()` store where `getInitialState()` reads LS and the setters write through to LS + `set()`.

Right now the one concern is split across two layers — reactive in-memory state (`useState` in `ConversationSection`) and persistence (`readStoredMessageFormat` / `writeStoredMessageFormat` over the `storage` util) — manually merged as `override ?? stored ?? default`. That split only exists because React state isn't reactive to localStorage, which is exactly what a persisted store solves in one place.

A `useMessageFormatStore` keyed by attribute name would:
- drop the `overrides` `useState` and the "lift state up so every message updates live" comment — store subscribers get that for free;
- drop the `read`/`write` module functions — the setter writes through to localStorage and init hydrates from it;
- match our `*-store.ts` / `useXxxStore` naming.

Net: less code, and consistent with the rest of the app. (This supersedes my earlier "we normally define these in separate files" note — the file is right, the shape isn't yet.)

**2. Message content was re-parsed as JSON on every render.**

`MessageBlock` called `tryParseJson(message.content)` in its render body, and the parent re-parsed the same string in its format lookup — so changing one message's format re-parsed every sibling message. I pushed a small fix directly onto the branch (723a5bff): memoize the parse, resolve only the override in the parent and derive the content-based default in `MessageBlock` from the memoized value, and fold the availability checks into a single `canRender` map. Please fold it into the store refactor above rather than treating it as final.

swetalin-10:
Both addressed in 4f67d60e.

**1. Zustand store** — new `message-format-store.ts` / `useMessageFormatStore`, matching `search-panel-store.ts`/`store.layout.ts`: `getInitialState()` hydrates the three known message attribute keys from `storage` at creation, `setFormat` writes through to storage then `set()`. `ConversationSection` dropped the `overrides` `useState` and the read-then-merge — it just reads `overrides[attributeKey] ?? null` off the store now. `messageFormat.ts` kept `MessageFormat`/`MARKDOWN_SIZE_LIMIT`, lost the two storage functions.

**2. 723a5bff** — folded in as-is, not treated as final. `MessageBlock`'s memoized `parsedJson`/`canRender` map is unchanged; only what feeds `formatOverride` into it moved from `useState` to the store.

One thing I caught while doing this: `index.test.tsx`'s `beforeEach` only cleared `localStorage`, which was correct for the old per-render `useState` but not for a module-level store singleton — added a `useMessageFormatStore.setState({ overrides: {} })` reset alongside it so state can't leak across test cases.

Verified against the real trace already on this PR: format choice survives a full page reload (proves `getInitialState()` hydration, not just live `set()`), and on a span with 5 messages sharing `gen_ai.input.messages`, changing one live-updates all 5 immediately while the sibling `gen_ai.output.messages` message stays independent.

93/93 tests passing (11 existing format tests + new store tests), `tsc-lint` clean, CI green.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/message-format-store.ts
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) 2026 The Jaeger Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+import { create } from 'zustand';
+import storage from '../../../../../utils/storage';
+import type { MessageFormat } from './messageFormat';
+
+const MESSAGE_FORMAT_STORAGE_PREFIX = 'jaeger.genaiTab.messageFormat.';
```
yurishkuro: we said we want to generalize this to regular span detail rendering as well
```suggestion
const MESSAGE_FORMAT_STORAGE_PREFIX = 'jaeger.spanDetail.attributeFormat.';
```

yurishkuro (approved):

PR #4221, opened 2026-07-18
fix(genai-tab): simplify and fix rendering of GenAI span icons

### LINKED ISSUE #4217: [bug]: Simplify GenAI span icons
@swetalin-10 we need to improve the rendering. I see several problems below:
  1. The first row shows both ✨ and 🤖, I am not clear what's the difference
  2. The second and other rows are even worse, they show two different sparkles, that just doesn't make sense
  3. We need the icons to have a hover tooltip that explains what they mean.
  4. The first and second icons don't seem to be v-aligned

I also question why we need a separate GenAI icon in the first place, just show one icon indicating the type of GenAI operation.

Image

swetalin-10:

Fixes #4217. Every GenAI span in the trace timeline rendered two icons instead of one:

1. `span-icons.ts`'s generic namespace-based icon (`IoSparkles`), shown for any span with `gen_ai.*` attributes.
2. `GenAISpanIcon`'s operation-kind icon (agent/LLM call/tool call/retrieval, or `MdAutoAwesome` - a different sparkle-style icon - for a GenAI span whose operation isn't in the classification list).

Both rendered side by side on every GenAI span, which produced exactly the four problems reported:
- An agent span showed both the generic sparkle and the robot icon, with no visible distinction between them.
- Every other GenAI span showed two different sparkle-style icons next to each other, conveying no additional information over one.
- Neither icon had a hover tooltip explaining what it meant, only an `aria-label` for screen readers.
- The two icons come from different icon libraries (Ionicons vs Material Design) with different internal glyph metrics, so despite identical `vertical-align`/`font-size` CSS on both, they did not line up with each other.

## Description of the changes

- `span-icons.ts`: removed the `gen_ai` entry from `NAMESPACE_PRIORITY`/`NAMESPACE_ICON`. `classifySpan` (in `utils/genai/detect.ts`) already falls back to `UNKNOWN_GENAI` for any span with `gen_ai.*` attributes it can't classify by operation name, so `GenAISpanIcon` alone already covers every case the generic namespace entry did.
- `SpanBarRow.tsx`: suppresses the generic namespace icon entirely whenever a span is already GenAI-classified (`span.genAIKind !== undefined`), instead of relying on `span-icons.ts` to individually exclude every namespace that might co-occur with `gen_ai` attributes on the same span. This was needed because removing only the `gen_ai` entry wasn't sufficient on its own: the reported trace's root `/api/ai/chat` span carries both `gen_ai.*` attributes (it's classified `AGENT`) and genuine `http.*` attributes (it's an HTTP server span - `http.request.method`, `http.response.status_code`, etc.). Once `gen_ai` stopped winning the generic icon's namespace-priority contest, `http` won instead, so the span still rendered two icons - just a globe next to the robot instead of a sparkle next to the robot. Found by testing the first version of this fix against the real trace end-to-end (see below), not by inspection.
- `GenAISpanIcon.tsx`: wrapped the icon in an antd `Tooltip`, reusing the same per-kind `label` already used for `aria-label` (e.g. "Agent", "LLM call", "Tool call", "Retrieval", "GenAI span"). Hovering any GenAI span icon now shows what it means.
- With the generic icon suppressed, a GenAI span now renders exactly one icon, which also resolves the alignment complaint as a direct consequence - verified visually, not just logically: rendered the real `IoSparkles`/`GenAISpanIcon` SVG output through the real CSS in a browser, before and after. Before, the sparkle icon's visual mass sat clearly above a guide line marking the row's true vertical center, while the robot icon straddled it evenly - a real, visible offset between the two. After, the single remaining icon sits centered on the line.

## Screenshots

4 real spans from the exact trace behind the issue's screenshot (trace `bde4caeb`), rendered through the real components with the real CSS - before/after. Row 1 is the root `/api/ai/chat` span specifically, since that's the one that exposed the deeper root cause described above: it also carries `http.*` attributes, which is why its "before" state shows a globe icon rather than a sparkle - same underlying double-icon problem, just a different second icon depending on what else the span happens to be instrumented with.

Before/after: 4 real spans from the reported trace, real components, real CSS

**Trace file** used for testing and for the screenshot above: [benchmark-trace-bde4caeb.json](https://github.com/user-attachments/files/30154342/benchmark-trace-bde4caeb.json) (the same trace already attached to PR #4213).

## Test changes

- `span-icons.test.ts`: updated the namespace-icon test to exclude `gen_ai` from the "returns an icon" cases, added an explicit test that `gen_ai`-only attributes return `null`, and replaced the `gen_ai`-priority test with an equivalent `db`-vs-`http` priority test.
- `GenAISpanIcon.test.tsx`: added a test that hovering the icon shows a tooltip with the same text as its accessible name.
- `SpanBarRow.test.jsx`: added a test asserting a span with `gen_ai.*` attributes renders only `GenAISpanIcon`'s kind-specific icon, and a second regression test for the http+gen_ai co-occurrence case described above (a span with both `gen_ai.operation.name` and `http.*` attributes still renders exactly one icon).

- `npm test` - `GenAISpanIcon.test.tsx`, `span-icons.test.ts`, `SpanBarRow.test.jsx`: all passing (46/46 across the three). Full `TraceTimelineViewer` suite (36 files, 644 tests): all passing, no regressions.
- `npx tsc -p packages/jaeger-ui/tsconfig.json` - no new type errors.
- `npm run lint` (oxlint + knip) - clean on all changed files.
- End-to-end verification against the actual trace behind the linked screenshot, through the real production pipeline (not synthetic data or a reimplemented check): loaded the trace's raw JSON through the real `transformTraceData` and `OtelTraceFacade`, rendered the real `SpanBarRow` for every one of its 13 real spans, and asserted the actual rendered DOM. Result: every GenAI-classified span (including the root `/api/ai/chat` span, `genAIKind=AGENT`) renders exactly one icon (`GenAISpanIcon`, zero generic icons); the one non-GenAI span (`mcp.discover_tools`) renders zero icons from either source. This test run is what caught the http+gen_ai co-occurrence gap in the first version of this fix.

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/GenAISpanIcon.test.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/GenAISpanIcon.test.tsx
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import * as React from 'react';
-import { render, screen } from '@testing-library/react';
+import { render, screen, fireEvent, waitFor } from '@testing-library/react';
 import { GenAISpanIcon } from './GenAISpanIcon';
 import type { IOtelSpan, GenAISpanKind } from '../../../types/otel';
 
@@ -44,5 +44,14 @@ describe('GenAISpanIcon', () => {
   it('falls back to the generic GenAI icon for an unrecognized genAIKind', () => {
     render(<GenAISpanIcon span={makeSpan('FUTURE_KIND' as GenAISpanKind)} />);
     expect(screen.getByRole('img', { name: 'GenAI span' })).toBeInTheDocument();
+  });
+
+  it('shows a hover tooltip explaining what the icon means (#4217)', async () => {
+    render(<GenAISpanIcon span={makeSpan('TOOL_CALL')} />);
+    expect(screen.queryByRole('tooltip')).not.toBeInTheDocument();
+    fireEvent.mouseEnter(screen.getByRole('img', { name: 'Tool call' }));
+    await waitFor(() => {
+      expect(screen.getByRole('tooltip')).toHaveTextContent('Tool call');
+    });
   });
 });
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/GenAISpanIcon.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/GenAISpanIcon.tsx
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import * as React from 'react';
+import { Tooltip } from 'antd';
 import { MdSmartToy, MdBolt, MdBuild, MdStorage, MdAutoAwesome } from 'react-icons/md';
 import type { IconType } from 'react-icons';
 import type { IOtelSpan, GenAISpanKind } from '../../../types/otel';
@@ -20,8 +21,10 @@ export function GenAISpanIcon({ span }: { span: IOtelSpan }): React.ReactElement
   if (kind === undefined) return null;
   const { icon: Icon, label } = KIND_META[kind] ?? KIND_META.UNKNOWN_GENAI;
   return (
-    <span role="img" aria-label={label} className="GenAISpanIcon">
-      <Icon aria-hidden="true" />
-    </span>
+    <Tooltip title={label}>
+      <span role="img" aria-label={label} className="GenAISpanIcon">
+        <Icon aria-hidden="true" />
+      </span>
+    </Tooltip>
   );
 }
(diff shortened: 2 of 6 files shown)


swetalin-10:
Attaching real evidence for this fix:

- **Screenshot**: 4 real spans from the exact trace behind this issue's screenshot (trace `bde4caeb`), rendered through the real components with the real CSS - before/after. Row 1 is the root `/api/ai/chat` span specifically, since that's the one that exposed the deeper bug (it also carries `http.*` attributes, which is why the "before" state shows a globe icon rather than a sparkle for that row - same underlying double-icon problem, just a different second icon depending on what else the span is instrumented with).
- **Trace file**: the actual trace this was tested against, same one already attached to PR #4213 for the same reason.

pr4221_evidence
[benchmark-trace-bde4caeb.json](https://github.com/user-attachments/files/30154342/benchmark-trace-bde4caeb.json)

yurishkuro (approved):

PR #4239, opened 2026-07-19
fix(genai): Fix span icon hover tooltip and refine type icons and labels

### LINKED ISSUE #4217: [bug]: Simplify GenAI span icons

swetalin-10:
Fixes #4217 (point 3 was reopened - the hover tooltip added in #4221 wasn't actually working).

## Root cause

`.span-name::after` is an absolutely-positioned pseudo-element that extends the row's click target across the rest of the name column:

```css
.span-name::after {
  background: transparent;
  bottom: 0;
  content: ' ';
  left: 0;
  position: absolute;
  top: 0;
  width: 1000px;
}
```

Because it's a positioned box generated after all of `.span-name`'s other content, it paints above `GenAISpanIcon`'s plain in-flow `<span>` within the same stacking context - so real mouse hover lands on the `::after` overlay instead of the icon, and the `antd` `Tooltip`'s `onMouseEnter` never fires. Confirmed via `document.elementFromPoint()` at the icon's own bounding-box center: it resolved to `A.span-name`, not the icon.

This is exactly the problem `.SpanBarRow--pillWrap` already solves for span pills, in the same file, with the same fix:

```css
.SpanBarRow--pillWrap {
  /* Sit above .span-name::after so hover tooltips receive pointer events. */
  position: relative;
  z-index: 1;
}
```

`GenAISpanIcon.css` never got the same treatment, so the icon's tooltip silently never worked, even though the isolated `GenAISpanIcon.test.tsx` unit test passes - it renders the icon alone, with no `.span-name::after` overlay to reproduce the bug.

## Fix

Apply the identical `position: relative; z-index: 1;` to `.GenAISpanIcon`.

## How this was verified

Unit tests don't catch this class of bug - `jsdom` doesn't do real hit-testing, so a synthetic `fireEvent.mouseEnter` on the icon bypasses the overlay regardless. Verified for real instead:

- Ran the actual dev server (`npm start`) against a clean `upstream/main` checkout.
- Uploaded the real trace behind this issue's original screenshot (`bde4caeb`, same file already attached to #4213/#4221) via the UI's own trace-upload feature - no synthetic data.
- Confirmed the bug first: `document.elementFromPoint()` at the icon's rect center resolved to the `.span-name` link, not the icon.
- Applied the fix, confirmed HMR picked it up, confirmed `elementFromPoint()` now resolves inside the icon's own SVG.
- Did a real mouse hover (not a synthetic event) over the icon and visually confirmed the tooltip renders - across multiple icon kinds (Agent, LLM call, GenAI span) on multiple spans in the same real trace.
- Re-verified the same real hover on a clean rebuild of this exact branch off latest `upstream/main`, after discarding an unrelated repo-wide formatting run.
- `npx vitest run` on `GenAISpanIcon.test.tsx` + `SpanBarRow.test.js`: 38/38 passing, no regressions.
- `npm run tsc-lint`: clean.

**Before / after**, same real trace (`bde4caeb`), same hover coordinates, only the CSS fix differs:

before-after

## Follow-up: icon and label refinements

Building on the same GenAI span-type icons, a few presentation tweaks:

- **LLM call icon**: swap the bolt (`MdBolt`) for a graduation cap (`RiGraduationCapFill`), a more fitting glyph for a model call.
- **Labels** (tooltip + `aria-label`): `Agent` → `AI Agent`, `Tool call` → `MCP Tool call` for clarity.

Unit tests updated accordingly (`GenAISpanIcon.test.tsx`, `SpanBarRow.test.jsx`).

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/GenAISpanIcon.css
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/GenAISpanIcon.css
@@ -8,4 +8,7 @@ SPDX-License-Identifier: Apache-2.0
   margin-right: 0.3rem;
   vertical-align: -0.18em;
   color: var(--interactive-primary);
+  /* Sit above .span-name::after so the hover tooltip receives pointer events. */
+  position: relative;
+  z-index: 1;
 }
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/GenAISpanIcon.test.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/GenAISpanIcon.test.tsx
@@ -23,12 +23,12 @@ describe('GenAISpanIcon', () => {
 
   it('renders a tool call icon for genAIKind=TOOL_CALL', () => {
     render(<GenAISpanIcon span={makeSpan('TOOL_CALL')} />);
-    expect(screen.getByRole('img', { name: 'Tool call' })).toBeInTheDocument();
+    expect(screen.getByRole('img', { name: 'MCP Tool call' })).toBeInTheDocument();
   });
 
   it('renders an agent icon for genAIKind=AGENT', () => {
     render(<GenAISpanIcon span={makeSpan('AGENT')} />);
-    expect(screen.getByRole('img', { name: 'Agent' })).toBeInTheDocument();
+    expect(screen.getByRole('img', { name: 'AI Agent' })).toBeInTheDocument();
   });
 
   it('renders a retrieval icon for genAIKind=RETRIEVAL', () => {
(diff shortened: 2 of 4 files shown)


swetalin-10:
before-after

yurishkuro (dismissed):

yurishkuro (approved):

PR #4244, opened 2026-07-20
feat(genai-tab): add LLM caption to the meta row and elevate agent attributes

### LINKED ISSUE #4237: [chore]: LLM and Agent attributes
### Requirement

(1) Similar to Tokens section we can have "LLM" caption for the top section

Image

(2) The agent attributes are not elevated at all, could be their own Agent section (name, version, whatever else semconv have for the agent)

Image

cc @swetalin-10

swetalin-10:
- Resolves #4237

## Description of the changes
- The provider/model row had no caption, unlike the Tokens row added in #4222. `LLMDetails` now gets its caption from the shared `AccordionAttributes` component's `label` prop - `LLM` for a span `classifySpan()` identifies as an actual LLM call, `Model` otherwise (per Ansh's review, so a provider/model reported on an Agent/Tool/Retrieval span isn't mislabeled as an LLM call) - open by default, with a `Provider`/`Model` table when expanded and a `Provider=... Model=...` summary when collapsed.
- `gen_ai.agent.*` attributes were falling into the raw "Other GenAI Attributes" key/value dump instead of being elevated like provider/model and token usage already are. Checked the current [OTel GenAI semconv registry](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-agent-spans.md) for the full set rather than assuming just name/version - the registry has four fields: `gen_ai.agent.id`, `.name`, `.version`, `.description`. Added a builder claiming all four and an `AgentDetails` component rendering them through the same shared `AccordionAttributes` component as `LLMDetails` (name first in the display order, so `AccordionAttributes`' collapsed preview leads with the highest-signal field), and excluded once claimed from Other GenAI Attributes.
- Per @yurishkuro's later request, `ToolCallDetails` and `UnknownDetails` (the defensive fallback for a section type not yet handled) also render through the same shared accordion, so every elevated GenAI section (Agent, LLM, Tool Call, Unknown) is visually and behaviorally consistent instead of a mix of one-off components. `Tokens:` stays an inline single-line row (not a collapsible section), and `Conversation` stays a bespoke component (its per-message Markdown/JSON/Plain toggle and copy button don't fit the accordion's flat key-value shape) - both intentionally out of scope, per discussion with Yuri.

- Real trace, not synthetic - uploaded the actual `bde4caeb` trace (`jaeger: /api/ai/chat`) via the dev server's own trace-upload feature, and compared directly against the screenshots in the issue:
  - The root `/api/ai/chat` span (`gen_ai.agent.name=jaeger-gemini-sidecar`, `.version=0.1.0`, no provider/model/tokens) - exactly the span in the issue's second screenshot - shows a standalone **Agent** section, and `gen_ai.agent.name`/`.version` no longer appear in Other GenAI Attributes.
  - The first `generate_content gemini-2.5-flash` span (`gen_ai.provider.name=gcp.gen_ai`, `.request.model`/`.response.model=gemini-2.5-flash`, input 3,803/output 45 tokens) - exactly the span and token counts in the issue's first screenshot - now shows an **LLM** accordion (open by default) with Provider/Model in a table, same collapsible pattern as **Agent**. `Tokens:` stays an inline row below it, unchanged.
  - LLM accordion open on the real bde4caeb trace's generate_content span - Provider gcp.gen_ai, Model gemini-2.5-flash, Tokens Input 3,803 Output 45 - matching the issue's screenshot exactly
- `vitest run` on `genAiData.test.ts` + `index.test.tsx`: 60/60 passing on `genAiData.test.ts`, 32/32 passing on `index.test.tsx` (92/92 total), including the accordion rendering for Agent/LLM/Tool Call/Unknown and the "Model" vs "LLM" label case for a non-LLM span with a provider/model.
- `tsc` and `oxlint` clean on the changed files.

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/genAiData.ts
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/genAiData.ts
@@ -24,6 +24,16 @@ export type GenAiToolCall = {
   result?: unknown;
 };
 
+// Fields per the gen_ai.agent.* registry (create_agent/invoke_agent spans): id, name,
+// version, description. All are individually optional/conditionally-required per spec,
+// so a span can carry any subset.
+export type GenAiAgent = {
+  id?: string;
+  name?: string;
+  version?: string;
+  description?: string;
+};
+
 export type GenAiTokenUsage = {
   inputTokens?: number;
   outputTokens?: number;
@@ -36,6 +46,7 @@ export type GenAiTokenUsage = {
 // so a generic fallback renderer can walk `Object.entries(data)` for a
 // section type it doesn't have a specific case for.
 export type GenAiSection =
+  | { type: 'agent'; data: GenAiAgent }
   | { type: 'meta'; data: { provider?: string; model?: string } }
   | { type: 'tokens'; data: GenAiTokenUsage }
   | {
@@ -246,6 +257,15 @@ type SectionBuilder = (get: GetAttr) => GenAiSection | undefined;
 // position; the loop below walks this array (developer-controlled, fixed
 // order), not the span's attributes (whose order is instrumentation-dependent).
 const REGISTRY: SectionBuilder[] = [
+  get => {
+    const id = asString(get('gen_ai.agent.id'));
+    const name = asString(get('gen_ai.agent.name'));
+    const version = asString(get('gen_ai.agent.version'));
+    const description = asString(get('gen_ai.agent.description'));
+    return id || name || version || description
+      ? { type: 'agent', data: { id, name, version, description } }
+      : undefined;
+  },
   get => {
     // gen_ai.provider.name/gen_ai.system is a genuine current/deprecated pair -
     // `||` short-circuits so gen_ai.system is only read (and claimed) when
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/genAiData.ts
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/genAiData.ts
@@ -262,7 +262,11 @@ const REGISTRY: SectionBuilder[] = [
     const name = asString(get('gen_ai.agent.name'));
     const version = asString(get('gen_ai.agent.version'));
     const description = asString(get('gen_ai.agent.description'));
-    return id || name || version || description
+    // Presence, not truthiness: get() already claimed these keys, so a
+    // legitimate empty-string value must still produce a section - otherwise
+    // it's dropped here AND excluded from "Other GenAI Attributes" (already
+    // claimed), silently losing the attribute entirely.
+    return id !== undefined || name !== undefined || version !== undefined || description !== undefined
       ? { type: 'agent', data: { id, name, version, description } }
       : undefined;
   },
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.css
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.css
@@ -32,6 +32,7 @@ SPDX-License-Identifier: Apache-2.0
   margin-right: 4px;
 }
 
+.GenAITab--metaPrefix,
 .GenAITab--tokensPrefix {
   font-size: 13px;
   font-weight: 600;
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.css
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.css
@@ -9,7 +9,6 @@ SPDX-License-Identifier: Apache-2.0
   gap: 16px;
 }
 
-.GenAITab--meta,
 .GenAITab--tokens {
   display: flex;
   flex-wrap: wrap;
@@ -20,19 +19,16 @@ SPDX-License-Identifier: Apache-2.0
   border-radius: 4px;
 }
 
-.GenAITab--metaItem,
 .GenAITab--tokenItem {
   font-size: 13px;
   color: var(--text-primary);
 }
 
-.GenAITab--metaLabel,
 .GenAITab--tokenLabel {
   color: var(--text-secondary);
   margin-right: 4px;
 }
 
-.GenAITab--metaPrefix,
 .GenAITab--tokensPrefix {
   font-size: 13px;
   font-weight: 600;
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.css
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.css
@@ -123,20 +123,6 @@ SPDX-License-Identifier: Apache-2.0
   white-space: pre-wrap;
 }
 
-.GenAITab--toolSubsection {
-  display: flex;
-  flex-direction: column;
-  gap: 4px;
-}
-
-.GenAITab--toolLabel {
-  font-size: 12px;
-  font-weight: 600;
-  color: var(--text-secondary);
-  text-transform: uppercase;
-  letter-spacing: 0.03em;
-}
-
 .GenAITab--json,
 .GenAITab--pre {
   border: 1px solid var(--border-default);
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.tsx
@@ -159,7 +159,7 @@ const AGENT_LABELS: Partial<Record<keyof GenAiAgent, string>> = {
 };
 
 function AgentSection({ id, name, version, description }: GenAiAgent) {
-  const agent: GenAiAgent = { name, version, id, description };
+  const agent: GenAiAgent = { id, name, version, description };
   return (
     <div className="GenAITab--section">
       <h3 className="GenAITab--sectionTitle">Agent</h3>
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.tsx
@@ -9,6 +9,7 @@ import {
   extractGenAiSections,
   formatTokenCount,
   tryParseJson,
+  GenAiAgent,
   GenAiMessage,
   GenAiTokenUsage,
   GenAiToolCall,
@@ -132,6 +133,7 @@ function MessageBlock({
 function MetaRow({ provider, model }: { provider?: string; model?: string }) {
   return (
     <div className="GenAITab--meta">
+      <span className="GenAITab--metaPrefix">LLM:</span>
       {provider && (
         <span className="GenAITab--metaItem">
           <span className="GenAITab--metaLabel">Provider</span> {provider}
@@ -142,6 +144,34 @@ function MetaRow({ provider, model }: { provider?: string; model?: string }) {
           <span className="GenAITab--metaLabel">Model</span> {model}
         </span>
       )}
+    </div>
+  );
+}
+
+// Cosmetic only, same rationale as TOKEN_LABELS below - a field missing from here
+// still renders under its own key, so a future gen_ai.agent.* field shows up
+// automatically without needing a matching entry added here.
+const AGENT_LABELS: Partial<Record<keyof GenAiAgent, string>> = {
+  name: 'Name',
+  version: 'Version',
+  id: 'ID',
+  description: 'Description',
+};
+
+function AgentSection({ id, name, version, description }: GenAiAgent) {
+  const agent: GenAiAgent = { name, version, id, description };
+  return (
+    <div className="GenAITab--section">
+      <h3 className="GenAITab--sectionTitle">Agent</h3>
+      {(Object.keys(agent) as Array<keyof GenAiAgent>).map(key => {
+        const value = agent[key];
+        if (!value) return null;
+        return (
+          <div key={key} className="GenAITab--toolSubsection">
+            <span className="GenAITab--toolLabel">{AGENT_LABELS[key] ?? key}</span> {value}
+          </div>
+        );
+      })}
     </div>
   );
 }
@@ -291,6 +321,8 @@ export default function GenAITab({ span }: Props): React.ReactElement {
     <div className="GenAITab">
       {sections.map(section => {
         switch (section.type) {
+          case 'agent':
+            return <AgentSection key="agent" {...section.data} />;
           case 'meta':
             return <MetaRow key="meta" {...section.data} />;
           case 'tokens':
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.tsx
@@ -130,10 +130,14 @@ function MessageBlock({
   );
 }
 
-function MetaRow({ provider, model }: { provider?: string; model?: string }) {
+function MetaRow({ provider, model, isLlmCall }: { provider?: string; model?: string; isLlmCall: boolean }) {
   return (
     <div className="GenAITab--meta">
-      <span className="GenAITab--metaPrefix">LLM:</span>
+      {/* classifySpan() can also resolve to AGENT/TOOL_CALL/RETRIEVAL, which can carry
+          their own backing provider/model per the OTel GenAI semconv - captioning the
+          row "LLM:" would misrepresent those as LLM calls, so the prefix is scoped to
+          spans classifySpan() actually identifies as one. */}
+      {isLlmCall && <span className="GenAITab--metaPrefix">LLM:</span>}
       {provider && (
         <span className="GenAITab--metaItem">
           <span className="GenAITab--metaLabel">Provider</span> {provider}
@@ -330,7 +334,7 @@ export default function GenAITab({ span }: Props): React.ReactElement {
           case 'agent':
             return <AgentSection key="agent" {...section.data} />;
           case 'meta':
-            return <MetaRow key="meta" {...section.data} />;
+            return <MetaRow key="meta" {...section.data} isLlmCall={span.genAIKind === 'LLM_CALL'} />;
           case 'tokens':
             return <TokensRow key="tokens" usage={section.data} />;
           case 'conversation':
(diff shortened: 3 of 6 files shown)


packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.css
```diff
@@ -32,6 +32,7 @@ SPDX-License-Identifier: Apache-2.0
   margin-right: 4px;
 }
 
+.GenAITab--metaPrefix,
 .GenAITab--tokensPrefix {
   font-size: 13px;
   font-weight: 600;
```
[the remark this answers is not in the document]
swetalin-10: Checked this directly - measured getBoundingClientRect() on the live rendered DOM rather than just reading the CSS. Both `.GenAITab--metaPrefix` and `.GenAITab--tokensPrefix` get an identical, real 16px gap from the parent's flex `gap: 16px` (`.GenAITab--meta, .GenAITab--tokens { gap: 16px; }`), not from `margin-right`. That `margin-right: 4px` rule you're pointing at belongs to `.GenAITab--metaLabel, .GenAITab--tokenLabel` (the inner field label, e.g. "Provider"), not the prefix - so it was never the source of the prefix-to-first-item spacing for either row. No change needed here; verified in-browser, not just in code.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.tsx
```diff
@@ -146,6 +148,34 @@ function MetaRow({ provider, model }: { provider?: string; model?: string }) {
...
+      <h3 className="GenAITab--sectionTitle">Agent</h3>
+      {(Object.keys(agent) as Array<keyof GenAiAgent>).map(key => {
+        const value = agent[key];
+        if (!value) return null;
+        return (
+          <div key={key} className="GenAITab--toolSubsection">
+            <span className="GenAITab--toolLabel">{AGENT_LABELS[key] ?? key}</span> {value}
+          </div>
+        );
+      })}
```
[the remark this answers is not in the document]
swetalin-10: Fixed in 38e03b73 - replaced `Object.keys(agent) as Array<keyof GenAiAgent>` with an explicit `AGENT_FIELD_ORDER: ReadonlyArray<keyof GenAiAgent> = ['id', 'name', 'version', 'description']` that the component now iterates directly. Render order is no longer tied to object literal insertion order and the unsafe cast is gone.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.tsx
```diff
@@ -146,6 +148,34 @@ function MetaRow({ provider, model }: { provider?: string; model?: string }) {
   );
 }
 
+// Cosmetic only, same rationale as TOKEN_LABELS below - a field missing from here
+// still renders under its own key, so a future gen_ai.agent.* field shows up
+// automatically without needing a matching entry added here.
+const AGENT_LABELS: Partial<Record<keyof GenAiAgent, string>> = {
```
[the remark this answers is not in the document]
swetalin-10: Good catch, reworded in 38e03b73 to make the distinction explicit: the comment now says a new gen_ai.agent.* attribute does NOT show up automatically - it has to be added to GenAiAgent, extracted in genAiData.ts, and listed in AGENT_FIELD_ORDER first. Only the label lookup for an already-extracted field is what's optional.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.tsx
```diff
@@ -146,6 +148,34 @@ function MetaRow({ provider, model }: { provider?: string; model?: string }) {
...
+
+function AgentSection({ id, name, version, description }: GenAiAgent) {
+  const agent: GenAiAgent = { id, name, version, description };
+  return (
+    <div className="GenAITab--section">
+      <h3 className="GenAITab--sectionTitle">Agent</h3>
+      {(Object.keys(agent) as Array<keyof GenAiAgent>).map(key => {
+        const value = agent[key];
+        if (!value) return null;
+        return (
```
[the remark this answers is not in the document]
swetalin-10: Fixed in 38e03b73 - changed `if (!value) return null` to `if (value == null) return null`, matching TokensRow's convention, so a legitimate empty-string field isn't silently dropped.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.tsx
```diff
@@ -146,6 +148,34 @@ function MetaRow({ provider, model }: { provider?: string; model?: string }) {
...
+  id: 'ID',
+  description: 'Description',
+};
+
+function AgentSection({ id, name, version, description }: GenAiAgent) {
+  const agent: GenAiAgent = { id, name, version, description };
+  return (
+    <div className="GenAITab--section">
+      <h3 className="GenAITab--sectionTitle">Agent</h3>
+      {(Object.keys(agent) as Array<keyof GenAiAgent>).map(key => {
```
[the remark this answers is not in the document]
swetalin-10: Same underlying issue as the other review pass flagged on this file - fixed together in 38e03b73 via the explicit AGENT_FIELD_ORDER array described in my reply on that thread.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/genAiData.ts
```diff
@@ -246,6 +257,15 @@ type SectionBuilder = (get: GetAttr) => GenAiSection | undefined;
...
 // order), not the span's attributes (whose order is instrumentation-dependent).
 const REGISTRY: SectionBuilder[] = [
+  get => {
+    const id = asString(get('gen_ai.agent.id'));
+    const name = asString(get('gen_ai.agent.name'));
+    const version = asString(get('gen_ai.agent.version'));
+    const description = asString(get('gen_ai.agent.description'));
+    return id || name || version || description
+      ? { type: 'agent', data: { id, name, version, description } }
+      : undefined;
```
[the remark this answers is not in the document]
swetalin-10: Good catch - fixed in 24879b45. The truthy check meant `get()` had already claimed the four keys, so a legitimate empty-string value was dropped by the section builder AND excluded from Other GenAI Attributes (already claimed), losing it entirely. Switched to a presence check (`!== undefined`) and added a regression test (`still produces an agent section for a legitimate empty-string field, instead of silently dropping it`) covering exactly this case.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.tsx
```diff
@@ -132,6 +133,7 @@ function MessageBlock({
 function MetaRow({ provider, model }: { provider?: string; model?: string }) {
   return (
     <div className="GenAITab--meta">
+      <span className="GenAITab--metaPrefix">LLM:</span>
```
anshgoyalevil: Verified via live UI rendering (uploaded a synthetic trace with an `invoke_agent` span carrying `gen_ai.provider.name=anthropic` but no `gen_ai.request.model`/`gen_ai.response.model`): the meta row renders `LLM: Provider anthropic` on this Agent span's GenAI tab, even though it isn't an LLM call.

`extractMetaSection` in `genAiData.ts` builds this section whenever *any* span has `gen_ai.provider.name`/`gen_ai.system` or a request/response model attribute, regardless of `classifySpan()`'s result (`AGENT`, `TOOL_CALL`, `RETRIEVAL`, etc. can all carry a provider attribute per the OTel GenAI semconv). Before this PR, `MetaRow` had no caption at all, so it was span-kind-agnostic and accurate for any kind; now the hardcoded `LLM:` prefix asserts something that isn't always true for non-LLM_CALL spans.

Consider gating the caption (or the whole meta section) on `classifySpan(span) === 'LLM_CALL'`, or using a more neutral label like `Model:` for the meta row so it isn't misleading on Agent/Tool/Retrieval spans that also report their backing provider/model.

swetalin-10: Fixed in ae479882. Gated the "LLM:" caption on span.genAIKind === 'LLM_CALL' rather than removing the whole meta section for non-LLM spans - the provider/model info itself is still useful context on Agent/Tool/Retrieval spans, so it stays visible, just without the misleading label.

Re-verified live using your exact repro: uploaded a synthetic invoke_agent span with gen_ai.provider.name=anthropic and no request/response model. The GenAI tab's Agent section now shows just "Provider anthropic" with no "LLM:" prefix (confirmed via DOM inspection: hasLlmPrefix: false). Also verified the positive case wasn't broken - a real chat/gpt-4o span still shows "LLM: Provider openai Model gpt-4o".

Added a regression test mirroring your repro (invoke_agent + provider, no LLM: prefix, provider still visible) and updated the existing LLM: caption test to use an actual LLM_CALL span (gen_ai.operation.name: chat) rather than an ambiguous one with no operation name. Also updated the test's makeSpan() helper to derive genAIKind via classifySpan() the same way production code does, so tests can't silently diverge from real span behavior. 196/196 tests pass in SpanDetail, lint and typecheck clean.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.tsx
```diff
@@ -146,6 +152,40 @@ function MetaRow({ provider, model }: { provider?: string; model?: string }) {
...
+  id: 'ID',
+  description: 'Description',
+};
+
+// Explicit rather than Object.keys(agent) - render order shouldn't depend on
+// object literal insertion order, and this avoids an unsafe keyof cast.
+const AGENT_FIELD_ORDER: ReadonlyArray<keyof GenAiAgent> = ['id', 'name', 'version', 'description'];
+
+function AgentSection({ id, name, version, description }: GenAiAgent) {
+  const agent: GenAiAgent = { id, name, version, description };
```
yurishkuro: why decompose the struct first and then re-compose? Would't this work?
```suggestion
function AgentSection(agent: GenAiAgent) {
```

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.tsx
```diff
@@ -146,6 +152,40 @@ function MetaRow({ provider, model }: { provider?: string; model?: string }) {
   );
 }
 
+// Cosmetic only: a key missing from here still renders under its raw field name
+// instead of a friendly label. This does NOT mean a new gen_ai.agent.* attribute
+// shows up automatically - it must first be added to GenAiAgent, extracted by the
+// agent builder in genAiData.ts, and listed in AGENT_FIELD_ORDER below before it
+// renders at all; only the label lookup itself is optional.
```
yurishkuro: a very heavy and confusing explanation.  How is it even relevant for a constant that maps keys to labels?

```suggestion
// Map attribute keys to more reader-friendly labels.
```

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.tsx
```diff
@@ -291,8 +331,10 @@ export default function GenAITab({ span }: Props): React.ReactElement {
     <div className="GenAITab">
       {sections.map(section => {
         switch (section.type) {
+          case 'agent':
+            return <AgentSection key="agent" {...section.data} />;
           case 'meta':
-            return <MetaRow key="meta" {...section.data} />;
+            return <MetaRow key="meta" {...section.data} isLlmCall={span.genAIKind === 'LLM_CALL'} />;
```
yurishkuro: why is it called MetaRow if seems LLM-only?

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.tsx
```diff
@@ -291,8 +331,10 @@ export default function GenAITab({ span }: Props): React.ReactElement {
     <div className="GenAITab">
       {sections.map(section => {
         switch (section.type) {
+          case 'agent':
+            return <AgentSection key="agent" {...section.data} />;
```
yurishkuro: section, row, ... -  I suggest using a single term for these components, like AgentDetails, LLMDetails, TokenDetails, etc.

yurishkuro:
image


Can't say I like that - why can't we use the normal accordion, just with a custom caption? And accordion also has the ability to give a preview when collapsed, so if we put `name` first it would give high signal without even expanding the section. 

In fact, why don't we use it for all similar elevated sections, why introduce custom components? Using accordion keeps both the style and the user experience consistent.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.tsx
swetalin-10: Fixed in af8a4342 - takes `agent: GenAiAgent` directly now, no destructure/recompose.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.tsx
swetalin-10: Fair - fixed in af8a4342, replaced with your suggested one-liner.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.tsx
swetalin-10: Good point - renamed to LLMDetails in af8a4342, since this only ever renders LLM provider/model.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.tsx
swetalin-10: Done in af8a4342 - unified all of them onto `Details`: AgentSection -> AgentDetails, MetaRow -> LLMDetails, TokensRow -> TokenDetails, ConversationSection -> ConversationDetails, ToolCallSection -> ToolCallDetails, UnknownSection -> UnknownDetails.

swetalin-10:
Fair, and done in 86d1a850. Agent (and Tool Call, and the defensive Unknown fallback) now render through AccordionAttributes instead of one-off components, with `name` moved first in the field order so the collapsed preview reads `Agent: Name = jaeger-gemini-sidecar | Version = 0.1.0` without expanding, same pattern as Other GenAI Attributes now.

One exception: I kept Conversation as its own component. It renders a per-message Markdown/JSON/Plain format toggle and a copy button per message - none of that fits AccordionAttributes/AttributesTable's flat key-value + JSON-tree model, and collapsing it into that shape would mean dropping the format switching. Happy to revisit if you see a way to keep that functionality through the accordion, but as-is it seemed like the wrong tradeoff to force.

LLM/Tokens stayed as inline chip rows too, since those are single-line labeled groups, not collapsible sections like Agent/Tool Call - not the same shape you were pointing at.

Verified against a real trace, not just unit tests: uploaded the same jaeger-gemini-sidecar trace and confirmed the Agent accordion collapses/expands and the preview line renders as expected.

yurishkuro:
> One exception: I kept Conversation as its own component. It renders a per-message Markdown/JSON/Plain format toggle and a copy button per message - none of that fits AccordionAttributes/AttributesTable'

I disagree but we can keep it out of scope - I mentioned elsewhere that the accordion should have the same rich media rendering capabilities that we're using in GenAI, instead of singling out GenAI traces only.

yurishkuro:
please update the screenshot

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.tsx
```diff
@@ -146,6 +161,48 @@ function MetaRow({ provider, model }: { provider?: string; model?: string }) {
...
+    () =>
+      makeAttributes(
+        AGENT_FIELD_ORDER.filter(key => agent[key] != null).map(
+          (key): IAttribute => ({ key: AGENT_LABELS[key] ?? key, value: agent[key] as AttributeValue })
+        )
+      ),
+    [agent]
+  );
+  return (
+    <AccordionAttributes
```
yurishkuro, line 200: only this one component changed, my comment applied to all sections in GenAI to be rendered as accordion. What's the plan?

yurishkuro: just LLMDetails is not using accordion (Conversation change is deferred)

swetalin-10: Fixed in 44d2e471 - `LLMDetails` now renders through `AccordionAttributes` too, same as `AgentDetails`/`ToolCallDetails`/`UnknownDetails`. Open by default (it's primary content, like Agent), label doubles as the caption: `LLM` for a span `classifySpan()` actually identifies as an LLM call, `Model` otherwise (per Ansh's suggestion) - collapsing shows the usual `LLM: Provider=... Model=...` preview.

`Tokens:` stays as its own inline row - it's a single-line labeled group, not a collapsible section, and wasn't part of "all sections" per your own follow-up here. `Conversation` stays out of scope as agreed.

Verified against a real trace via the dev server's upload feature, not just the rewritten unit tests - uploaded a synthetic-but-schema-accurate trace with an `invoke_agent` root span and a child `chat gpt-4o` span, confirmed the LLM accordion opens/collapses correctly and `Tokens:` renders unchanged below it. Updated the PR screenshot to show this.

swetalin-10:
Understood on the broader point - agreed the accordion should eventually support the same rich-media rendering (Markdown/JSON toggle, per-message actions) generally, rather than GenAI building its own one-off for it. Keeping that out of scope here as you said; happy to pick it up as a follow-up if you'd like one filed.

On the actual actionable part of your comment - `LLMDetails` now uses the shared accordion too, fixed in 44d2e471 (details in the inline thread). Also updated the PR description's screenshot to reflect the current UI.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/index.tsx
swetalin-10: Follow-up correction on my note above: I got access to the actual `bde4caeb` trace after posting that comment. Re-verified with it instead of the synthetic one - the first `generate_content gemini-2.5-flash` span (Provider `gcp.gen_ai`, Model `gemini-2.5-flash`, input 3,803/output 45 tokens) matches the issue's first screenshot exactly, and the LLM accordion renders correctly against it. Updated the PR description and screenshot to reflect this; no more synthetic substitute in the testing narrative.

yurishkuro (approved):

PR #4247, opened 2026-07-21
fix(date): Prevent incorrect partial rounding-up of durations

### LINKED ISSUE #4246: formatDuration renders malformed output like "1h 60m" at unit boundaries
Image

## Description

`formatDuration` (`packages/jaeger-ui/src/utils/date.ts`) can emit an overflowed secondary unit — e.g. `1h 60m` or `23h 60m` — instead of carrying the rounded-up remainder into the next unit (`2h`, `1d`).

## Reproduction

```js
formatDuration(7176000000);  // 1h 59m 36s   -> "1h 60m"  (expected "2h")
formatDuration(86380000000); // 1 day - 20s  -> "23h 60m" (expected "1d")
```

## Root cause

The secondary value is computed as `Math.round((duration / secondaryUnit.microseconds) % ofPrevious)`. `Math.round` runs *after* the modulo, so a remainder such as `59.6` rounds up to `60` and is printed verbatim rather than carrying into the primary unit.

## Severity

Low. It only triggers within roughly the last 30 seconds before a unit rollover (e.g. `59m30s`–`59m59s`) and only for durations of a minute or more, so it is an uncommon, cosmetic issue. It is still user-visible in duration tooltips and relative-offset displays.

thisis-manan:
Before
Screenshot 2026-07-21 at 9 07 57 PM
After
Screenshot 2026-07-21 at 9 20 55 PM

- Resolves #4246

`formatDuration` could render an overflowed secondary unit such as `1h 60m` or `23h 60m` instead of carrying into the next unit.

Root cause: `secondaryValue` applied `Math.round` after the modulo, so a remainder that rounded up to a full 60 (or 24) was emitted as-is.

## Description of the changes

Round the duration to the nearest secondary unit first, then re-select the units on the rounded value so the remainder carries into the primary unit and the primary/secondary values stay consistent.

| input | before | after |
| --- | --- | --- |
| 1h 59m 36s | `1h 60m` | `2h` |
| 1m 59.6s | `1m 60s` | `2m` |
| 1 day − 20s | `23h 60m` | `1d` |

Existing formatted output (`2h 31m`, `10d 14h`, decimal ms/s, etc.) is unchanged.

## Severity

Low. The bug only triggers within ~30s of a unit rollover and only for durations of a minute or more, so it is an uncommon, cosmetic rendering issue — but it is user-visible where durations are shown.

Added carry-case assertions to `date.test.js` covering the hour, minute, and day boundaries; the existing cases are untouched and still pass.

## Checklist

- [x] I have signed all commits (DCO)

### DIFF
--- a/packages/jaeger-ui/src/utils/date.test.js
+++ b/packages/jaeger-ui/src/utils/date.test.js
@@ -65,6 +65,20 @@ describe('formatDuration', () => {
     const input = 0;
     expect(formatDuration(input)).toBe('0μs');
   });
+
+  it('carries a rounded-up secondary unit into the primary unit', () => {
+    expect(formatDuration(ONE_HOUR + 59 * ONE_MINUTE + 36 * ONE_SECOND)).toBe('2h');
+    expect(formatDuration(ONE_MINUTE + 59 * ONE_SECOND + 600 * ONE_MILLISECOND)).toBe('2m');
+  });
+
+  it('does not carry when the secondary unit rounds down', () => {
+    expect(formatDuration(ONE_HOUR + 59 * ONE_MINUTE + 29 * ONE_SECOND)).toBe('1h 59m');
+    expect(formatDuration(ONE_MINUTE + 59 * ONE_SECOND + 400 * ONE_MILLISECOND)).toBe('1m 59s');
+  });
+
+  it('carries across the day boundary', () => {
+    expect(formatDuration(ONE_DAY - 20 * ONE_SECOND)).toBe('1d');
+  });
 });
 
 describe('getSuitableTimeUnit', () => {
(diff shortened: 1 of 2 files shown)


thisis-manan:
Hi @yurishkuro, 

This PR is ready for review whenever you have a moment! It fixes the `formatDuration` overflow bug (#4246) and all core unit tests are passing. 

I noticed the `Verify PR Label` check is failing, but I don't have permissions to add labels to the PR. Could you please help apply the correct label or let me know if I need to change anything else when ever you have time .

Thank you!

yurishkuro (commented):
please restore the comments

yurishkuro (approved):

PR #4248, opened 2026-07-21
fix(ui): use Select for AltViewOptions to fix keyboard navigation (#4…

### LINKED ISSUE #4178: [Bug]: Arrow keys in Trace view selector dropdown resize minimap instead of navigating options.

Myself-Praveen:
- Resolves #4178

## Description of the changes
- Replaced the Ant Design `Dropdown` and `Button` components with `Select` inside `AltViewOptions.tsx`. This aligns the view-switching UI with the existing `SearchableSelect` components, allowing it to natively intercept arrow key events so they do not propagate down to the SpanGraph minimap.
- Migrated JSON view links from embedded `<a>` tags into the `Select`'s `onChange` handler via `window.open` to maintain keyboard accessibility.
- Cleaned up conflicting custom button styling in `AltViewOptions.css`.
- Updated `AltViewOptions.test.jsx` to mock the `Select` component and reliably assert on the updated `window.open` behavior.

- Updated the existing Vitest unit tests in `AltViewOptions.test.jsx` to accommodate the `Select` UI paradigm. All tests pass locally.

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TracePageHeader/AltViewOptions.css
+++ b/packages/jaeger-ui/src/components/TracePage/TracePageHeader/AltViewOptions.css
@@ -4,16 +4,5 @@ SPDX-License-Identifier: Apache-2.0
 */
 
 .AltViewOptions {
-  border: 1px solid var(--border-default);
-  border-radius: 4px;
-  height: 32px;
-  line-height: 30px;
   margin-right: 1rem;
-  padding: 0 8px;
-  display: flex;
-  align-items: center;
-}
-
-.AltViewOptions svg {
-  margin-left: 3px;
 }
--- a/packages/jaeger-ui/src/components/TracePage/TracePageHeader/AltViewOptions.test.jsx
+++ b/packages/jaeger-ui/src/components/TracePage/TracePageHeader/AltViewOptions.test.jsx
@@ -126,7 +126,11 @@ describe('AltViewOptions', () => {
     fireEvent.click(screen.getByTestId('menu-item-trace-json'));
 
     expect(trackJsonView).toHaveBeenCalled();
-    expect(windowOpenSpy).toHaveBeenCalledWith('/api/traces/test trace ID?prettyPrint=true', '_blank');
+    expect(windowOpenSpy).toHaveBeenCalledWith(
+      '/api/traces/test trace ID?prettyPrint=true',
+      '_blank',
+      'noopener,noreferrer'
+    );
   });
 
   it('handles unadjusted JSON view correctly', () => {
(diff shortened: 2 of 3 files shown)


Myself-Praveen:
I've updated the PR to address the AI reviewer's feedback! Let me know if there's anything else you need me to change.

PR #4254, opened 2026-07-24
refactor(trace): Combine span decoration matching into one registry

### LINKED ISSUE #4223: [chore]: Combine span decoration mechanisms
### Requirement

We currently have two mechanisms for span decorations in the trace timeline: the icons (e.g. #4221) and the span pills.

They both derive the signals by detecting various span attributes. I propose we consolidate the configuration of matching rules into a single place, the icons vs. pills only need to differ in the target outcome they produce (icon or a pill), but the matching configuration can be nearly identical and in one place.

cc @swetalin-10 @udita-0707 

## Update on 2026-08-13: proposed design

The shorthand above left too much open — #4254 read "single place" as two lists in one file, which is a fair reading and does not achieve the goal. Concretely, here is the shape I am after. Working on #4254 is what surfaced the sharp edge in it (the GenAI axis below), so this is a refinement of the ask, not a rejection of that work.

### One entry per attribute namespace, owning both outcomes

In `packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/spanDecorations.ts`:

```ts
type IPillSource = {
  label: string;
  attrKeys: readonly string[];       // first key present wins; aliases cover OTel renames
  isError?: (value: string) => boolean;
};

type ISpanDecoration = {
  namespace: string;                 // matches any attribute key with this prefix
  icon?: IconType;                   // category glyph, where the namespace has one
  pills?: readonly IPillSource[];    // value-bearing keys, in emission order
};

export const SPAN_DECORATIONS: readonly ISpanDecoration[] = [
  { namespace: 'db', icon: DbIcon, pills: [{ label: 'db.system', attrKeys: ['db.system.name', 'db.system'] }] },
  { namespace: 'http', icon: HttpIcon, pills: [
      { label: 'http.status_code', attrKeys: ['http.status_code', 'http.response.status_code'], isError: is5xx },
      { label: 'http.method', attrKeys: ['http.method', 'http.request.method'] },
  ] },
  { namespace: 'messaging', icon: MessagingIcon },
  { namespace: 'rpc', icon: RpcIcon, pills: [{ label: 'rpc.system', attrKeys: ['rpc.system.name', 'rpc.system'] }] },
  { namespace: 'gen_ai', pills: [{ label: 'gen_ai.request.model', attrKeys: [GEN_AI_REQUEST_MODEL] }] },
];
```

List position carries the ordering for both outcomes: the first matching entry supplies the icon, and pills emit in list order (within an entry, in its own order). No separate priority or weight fields.

This changes pill order in one case: on a span carrying both `db.*` and `http.*` values, `db.system` now renders ahead of the http pills. That is accepted — no other user-visible change.

### GenAI kind is a second axis, and stays separate

Do not fold GenAI into the namespace table. The two work differently, at different times:

| | how the icon is chosen | when |
| --- | --- | --- |
| `db` / `http` / `messaging` / `rpc` | presence of an attribute-key namespace | render time, from `attributes` |
| `gen_ai` | `gen_ai.operation.name` (plus secondary signals) → kind | transform time, stored as `span.genAIKind` |

Consequences:

- `operation → kind` is classification, not decoration. It stays private to `utils/genai/detect.ts`, which `OtelSpanFacade`'s constructor already calls per span.
- `kind → icon + label` is presentation. That table (`GEN_AI_KIND_META`) belongs next to the registry.
- The registry has no `gen_ai` icon entry — GenAI spans get their glyph from the kind axis, and the resolver prefers it when both would match (#4217).
- `utils/` and `model/` must not import from `components/`. Dependencies run one way.

### Resolvers and rendering

`spanDecorations.ts` owns the data and both resolvers: `getSpanDecorationIcon(span)` (GenAI kind first, else first matching namespace) and `getSpanPillsForSpan(span)` (walks the registry in order, extracting values). `spanPills.tsx` keeps only the `SpanPill` component.

One presentational component, `SpanDecorationIcon.tsx`, renders either resolver outcome: wrapped in a `Tooltip` with `role="img"` and an `aria-label` when a label is present (GenAI today), a bare `aria-hidden` glyph otherwise.

Files added: `spanDecorations.ts`, `SpanDecorationIcon.tsx` (+ its CSS, from `GenAISpanIcon.css`).
Files deleted: `span-icons.ts`, `GenAISpanIcon.tsx`.

### Tests

Assert behavior, not the tables: a span with both `db.*` and `http.*` resolves to the db icon; a 5xx status yields an error pill; `gen_ai` plus `http` yields exactly one icon; a span with both http pills emits both, in order. Assertions that restate the registry literals are not useful — they fail on every intentional edit and pass through real regressions.

### Non-goals

- The `messaging.system` pill — separate follow-up.
- Any user-visible change beyond the `db.system` pill ordering noted above.
- Moving GenAI classification, or the `spanPillsEnabled` config toggle.

udita-0707:

- Resolves #4223
- Timeline icons and pills each had their own attribute-matching setup; adding a signal meant touching multiple places and risked redundant decorations

## Description of the changes

Implements the design in #4223:

- One `SPAN_DECORATIONS` registry in `spanDecorations.ts`: each namespace owns its icon and/or pills; **list position** is the only ordering (no separate priority / `PILL_SOURCES` list)
- Both resolvers live there: `getSpanDecorationIcon` (GenAI kind first, else first matching namespace with an icon) and `getSpanPillsForSpan`
- `SpanDecorationIcon` is the single presentational component (replaces `span-icons.ts` + `GenAISpanIcon`)
- `spanPills.tsx` keeps only `SpanPill` + `spanPillsEnabled`
- GenAI stays a second axis: `OPERATION_TO_KIND` stays private in `utils/genai/detect.ts` (classification at transform time); `GEN_AI_KIND_META` stays next to the registry (presentation)
- `utils/` / `model/` do not import from `components/`

**Intentional user-visible deltas** (called out in #4223 / review):

1. On a span with both `db.*` and `http.*` values, `db.system` pills render **before** http pills
2. Decoration icon renders **before** the error icon (was after for GenAI on main)

**Out of scope:** `messaging.system` pill; moving GenAI classification; changing `spanPillsEnabled`

1. Unit: `spanDecorations.test.ts`, `SpanDecorationIcon.test.tsx`, `spanPills.test.ts`, `SpanBarRow.test.jsx` (incl. #4217 one-icon / http+gen_ai, decoration-before-error), `detect.test.ts`
2. `pnpm run fmt`, `pnpm run lint`, `pnpm test` (as exercised in CI)

## Checklist

- [ ] **None**


udita-0707:
@yurishkuro @swetalin-10 PTAL

swetalin-10:
Went through this — checked out the current head, ran it, and it renders correctly. Solid work on the consolidation itself.

Still the one open thing though — on the issue you'd asked about `messaging.system` and we agreed to add it there. It's not in this PR and there's been no explanation on the issue or the PR for why. What's the reason you dropped it? If it's a "do it after" call, that's fine. Otherwise happy to see it added here.

udita-0707:
Thank you @swetalin-10 for taking a look.
Actually it is planned as a follow-up. I wanted to keep the consolidation work separate from the `message.system` pill changes.
I have mentioned that in the PR description already that it is out of scope for this PR.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/spanDecorations.ts
```diff
@@ -0,0 +1,136 @@
...
+];
+
+/**
+ * Value-bearing pills with no matching namespace icon in this registry.
+ * GenAI category icons come from GenAISpanIcon / classifySpan instead.
+ */
+export const PILL_ONLY_SOURCES: readonly IPillSource[] = [GEN_AI_MODEL_PILL];
+
+/** Namespace → icon lookups derived from {@link DECORATION_FAMILIES}. */
+export const NAMESPACE_ICONS: readonly INamespaceIcon[] = DECORATION_FAMILIES.map(
```
yurishkuro: this just renames one array for another, why not define directly? Having the indirection increases cognitive overhead, what is "decoration families"?

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/spanDecorations.ts
```diff
@@ -0,0 +1,136 @@
...
+ * Shared span decoration registry.
+ *
+ * Icons and pills derive signals from the same attribute families; they only
+ * differ in outcome:
+ * - icon: namespace presence → one category glyph (priority when several match)
+ * - pill: specific keys → attribute value (only when it adds info the icon cannot)
+ *
+ * Rules encoded here:
+ * - No pill for a pure category signal already covered by an icon (e.g. span.kind).
+ * - gen_ai has no namespace icon here: GenAISpanIcon + classifySpan handle that (#4217).
```
yurishkuro: what is the motivation for splitting out the GenAI mechanism?

udita-0707: keeping classifySpan / GenAISpanIcon separate on purpose. Namespace icons are presence-only; GenAI picks a kind from`gen_ai.operation.name` and that kind is also used outside the timeline (GenAI tab / isGenAITrace). A generic gen_ai namespace icon here is what caused the double-icon bug in #4217. The model name still belongs in PILL_SOURCES as a value pill.

yurishkuro: > Namespace icons are presence-only

key+value check is just a generalization of "key exists" check. I am not convinced by the argument.

> A generic gen_ai namespace icon here is what caused the double-icon bug in https://github.com/jaegertracing/jaeger-ui/issues/4217

Even more reason to have a single mechanism instead of two.

udita-0707: I see your point. Presence-only vs key+value doesn't really feel like a meaningful distinction.
What I'm thinking is, if we're moving towards a single mechanism in `spanDecorations.ts`, we could structure it like this:

1. Have a single ordered list of rules, where each rule defines both the match and the outcome (icon and/or pill):

   * `anyKeyPrefix` (`db.`, `http.`, etc.) → namespace icon
   * key + value (`gen_ai.operation.name` → `chat`, `invoke_agent`, etc.) → GenAI-specific icon
   * `gen_ai.` prefix as a fallback → generic GenAI icon
   * specific keys → pills (`http.status_code`, `db.system`, `gen_ai.request.model`, etc.)

2. Use a single icon resolver in `SpanBarRow`, with GenAI-specific rules taking priority over namespace rules. That way we avoid rendering `getSpanIconComponent` and `GenAISpanIcon` together.

3. `classifySpan` can continue handling the GenAI tab and `isGenAITrace`, but instead of maintaining its own mapping, it can read the same operation → kind registry so there's only one source of truth.

here's what I plan:

* In this PR, flatten things to `NAMESPACE_ICONS` + `PILL_SOURCES` and remove `DECORATION_FAMILIES` / `PILL_ONLY_SOURCES`.
* Then either as an immediate follow-up (or in this PR if you'd prefer), fold the GenAI icon logic into the same matcher and remove the dual-render path entirely.

If you'd prefer, I can also include the GenAI unification in this PR so everything is reviewed together.

@yurishkuro ,would this align with what you had in mind?

yurishkuro: just include unification

udita-0707: okay noted!

udita-0707:
@yurishkuro this is up for review, please have a look

yurishkuro (changes requested):
Thanks for this, and sorry for the moving target — reviewing it made clear my issue text was too vague, so I have written an actual design into #4223. Two lists in one file was a fair reading of what I originally wrote.

The deltas from what you have: one entry per namespace owning both its icon and its pills (list position replaces the `priority` field and the separate `PILL_SOURCES` order), `GEN_AI_OPERATION_TO_KIND` back to being private in `utils/genai/detect.ts`, and `span-icons.ts` plus `GenAISpanIcon.tsx` collapsing into one presentational component. The reasoning for the middle one is worth a look: `classifySpan` runs in `OtelSpanFacade`'s constructor at transform time, so that map is classification rather than decoration, and importing it from a component makes every trace transform depend on `TraceTimelineViewer` — and on three `react-icons` packages.

Worth keeping as-is: the resolver preferring GenAI kind over namespace icons is exactly right and is now written into the design, `GEN_AI_KIND_META` is in the right place, and the three `getSpanDecorationIcon` tests are the ones that earn their keep.

Two things the design does not cover, both from this diff:

- The GenAI icon now renders *before* the error icon; on `main` it came after. I think that is an improvement, but it needs saying in the description and pinning with a test rather than shipping under "no user-visible behavior change".
- Most of `spanDecorations.test.ts` restates the literals above it — priorities unique and ascending, pill labels in a fixed order, no category-only pill. Those pass through any real regression, and several stop meaning anything once the two lists merge. #4223 lists the behavior cases I would rather have.

Nits: title should be `refactor(trace): …` so it maps to `changelog:refactoring`; it needs a `changelog:` label, which I will add once the title settles; and `if (bestPriority === 0) break;` is unreachable now that priorities start at 1.

---

udita-0707:
PTAL @yurishkuro

JnRouvignac:
@yurishkuro if you had some time, that would be great if you could take a look at this PR.
This would unblock progress on #4017.

Thank you for your guidance and reviews! And thanks @udita-0707 for pushing through.

udita-0707:
@JnRouvignac I’m also still waiting for the review process to move forward. Thanks for flagging this!

cc: @yurishkuro

yurishkuro (approved):

JnRouvignac:
> @JnRouvignac I’m also still waiting for the review process to move forward. Thanks for flagging this!

I am following your PRs along until the last one. Thank you very much for pushing through 💪🏻

PR #4257, opened 2026-07-24
fix(deps): Update brace-expansion to 5.0.8

jkowall:
## Summary

- Merge the current `main`, including #4259's removal of `depcheck`.
- Resolve the one remaining `brace-expansion` dependency to `5.0.8` with a single pnpm override.
- Regenerate `pnpm-lock.yaml`.

## Why

After #4259 removed `depcheck`, the old `minimatch` 3.x and 7.x dependency paths disappeared. The remaining path is:

```text
minimatch@10.2.5 -> brace-expansion@5.0.6
```

`brace-expansion@5.0.8` fixes [GHSA-3jxr-9vmj-r5cp](https://github.com/advisories/GHSA-3jxr-9vmj-r5cp) and [GHSA-mh99-v99m-4gvg](https://github.com/advisories/GHSA-mh99-v99m-4gvg). Since `minimatch@10` natively uses the current `brace-expansion` API, no package patch or compatibility shim is needed.

This is a transitive development-tool dependency. Application runtime dependencies and source code are unchanged.

Additional audit findings remain separate from this PR.

## Validation

- `pnpm install --frozen-lockfile`
- `pnpm why brace-expansion -r`, all paths resolve through `minimatch@10.2.5` to `brace-expansion@5.0.8`
- `pnpm run fmt`
- `pnpm run lint`
- `pnpm test`, 3,603 tests passed
- `pnpm run build`
- `pnpm audit --json`, no `brace-expansion` advisories reported

PR #4258, opened 2026-07-24
fix(deps): Patch transitive advisories

jkowall:
## Summary

- Refresh `pnpm-lock.yaml` to select patched transitive dependency versions.
- Resolve `js-yaml` to 4.3.0, `shell-quote` to 1.9.0, and `fast-uri` to 3.1.4.
- Keep `package.json` unchanged; no `pnpm.overrides` are needed.

## Why

After #4259 removed `depcheck` and #4257 landed, four high-severity Dependabot alerts remain in the development dependency graph:

- `js-yaml`: [GHSA-52cp-r559-cp3m](https://github.com/advisories/GHSA-52cp-r559-cp3m)
- `shell-quote`: [GHSA-395f-4hp3-45gv](https://github.com/advisories/GHSA-395f-4hp3-45gv)
- `fast-uri`: [GHSA-v2hh-gcrm-f6hx](https://github.com/advisories/GHSA-v2hh-gcrm-f6hx)
- `fast-uri`: [GHSA-4c8g-83qw-93j6](https://github.com/advisories/GHSA-4c8g-83qw-93j6)

All three patched versions satisfy their existing parent dependency ranges, so a lockfile-only refresh fixes the vulnerable paths without the fragile override approach from the original revision.

Dependabot alert #290 is intentionally separate. GHSA-qwww-vcr4-c8h2 only affects unstable React Router RSC APIs, which Jaeger UI does not use.

## Validation

- `pnpm install --frozen-lockfile`
- `pnpm why fast-uri -r`
- `pnpm why shell-quote -r`
- `pnpm why js-yaml -r`
- `pnpm audit --json` — only the non-applicable React Router RSC advisory remains
- `pnpm test` — 267 Jaeger UI files / 3,497 tests and 9 Plexus files / 115 tests passed
- `pnpm run build`

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -3709,8 +3709,8 @@ packages:
   fast-levenshtein@2.0.6:
     resolution: {integrity: sha512-DCXu6Ifhqcks7TZKY3Hxp3y6qphY5SJZmrWMDrKcERSOXWQdMhU9Ig/PYrzyw/ul9jOIyh0N4M0tbC5hodg8dw==}
 
-  fast-uri@3.1.2:
-    resolution: {integrity: sha512-rVjf7ArG3LTk+FS6Yw81V1DLuZl1bRbNrev6Tmd/9RaroeeRRJhAt7jg/6YFxbvAQXUCavSoZhPPj6oOx+5KjQ==}
+  fast-uri@3.1.4:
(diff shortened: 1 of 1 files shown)


yurishkuro:
I removed the major source of CVAs by deleting depcheck in https://github.com/jaegertracing/jaeger-ui/pull/4259. Let's see if any advisories still stick around after that before we do this invasive & fragile patching.

yurishkuro (approved):

PR #4260, opened 2026-07-25
perf(tests): reduce redundant renders in operationDetailsTable tests

Harshbansal06:
- Partially addresses item 4 of #3996 (`<OperationTableDetails>` half only — `<TraceTagOverview>` left for a separate PR per the issue's one-PR-per-item guidance)

## Description of the changes
- Reviewed every test in `operationDetailsTable/index.test.jsx` to identify which ones share identical render props with no interactions in between (safe to consolidate) vs. which ones use unique/cloned data or fire interactions like hover/click/sort (must stay isolated to avoid state leaking between tests).
- Found two genuine consolidation opportunities and merged each pair into a single `it()` with one render:
  - `does not explode` + `Loading indicator is displayed` — identical props (loading=true, default), zero interactions.
  - `Table rendered successfully` + `render No data table` — these rendered with identical props (loading=false, no data override) and asserted the exact same thing (`container.querySelector('table')`), so they were duplicate coverage rather than distinct cases.
- All other tests in the file (per-case data mutations, `highlight the row`, `sort row`, `Should track all events`) are untouched — sharing renders across those would risk order-dependent flakiness since they mutate component state via interactions.
- Note: initially tried sharing a render across multiple `it` blocks via `beforeAll`, but React Testing Library's automatic cleanup unmounts the rendered component after every individual `it`, not just at the end of the describe block — so the second test in each pair saw a detached container and failed. Switched to merging into single `it()` blocks with one inline render each, which achieves the same render-count reduction without conflicting with RTL's cleanup behavior.

- Ran `npx vitest run src/components/Monitor/ServicesView/operationDetailsTable/` before and after the change — 18 tests passing before, 16 passing after (2 consolidated, 0 dropped), full directory (27 tests across all 3 files) green with no regressions.
- Measured before/after timing per the issue's suggested method (`--reporter=json --outputFile=...`, compared via `jq`):

  | | Before | After |
  |---|---|---|
  | `operationDetailsTable/index.test.jsx` | 6384ms | 3756ms |
  | Test count | 18 | 16 |

## Checklist

See [[AI Usage Policy](https://github.com/jaegertracing/jaeger/blob/main/CONTRIBUTING_GUIDELINES.md#ai-usage-policy)](https://github.com/jaegertracing/jaeger/blob/main/CONTRIBUTING_GUIDELINES.md#ai-usage-policy).


jkowall (changes requested):
Test consolidation LGTM (Copilot Balanced also approved the coverage merge).

Blocked on DCO: the merge commit into this branch has no Signed-off-by. Please amend/re-sign so DCO is green (the feature commit may already be signed; the merge commit is the one failing).

PR #4261, opened 2026-07-26
fix(date): carry a rounded-up remainder instead of rendering a full unit

Qalipso:

`formatDuration` rounds the secondary unit, so the rounded value can reach the primary unit's base — a value that unit cannot hold:

| input | current | expected |
|---|---|---|
| `1h 59m 40s` | `1h 60m` | `2h` |
| `1m 59.7s` | `1m 60s` | `2m` |
| `1d 23h 40m` | `1d 24h` | `2d` |
| `59m 59.6s` | `59m 60s` | — |

The rounding itself is intended — `date.test.js` documents it as "displays a maximum of 2 units and rounds the last one". What is missing is the carry when that rounding overflows.

`formatDuration` backs span duration and relative-start tooltips in `TraceSpanView`, `TraceLogsView`, `FlamegraphTable` and `FlamegraphTooltip`, so the bad string is user-visible whenever a duration lands near a unit boundary.

## Description of the changes

Carry the overflow into the primary unit and drop the secondary — the path the function already takes when the remainder rounds to zero (`2d`). Rounding is preserved for everything that does not overflow: `2h 30m 30s` still renders as `2h 31m`.

- `vitest run src/utils/date.test.js` → **49 passed** (47 on `main`; the 2 added tests are the difference).
- Red→green: reverting only `date.ts` fails `carries a rounded-up remainder instead of printing a full unit` and passes the rest. The companion test `still rounds the secondary unit when it does not overflow` passes both before and after — it pins the behavior this change must not alter.
- Full `pnpm test`: **2951 passed** with this change vs **2949** on `main`, with an identical 144 failures / 24 failed files in both runs. Those failures are pre-existing in my local environment (`localStorage` undefined under the test environment in unrelated files) and are untouched by this change.
- `vp fmt --check` clean; `tsc -p packages/jaeger-ui/tsconfig.json` exits 0.

## Checklist

- [x] I have signed all commits (DCO)

## Note

I could not add the required `changelog:` label — that needs write access. Happy to adjust the title or scope if you would prefer a different label. I can also open a tracking issue first if that is the preferred flow here.

One residual worth naming: `23h 59m 40s` now renders as `24h` rather than the previous `23h 60m`. Promoting it to `1d` would mean re-running unit selection after the carry, which is a larger change than this fix — say the word if you would rather have that.

### DIFF
--- a/packages/jaeger-ui/src/utils/date.test.js
+++ b/packages/jaeger-ui/src/utils/date.test.js
@@ -65,6 +65,19 @@ describe('formatDuration', () => {
     const input = 0;
     expect(formatDuration(input)).toBe('0μs');
   });
+
+  it('carries a rounded-up remainder instead of printing a full unit', () => {
+    // Rounding the secondary unit up can reach the primary unit's base, which
+    // the secondary unit cannot hold ("1h 60m", "1m 60s", "1d 24h").
+    expect(formatDuration(1 * ONE_HOUR + 59 * ONE_MINUTE + 40 * ONE_SECOND)).toBe('2h');
+    expect(formatDuration(1 * ONE_MINUTE + 59.7 * ONE_SECOND)).toBe('2m');
+    expect(formatDuration(1 * ONE_DAY + 23 * ONE_HOUR + 40 * ONE_MINUTE)).toBe('2d');
+  });
+
+  it('still rounds the secondary unit when it does not overflow', () => {
+    expect(formatDuration(2 * ONE_HOUR + 30 * ONE_MINUTE + 30 * ONE_SECOND)).toBe('2h 31m');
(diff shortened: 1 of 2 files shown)


Qalipso:
Superseded by #4247, which landed the same fix for #4246 with a better approach — rounding the duration to the nearest secondary unit and re-selecting the units on the rounded value, rather than special-casing the carry. That also handles the case where the carry cascades past the primary unit (1 day − 20s → `1d`), which my version did not.

Closing this. Nothing here is worth keeping on top of `main`; the only test case not already covered there is a cascading carry (`1d 23h 40m` → `2d`) — happy to send that as a one-line test addition if it is useful, otherwise consider this resolved.

PR #4281, opened 2026-07-29
fix(trace): offset timeline header below page header

### LINKED ISSUE #4279: [bug]: Span Details header renders in the wrong place
### Requirement

Original view:

Image

Switch view to Graph

Image

Switch back to timeline

Image

In the last pic the span details are incorrectly rendered on the minimap

workeainc:
## Summary
- expose the measured trace-page header height to the rendered view section
- position the fixed timeline header below the trace-page header
- add a regression assertion for the inherited offset

Closes #4279

## Validation
- `pnpm run fmt`
- `pnpm run lint`
- `pnpm test` (3,620 passed)
- `pnpm run build`


yurishkuro:
sorry I didn't see this and went with a different fix. Thanks for contributing.

PR #4300, opened 2026-07-30
fix(graph): rebuild parent-child map per trace to fix stale self-time

anusha19murthy:
- Graph view shows wrong self-time for every trace after the first one viewed in a session.

## Description of the changes
- `calculateTraceDagEV.ts` had a module level cache, `parentChildOfMap`, that was only built once:

```ts
let parentChildOfMap: Record<string, IOtelSpan[]>;

function getBlockingChildSpans(parentID: string, trace: IOtelTrace): IOtelSpan[] {
  if (!parentChildOfMap) {
    parentChildOfMap = {};
    // ...builds map from trace.spans
  }
  return parentChildOfMap[parentID] || [];
}
```

  The `if (!parentChildOfMap)` guard only fires on the very first call. After that it stays truthy forever, so every trace opened afterward in the same session reuses the first trace's parent to child span data instead of its own. This silently produces wrong `selfTime` and `percentSelfTime` values in the graph view for every trace after the first.

- Fix: removed the module level variable and the two standalone functions (`getBlockingChildSpans`, `getChildOfDrange`). The parent to child map is now built fresh inside `calculateTraceDag` on every call, as a local const, so it can never leak between traces.

```ts
function calculateTraceDag(trace: IOtelTrace): TraceDag<TSumSpan & TDenseSpanMembers> {
  const baseDag = TraceDag.newFromTrace(trace);

  const blockingChildMap: Record<string, IOtelSpan[]> = {};
  trace.spans.forEach(s => {
    if (s.parentSpanID && s.kind !== SpanKind.CONSUMER) {
      (blockingChildMap[s.parentSpanID] ??= []).push(s);
    }
  });

  const getChildOfDrange = (parentID: string) => {
    const childrenDrange = new DRange();
    (blockingChildMap[parentID] || []).forEach(s => {
      childrenDrange.add(s.startTime, s.startTime + (s.duration <= 0 ? 0 : s.duration - 1));
    });
    return childrenDrange;
  };
  // ... rest unchanged
}
```

  The logic itself is unchanged, same CONSUMER filtering, same DRange boundary handling. Only the scope of the map changed, from module level to per call.

- Ran the existing test suite for this file, all 8 tests pass 

pnpm test -- calculateTraceDagEV

- Ran the full typecheck for the package, no errors:

pnpm tsc-lint

- Manually reproduced the bug before the fix: opened trace A in graph view (correct self-time), opened trace B in graph view without reloading (wrong, near 100% self-time on every node), confirmed against trace B's timeline view that it does have real child span duration. After the fix, repeated the same steps and trace B's self-time is correct.

## Checklist

- [x] **None**
- [ ] **Light**
- [ ] **Moderate**
- [ ] **Heavy**: AI helped identify the root cause and draft the fix, based on reading the actual source. I reviewed the diff line by line, ran the existing tests and typecheck locally, and manually reproduced the bug before and after the fix in a local Jaeger instance.

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceGraph/calculateTraceDagEV.ts
+++ b/packages/jaeger-ui/src/components/TracePage/TraceGraph/calculateTraceDagEV.ts
@@ -11,8 +11,6 @@ import { TDenseSpanMembers } from '../../../model/trace-dag/types';
 import { IOtelSpan, IOtelTrace, SpanKind, StatusCode } from '../../../types/otel';
 import { TSumSpan, TEv } from './types';
 
-let parentChildOfMap: Record<string, IOtelSpan[]>;
-
 export function mapNonBlocking(
   edges: TEdge[],
   nodes: TDagNode<TSumSpan & TDenseSpanMembers>[]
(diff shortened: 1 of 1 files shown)


jkowall:
Closing and reopening to refresh stale fork CI workflows (>~30d / stuck queued).

jkowall (changes requested):
The stale module-level map fix looks correct.

Please add a unit test that calls calculateTraceDag twice with two different traces in one run and asserts the second selfTime/percentSelfTime is not polluted by the first. While you are there, fix the comment typo "incluse" → "inclusive".

PR #4301, opened 2026-07-30
fix(link-formatting): Add missing return in pad_start NaN guard

### LINKED ISSUE #4284: [Bug]: Out-of-range search timestamps cause trace summary requests to fail
### What happened?

A trace-summary search fails before making an API request when the search URL contains a timestamp that is finite and positive but outside JavaScript's supported `Date` range.

For example, this value passes the current numeric checks:

```text
100000000000000000000
```

Converting it to an ISO timestamp throws:

```text
RangeError: Invalid time value
```

This can be triggered by a malformed, manually edited, or corrupted shared search URL.

### Steps to reproduce

1. Construct a trace search query with an out-of-range microsecond timestamp:

```ts
const query = {
  service: 'my-svc',
  operation: undefined,
  start: '100000000000000000000',
  end: '1700000060000000',
  limit: 20,
  lookback: 'custom',
};
```

2. Call:

```ts
client.fetchTraceSummaries(query);
```

3. Observe that the promise rejects with:

```text
RangeError: Invalid time value
```

4. Observe that `fetch()` is never called.

The same behavior occurs independently when `end` contains an out-of-range value.

### Expected behavior

Out-of-range timestamps should be handled as malformed URL parameters instead of being passed to `Date.prototype.toISOString()`.

The API client should not throw an unexpected `RangeError`. Consistent with the existing comment in `fetchTraceSummaries()`, an invalid timestamp can be omitted from the generated request so malformed URL parameters are handled gracefully.

### Relevant log output

```text
RangeError: Invalid time value
```

### Screenshot

Not applicable.

### Additional context

The current validation checks only whether the converted value is finite and positive:

https://github.com/jaegertracing/jaeger-ui/blob/9da77d02b7150354d8eab3da84255c33b111d8c9/packages/jaeger-ui/src/api/v3/client.ts#L65-L72

```ts
const startUs = Number(query.start);
const endUs = Number(query.end);

if (Number.isFinite(startUs) && startUs > 0)
  params.set('query.startTimeMin', new Date(startUs / 1000).toISOString());

if (Number.isFinite(endUs) && endUs > 0)
  params.set('query.startTimeMax', new Date(endUs / 1000).toISOString());
```

A finite number is not necessarily representable as a JavaScript `Date`.

A possible fix is to validate the constructed `Date` before calling `toISOString()`. Regression tests should cover out-of-range `start` and `end` values and verify that request construction does not throw.

I searched existing issues and pull requests for `Invalid time value`, out-of-range timestamps, timestamp overflow, `toISOString`, and `fetchTraceSummaries`, but did not find an existing report. I would be happy to submit a focused PR with regression tests if this behavior is confirmed.

### Jaeger backend version

Not applicable; this occurs while Jaeger UI constructs the API v3 request.

### SDK

Not applicable.

### Pipeline

Not applicable.

### Storage backend

Not applicable.

### Operating system

Not applicable.

### Deployment model

Not applicable.

### Deployment configs

Not applicable.

okxint:
**What**

The `pad_start` formatter in `link-formatting.ts` logs an error when `desiredLength` is `NaN` ("ignoring formatting") but then falls through and calls `value.padStart(NaN, padCharacter)` instead of returning early.

The sibling `add` formatter in the same file correctly `return value`s in the equivalent invalid-input branch. These two formatters are written in the same shape but `pad_start` is missing the early-return.

**Why it matters**

The current output happens to be correct by accident: `padStart(NaN, ...)` coerces `NaN` to `0` and returns the original string unchanged. But:
- The error message says "ignoring formatting" while the code keeps executing — inconsistent contract.
- The `add` formatter already returns early in the same situation — inconsistent control flow in the same file.
- If `padStart`'s `NaN` handling ever changes or the surrounding code is refactored, the bug becomes visible.

The existing test `Invalid desired length` in `link-formatting.test.js` passes today because the accidental output equals the correct output. It continues to pass after this fix.

**Fix**

Add the missing `return value;` after logging the NaN error.

Fixes #4284

### DIFF
--- a/packages/jaeger-ui/src/model/transform-trace-data.ts
+++ b/packages/jaeger-ui/src/model/transform-trace-data.ts
@@ -13,14 +13,17 @@ import OtelTraceFacade from './OtelTraceFacade';
 // exported for tests
 export function deduplicateTags(spanTags: ReadonlyArray<KeyValuePair>) {
   const warningsHash: Map<string, string> = new Map<string, string>();
-  const tags: KeyValuePair[] = spanTags.reduce<KeyValuePair[]>((uniqueTags, tag) => {
-    if (!uniqueTags.some(t => t.key === tag.key && t.value === tag.value)) {
-      uniqueTags.push(tag);
+  const seen = new Set<string>();
+  const tags: KeyValuePair[] = [];
+  for (const tag of spanTags) {
+    const key = `${tag.key}:${tag.value}`;
+    if (!seen.has(key)) {
+      seen.add(key);
+      tags.push(tag);
     } else {
-      warningsHash.set(`${tag.key}:${tag.value}`, `Duplicate tag "${tag.key}:${tag.value}"`);
+      warningsHash.set(key, `Duplicate tag "${tag.key}:${tag.value}"`);
     }
-    return uniqueTags;
-  }, []);
+  }
   const warnings = Array.from(warningsHash.values());
   return { tags, warnings };
 }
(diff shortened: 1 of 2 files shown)


okxint:
Closing — targeting higher-traffic repos.

PR #4302, opened 2026-07-30
fix(v3-client): Guard trace summary timestamp params against out-of-range Date

### LINKED ISSUE #4284: [Bug]: Out-of-range search timestamps cause trace summary requests to fail

okxint:
**Problem**

`fetchTraceSummaries` guards startUs/endUs with `Number.isFinite`, but `Number.isFinite(1e20)` is `true`. Values that large (beyond Date's maximum of 8.64×10¹⁵ ms) cause `new Date(startUs / 1000).toISOString()` to throw `RangeError: Invalid time value`, crashing the function before `fetch()` is ever called.

**Fix**

Introduce an intermediate `Date` object and guard with `Number.isFinite(date.getTime())` before serialising. `new Date()` with an out-of-range argument produces an Invalid Date whose `getTime()` returns `NaN`, so `Number.isFinite(NaN) === false` silently drops the parameter — matching the existing comment "drop malformed URL params gracefully" and consistent with how the `Number.isFinite(startUs)` check above it works for non-finite inputs.

```diff
-    if (Number.isFinite(startUs) && startUs > 0)
-      params.set('query.startTimeMin', new Date(startUs / 1000).toISOString());
-    if (Number.isFinite(endUs) && endUs > 0)
-      params.set('query.startTimeMax', new Date(endUs / 1000).toISOString());
+    const startDate = Number.isFinite(startUs) && startUs > 0 ? new Date(startUs / 1000) : null;
+    if (startDate && Number.isFinite(startDate.getTime()))
+      params.set('query.startTimeMin', startDate.toISOString());
+    const endDate = Number.isFinite(endUs) && endUs > 0 ? new Date(endUs / 1000) : null;
+    if (endDate && Number.isFinite(endDate.getTime()))
+      params.set('query.startTimeMax', endDate.toISOString());
```

Fixes #4284

### DIFF
--- a/packages/jaeger-ui/src/api/v3/client.ts
+++ b/packages/jaeger-ui/src/api/v3/client.ts
@@ -66,10 +66,12 @@ export class JaegerClient {
     // Guard with Number.isFinite to drop malformed URL params gracefully.
     const startUs = Number(query.start);
     const endUs = Number(query.end);
-    if (Number.isFinite(startUs) && startUs > 0)
-      params.set('query.startTimeMin', new Date(startUs / 1000).toISOString());
-    if (Number.isFinite(endUs) && endUs > 0)
-      params.set('query.startTimeMax', new Date(endUs / 1000).toISOString());
+    const startDate = Number.isFinite(startUs) && startUs > 0 ? new Date(startUs / 1000) : null;
+    if (startDate && Number.isFinite(startDate.getTime()))
+      params.set('query.startTimeMin', startDate.toISOString());
+    const endDate = Number.isFinite(endUs) && endUs > 0 ? new Date(endUs / 1000) : null;
+    if (endDate && Number.isFinite(endDate.getTime()))
+      params.set('query.startTimeMax', endDate.toISOString());
     if (query.limit) params.set('query.searchDepth', String(query.limit));
     if (query.minDuration) params.set('query.durationMin', query.minDuration);
     if (query.maxDuration) params.set('query.durationMax', query.maxDuration);
--- a/packages/jaeger-ui/src/model/transform-trace-data.ts
+++ b/packages/jaeger-ui/src/model/transform-trace-data.ts
@@ -13,14 +13,17 @@ import OtelTraceFacade from './OtelTraceFacade';
 // exported for tests
 export function deduplicateTags(spanTags: ReadonlyArray<KeyValuePair>) {
   const warningsHash: Map<string, string> = new Map<string, string>();
-  const tags: KeyValuePair[] = spanTags.reduce<KeyValuePair[]>((uniqueTags, tag) => {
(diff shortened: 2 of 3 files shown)


okxint:
Closing — targeting higher-traffic repos.

PR #4315, opened 2026-08-02
perf(trace-diff): Stabilize TraceDiffHeader props to avoid unnecessar…

### LINKED ISSUE #4314: [Bug]: Trace Comparison page performs unnecessary re-renders in TraceDiffHeader
### What happened?

The Trace Comparison page performs unnecessary re-renders of the `TraceDiffHeader` even when the displayed trace data remains unchanged. As a result, updates in the parent component trigger additional rendering work that does not affect the visible output.

This can reduce the responsiveness of the Trace Comparison page, particularly when comparing multiple traces or traces with a large amount of data.

### Steps to reproduce

1. Start the Jaeger UI development environment.
2. Navigate to the **Trace Comparison** page and compare two or more traces.
3. Open **React DevTools Profiler** and start recording.
4. Trigger a re-render of the `TraceDiff` page without changing the selected traces (for example, by resizing the browser window or performing another action that updates the parent component).
5. Stop the profiler recording and inspect the rendered components.
6. Observe that `TraceDiffHeader` and its child components re-render even though their displayed data has not changed.

### Expected behavior

`TraceDiffHeader` should only re-render when its displayed data changes. If the selected traces and comparison data remain unchanged, updates in the parent component should not trigger additional renders of `TraceDiffHeader` or its child components. This allows React's memoization to avoid unnecessary rendering work and helps maintain a responsive Trace Comparison page.

### Relevant log output

```shell

```

### Screenshot

_No response_

### Additional context

`TraceDiffHeader` is wrapped in `React.memo`, but the parent component (`TraceDiffImpl`) recreates the `traceA`, `traceB`, and `cohortData` objects and arrays on every render. Since these new references are passed as props, the shallow comparison performed by `React.memo` always fails, causing `TraceDiffHeader` and its child components to re-render even when their displayed data has not changed.

A React DevTools Profiler recording confirms that these unnecessary re-renders occur during parent updates that do not modify the comparison data.

### Jaeger backend version

_No response_

### SDK

_No response_

### Pipeline

_No response_

### Stogage backend

_No response_

### Operating system

macOS 

### Deployment model

_No response_

### Deployment configs

```shell

```

kunaldevsahu:

* Resolves #4314 

## Description of the changes

This PR reduces unnecessary re-renders of `TraceDiffHeader` in the Trace Comparison page.

The derived props passed to `TraceDiffHeader` are now memoized, allowing `React.memo` to skip redundant renders when the comparison data has not changed. Unit tests have also been added to verify the updated behavior.

* Added unit tests covering the updated behavior.
* Verified the Trace Comparison page works as expected.
* Confirmed that unnecessary re-renders are reduced.
* Ran lint and tests successfully.

## Checklist

* [x] I have added unit tests for the new functionality

## AI Usage in this PR

* [x] **Light**: AI provided minor assistance (formatting, simple suggestions)
* [ ] **None**
* [ ] **Moderate**
* [ ] **Heavy**

### DIFF
--- a/packages/jaeger-ui/src/components/TraceDiff/TraceDiff.test.jsx
+++ b/packages/jaeger-ui/src/components/TraceDiff/TraceDiff.test.jsx
@@ -23,8 +23,9 @@ vi.mock('react-router-dom', async () => {
   };
 });
 
-const { useTracesMock } = vi.hoisted(() => ({
+const { useTracesMock, mockHeaderProps } = vi.hoisted(() => ({
   useTracesMock: jest.fn(),
+  mockHeaderProps: [],
 }));
 
 vi.mock('../../hooks/useTraceLoading', () => ({
@@ -33,6 +34,7 @@ vi.mock('../../hooks/useTraceLoading', () => ({
 
 vi.mock('./TraceDiffHeader', async () => {
   return mockDefault(function MockTraceDiffHeader(props) {
+    mockHeaderProps.push(props);
     return (
       <div data-testid="trace-diff-header">
         <button type="button" data-testid="diff-set-a-btn" onClick={() => props.diffSetA('newAValue')}>
@@ -84,6 +86,7 @@ describe('TraceDiff', () => {
   });
 
   beforeEach(() => {
+    mockHeaderProps.length = 0;
     useTracesMock.mockReturnValue(new Map(defaultCohort.map(id => [id, { id, state: fetchedState.DONE }])));
     getUrlSpy.mockClear();
     mockNavigate.mockClear();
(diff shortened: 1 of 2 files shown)


jkowall (changes requested):
Memoizing the derived trace props is the right idea for React.memo(TraceDiffHeader), but mapStateToProps still allocates a fresh cohort via Array.from(fullCohortSet) on every run, which defeats the memo (and recreates the callback props via diffSetUrl). Please stabilize cohort by ID content (or memoize at the mapStateToProps boundary) and extend the test to cover an equivalent newly allocated cohort array.

PR #4319, opened 2026-08-03
chore(plexus): Use DedicatedWorkerGlobalScope in layout worker

### LINKED ISSUE #3860: [chore]: use DedicatedWorkerGlobalScope instead of Worker in layout.worker.ts
## Problem
`layout.worker.ts` types `ctx` as `Worker`, which is the type used when 
controlling a worker from the main thread. Inside a dedicated worker, 
the correct type is `DedicatedWorkerGlobalScope`, which includes `postMessage` 
and other worker-specific APIs.

## Solution
Replace `Worker` with `DedicatedWorkerGlobalScope` in both the type annotation 
and the `TMessageErrorTarget` callback type.

This resolves a TODO comment

021nirav-blip:
Fix the type annotation in the layout worker (packages/plexus/src/LayoutManager/layout.worker.ts). ctx was typed as Worker (the main‑thread controller type), but inside a dedicated worker the right type is DedicatedWorkerGlobalScope – which already covers postMessage, onmessage, and company.

Updated TMessageErrorTarget callback this and the ctx annotation accordingly. Dropped a stale 2017 TODO comment.

Coordinator.ts (which creates the worker from the main thread) keeps the Worker type – unchanged and correct.

Type‑only fix. tsc-lint and tests (115/115) pass. No test file exists for this component, so no unit tests added.

Closes #3860.

### DIFF
--- a/packages/plexus/src/LayoutManager/layout.worker.ts
+++ b/packages/plexus/src/LayoutManager/layout.worker.ts
@@ -12,12 +12,10 @@ import {
 } from './types';
 
 type TMessageErrorTarget = {
-  onmessageerror: ((this: Worker, ev: ErrorEvent) => any | void) | null;
+  onmessageerror: ((this: DedicatedWorkerGlobalScope, ev: ErrorEvent) => any | void) | null;
 };
 
-// TODO: Use WorkerGlobalScope instead of Worker
-
-const ctx: Worker & TMessageErrorTarget = self as any;
+const ctx: DedicatedWorkerGlobalScope & TMessageErrorTarget = self as any;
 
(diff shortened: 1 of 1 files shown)


021nirav-blip:
@yurishkuro @parshipcy
can you check my changes

jkowall (approved):
LGTM. Using `DedicatedWorkerGlobalScope` matches what `self` actually is in the layout worker.

jkowall:
Temporarily closing to re-trigger Actions workflows (prior runs are >30d and cannot be re-run).

parshipcy (approved):
lgtm

Issue #4106, opened 2026-06-16
[Bug]: Deeply nested traces crash trace view with RangeError (recursive processSpan)


aprv10:
### What happened?

`transformTraceData` normalizes every loaded trace and orders spans using a
recursive depth-first helper, `processSpan`, with no depth guard:

`packages/jaeger-ui/src/model/transform-trace-data.ts` (~lines 161-194)

```ts
const processSpan = (span, depth) => {
  ...
  span.childSpans.forEach(child => processSpan(child, depth + 1));
};
rootSpans.forEach(root => processSpan(root, 0));
```
For a trace whose span tree is deep (a long parent→child chain), this recursion
exceeds the JS call-stack limit and throws
RangeError: Maximum call stack size exceeded. The error is uncaught, so the
whole trace fails to render rather than degrading.

Reproducing the link + DFS logic in isolation, a chain depth of ~5000 already
crashes (each level uses two stack frames: processSpan plus the forEach
callback). Jaeger UI supports traces up to 80k spans, so deep chains are within
the supported range.

### Steps to reproduce

1. Load a trace whose spans form a deep parent → child chain (depth in the low
thousands).
2. The trace view fails to render; the console shows
RangeError: Maximum call stack size exceeded.

### Expected behavior

Large/deep traces should render without crashing. processSpan should use an
explicit stack-based iterative traversal instead of recursion (pushing children
in reverse to preserve start-time ordering), which removes the stack-depth limit
while producing identical output.

### Relevant log output

```shell

```

### Deployment configs

```shell

```

aprv10:
I'll be working on it

PR #4321, opened 2026-08-04
fix(TracePage): Render traces with a parent cycle instead of nothing

### LINKED ISSUE #1802: [Bug]: Backwards rendering of CHILD_OF vs. FOLLOWS_FROM
### What happened?

When creating a new span with a single link to a `followed` span in the same trace, the UI renders the new span as if the `followed` span were actually the `parent` span, and renders a link icon for the true `parent` of the new span.  In other words, the UI seems to flip the meaning of `CHILD_OF` and `FOLLOWS_FROM` when rendering, even though the references are correctly expressed in the trace.

### Steps to reproduce

1. Create a new trace with two immediate child spans logically referred to as `followed` and `parent`
2. Create a third span as a child of `parent` and specify one link to the `followed` span
3. View the trace and note that the third span is rendered as a child of `followed` with a link to `parent` instead of the other way round.

### Expected behavior

I expected child spans to be rendered beneath their parent, not beneath a linked span.

### Relevant log output

_No response_

### Screenshot

incorrect-span-link-annotated

### Additional context

Hopefully the annotated screen shot depicts the issue clearly.

The red-circled span shows a parent spanID ending in `dc34` and a follows-from spanId ending in `de6f`.  Yet the red-circled span named `remote_client::send_response` is rendered as if the parent is the follows-from span of `de6f`.

### Jaeger backend version

v1.49.0

### SDK

https://crates.io/crates/opentelemetry_sdk

OTEL Rust SDK -> [OTLP] -> Jaeger-all-in-one

### Pipeline

[OTEL Rust SDK](https://crates.io/crates/opentelemetry_sdk) -> [OTLP](https://crates.io/crates/opentelemetry-otlp) -> Jaeger-all-in-one

```rust
    // export OTEL_EXPORTER_OTLP_ENDPOINT=http://172.17.0.3:4317
    // where IP is that of jaeger container for docker inspect network bridge
    let _tracer = opentelemetry_otlp::new_pipeline()
        .tracing()
        .with_exporter(opentelemetry_otlp::new_exporter().tonic())
        .with_trace_config(config().with_resource(Resource::new(service_attributes(service_name))))
        .install_batch(opentelemetry_sdk::runtime::TokioCurrentThread)?;
```

### Stogage backend

Jaeger all-in-one

### Operating system

Linux

### Deployment model

Docker

### Deployment configs

_No response_

### LINKED ISSUE #4320: [Bug]: Span link read as parentage creates a parent cycle and a blank trace
### What happened?

`/api/traces/{id}` renders an OTLP span Link as a `FOLLOWS_FROM` reference, and
`transformTraceData` accepts `FOLLOWS_FROM` as a parent. When that fabricated parent edge
closes a cycle, every span in the cycle has its parent inside the cycle, so none is a root.
The traversal at
[`transform-trace-data.ts:218`](https://github.com/jaegertracing/jaeger-ui/blob/bcddf65b31a61d2fd5f5edd3e93944315657eb3f/packages/jaeger-ui/src/model/transform-trace-data.ts#L218)
is seeded exclusively from `rootSpans`, so nothing is visited and `spans` comes back empty.

Nothing signals the failure. The timeline is empty
([`VirtualizedTraceView.tsx:550`](https://github.com/jaegertracing/jaeger-ui/blob/bcddf65b31a61d2fd5f5edd3e93944315657eb3f/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx#L550)
receives `dataLength=0`, and the timeline viewer has no empty-state branch), the header title
is blank
([`trace-viewer.ts:57-61`](https://github.com/jaegertracing/jaeger-ui/blob/bcddf65b31a61d2fd5f5edd3e93944315657eb3f/packages/jaeger-ui/src/model/trace-viewer.ts#L57-L61)
finds no candidate root) and the duration is zero (the #4245 guard at
[`transform-trace-data.ts:220-227`](https://github.com/jaegertracing/jaeger-ui/blob/bcddf65b31a61d2fd5f5edd3e93944315657eb3f/packages/jaeger-ui/src/model/transform-trace-data.ts#L220-L227)).
No error is shown:
[`TracePage/index.tsx:377`](https://github.com/jaegertracing/jaeger-ui/blob/bcddf65b31a61d2fd5f5edd3e93944315657eb3f/packages/jaeger-ui/src/components/TracePage/index.tsx#L377)
sees no error and
[`:380`](https://github.com/jaegertracing/jaeger-ui/blob/bcddf65b31a61d2fd5f5edd3e93944315657eb3f/packages/jaeger-ui/src/components/TracePage/index.tsx#L380)
sees a valid `traceData`, so the full page chrome draws around an empty timeline, with no
spinner and no hang. There is no error boundary anywhere in `src/`, and the console is
silent. `orphanSpanCount` is `0` in every case, so
[`getIncompleteTraceTooltip`](https://github.com/jaegertracing/jaeger-ui/blob/bcddf65b31a61d2fd5f5edd3e93944315657eb3f/packages/jaeger-ui/src/model/trace-viewer.ts#L69)
never fires either.

Two shapes reach the blank render. **Shape 1**: a root span linking to its own descendant.
**Shape 2**: a non-root span with a same-trace link whose ancestor is missing from the
returned set, as with sampling loss or an unreturned batch, where the fall-through at
[`transform-trace-data.ts:127-135`](https://github.com/jaegertracing/jaeger-ui/blob/bcddf65b31a61d2fd5f5edd3e93944315657eb3f/packages/jaeger-ui/src/model/transform-trace-data.ts#L127-L135)
skips the unresolvable `CHILD_OF` and lands on the link. A milder **shape 3** misparents a
span under the span it merely links to, without blanking.

A cycle can also coexist with a valid root, and then the trace renders normally while
silently omitting every span inside the cycle. Measured on that fixture, 2 of 4 spans render,
`rootSpans` is 1, and `orphanSpanCount` is still `0`, so nothing indicates half the trace is
missing.

Reachability, stated plainly: not reachable from stock instrumentation. Both blank-render
shapes need a link whose target is in the same trace *and* a descendant of the linking span.
Links are supplied at span creation, before any descendant exists, so this requires `AddLink`
after creation or hand-built data. Every link pattern in the semantic conventions is
cross-trace (messaging, .NET connection setup, Lambda SQS) and the trace-ID gate rejects
those; the one same-trace source, the OpenTracing Shim, points at ancestors, which is correct
parentage. I also searched the four OTel instrumentation repos for link creation and found no
same-trace non-ancestor link, though that was one pass of code search, not an exhaustive
review. So this is rare. **It is worth fixing because when it does happen the failure is
completely silent, and the one diagnostic the UI already has is suppressed.**

### Steps to reproduce

Send a two-span trace to a stock `jaeger:2.19.0` all-in-one where `A` is the root carrying
`links: [B]` and `B` has `parentSpanId: A`, then open `/trace/{id}`. Zero spans render and no
error appears. Default memory storage is enough. It also reproduces on jaeger `1.62.0` and
`2.5.0`, so this is not a recent regression.

<details>
<summary>OTLP payload, and what each route returns for that identical input</summary>

```bash
docker run --rm -p 16686:16686 -p 4318:4318 cr.jaegertracing.io/jaegertracing/jaeger:2.19.0

curl -X POST http://localhost:4318/v1/traces -H 'Content-Type: application/json' -d '{
 "resourceSpans":[{"resource":{"attributes":[
   {"key":"service.name","value":{"stringValue":"cycle-repro"}}]},
  "scopeSpans":[{"scope":{"name":"repro"},"spans":[
   {"traceId":"00000000000000000000000000000042","spanId":"000000000000000a",
    "name":"root-with-link","kind":2,
    "startTimeUnixNano":"1754300000000000000","endTimeUnixNano":"1754300010000000000",
    "links":[{"traceId":"00000000000000000000000000000042","spanId":"000000000000000b"}]},
   {"traceId":"00000000000000000000000000000042","spanId":"000000000000000b",
    "parentSpanId":"000000000000000a","name":"child","kind":3,
    "startTimeUnixNano":"1754300001000000000","endTimeUnixNano":"1754300005000000000"}
  ]}]}]}'
```

| route | memory | badger | Elasticsearch 8.19 |
|---|---|---|---|
| `/api/traces/{id}` | cycle, root count 0 | cycle, root count 0 | cycle, root count 0 |
| `/api/v3/traces/{id}` | *control:* clean, root has no `parentSpanId` | cycle | cycle |

The memory + `/api/v3` cell is the control: same instance, same input, correct structure. The
legacy route there returns `root-with-link references=[FOLLOWS_FROM -> 000000000000000b]`,
root count 0. Through `transformTraceData` on `main` this gives `rendered=0, rootSpans=0`.
</details>

<details>
<summary>Shape 2 and the disjoint-cycle case, measured through the real transform</summary>

`R -> A -> B`, where `A` also carries a same-trace link to `B`, giving references
`[CHILD_OF -> R, FOLLOWS_FROM -> B]`. Then `R` is removed from the returned set. The third
row is a separate 4-span fixture: a normal `realRoot -> realChild` subtree alongside a
disjoint `cyc1 <-> cyc2` cycle.

| | rendered | rootSpans | orphanSpanCount | resulting parents |
|---|---|---|---|---|
| **control**, `R` present | 3 | 1 | 0 | `R->ROOT`, `A->R`, `B->A` |
| **partial**, `R` removed | 0 | 0 | 0 | `A->B`, `B->A` |
| **disjoint cycle**, valid root present | 2 of 4 | 1 | 0 | cycle members omitted entirely |

`A` does not merely misparent under `B`. `B` already parents to `A`, so this is a two-span
cycle and the trace blanks. It reproduces in both reference orders.
</details>

### Expected behavior

The trace renders. #4245 set the precedent in this very function, keeping spans with an
unusable `startTime` and repairing them instead of dropping them so that the whole span tree
stays visible. The same principle applies to spans unreachable from any root: keep them and
give them an entry point rather than silently omitting the entire trace. Failing that, an
explicit error would be better than a blank page.

### Additional context

**Root cause here.**
[`transform-trace-data.ts:127-135`](https://github.com/jaegertracing/jaeger-ui/blob/bcddf65b31a61d2fd5f5edd3e93944315657eb3f/packages/jaeger-ui/src/model/transform-trace-data.ts#L127-L135)
walks references in array order and takes the first that is `CHILD_OF` **or** `FOLLOWS_FROM`
and resolves in `spanMap`. It does not prefer `CHILD_OF`, does not check the trace ID, and
skips unresolvable references, which is what produces shape 2. The `FOLLOWS_FROM` reference
itself is created upstream of the UI, when the OTLP Link is converted to the Jaeger v1 model,
so the input is already wrong on arrival; fixing the render here does not depend on that
changing.

**Prior art.** #1802 (open, `help wanted`) appears to be the same root cause at
[`transform-trace-data.ts:127-135`](https://github.com/jaegertracing/jaeger-ui/blob/bcddf65b31a61d2fd5f5edd3e93944315657eb3f/packages/jaeger-ui/src/model/transform-trace-data.ts#L127-L135),
reporting the milder shape 3 symptom. #2750 asks for true DAG rendering, the general form.
#4245 added the guard at
[`transform-trace-data.ts:220-227`](https://github.com/jaegertracing/jaeger-ui/blob/bcddf65b31a61d2fd5f5edd3e93944315657eb3f/packages/jaeger-ui/src/model/transform-trace-data.ts#L220-L227)
that already anticipates a rootless trace, collapsing the time range but leaving the blank
render. #4106 notes `processSpan` is recursive with no depth guard, so any change here should
avoid deepening recursion. #3265 Phase 4 eventually removes `transformTraceData`, but its
replacement parser is not on `main` (#4129 is the open attempt), so this path is load-bearing
meanwhile.

**Two candidate fixes, neither covering all three shapes.** *A*: prefer `CHILD_OF`, gate on
trace ID, stop falling through, matching
[`OtelSpanFacade.ts:54-57`](https://github.com/jaegertracing/jaeger-ui/blob/bcddf65b31a61d2fd5f5edd3e93944315657eb3f/packages/jaeger-ui/src/model/OtelSpanFacade.ts#L54-L57)
from #3268. Measured on shape 2, the partial fixture goes from
`rendered=0, rootSpans=0, orphans=0` to `rendered=2, rootSpans=1, orphans=1`, so the trace
renders and `orphanSpanCount` correctly drives the incomplete-trace tooltip; it also fixes
shape 3 and would close #1802, but not shape 1, where with no `CHILD_OF` present the link is
selected anyway. *B*: after traversing from the real roots, promote the earliest unreached
span to a root and traverse again until every span is placed, with a visited set that is
required rather than defensive since a promoted cycle member becomes reachable; this fixes
shapes 1 and 2 but not shape 3, where no cycle forms.

**Follow-up, out of scope for both proposed PRs.** `getTraceName` runs its own root heuristic
that skips any span holding an internal reference, so a wholly cyclic trace still shows an
empty header name even once the timeline renders.

**Environment.** jaeger `2.19.0`, storage memory (default), badger and Elasticsearch 8.19;
jaeger-ui `main` at `bcddf65b`; Linux; Node v24.15.0.

Asking for confirmation that a trace with no root should still render, and for clearance to
open two PRs, one adopting the `CHILD_OF`-preferring rule and one adding the root-promotion
guard, since neither alone covers all three shapes.

u7k4rs6:

Closes #4320.

A trace can come back from `jaeger-query` with every span holding a resolvable parent
reference and still have no root, because each member of a cycle has its parent inside that
cycle. `transformTraceData` seeds its depth-first traversal exclusively from `rootSpans`, so
nothing is visited and the trace renders as an empty timeline with no error. `orphanSpanCount`
is `0` as well, because every parent reference did resolve, so the incomplete-trace tooltip
does not fire either.

This is reachable from ordinary data: `/api/traces/{id}` renders an OTLP span Link as a
`FOLLOWS_FROM` reference and the parent search accepts `FOLLOWS_FROM`, so a span carrying a
link to one of its own descendants becomes that descendant's child.

## Description of the changes

Same repair-rather-than-drop shape #4245 established in this function. The cycle is cut
**while the graph is built**, so that what the later phases receive is already a forest.

- Extract the existing "first `CHILD_OF` or `FOLLOWS_FROM` reference that resolves" search
  into `findParent`, shared by the linking pass and the new pass below.
- Add a third pass, immediately after parents and children are linked and before anything
  traverses the graph. It climbs parent pointers from each span, stopping at spans an earlier
  climb already settled. A span has at most one parent, so a climb either reaches a root or
  closes exactly one loop; when it closes a loop, the earliest member of that loop by
  `startTime` is cut free of its parent and becomes a root, and the rest of the loop hangs
  below it as an ordinary chain. Every span is climbed from at most once, so an acyclic trace
  pays one set lookup per span.
- Cut the edge in **both** directions: the promoted span is removed from its parent's
  `childSpans`, and `parentCycleBroken` is set on the span itself. The flag is load-bearing —
  `OtelSpanFacade` re-derives `parentSpanID` from `references` under its own rule, so without
  it the facade's `parentSpan` chain stays cyclic even though `childSpans` is a tree.
  Tolerating the cycle instead is not an option: `SpanTreeOffset` climbs `parentSpan` to draw
  indent guides, and the flamegraph and DDG descend `childSpans`, none of them with a visited
  set of its own.
- The reference itself stays on the span and surfaces in `OtelSpanFacade` as a link, which is
  where a `FOLLOWS_FROM` standing in for an OTLP span link belonged in the first place.
- Attach a warning to each promoted span so the condition is visible rather than silent.

## Changes since the first review

@yurishkuro's [review](https://github.com/jaegertracing/jaeger-ui/pull/4321#discussion_r3980397089)
objected to the position of the fix: the original version ran after the graph-building phase
was supposed to be finished, and did graph building and data cleanup together. The fix now
sits in the graph-building phase, immediately after the linking pass, and the later phases are
unchanged.

Two things fell out of that move. The `visited` set the earlier version added to `processSpan`
is gone — each span has at most one parent, so the climb detects a cycle on its own and the
traversal never needs to guard against one. And the earlier version only made a cycle member
*reachable* without actually breaking the cycle, which left `SpanTreeOffset` looping and the
flamegraph and DDG overflowing; cutting the edge in both directions fixes that, and the new
`expectAcyclicSpanGraph` helper asserts both directions rather than just the flat span list.

On the second question in that review — whether this is worth fixing before the OTEL
conversion — I am happy to park it if you prefer. `transformTraceData` is still the only place
the graph is built, so the fix fits there cleanly today. A `parentSpanId` cycle is still
representable in OTLP even though the link-based trigger goes away, so whatever builds the
graph after the conversion will need the same guard.

## What this covers, and what it does not

Using the shape numbering from the issue:

- **Shape 1**, a root span linking to its own descendant: fixed.
- **Shape 2**, a non-root span with a same-trace link whose ancestor is missing from the
  returned set: fixed.
- **Shape 3**, a span misparented under the span it merely links to, with no cycle: **not
  fixed**, and out of scope here. There is no cycle to break, so nothing gets promoted.
- **A cycle alongside a valid root**, where `main` renders the healthy subtree and silently
  drops the cycle members: fixed. Measured on the 4-span disjoint fixture, `main` renders 2
  of 4 with `orphanSpanCount` 0; this PR renders all 4 and promotes one root per cycle.

The incomplete-trace signal is deliberately not reused here: the count behind it is
recomputed in `OtelTraceFacade` from `parentSpanID` membership, and in a cycle every
`parentSpanID` resolves, so surfacing this condition would need either a reworded tooltip or
a separate count.

**This does not close #1802.** That is shape 3, and closing it needs the separate change to
prefer a same-trace `CHILD_OF` over `FOLLOWS_FROM` in the reference search, matching the rule
already used by `OtelSpanFacade`. The two changes are independent and do not overlap
textually: this PR inserts a pass around the linking loop, that one rewrites the selection
block.

Also out of scope: `getTraceName` runs its own root heuristic that skips any span holding an
internal reference, so a wholly cyclic trace still shows an empty header name even once the
timeline renders. Asserted in a test here so the residue is visible rather than forgotten.

## Things to know before reviewing

**An existing test asserted the old behaviour.** `should not produce a negative duration for
a trace with spans but no root` built a two-span cycle and asserted `result.spans.length === 0`
alongside `duration === 0`. It now asserts the trace renders with a correct duration. The
test's stated intent, in its own comment, is that the time range must not be left at its
sentinel value; that intent is preserved and still asserted. The `spans.length === 0` line was
incidentally encoding the bug.

New and updated tests in `transform-trace-data.test.js`. Every cycle case also runs through
`expectAcyclicSpanGraph`, which walks `parentSpan` upward and `childSpans` downward from the
roots and fails on a loop in either direction:

- a two-span cycle, the rewritten existing test, asserting a correct duration and time range
- a root span with a `FOLLOWS_FROM` link back to its own child, the reported shape
- a three-span cycle, asserting each span is placed exactly once, that the cut edge is gone
  from both `childSpans` and `parentSpanID`, and that the reference survives as a link
- a disjoint cycle alongside a normal subtree, asserting the normal subtree is untouched and
  one root is promoted per cycle
- a cycle with a descendant hanging below it, asserting the descendant keeps its own parent
- a span that references itself, the one-span cycle
- a 500-span ring where no span is reachable
- an acyclic trace, asserting identical output and zero promotions
- a 2000-deep chain as a regression tripwire, kept below the recursion ceiling tracked in #4106

Run against `origin/main` at `2386661c`, which this branch is rebased onto:

| gate | result |
|---|---|
| `transform-trace-data.test.js` | 36/36 pass |
| jaeger-ui suite | 273 files, 3749 tests pass |
| plexus suite | 9 files, 115 tests pass |
| `pnpm run tsc-lint` | clean |
| `pnpm run fmt-lint` | clean |
| `pnpm run oxlint` | no findings in changed files |
| `pnpm run build` | clean |

### DIFF
--- a/packages/jaeger-ui/src/model/transform-trace-data.ts
+++ b/packages/jaeger-ui/src/model/transform-trace-data.ts
@@ -11,6 +11,10 @@ import { IOtelTrace } from '../types/otel';
 import OtelTraceFacade from './OtelTraceFacade';
 
 // exported for tests
+export const CYCLIC_PARENT_WARNING =
+  'Cyclic parent reference detected; this span was promoted to a root span so the trace can render';
+
+// exported for tests
 export function deduplicateTags(spanTags: ReadonlyArray<KeyValuePair>) {
   const warningsHash: Map<string, string> = new Map<string, string>();
   const tags: KeyValuePair[] = spanTags.reduce<KeyValuePair[]>((uniqueTags, tag) => {
@@ -149,6 +153,10 @@ export default function transformTraceData(data: TraceData & { spans: SpanData[]
 
   const spans: Span[] = [];
   const svcCounts: Record<string, number> = {};
+  // Tracks which spans the traversal has already emitted. Besides keeping a span out of
+  // the flat list twice, this makes the traversal safe against parent cycles, where a
+  // span can appear inside its own subtree.
+  const visited = new Set<string>();
 
   // A startTime that is missing/NaN or 0 (the Unix epoch, which no real span
   // emits) carries no usable timestamp. Repair it by inheriting the parent's
@@ -173,6 +181,10 @@ export default function transformTraceData(data: TraceData & { spans: SpanData[]
   // Depth-first traversal to order spans, populate the flat array, and compute
   // the trace's time range from the (already-repaired) start times.
   const processSpan = (span: Span, depth: number) => {
+    if (visited.has(span.spanID)) {
+      return;
+    }
+    visited.add(span.spanID);
     span.depth = depth;
     span.hasChildren = span.childSpans.length > 0;
 
@@ -217,9 +229,40 @@ export default function transformTraceData(data: TraceData & { spans: SpanData[]
   rootSpans.sort((a, b) => a.startTime - b.startTime);
   rootSpans.forEach(root => processSpan(root, 0));
 
+  // A span whose parent reference resolves can still be unreachable from every root,
+  // because each member of a parent cycle has its parent inside that cycle and so is
+  // never itself a root. A trace made entirely of such spans has no root at all and
+  // would render as an empty timeline.
+  //
+  // This is reachable from ordinary data: the v1 rendering of an OTLP span link is a
+  // FOLLOWS_FROM reference, and the parent search above accepts FOLLOWS_FROM, so a root
+  // span carrying a link to one of its own descendants becomes that descendant's child.
+  //
+  // Promote the earliest unreached span to a root and traverse again, repeating until
+  // every span has been placed. Sorting once and skipping spans that a previous pass
+  // already reached is equivalent to repeatedly picking the earliest unreached span.
+  // For a normal forest nothing is unreached and this costs one size comparison.
+  if (visited.size < spanMap.size) {
+    const unreached: Span[] = [];
+    for (const span of spanMap.values()) {
+      if (!visited.has(span.spanID)) {
+        unreached.push(span);
+      }
+    }
+    unreached.sort((a, b) => a.startTime - b.startTime);
+    for (const span of unreached) {
+      if (visited.has(span.spanID)) {
+        continue;
+      }
+      repairStartTime(span);
+      (span.warnings as string[]).push(CYCLIC_PARENT_WARNING);
+      rootSpans.push(span);
+      processSpan(span, 0);
+    }
+  }
+
   // traceStartTime/traceEndTime are only updated while visiting spans reachable
-  // from a root. If the trace has spans but no root (e.g. every span is part of
-  // a reference cycle), nothing is visited and traceStartTime keeps its sentinel
+  // from a root. A trace with no spans at all leaves traceStartTime at its sentinel
   // value, which would produce a negative duration. Collapse to a zero range.
   if (spans.length === 0) {
     traceStartTime = 0;
--- a/packages/jaeger-ui/src/model/transform-trace-data.test.js
+++ b/packages/jaeger-ui/src/model/transform-trace-data.test.js
@@ -2,7 +2,11 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import traceGenerator from '../demo/trace-generators';
-import transformTraceData, { orderTags, deduplicateTags } from './transform-trace-data';
+import transformTraceData, {
+  orderTags,
+  deduplicateTags,
+  CYCLIC_PARENT_WARNING,
+} from './transform-trace-data';
 
 describe('orderTags()', () => {
   it('correctly orders tags', () => {
@@ -310,10 +314,11 @@ describe('transformTraceData()', () => {
     expect(result.duration).toBe(1000);
   });
 
-  it('should not produce a negative duration for a trace with spans but no root', () => {
-    // Two spans referencing each other form a cycle, so neither is a root and
-    // nothing is reachable by the traversal. The time range must not be left at
-    // its sentinel value, which would yield a negative duration.
+  it('should render a trace whose spans form a reference cycle, without a negative duration', () => {
+    // Two spans referencing each other form a cycle, so neither is a root. The
+    // earliest span is promoted to a root so the trace still renders, and the time
+    // range must not be left at its sentinel value (which would yield a negative
+    // duration).
     const spanA = {
       traceID,
       spanID: 'a',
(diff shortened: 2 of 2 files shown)


packages/jaeger-ui/src/model/transform-trace-data.ts
```diff
@@ -217,9 +229,40 @@ export default function transformTraceData(data: TraceData & { spans: SpanData[]
   rootSpans.sort((a, b) => a.startTime - b.startTime);
   rootSpans.forEach(root => processSpan(root, 0));
 
+  // A span whose parent reference resolves can still be unreachable from every root,
```
yurishkuro, line 232: the position of this code doesn't feel right to me. By the time reach the above three steps we are supposed to be done with the graph structure and the focus is cleaning up the data. But you are introducing another graph-building step right after it, which does both graph building AND data clean up. I would prefer to find a way to be done with graph building in the first phase.

Separately, do we even need to fix this now, before we do OTEL conversion? In OTEL the graph building is more straightforward, no follows-from edge cases mentioned here.

u7k4rs6:
@yurishkuro Added a cycle check after linking parents and children and cut one parent edge per cycle.
Made the earliest span in each cycle the root to keep downstream SpanTreeOffset, flamegraph, and DDG traversals safe.

packages/jaeger-ui/src/model/transform-trace-data.ts
u7k4rs6: Done. I moved the cycle check into a third pass immediately after linking parents and children, so the graph is fully fixed before anything traverses it. I also removed the `visited` set since each span has at most one parent, so we can detect a cycle just by walking parent references and cut one edge to make the cycle a root.

While rewriting this, I also found that the previous version only made the cycle reachable without actually breaking it. That caused `SpanTreeOffset` to loop and the flamegraph/DDG to overflow, so I now cut the edge in both directions.

u7k4rs6: I’m fine with parking this until the OTEL conversion if you prefer. Right now `transformTraceData` is still the only place where the graph is built, so the fix fits there cleanly.

The OTLP case can still represent a `parentSpanId` cycle, even though the original link-based trigger goes away, so the same issue may need to be handled by the future OTLP graph builder.

PR #4326, opened 2026-08-05
fix(span-detail): Truncate collapsed attribute summary values

### LINKED ISSUE #3798: [Bug]: Collapsed span attributes with massive text values cause severe DOM bloat and layout anomalies
### What happened?

When expanding a span's inline detail row, the "Tags" (Attributes) accordion is collapsed by default. In this collapsed state, Jaeger renders a one-line summary of all the span's tags. However, the `AttributesSummary` component injects the *entire* raw string value of every attribute directly into the DOM, relying solely on CSS `text-overflow: ellipsis` to hide the excess text from the user's viewport.

While this appears fine visually on the surface, the browser still parses and holds the massive text node in memory. In the context of GenAI traces, where attributes like `gen_ai.prompt` or `llm.request.messages` routinely contain multi-kilobyte strings, this causes severe DOM bloat. If a trace contains dozens of agentic loop spans with massive prompts, rendering these collapsed headers forces hundreds of thousands of invisible words into the DOM, causing the trace timeline to become highly sluggish. 

Furthermore, in the side-panel mode, the `white-space: nowrap` constraint on massive strings causes the panel's layout to overflow horizontally.

### Steps to reproduce

1. Generate a trace with a span containing a string attribute over 4,000 characters long (e.g., an LLM prompt or large JSON payload).
2. Open the Jaeger UI and load the trace in the timeline view.
3. Click on the span to expand its inline detail row.
4. Ensure the "Tags" (Attributes) accordion is **collapsed**.
5. Observe the DOM using browser developer tools: the entire massive string is fully rendered inside the `<li>` element.
6. Open the span in the side-panel view and observe the layout being pushed out of bounds horizontally.

### Expected behavior

The raw string value should be truncated at the JavaScript/React level (e.g., to 60 characters) *before* it is injected into the DOM for the collapsed summary view. The full, un-truncated string should only be rendered when the user explicitly clicks to expand the accordion.

### Relevant log output

```shell

```

### Screenshot

_No response_

### Additional context

I noticed this issue while diving into the UI architecture in preparation for the LFX Term 2 "Jaeger for GenAI Observability" mentorship proposal.

I have a fix in mind and happy to open a PR and discuss it furtther with the maintainers.

### Jaeger backend version

`2.17.0` (UI version)

### SDK

_No response_

### Pipeline

Frontend issue

### Stogage backend

_No response_

### Operating system

EndeavourOS(Arch)

### Deployment model

_No response_

### Deployment configs

```shell

```

021nirav-blip:
Fix DOM bloat in the collapsed AttributesSummary (SpanDetail/AccordionAttributes.tsx).

We were rendering multi‑KB GenAI strings like gen_ai.prompt and llm.request.messages in full on collapsed headers, then hiding them with CSS text-overflow: ellipsis. That meant the DOM still held all that data, which bloated the side panel and could cause overflow.

Now we truncate each attribute value to 60 characters (with ...) before injecting it into the DOM when collapsed. The expanded AttributesTable still shows the full value.

Closes #3798.

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/AccordionAttributes.test.jsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/AccordionAttributes.test.jsx
@@ -53,6 +53,13 @@ describe('<AttributesSummary />', () => {
     const expectedTexts = tagsArray.map(tag => `${tag.key}=${tag.value}`);
     expect(texts).toEqual(expectedTexts);
   });
+
+  it('truncates values longer than 60 characters in the summary', () => {
+    const longValue = 'x'.repeat(100);
+    const data = makeAttributes([{ key: 'long', value: longValue }]);
+    const { container } = render(<AttributesSummary data={data} />);
+    expect(container.querySelector('li').textContent).toBe(`long=${'x'.repeat(60)}...`);
+  });
 });
 
 describe('<AccordionAttributes />', () => {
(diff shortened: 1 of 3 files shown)


021nirav-blip:
@yurishkuro @parshipcy 
can you review my changes

Issue #3604, opened 2026-03-13
[chore]: Remove unsafe use of _get()


yurishkuro:
### Requirement

The _get() function from lodash succeeds even when the path refers to properties that do not exist on the object (e.g. a bug fixed in #3592). There is no need for it when we can do a type-safe compiler-checked access, e.g.
```diff
- _get(state, 'config.search.maxLookback')
+ state?.config?.search?.maxLookback
```
Examples of usage:
```
$ rg '_get\(' packages/jaeger-ui/src
packages/jaeger-ui/src/reducers/embedded.ts
11:    const search = _get(window, 'location.search');

packages/jaeger-ui/src/components/SearchTracePage/SearchForm.tsx
851:    searchMaxLookback: _get(state, 'config.search.maxLookback'),
852:    searchAdjustEndTime: _get(state, 'config.search.adjustEndTime'),

packages/jaeger-ui/src/model/path-agnostic-decorations/index.tsx
34:  let decorationValue = _get(state, `pathAgnosticDecorations.${decorationID}.withOp.${service}.${operation}`);
35:  let decorationMax = _get(state, `pathAgnosticDecorations.${decorationID}.withOpMax`);
37:    decorationValue = _get(state, `pathAgnosticDecorations.${decorationID}.withoutOp.${service}`);
38:    decorationMax = _get(state, `pathAgnosticDecorations.${decorationID}.withoutOpMax`);

packages/jaeger-ui/src/utils/plexus/set-on-graph.ts
22:  const matchSize = BASE_MATCH_SIZE + SCALABLE_MATCH_SIZE / _get(zoomTransform, 'k', 1);

packages/jaeger-ui/src/actions/path-agnostic-decorations.ts
80:      return _get(res, getPath, `\`${getPath}\` not found in response`);

packages/jaeger-ui/src/utils/DraggableManager/DraggableManager.ts
103:    const style = _get(document, 'body.style');
177:      const style = _get(document, 'body.style');

packages/jaeger-ui/src/utils/config/process-deprecation.ts
33:      const value = _get(config, formerKey);

packages/jaeger-ui/src/utils/tracking/ga.ts
272:    isTruish(parseQuery(_get(window, 'location.search'))['ga-debug']);
273:  const gaID = _get(config, 'tracking.gaID');
274:  const isErrorsEnabled = isDebugMode || Boolean(_get(config, 'tracking.trackErrors'));
[...]

PR #4338, opened 2026-08-07
chore: remove unsafe use of _get() for static paths

DsThakurRawat:
Ref #3604

This PR incrementally removes the `lodash/get` usage for static string paths in uncontested files, replacing them with modern optional chaining (`?.`) and nullish coalescing (`??`) operators. 

**Scope of this PR:**
- `utils/DraggableManager/DraggableManager.ts`
- `utils/plexus/set-on-graph.ts` (using `??` instead of `||` to safely preserve `0` as a legitimate value for `zoomTransform.k`)
- `components/TraceDiff/TraceDiffGraph/traceDiffGraphUtils.ts`

**Deliberate Omissions (Why some `_get` calls remain):**
As noted in the issue discussion, *"The answer might be different depending on the situation."* I have deliberately left `_get` in place where the path is a *dynamic string* or constructed at runtime, as optional chaining cannot express those dynamically resolved paths safely without type-casting or breaking abstractions. These include:
- `utils/config/process-deprecation.ts:33` (`_get(config, formerKey)`)
- `model/path-agnostic-decorations/index.tsx:34-38` (template-literal paths)
- `actions/path-agnostic-decorations.ts:80` (runtime `getPath`)
- `DeepDependencies/SidePanel/DetailsPanel.tsx:70,77` (runtime `getPath`)

Lodash `get` remains the correct tool for these dynamic situations. 

*Note: The remaining static path replacements are deliberately excluded here as they are located in highly contested files (e.g. `TracePage/index.tsx`) where multiple PRs are currently open, avoiding unnecessary merge conflicts.*

Testing:
- Covered by existing tests (`DraggableManager.test.js` and `set-on-graph.test.js`). Behavior preservation is verified.
- `pnpm lint` and `pnpm test` pass.

### DIFF
--- a/packages/jaeger-ui/src/components/TraceDiff/TraceDiffGraph/traceDiffGraphUtils.ts
+++ b/packages/jaeger-ui/src/components/TraceDiff/TraceDiffGraph/traceDiffGraphUtils.ts
@@ -2,7 +2,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { TVertexKey } from '@jaegertracing/plexus/lib/types';
-import _get from 'lodash/get';
 import _map from 'lodash/map';
 import memoizeOne from 'memoize-one';
 
@@ -20,7 +19,7 @@ function getUiFindVertexKeysFn(
   if (!uiFind) return new Set<TVertexKey>();
   const newVertexKeys: Set<TVertexKey> = new Set();
   vertices.forEach(({ key, data: { members } }) => {
-    if (_get(filterSpans(uiFind, _map(members, 'span')), 'size')) {
+    if (filterSpans(uiFind, _map(members, 'span'))?.size) {
       newVertexKeys.add(key);
     }
   });
--- a/packages/jaeger-ui/src/utils/DraggableManager/DraggableManager.ts
+++ b/packages/jaeger-ui/src/utils/DraggableManager/DraggableManager.ts
@@ -1,7 +1,6 @@
 // Copyright (c) 2017 Uber Technologies, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-import _get from 'lodash/get';
 import type React from 'react';
 
 import EUpdateTypes from './EUpdateTypes';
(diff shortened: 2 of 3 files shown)


DsThakurRawat:
closing this for now to free my PR slot for #4373, the RFC 0008 follow-up @yurishkuro asked for on #3972. the new-contributor limit is 1 concurrent PR and that one's on hold behind this. will reopen this once #4373 is through.

DsThakurRawat:
reopening this now that #4373 has merged, since that is what was holding the slot.

rebased onto current `main`, it was 43 commits behind. the three call sites are still there on
`main`, so the change is still needed and nothing has superseded it.

tests pass on the rebased tree: `set-on-graph` and `DraggableManager` 26 of 26, `TraceDiffGraph`
20 of 20.

jkowall (approved):
These three `_get` sites are all static paths, so optional chaining is the right replacement. Same pattern as #3604 and as SearchForm already using `?.` instead of `_get(state, 'config.search.maxLookback')`.

- `filterSpans(...)?.size` is equivalent to `_get(..., 'size')` in a truthy check (0 / missing still falsy).
- `document.body?.style` matches `_get(document, 'body.style')`. The existing `if (style)` guard still covers a missing body.
- `zoomTransform?.k ?? 1` matches `_get(..., 'k', 1)` for numeric `k`. `??` keeps `0`, which `||` would not.

Unused `lodash/get` imports are gone. Remaining `_getBounds` / `_getPosition` on DraggableManager are class methods, not lodash. Dynamic `_get` call sites correctly left alone.

LGTM.

jkowall:
Copilot's follow-up on `setOnNodesContainer` (`k === 0` → divide by zero / `Infinity` px outline) is real, but it is pre-existing: lodash `_get(zoomTransform, 'k', 1)` also returns `0` when `k` is 0. This PR did not change that. Leaving the approve; worth a follow-up if zoom can actually hit 0.

Issue #4339, opened 2026-08-07
[Bug]: Unhandled "window is not defined" from React scheduler fails the run on Node 26


Animesh-Parashar:
### What happened?

On Node 26 every test passes and the run still exits 1, from one uncaught
exception after a test file finishes:

```
ReferenceError: window is not defined
 ❯ performWorkOnRootViaSchedulerTask react-dom-client.development.js:18936:7
 ❯ Immediate.performWorkUntilDeadline scheduler.development.js:45:48
 ❯ processImmediate node:internal/timers:504:21
```

React 19 reads `window.event` unconditionally at that line and schedules the work
through `setImmediate`, so a task still queued when Vitest tears down the jsdom
environment fires with no `window`. Node 24 produces none of these, and `engines`
permits Node 26 (`>=24.16.0`).

This is separate from the `localStorage` failures on Node 25+; it reproduces with
and without a fix for those.

### Steps to reproduce

1. Use Node 26 (v26.4.0 here).
2. `pnpm install --frozen-lockfile`
3. `pnpm --filter @jaegertracing/jaeger-ui test`

### Expected behavior

Either the run is green on a Node version `engines` allows, or `engines` is
bounded so it is rejected up front.

### Relevant log output

```shell
Test Files  267 passed (267)
     Tests  3527 passed (3527)
    Errors  1 error
```

### Additional context

It only reproduces on the full suite. `DownloadResults.test.jsx` alone,
`src/components/**` (182 files), and everything outside it (85 files) are all
clean, so the file Vitest names is probably not the culprit.

Draining pending immediates in an `afterAll` does not help, and it breaks
`utils/date.test.js` and `utils/tracking/ga.test.js`, which use
`vi.useFakeTimers()`. I could not work out why it is Node 26 only, so I have not
guessed at a fix. Happy to keep digging if it is worth fixing rather than
bounding `engines`.

Aravind226:
Hi! I'd like to work on this issue. From what I’ve looked into so far, I suspect the issue may be related to a stale `window.event` being accessed after the JSDOM teardown. I’m still investigating why this appears to fail on Node 26 while working on Node 24. Would it be okay if I take this up?

Animesh-Parashar:
Please do, go ahead.

Your read matches mine: React 19 reads `window.event` unconditionally in `performWorkOnRootViaSchedulerTask`, and the scheduler queues that work through `setImmediate`, so a task still pending when the environment tears down fires with no `window`.

[...]

PR #4341, opened 2026-08-07
fix(tests): provide Storage when Node 25+ shadows jsdom's implementation

### LINKED ISSUE #4340: [Bug]: Test suite fails on Node 25+ because built-in Storage shadows jsdom's
### What happened?

On Node 25 and newer the test suite fails with 24 failing files and 145 failing tests, all `TypeError: Cannot read properties of undefined` on `localStorage.clear`, `.getItem` or `.setItem`. On Node 24 the same suite is
green.

`engines` declares `"node": ">=24.16.0"`, which permits Node 26, so a clean checkout fails for anyone whose system Node is newer than `.nvmrc`. CI does not catch it because `setup-node` reads that same unbounded range and resolves below 26.

Node 25+ exposes `localStorage`/`sessionStorage` as built-in globals, and `localStorage` is an own getter on `globalThis` that evaluates to `undefined` unless `--localstorage-file` is passed:

```
$ node -v && node -e "console.log(typeof localStorage, typeof sessionStorage)"
v26.4.0
undefined object
ExperimentalWarning: localStorage is not available because --localstorage-file was not provided.
```

Vitest's `getWindowKeys` only copies a jsdom key onto `globalThis` if it is absent there or in its allowlist (`if (k in global) return keysArray.includes(k)`). `Storage` and `StorageEvent` are on that list, the `localStorage`/`sessionStorage` instances are not, so jsdom's implementations are skipped and Node's `undefined` is what tests get.

### Steps to reproduce

1. Use Node 25 or newer (v26.4.0 here).
2. `pnpm install --frozen-lockfile`
3. `pnpm --filter @jaegertracing/jaeger-ui test`

### Expected behavior

The suite passes on a Node version `engines` allows, or `engines` is bounded so unsupported versions are rejected up front.

### Relevant log output

```shell
Test Files  24 failed | 243 passed (267)
      Tests  145 failed | 2980 passed (3125)

TypeError: Cannot read properties of undefined (reading 'clear')
 ❯ src/components/TracePage/TraceTimelineViewer/SpanDetail/GenAITab/message-format-store.test.ts:47:18
     46|   beforeEach(() => {
     47|     localStorage.clear();
```

### Additional context

This is vitest-dev/vitest#10867, fixed in Vitest 5 by vitest-dev/vitest#10293. The Vitest maintainers declined to backport it to the 4.x line because it needs a breaking change, and Vitest 5 is still in beta, so there is no upstream fix for the pinned version. `environmentOptions` does not help either: `additionalKeys` is hardcoded in Vitest's jsdom environment.

I have a fix in `test/vitest-setup.ts` and will open a PR.

Animesh-Parashar:
- Resolves #4340 
- On Node 25+ the test suite fails with 24 failing files and 145 failing tests, all `TypeError: Cannot read properties of undefined` on `localStorage.clear`, `.getItem` or `.setItem`. On Node 24 the same suite is green. `engines` declares `"node": ">=24.16.0"`, which permits Node 26, so a clean checkout fails for anyone whose system Node is newer than `.nvmrc`. CI does not catch it because `setup-node` reads the same unbounded range and resolves below 26.

## Description of the changes
- Node 25+ defines `localStorage`/`sessionStorage` as globals, so Vitest's `getWindowKeys` keeps those over jsdom's, and Node's are `undefined` without `--localstorage-file`. Fixed upstream only in Vitest 5 (vitest-dev/vitest#10867), which is still beta.
- Installs an in-memory `Storage` in `test/vitest-setup.ts`, next to the existing `TextEncoder` and `ResizeObserver` shims. Methods go on `Storage.prototype` because tests spy on it.
- Guarded by `if (!globalThis.localStorage || !globalThis.sessionStorage)`, so nothing runs on Node 24.
 

- Full suite, same flags, only the Node binary changed:
 
 | | Node 24.19.0 | Node 26.4.0 |
  |---|---:|---:|
  | before | 3527 pass | **145 fail** |
  | after | 3527 pass | **3527 pass** |

- `make lint test` passes on Node 24.
- No new tests: this is harness setup, covered by the 3527 existing tests, 145 of which fail without it.
- Node 26 still exits non-zero from an unrelated `ReferenceError: window is not defined` in the React scheduler after teardown, reported separately in #4339 . It reproduces with and without this change. If you would rather bound `engines` to `<26`, happy to send that instead.

## Checklist

### DIFF
--- a/packages/jaeger-ui/test/vitest-setup.ts
+++ b/packages/jaeger-ui/test/vitest-setup.ts
@@ -12,6 +12,55 @@ rafPolyfill();
 // TextEncoder is not provided by JSDOM
 global.TextEncoder = TextEncoder as typeof global.TextEncoder;
 
+// Node 25+ defines localStorage/sessionStorage as globals, so Vitest keeps those
+// over jsdom's, and Node's are undefined without --localstorage-file (vitest#10867).
+if (!globalThis.localStorage || !globalThis.sessionStorage) {
+  const backings = new WeakMap<Storage, Map<string, string>>();
+  const backing = (self: Storage) => {
+    let map = backings.get(self);
+    if (!map) {
+      map = new Map();
+      backings.set(self, map);
+    }
+    return map;
+  };
+
+  // On the prototype, not the instance, because tests spy on Storage.prototype.
+  const def = (name: string, value: unknown) =>
+    Object.defineProperty(Storage.prototype, name, { value, configurable: true, writable: true });
+
+  Object.defineProperty(Storage.prototype, 'length', {
+    configurable: true,
+    get(this: Storage) {
+      return backing(this).size;
+    },
+  });
+  def('key', function (this: Storage, index: number) {
+    return [...backing(this).keys()][index] ?? null;
+  });
+  def('getItem', function (this: Storage, key: string) {
+    const map = backing(this);
+    return map.has(String(key)) ? (map.get(String(key)) as string) : null;
+  });
+  def('setItem', function (this: Storage, key: string, value: string) {
+    backing(this).set(String(key), String(value));
+  });
+  def('removeItem', function (this: Storage, key: string) {
+    backing(this).delete(String(key));
+  });
+  def('clear', function (this: Storage) {
+    backing(this).clear();
+  });
+
+  for (const key of ['localStorage', 'sessionStorage'] as const) {
+    Object.defineProperty(globalThis, key, {
+      value: Object.create(Storage.prototype) as Storage,
+      configurable: true,
+      writable: true,
+    });
+  }
+}
+
 // Alias jest → vi so existing test files work without modification.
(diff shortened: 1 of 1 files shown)


jkowall (changes requested):
Thanks for the Node 25+ Storage shim. The in-memory prototype implementation is the right idea.

One fix before merge: Vitest worker reuse can leave these globals around while installing a new jsdom Storage constructor, so the current !localStorage guard skips re-init and spies on Storage.prototype stop applying. Please recreate the shim whenever localStorage/sessionStorage are missing or not instanceof Storage.

jkowall (changes requested):
Shim direction LGTM (prototype methods for the spies, Node 24 left alone). Please recreate the shim whenever the objects are not instanceof the current environment Storage — the truthy localStorage/sessionStorage guard can leave stale instances across reused Vitest workers and break prototype spies. Keep only changelog:skip (already applied).

Issue #4351, opened 2026-08-09
Trace timeline keeps dead hoverIndentGuideIds state (highlight removed in #3302)


thisis-manan:
## What I found

The trace timeline still keeps `hoverIndentGuideIds` in the Redux `traceTimeline` slice, but nothing consumes it anymore. The indent-guide hover highlight it was built for is gone.

- **Write path (still runs):** `SpanTreeOffset` dispatches `addHoverIndentGuideId` / `removeHoverIndentGuideId` on mouse enter/leave over every indent guide.
- **Read path (dead):** `VirtualizedTraceView` reads the set with `useSelector` and passes it into `VirtualizedTraceViewImpl` as a prop that is never referenced. No component applies a class from it, and there is no matching CSS.

So each hover over an indent guide dispatches Redux actions and clones a `Set` to feed state that paints nothing. On large traces that is wasted work on a hot path.

## Where it regressed

#3302 ("Better tree hierarchy for trace view") redrew the tree and removed the `is-active` class that consumed this state:

```diff
-          className={cx('SpanTreeOffset--indentGuide', {
-            'is-active': hoverIndentGuideIds.has(ancestorId),
```

It also deleted the matching CSS. The Redux field, the two actions, the writers, and the read were all left behind. `git log -S "is-active" -- packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanTreeOffset.tsx` points at #3302.

## Verification

Rendering `SpanTreeOffset` with `hoverIndentGuideIds` populated with every ancestor produces DOM byte-identical to an empty set, and a real `mouseEnter` adds no highlight class. Nothing reads the state.

## What the highlight used to do

Hovering an indent guide added that ancestor's id to the set, and every guide sharing the id lit up across rows — so you could see, at a glance, which ancestor a span descended from. In deep or wide traces that is a genuine navigation aid.

Visually (pre-#3302) it just darkened the guide line:

```css
.SpanTreeOffset--indentGuide.is-active {
  border-left-color: var(--border-strongest);
}
.SpanTreeOffset--indentGuide.is-active:before {
  background-color: var(--border-strongest);
}
```

## The decision: remove or restore?

**Was dropping the highlight in #3302 intentional?** That decides the direction, and I would rather ask than guess.

[...]

PR #4352, opened 2026-08-09
refactor(timeline): Remove dead hover-guide state

### LINKED ISSUE #4328: [Bug]: Indent guide hover state is written on every mouse move but nothing reads it
### What happened?

Hovering an indent guide in the trace timeline dispatches Redux actions that update `hoverIndentGuideIds`, but nothing reads that state anymore, so nothing renders from it.

It used to drive an ancestor highlight: hovering one guide lit up the same ancestor's guide in the other rows. #3302 rewrote the guides and removed the `is-active` class and its CSS, but left the state, the actions, the reducers, `mapStateToProps` and a `useSelector` in place.

So every mouse move across a guide costs a dispatch, a new Set, and a re-render of `VirtualizedTraceViewWrapper`, for no visual effect. Deeper traces have more guides per row, so it dispatches more.

### Steps to reproduce

1. Open any trace with nested spans.
2. Move the mouse across the vertical indent guides in the left column.
3. Nothing highlights in the other rows. Redux actions still fire on every guide.

### Expected behavior

Either the hover should highlight the ancestor's guides across rows as it used to, or the state should not exist.

### Relevant log output

```shell

```

### Screenshot

_No response_

### Additional context

Before #3302, SpanTreeOffset.tsx had:

    'is-active': hoverIndentGuideIds.has(ancestorId),

#3302 removed that and the CSS behind it:

    -.SpanTreeOffset--indentGuide.is-active {
    -.SpanTreeOffset--indentGuide.is-active:before {

#3753 later removed the unused destructure, which was correct — oxlint flagged it because nothing used it by then. The write path was untouched.

Still on main:
- SpanTreeOffset.tsx — handleMouseEnter/handleMouseLeave dispatch addHoverIndentGuideId/removeHoverIndentGuideId
- duck.ts:416-427 — both reducers clone the Set
- SpanTreeOffset.tsx:209 — mapStateToProps returns it, TProps declares it, the component never destructures it
- VirtualizedTraceView.tsx:580 — read via useSelector, spread into combinedProps at 702, unused by the class component. Only type-checks because VirtualizedTraceViewProps includes TTraceTimeline.

SpanTreeOffset.css has no class keyed on the set.

RFC-0004 Phase 1c step 3 lists this as "hover state still lives only in the duck", pending a move to Zustand. If it's dead there's nothing to migrate — deleting it also removes the last useSelector in VirtualizedTraceView and the connect() in SpanTreeOffset.

Was dropping the highlight in #3302 intentional? Two options, and I'd rather ask before opening a PR:
1. Delete the state, the actions, the reducers, the connect(), the useSelector and the handlers.
2. Restore the feature and migrate the state to useTraceTimelineStore as Phase 1c.3 planned.

Happy to do either.

UI-only; reproducible against any backend since it's a client-side state question.

### Jaeger backend version

_No response_

### SDK

_No response_

### Pipeline

_No response_

### Stogage backend

_No response_

### Operating system

_No response_

### Deployment model

_No response_

### Deployment configs

```shell

```

thisis-manan:

The trace timeline keeps `hoverIndentGuideIds` in the Redux `traceTimeline` slice, but nothing reads it to render anything. The indent-guide hover highlight that used it (the `is-active` class and its CSS) was removed in #3302, leaving the field, its two actions, the `SpanTreeOffset` mouse-enter/leave writers, and the `VirtualizedTraceView` `useSelector` read as orphaned code. Every hover still dispatches Redux actions and clones a `Set` for no visible effect.

Full analysis and the git bisect are in #4351.

## Description of the changes

- Remove `hoverIndentGuideIds` and the `ADD_/REMOVE_HOVER_INDENT_GUIDE_ID` actions and reducers from `TraceTimelineViewer/duck.ts`; drop the field from `TTraceTimeline`.
- Remove the mouse-enter/leave handlers from `SpanTreeOffset`. It no longer needs Redux, so it becomes a plain functional component instead of a `connect()`ed one — one fewer connected component on the way to removing Redux (RFC 0004 Phase 1c/4).
- Remove the dead `useSelector` read and prop threading from `VirtualizedTraceView`.
- Drop the now-unnecessary Redux `Provider` from the `SpanDetailRow` integration test.
- Add a regression test asserting hover paints no highlight class, so the dead state cannot quietly return.
- Update RFC 0004 to record Phase 1c step 3 as resolved by deletion.

Hover is not tracked by the analytics middleware, so no dual-write is involved and nothing needs to move off `duck.track.ts`.

If the highlight removal in #3302 was not intentional and it should come back instead, I am happy to restore it on the Zustand store rather than deleting — see the options in #4351.

`pnpm run lint`, `pnpm test`, and `pnpm run build` all pass. Rendering `SpanTreeOffset` with the hover set populated produces DOM identical to an empty set, confirming nothing consumed it.

 Resolves #4328

### DIFF
--- a/docs/rfc/0004-state-management-strategy.md
+++ b/docs/rfc/0004-state-management-strategy.md
@@ -978,12 +978,12 @@ This is the **largest and most performance-sensitive** slice. Split across multi
 **Suggested PR split** (status as of 2026-07-28):
 1. ✅ Layout prefs: `spanNameColumnWidth`, `sidePanelWidth`, `detailPanelMode`, `timelineBarsVisible`.
 2. ✅ Collapse/expand + detail: `childrenHiddenIDs`, `detailStates`, `shouldScrollToFirstUiFindMatch`.
-3. ⬜ Hover: `hoverIndentGuideIds` still lives only in the duck. `VirtualizedTraceView` reads it with `useSelector`, and `SpanTreeOffset` is still a `connect()`ed component whose `mapStateToProps` reads `state.traceTimeline`.
+3. ✅ Hover: resolved by deletion rather than migration. `hoverIndentGuideIds` was dead state — the indent-guide highlight that consumed it (the `is-active` class and its CSS) was removed in [#3302](https://github.com/jaegertracing/jaeger-ui/pull/3302), leaving the Redux field, its two actions, the `SpanTreeOffset` mouse-enter/leave writers, and the `VirtualizedTraceView` `useSelector` read as orphaned code that fed nothing. All removed; `SpanTreeOffset` is now a plain (unconnected) functional component. Hover is not tracked by the analytics middleware, so no dual-write was involved.
 4. ⬜ Rewire the Redux **tracking middleware** hooks that listen to timeline action types.
 
 **Why the duck is still alive**: `duck.track.ts` supplies `middlewareHooks` to `src/middlewares/track.ts`, so analytics for timeline interactions are emitted by a Redux middleware keyed on action types. Every migrated interaction therefore **dual-writes** — the components dispatch the Redux action *first*, so the middleware observes the pre-update state, then call the Zustand action. Severing that dual write requires relocating the analytics call sites off the middleware, and until then neither the duck nor `configure-store.ts` can be deleted.
 
-**Components to rewire**: `TraceTimelineViewer`, `SpanBarRow`, `SpanDetailRow`, `TimelineHeaderRow`. Rewired so far, all still dual-writing to Redux: `TracePage`, `TraceTimelineViewer`, `VirtualizedTraceView`, `SpanTreeOffset`, `SpanDetailSidePanel`.
+**Components to rewire**: `TraceTimelineViewer`, `SpanBarRow`, `SpanDetailRow`, `TimelineHeaderRow`. Rewired so far, all still dual-writing to Redux: `TracePage`, `TraceTimelineViewer`, `VirtualizedTraceView`, `SpanDetailSidePanel`. `SpanTreeOffset` no longer touches Redux (its only Redux use was the dead hover state removed in step 3).
 
 #### ✅ 1d. Deep Dependencies view modifiers (`ddg` duck - client-only fields)
 
@@ -1212,7 +1212,7 @@ The residual surface is now small enough to enumerate:
 | :--- | :--- | :--- |
 | `reducers/metrics.ts` | `Monitor/ServicesView` (`connect`) | Phase 2f — implementation open in [#4048](https://github.com/jaegertracing/jaeger-ui/pull/4048) |
 | `reducers/path-agnostic-decorations.ts`, `actions/path-agnostic-decorations.ts` | `DdgNodeContent`, `DeepDependencies/SidePanel/DetailsPanel` | Phase 2g |
-| `TraceTimelineViewer/duck.ts`, `duck.track.ts`, `middlewares/track.ts` | `TracePage`, `TraceTimelineViewer`, `VirtualizedTraceView`, `SpanTreeOffset`, `SpanDetailSidePanel` | Phase 1c steps 3–4 |
+| `TraceTimelineViewer/duck.ts`, `duck.track.ts`, `middlewares/track.ts` | `TracePage`, `TraceTimelineViewer`, `VirtualizedTraceView`, `SpanDetailSidePanel` | Phase 1c step 4 (step 3 done) |
 | Vestigial `connect()` wrappers whose `mapStateToProps` ignores state entirely | `SearchTracePage/SearchForm.tsx`, `TraceDiff/TraceDiff.tsx` | Phase 4b (mechanical) |
 | `utils/configure-store.ts`, `<Provider>` in `components/App/index.tsx`, `ReduxState` in `types/index.ts`, `types/TTraceTimeline.ts` | app shell, ~14 test files | Phase 4b/4c |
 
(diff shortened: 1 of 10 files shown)


docs/rfc/0004-state-management-strategy.md
```diff
@@ -978,12 +978,12 @@ This is the **largest and most performance-sensitive** slice. Split across multi
...
 
 **Why the duck is still alive**: `duck.track.ts` supplies `middlewareHooks` to `src/middlewares/track.ts`, so analytics for timeline interactions are emitted by a Redux middleware keyed on action types. Every migrated interaction therefore **dual-writes** — the components dispatch the Redux action *first*, so the middleware observes the pre-update state, then call the Zustand action. Severing that dual write requires relocating the analytics call sites off the middleware, and until then neither the duck nor `configure-store.ts` can be deleted.
 
-**Components to rewire**: `TraceTimelineViewer`, `SpanBarRow`, `SpanDetailRow`, `TimelineHeaderRow`. Rewired so far, all still dual-writing to Redux: `TracePage`, `TraceTimelineViewer`, `VirtualizedTraceView`, `SpanTreeOffset`, `SpanDetailSidePanel`.
+**Components to rewire**: `TraceTimelineViewer`, `SpanBarRow`, `SpanDetailRow`, `TimelineHeaderRow`. Rewired so far, all still dual-writing to Redux: `TracePage`, `TraceTimelineViewer`, `VirtualizedTraceView`, `SpanDetailSidePanel`. `SpanTreeOffset` no longer touches Redux (its only Redux use was the dead hover state removed in step 3).
```
yurishkuro:
```suggestion
**Components to rewire**: `TraceTimelineViewer`, `SpanBarRow`, `SpanDetailRow`, `TimelineHeaderRow`. Rewired so far, all still dual-writing to Redux: `TracePage`, `TraceTimelineViewer`, `VirtualizedTraceView`, `SpanDetailSidePanel`.
```

yurishkuro (changes requested):
Thanks — the premise holds up and the removal is complete. Two things to fix before merge.

**What I verified first:**

- **The dead-state claim is true.** On `main`, `SpanTreeOffset` writes the hover set but never destructures `hoverIndentGuideIds`; `VirtualizedTraceView` reads it (line 580) and puts it in `combinedProps` (line 702) where nothing consumes it; `SpanTreeOffset.css` has no `is-active` rule. The only other reference anywhere is a test fixture.
- **The removal in #3302 was intentional, so deleting is the right call.** `git log -S "indentGuide.is-active"` points at 19e3bf62 ("Better tree hierarchy for trace view"), which removed `'is-active': hoverIndentGuideIds.has(ancestorId)` together with the CSS, as part of a deliberate redesign to boxes and service-colored lines. That answers the open question in your description: don't restore it on Zustand.
- **Nothing was missed.** No reference to the removed action types in `duck.track.ts` or `middlewares/track.ts`, so there is no dual-write to unwind, and `UnconnectedSpanTreeOffset` had no importer outside its own test.

**1. The new test cannot fail — please drop it.** I copied its assertion onto unmodified `main`, where the entire machinery this PR removes is still present, and it passes:

```
Test Files  1 passed (1)     Tests  1 passed (1)
```

That is the problem: it cannot distinguish "dead state present" from "dead state removed". It pins the absence of a class name that nothing in the tree produces, and after this PR `SpanTreeOffset` has no `onMouseEnter` at all, so `fireEvent.mouseEnter` fires at a component with no hover behavior. The real guard against the state returning is that the field no longer exists on `TTraceTimeline`, which TypeScript enforces for free. I would rather delete the test and its comment than keep a green assertion that implies coverage it does not have.

**2. `data-ancestor-id` is now dead as well.** Its only production consumer was the `relatedTarget.dataset.ancestorId` comparison in the handlers being deleted. It survives, and the remaining tests both assert its value and use it as a query selector — so an attribute shipped on every indent guide in an 80k-span timeline now exists only to be tested. The elements already carry `data-testid={`indent-guide-${ancestor.spanID}`}`, which covers the selector need. Dropping the attribute and pointing those queries at the existing testid would finish the removal this PR is about.

**3. `check-label` is red** — it needs a `changelog:` label. I have added `changelog:refactoring`; contributors cannot set labels themselves.

Otherwise this is a clean, well-scoped deletion, and the RFC 0004 update is exactly right: status tracking only, ✅ with the PR link, and no rewriting of the surrounding prose.

---

thisis-manan:
All three addressed: dropped the untestable test and the dead data-ancestor-id (tests now select via data-testid), and applied your RFC line-986 suggestion. Re-requesting review.

thisis-manan:
@yurishkuro i think this is good for another look whenever you get a sec.

Everything from your review should be sorted:

- Dropped that regression test (it also passed on plain main, so it wasn't testing the real issue).
- Changed data-ancestor-id to the existing data-testid in the tests.
- Took your wording on the components-to-rewire line.

While I was in there, I also noticed a few docs were out of date after the removal, so I fixed those:

- ADR-0005 still listed VirtualizedTraceView under useSelector (that was its last occurrence) and SpanTreeOffset in the connect() list, even though it's a plain functional component now.
- RFC-0006 still said hoverIndentGuideIds was "still in Redux."
- Left the target/original store-shape tables in RFC-0004 and RFC-0006 as-is, since those are design snapshots rather than current-state docs.

Happy to split the doc changes into a separate PR if you'd prefer that.

Thanks!

thisis-manan:
Hey @yurishkuro, gentle nudge on this everything from your review should be addressed and it's still green + conflict-free. Happy to keep it moving, but if the priorities have shifted just let me know. Thanks

PR #4356, opened 2026-08-09
[chore]: Eliminate as any cast on Redux Store in App Provider

### LINKED ISSUE #4333: [chore]: Eliminate as any cast on Redux Store in App Provider
## What happened?

In `packages/jaeger-ui/src/components/App/index.tsx`, the Redux store is cast to `any` to satisfy React 19's stricter `Provider` type:

```tsx
<Provider store={store as any}>
```

This bypasses type checking on the store prop, which defeats the purpose of TypeScript.

## Steps to reproduce

1. Open `packages/jaeger-ui/src/components/App/index.tsx`
2. Locate the `<Provider>` component
3. Observe the `as any` cast on the `store` prop

## Expected behavior

The store should be properly typed without casting. Options include:
- Creating a typed `Provider` wrapper using `React.createContext`
- Using `react-redux`'s type-compatible exports for React 19
- Widening the store type to match what `Provider` expects

## Proposed fix

The exact approach depends on whether `react-redux` has released React 19-compatible types. If not, a thin typed wrapper is the cleanest path:

```tsx
// Option 1: Cast to the specific store type instead of any
<Provider store={store as ReturnType<typeof createStore>}>

// Option 2: Use a typed context wrapper
const ReduxContext = createContext<Store | null>(null);
```

## Additional context

The linter config (`vite.config.ts`) already downgrades `typescript/no-explicit-any` to `warn` because of cases like this. Eliminating this cast moves the codebase toward stricter type safety, which aligns with the ongoing migration work (RFC 0004).

A-Chronicle:
## Summary

Eliminates the `as any` cast on the Redux Store in the App Provider component by introducing a typed wrapper.

## Changes

- Create `TypedReduxProvider` wrapper component that properly types the store prop
- Remove `as any` cast on the store prop in `packages/jaeger-ui/src/components/App/index.tsx`

## Motivation

The Redux store was cast to `any` to satisfy React 19's stricter `Provider` type:

```tsx
<Provider store={store as any}>
```

This bypasses type checking on the store prop, which defeats the purpose of TypeScript. The linter config (`vite.config.ts`) already downgrades `typescript/no-explicit-any` to `warn` because of cases like this.

The fix creates a typed wrapper component that accepts `Store` from `react-redux`, eliminating the need for the `as any` cast while maintaining the same runtime behavior.

## Testing

- Existing tests pass without modification
- The typed wrapper maintains identical runtime behavior to the previous implementation

Closes #4333

### DIFF
--- a/packages/jaeger-ui/src/components/App/index.tsx
+++ b/packages/jaeger-ui/src/components/App/index.tsx
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import React from 'react';
-import { Provider } from 'react-redux';
+import { Provider, type Store } from 'react-redux';
 import { Route, Routes, Navigate } from 'react-router-dom';
 import { AppQueryClientProvider } from '../../query/app-query-client';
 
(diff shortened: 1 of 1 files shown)


jkowall:
Thanks for the cleanup — this is the same goal as #4456 (drop the redundant `as any` on `<Provider store={...}>`), which is already merged. That fix was the smaller change (`configureStore()` already returns `Store<any>`, so no cast or wrapper is needed). Closing this one as a duplicate. Appreciate the contribution.

PR #4357, opened 2026-08-09
fix(span-detail): handle null attribute values in AttributesTable (#4354)

### LINKED ISSUE #4354: [Bug]: AttributesTable crashes with TypeError when span attribute value is null or "null"
## Bug Description
In `AttributesTable.tsx`, `formatValue()` parses span attribute values and checks:
`else if (typeof parsed === 'object')`
Because `typeof null === 'object'` in JavaScript, when a span attribute has a value of `null` or a JSON string `"null"`, `formatValue()` invokes `Object.keys(null)`. This throws a `TypeError: Cannot convert undefined or null to object` and crashes the span detail view.

## Expected Behavior
`null` attribute values should be safely formatted as scalar values (`json-markup-null`) without throwing a `TypeError`.

### Steps to reproduce

1. Render a trace containing a span with a null attribute value or a JSON string "null".
2. Open or expand the span detail view to display AttributesTable.
3. Observe the TypeError: Cannot convert undefined or null to object in the browser console.

### Expected behavior

Span attributes with null values or "null" strings should render safely as null scalar markup without throwing a TypeError.

### Relevant log output

```shell
TypeError: Cannot convert undefined or null to object
    at Object.keys (<anonymous>)
    at formatValue (AttributesTable.tsx:85)
```

### Screenshot

_No response_

### Additional context

_No response_

### Jaeger backend version

Latest / Dev

### SDK

_No response_

### Pipeline

_No response_

### Stogage backend

_No response_

### Operating system

Windows / Linux

### Deployment model

Docker

### Deployment configs

```shell

```

RamavathChanti:
Resolves #4354

## Description of the changes
In `AttributesTable.tsx`, `formatValue()` parses span attribute values and checks `else if (typeof parsed === 'object')`. Because `typeof null === 'object'` in JavaScript, a span attribute with value `null` or a JSON string `"null"` causes `formatValue()` to invoke `Object.keys(null)`. This throws an unhandled `TypeError: Cannot convert undefined or null to object` and crashes the span detail view.

This PR:
- Adds a `parsed !== null` guard in `formatValue()` before evaluating `typeof parsed === 'object'`.
- Updates `scalarMarkup()` to handle `null` and `undefined` safely with `json-markup-null` class styling.
- Adds unit test coverage for `{ key: 'nullVal', value: null }` and `{ key: 'nullStr', value: 'null' }` in `AttributesTable.test.jsx`.

Unit tests added in `AttributesTable.test.jsx` covering null and "null" string attributes; verified test suite passes.

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/AttributesTable.test.jsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/AttributesTable.test.jsx
@@ -180,11 +180,11 @@ describe('<AttributesTable>', () => {
         linksGetter={(array, i) =>
           array.entries()[i].key === 'span.kind'
             ? [
-                {
-                  url: `http://example.com/?kind=${encodeURIComponent(array.entries()[i].value)}`,
-                  text: `More info about ${array.entries()[i].value}`,
-                },
-              ]
+              {
+                url: `http://example.com/?kind=${encodeURIComponent(array.entries()[i].value)}`,
+                text: `More info about ${array.entries()[i].value}`,
+              },
+            ]
             : []
         }
       />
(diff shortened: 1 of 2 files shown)


jkowall (changes requested):
The `parsed !== null` guard in `formatValue` is the right fix (`typeof null === 'object'` was calling `Object.keys(null)`). `scalarMarkup` now renders null/undefined with `json-markup-null` instead of `.toString()`. Existing scalar, header-list, and JsonView paths look unchanged.

Please add the tests the PR description claims. `AttributesTable.test.jsx` only changes indentation in the `linksGetter` fixtures. The existing `jsonValue.null` case is nested inside a JSON object (JsonView), so it never hits the crash path. Add a top-level `{ key: 'nullVal', value: null }` render and assert it shows as `json-markup-null` without throwing.

The string `'null'` does not go through `JSON.parse` (`tryParseJson` only parses values that start with `{` or `[`), so that case never crashed. A dedicated test is still useful, but the top-level JS `null` is the one that matters.

PR #4373, opened 2026-08-11
docs(rfc): replace RFC 0008 layout timings with reproducible topology benchmarks

DsThakurRawat:
Opening this follow-up to RFC 0008 (#3972) per Yuri's suggestion after I posted benchmark results on that thread.

The RFC's dagre vs dot table came from dense random graphs and the text flags it as order-of-magnitude only. I reran the comparison with a seeded harness that mirrors what plexus `getLayout.ts` actually does (same dot options `DAG.tsx` uses: `nodesep 1.5, rankdir TB, ranksep 1.6, splines polyline`, viz `renderString` with plain output), dagre given the same node sizes and separations, warmup plus median of 3 runs, at 200/500/1200 nodes, across three topology shapes: the RFC's dense random generator, a sparse layered service shape, and the same shape with shared-infra hubs.

The headline: topology shape moves layout cost by 20-80x at a fixed node count and flips which engine is faster. `dot` wins the dense shape, `dagre` wins the sparse layered ones, and both are unusable at 500 nodes on realistic shapes. That does not weaken the RFC's conclusions, it makes the focal-neighborhood and fallback design more necessary, so the recommendations are untouched.

What this changes in the doc:
- The timing table is replaced with the measured one and the harness is described so the numbers are reproducible.
- Spike 3 gets an extra acceptance step: record topology stats (node and edge counts, out-degree distribution, hub fan-in) from real exported data, since shape dominates cost.
- The viz version reference is corrected from 3.29.0 to 3.28.0, which is what package.json and the lockfile pin.

| Topology | Nodes (n) | Edges | Graphviz `dot` (viz WASM 3.28.0) | `@dagrejs/dagre` 3.1.1 |
|---|---|---|---|---|
| Dense random (spanning tree + 0.6n extra forward edges) | 200 | 319 | 77 ms | 458 ms |
| Dense random | 500 | 799 | 483 ms | 3,390 ms |
| Dense random | 1,200 | 1,919 | 4,543 ms | 59,688 ms |
| Layered sparse (7-layer service DAG) | 500 | 1,062 | 9,302 ms | 7,042 ms |
| Layered sparse | 1,200 | 2,545 | did not finish (150 s timeout) | 55.4 s |
| Layered + shared-infra hubs ("realistic" microservices) | 200 | 488 | 667 ms | 1,163 ms |
| Layered + shared-infra hubs | 500 | 1,234 | 37,576 ms | 10,179 ms |
| Layered + shared-infra hubs | 1,200 | 2,915 | did not finish (150 s timeout) | 63.7 s |

Seed sensitivity was checked on the realistic shape (seeds 42, 7, 99) and the ranking holds.

### DIFF
--- a/docs/rfc/0008-graph-visualization-stack.md
+++ b/docs/rfc/0008-graph-visualization-stack.md
@@ -409,8 +409,8 @@ Download size is not scored in either matrix. Jaeger already ships 468 KB gzippe
 | **License** | 🟢 MIT | 🟢 MIT | 🟢 MIT |
 | **Layered DAG quality** | 🟢 the reference implementation | 🟢 same Sugiyama family, with `network-simplex`, `tight-tree`, and `longest-path` rankers | 🟢 Sugiyama with pluggable decross and coord operators |
 | **Trees, which is what most traces are** | 🟢 | 🟢 | 🟢 ¹ |
-| **Layered layout at 500 services** | 🟡 1.9 s, which is why `dot` is disabled there ² | 🟡 1.1 s — faster than `dot`, still too slow to feel interactive | 🟡 untested |
-| **Beyond ~1,200** | 🔴 `dot` did not finish; `sfdp` is the fallback ² | 🔴 6.3 s at 1,200, out of memory at 5,000 | 🔴 untested |
+| **Layered layout at 500 services** | 🔴 483 ms to 37.6 s depending on topology, which is why `dot` is disabled there ² | 🔴 3.4 s to 10.2 s over the same graphs, ahead of `dot` on service-shaped ones and behind it on dense ones | 🟡 untested |
+| **Beyond ~1,200** | 🔴 4.5 s on dense graphs, no finish inside 150 s on service-shaped ones; `sfdp` is the fallback ² | 🔴 55 s to 64 s across all three topologies, out of memory at 5,000 | 🔴 untested |
 | **Edge routing** | 🟢 splines available, polylines in use | 🟡 point lists, no spline routing | 🟡 point lists |
 | **Layout-direction toggle** | 🟢 `rankdir` | 🟢 `rankdir` | 🟡 rotate the result yourself |
 | **Compound (nested) graphs** | 🟡 clusters exist; Plexus's pipeline does not carry them as written ³ | 🟢 native, via `setParent` | 🔴 |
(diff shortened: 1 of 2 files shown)


yurishkuro:
how were you generating the graphs for testing, using some script we already have or using something else? If the latter I would like to commit the generator to make these findings reproducible.

DsThakurRawat:
It was a custom script I wrote to test the different topologies. Just pushed it to the PR at `scripts/rfc0008-bench.js`.
It has the seeded generators for all three shapes, so running it with `node` reproduces the exact edge counts and rankings from the table. 

Two small notes if you run it:
- `@viz-js/viz` is resolved from `packages/plexus` so it tests the version pinned in the repo.
- `@dagrejs/dagre` isn't a repo dependency, but the header comment has the quick command to run it.

DsThakurRawat:
hi @yurishkuro , no rush on this one , i mostly want to check its still the shape
that you had in mind. the generator you asked for is in  `scripts/rfc0008-bench.js`  and ci is green, so its ready from my side.
the part  i am least sure about is that it edits the merged rfc in place rather than 
leaving the  original numbers and putting the measured ones in follow-up section
underneath. if you'd rather it read that way i am happy to restructure it.

yurishkuro (dismissed):

DsThakurRawat:
pushed two commits on top of your approval, which i realise undoes it, so say the word if you'd rather have them separately.

copilot caught something real in the doc. the criteria matrix above the timing table still had 1.9 s, 1.1 s and 6.3 s in it, and those are the old synthetic figures this pr deleted, so the document was ranking the two engines one way in that table and the other way twelve lines further down. the two rows now carry the ranges the harness actually produced and say that the spread is topology rather than size.

the other one is a genuine bug in the harness i committed for you. `dense 2` never returns: gendense lays a spanning tree and then wants 0.6n more acyclic edges, and at n=2 the only edge that could be added is already in the tree, so the retry loop spins. n now has to be a whole number and at least the minimum the chosen topology can build. generation itself is untouched, and the STATS output is identical for (dense, 200, 42), (realistic, 500, 42) and (layered, 1200, 7).

the third thing it raised was fair too, so the protocol sentence now says which rows are single runs. the header comment in the harness has always said the n=1200 and extra-seed rows use RUNS=1 while the sentence above the table promised a median of three.

yurishkuro (approved):

yurishkuro (approved):

PR #4380, opened 2026-08-13
chore(scripts): remove jq dependency from check-overrides (#4345)

### LINKED ISSUE #4345: [chore]: `pnpm run lint` fails on a clean setup because `jq` is an undocumented prerequisite
### Requirement

Either document `jq` as a prerequisite, or remove the dependency on it.

## What happens now

On a clean WSL2 / Ubuntu setup, following the CONTRIBUTING.md steps (nvm install, `corepack enable pnpm`, `pnpm install --frozen-lockfile`), running `pnpm run lint` produces:

```
[check-overrides     ] ./scripts/check-overrides.sh: line 30: jq: command not found
[check-overrides     ] [ELIFECYCLE] Command failed.
[oxlint              ] [ELIFECYCLE] Command failed.
[check-tsx-naming    ] [ELIFECYCLE] Command failed.
[tsc-lint            ] [ELIFECYCLE] Command failed.
[fmt-lint            ] [ELIFECYCLE] Command failed.
[check-license       ] [ELIFECYCLE] Command failed.
[knip                ] [ELIFECYCLE] Command failed.
ERROR: "check-overrides" exited with 1.
```

Seven of those eight tasks are fine. They were killed by `npm-run-all --parallel` when `check-overrides` exited. Nothing in the output indicates that, so the apparent symptom is that linting is entirely broken.

**Confirmed:** `sudo apt install -y jq` resolves it, and `pnpm run lint` then completes cleanly (`check-overrides` reports "All overrides are active").

## Why it isn't caught

`scripts/check-overrides.sh:25` pipes into `jq` with no guard, under `set -euo pipefail`. `jq` is not listed in the CONTRIBUTING.md prerequisites (nvm, Node.js, pnpm), and no workflow installs it, so CI relies on `jq` being preinstalled on GitHub's `ubuntu-latest` runners. It always is, so this cannot surface in CI.

CONTRIBUTING.md also has a Windows section directing users to set up via WSL and follow the standard installation steps, which is the path that hits this.

## Possible approaches

1. Add a guard so the script fails with a clear message naming the missing binary, and add `jq` to the prerequisites. `check-tsx-naming.sh` already uses this idiom for its Node version requirement, so it would be consistent rather than new.
2. Drop `jq` and do the JSON parsing with `node -e`, since Node is already a pinned dependency. The filter is small enough to port.

I lean toward 2, as it makes the existing prerequisite list correct rather than longer, but happy to do either. Would submit a PR.

Jeevanm2004:
Fixes #4345

This PR removes `jq` as a dependency in `scripts/check-overrides.sh` by replacing `jq` JSON queries with standard Node.js inline execution (`node -e`).

### Problem
On a clean development environment, running `pnpm run lint` failed at the `check-overrides` step because `jq` was an undocumented prerequisite and not installed by default setup.

### Changes
- Replaced `jq` parsing of `pnpm config get overrides --json` with Node.js (`fs.readFileSync(0, "utf-8")` + `JSON.parse`).
- Replaced `jq` evaluation of `pnpm why` JSON output with Node.js inline check.
- Retained exact filtering and error exit behavior.

## Test Plan

- Ran `bash scripts/check-overrides.sh` locally — verified output: `All overrides are active.`
- Ran `pnpm run lint` — verified clean completion with 0 errors across all lint steps.

### DIFF
--- a/scripts/check-overrides.sh
+++ b/scripts/check-overrides.sh
@@ -22,11 +22,16 @@ set -euo pipefail
 
 # Extract simple (string-valued, non-reference, non-nested) override package
 # names from pnpm's active overrides configuration, one per line.
-overrides=$(pnpm config get overrides --json | jq -r '
-  (. // {}) | to_entries[]
-  | select((.key | contains(">")) | not)
-  | select((.value | type == "string") and (.value | startswith("$") | not))
-  | .key
+overrides=$(pnpm config get overrides --json | node -e '
+  const fs = require("fs");
+  try {
+    const data = JSON.parse(fs.readFileSync(0, "utf-8")) || {};
+    for (const [key, val] of Object.entries(data)) {
+      if (!key.includes(">") && typeof val === "string" && !val.startsWith("$")) {
+        console.log(key);
+      }
+    }
+  } catch {}
 ')
 
 if [ -z "$overrides" ]; then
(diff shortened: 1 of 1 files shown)


yurishkuro:
not a problem that needs to be fixed. The error already says it all "jq: command not found"

PR #4385, opened 2026-08-14
fix(tooltip): update flamegraph tooltip position with cursor

### LINKED ISSUE #4379: [Bug]: Flamegraph tooltip does not follow the cursor within the same frame
### What happened?

As a Jaeger user, I want the Flamegraph tooltip to follow my cursor as I move within a frame, so I can easily read it while exploring different parts of the trace.

Currently, when hovering over a Flamegraph frame, the tooltip appears at the cursor position. However, when I move the cursor within the same frame, the tooltip stays at its original position.

This behavior can be seen in:

`packages/jaeger-ui/src/components/TracePage/TraceFlamegraph/index.tsx
`
The mouse handler contains an early return:

`if (target === hoveredFrameRef.current) return;`

The tooltip position is based on the mouse coordinates, but when the same frame remains hovered, the handler returns early and the tooltip position is not updated.

### Steps to reproduce

1. Open Jaeger and select a trace.
2. Open the Flamegraph view.
3. Hover over any frame until its tooltip appears.
4. Keep the cursor inside the same frame.
5. Move the cursor to different positions within that frame.
6. Observe the tooltip position.

### Expected behavior

The tooltip should follow the cursor as I move within the same frame, or at least update its position to remain close to the cursor.

In the video, first display a fixed tooltip. After the code improvements, at the end, show the expected solutions.

https://github.com/user-attachments/assets/fd1947b5-aa30-401e-995d-386f9f085c68

### Screenshot

_No response_

### Additional context

This appears to be a UI behavior issue rather than a backend or trace-data issue. It can be reproduced using the local Jaeger UI with a trace containing a Flamegraph.

A regression test should also verify that:

- the tooltip appears when entering a frame;
- the tooltip position updates while moving within the same frame;
- tooltip content still updates when moving to another frame;
- the tooltip disappears when leaving the Flamegraph/frame.

Syed-Bilal-Haider-Engineer:
-  Resolves #4379 

## Description 
- Fixed the Flamegraph tooltip so its position is updated when the cursor moves within the same frame.
- Updated the mouse handling logic in `packages/jaeger-ui/src/components/TracePage/TraceFlamegraph/index.tsx` so the tooltip position is not skipped when the hovered frame has not changed.
- The **tooltip** now follows the cursor while moving inside the same **Flamegraph frame.**

## Changes
- Reproduced the issue by opening a trace and hovering over a Flamegraph frame.
- Moved the cursor within the same frame and confirmed that the tooltip follows the cursor.
- Verified that the tooltip still appears correctly when moving between different frames.
- Ran the relevant tests and lint checks.

## screenshot
https://github.com/user-attachments/assets/fd1947b5-aa30-401e-995d-386f9f085c68

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceFlamegraph/index.test.jsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceFlamegraph/index.test.jsx
@@ -412,19 +412,35 @@ describe('<TraceFlamegraph />', () => {
       expect(mockChart.search).toHaveBeenCalledWith('svc-a: myOp');
     });
 
-    it('mousemove over g.frame shows tooltip', async () => {
-      render(<TraceFlamegraph trace={otelTrace} />);
+    it('updates tooltip position when mouse moves within the same g.frame', async () => {
+      const { container } = render(<TraceFlamegraph trace={otelTrace} />);
+
       const svg = screen.getByTestId('flamegraph-chart').querySelector('svg');
       const g = document.createElementNS('http://www.w3.org/2000/svg', 'g');
       g.classList.add('frame');
-      g.__data__ = { data: { name: 'svc: op', value: 500000, duration: 500000, count: 3 } };
+      g.__data__ = {
+        data: {
+          name: 'svc: op',
+          value: 100,
+          duration: 100,
+          count: 1,
+        },
+      };
       svg.appendChild(g);
 
       await act(() => {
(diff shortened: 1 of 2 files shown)


jkowall (changes requested):
DCO is failing on the commits. Please rebase with sign-off (`git rebase HEAD~N --signoff`) and force-push.

The tooltip change itself looks right: dropping the same-frame early return so `mousemove` updates position.

PR #4406, opened 2026-08-24
test(genai): Verify icons, Logical View pruning, and the GenAI tab wire together on a real trace

swetalin-10:
Icons, the Logical View filter, and the GenAI tab each ship with their own isolated unit tests, but nothing exercises them together through the real `VirtualizedTraceView -> SpanBarRow -> SpanDetail -> GenAITab` wiring on one trace. This is the "Integration Tests" item from the original proposal's Testing Strategy (5.1) that was still outstanding going into Milestone 4.

## Description of the changes
New `GenAIFlow.integration.test.jsx`, mounted against the real `VirtualizedTraceViewImpl` using the existing `genaiTestTrace.json` fixture - no new fixture, same `itemRenderer`/mocked-`ListView` pattern `VirtualizedTraceView.test.jsx` already uses, same `.integration.test.jsx` naming `SpanDetailRow.integration.test.jsx` already established.

- Icon renders correctly for each GenAI span kind (AI Agent, LLM call, MCP Tool call, Retrieval) on the real multi-agent trace, and no icon at all for a plain infra span (`GET /validate`)
- `prunedServices` actually removes a filtered service and its whole subtree from the *rendered* rows - pruning `auth-service` and checking `dataLength`/`getKeyFromIndex` directly, not just `generateRowStates`'s return value, which is all the existing D4 tests check
- Expanding the `chat model-a` span reaches the real `SpanDetail -> GenAITab` wiring and shows its actual fixture data: provider `example-ai`, model `model-a`, the `Tokens:` row, and the real prompt/completion text ("Why did the checkout latency regress this afternoon?" / "Latency rose with a slow pod rollout...") - not a mocked span

While writing this I initially assumed Logical View's pruning checks each span individually for GenAI descendants before hiding it. Reading `generateRowStates.ts` directly showed that's wrong - pruning is a flat per-service-name match that drops the whole subtree; the "don't hide a real agent's infra" logic lives upstream in whatever computes the effective `prunedServices` set, not here. Adjusted the test to match what the code actually does.

4 new tests, all passing. Full `TraceTimelineViewer` suite (735 tests) still green after, no regressions. `fmt`, `oxlint`, `tsc-lint`, `knip`, `check-license`, `check-copyright-year`, `check-tsx-naming` all clean, production build succeeds.

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/GenAIFlow.integration.test.jsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/GenAIFlow.integration.test.jsx
@@ -1,0 +1,166 @@
+// Copyright (c) 2026 The Jaeger Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * Integration test for the GenAI rendering pipeline on one real trace: span-type
+ * icons, the Logical View service-pruning filter, and the SpanDetail -> GenAI tab
+ * wiring reached by actually expanding a row. Each deliverable already has its own
+ * unit tests (classification, icon component, generateRowStates, GenAITab), but
+ * those mount each piece in isolation - nothing currently renders them together
+ * through the real VirtualizedTraceView -> SpanBarRow -> SpanDetail -> GenAITab
+ * wiring on an actual multi-agent trace, which is what this file checks.
+ */
+
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import '@testing-library/jest-dom';
+
+import DetailState from './SpanDetail/DetailState';
+import { VirtualizedTraceViewImpl } from './VirtualizedTraceView';
+import transformTraceData from '../../../model/transform-trace-data';
+import memoizedTraceCriticalPath from '../CriticalPath/index';
+import genaiTestTrace from './genaiTestTrace.json';
+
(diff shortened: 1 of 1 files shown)


yurishkuro (commented):
.

yurishkuro:
Thanks for picking up the outstanding Testing Strategy item — the pruning case you found is a real gap, and it's now covered.

I'm landing that one assertion in #4436, as an addition to the existing `service filter pruning` block in `VirtualizedTraceView.test.jsx`. You're right that nothing checked that a pruned service actually disappears from the rows handed to `ListView`; the existing tests there only assert on the placeholder row's key, index, and height.

On the rest of the file, two things don't hold up:

- **It isn't a tier above the unit tests.** Mocking `ListView`, capturing its props, and calling `itemRenderer` by hand is the same seam `VirtualizedTraceView.test.jsx` already uses, and each row is still `render`ed on its own. So the pruning case needed no new file — in the existing block it reuses `makeSpansWithServices` and costs three lines instead of a duplicated 60-line `mockProps`.
- **`genaiTestTrace.json` isn't real instrumentation output.** It is generated by `scripts/utils/make-genai-trace.js`. Asserting classification against it proves what a hand-built span object proves, so it doesn't add coverage over `GenAISpanIcon.test.tsx` — while pinning the tests to hex span IDs and fixture prose that the generator can change under them.

PR #4408, opened 2026-08-29
refactor(plexus): migrate Digraph to functional component

### LINKED ISSUE #3386: [UI Refactor] Migrate Digraph to Functional Component
This issue is part of the larger effort to modernize the Jaeger UI codebase (Parent Issue #2610). The goal is to refactor the `Digraph` component from a React Class Component to a React Functional Component using Hooks.

**Target Component:** `packages/plexus/src/Digraph/index.tsx`

### Acceptance Criteria

### Plexus-Specific Guidance

| --- | --- | --- |

| --- | --- | --- |

### Resources

amarkdotdev:
Re-applies the Digraph functional migration with a fix for the callback stability issue that led to the earlier revert.

`setSizeVertices` now reads `edges`, `layoutManager`, and `measurableNodesKey` from refs so MeasurableNodesLayer does not re-render when the parent passes a new edges array reference.

Fixes #3386

## Test plan
- [x] `pnpm --filter @jaegertracing/plexus test`
- [ ] Smoke-test a graph view with zoom/minimap enabled

### DIFF
--- a/packages/plexus/src/Digraph/index.test.tsx
+++ b/packages/plexus/src/Digraph/index.test.tsx
@@ -1,0 +1,107 @@
+// Copyright (c) 2026 Uber Technologies, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+import React from 'react';
+import { render } from '@testing-library/react';
+import Digraph from './index';
+
+// Minimal layer for tests that don't need measurable nodes
+const renderNodeLayer = {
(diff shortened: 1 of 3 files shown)


amarkdotdev:
Hi team, soft bump for maintainer review when you get a moment.

amarkdotdev:
Rebased on latest main. CI should pick it up shortly.

PR #4412, opened 2026-08-30
fix(trace-timeline): make span-details divider draggable from the header

### LINKED ISSUE #3925: [Bug]: Column width adjustment on Timeline view is inconsistent

KKamJi98:

Resolves #3925

## Description of the changes

The trace timeline renders two column dividers, but only the service-name one was draggable from the fixed header row. The span-details boundary existed solely as a `VerticalResizer` inside the body layout (`index.tsx`), so dragging it from the header did nothing.

The fix follows the name-column pattern already in the header: `TimelineHeaderRow` is `position: fixed` and a `VerticalResizer` dragger's height reaches down through the table, so a single header-owned instance serves both the header row and the table rows.

- Move the span-details resizer into `TimelineHeaderRow`, wired to `setSidePanelWidth` through a new `onSidePanelWidthChange` prop, and remove the body instance from `index.tsx`. One divider owns the boundary instead of two instances coordinated by CSS.
- Keep left-origin coordinates (`position = 1 - sidePanelWidth`, flipped back in `onChange`) instead of `rightSide`. The `rightSide` variant mirrors the dragger with `scaleX(-1)`, which pushed the grip away from the actual column border at asymmetric widths (about 540px off at `sidePanelWidth = 0.56`).
- Preserve the body resizer's bounds, so the timeline column keeps `MIN_TIMELINE_COLUMN_WIDTH` and the panel keeps `SIDE_PANEL_WIDTH_MIN` / `SIDE_PANEL_WIDTH_MAX`.
- Gate the divider on `timelineBarsVisible` as well as `sidePanelVisible`. In tree-only mode the panel width is derived from `spanNameColumnWidth`, so the divider would sit on top of the name-column resizer and intercept drags without any visible effect.
- Render the side-panel header cell before the dividers, so a drag starting exactly on the boundary is not intercepted by the cell.
- Memoize the width adapter passed to the resizer, so hover-driven rerenders of the header no longer recreate the `DraggableManager` and cancel an in-flight drag.
- Drop the `.TraceTimelineViewer--sidePanelLayout > .VerticalResizer` z-index rule in `index.css`, which existed only to lift the body resizer above the fixed side panel.

Tests: added regressions for divider/cell ordering, boundary placement at asymmetric widths (0.3 and 0.56), bounds, tree-only visibility, and drag continuity across a rerender. Removed the `VerticalResizer` mock and its assertions from `index.test.jsx`, which became vacuous once `index.tsx` stopped importing the component.

## Testing done

- `vitest run` on `index.test.jsx` + `TimelineHeaderRow.test.jsx`: 40 passed.
- Full `jaeger-ui` suite: 271 files, 3706 tests passed.
- `pnpm run lint` (fmt-lint, tsc-lint, oxlint, check-license, check-copyright-year, check-tsx-naming, check-overrides, knip): clean.
- Manual check against the issue repro in side-panel mode: the grip, the header cell edge, the side panel's left edge and the main column's right edge line up on the same x, and dragging works from both the header row and the table rows.

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow/TimelineHeaderRow.css
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow/TimelineHeaderRow.css
@@ -35,3 +35,10 @@ SPDX-License-Identifier: Apache-2.0
 .TimelineHeaderRow--sidePanelCell {
   border-left: 1px solid var(--border-strong);
 }
+
+/* Cap header resizers to the header row so they do not overlay the table's
+   own dividers (issue #3925: the span-details divider was only draggable in
+   the table because the header's full-height dragger intercepted it). */
+.TimelineHeaderRow .VerticalResizer--dragger {
+  height: var(--timeline-header-row-height);
+}
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow/TimelineHeaderRow.css
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow/TimelineHeaderRow.css
@@ -36,9 +36,11 @@ SPDX-License-Identifier: Apache-2.0
   border-left: 1px solid var(--border-strong);
 }
 
-/* Cap header resizers to the header row so they do not overlay the table's
+/* Cap header draggers to the header row so they do not overlay the table's
    own dividers (issue #3925: the span-details divider was only draggable in
-   the table because the header's full-height dragger intercepted it). */
-.TimelineHeaderRow .VerticalResizer--dragger {
-  height: var(--timeline-header-row-height);
+   the table because the header's full-height dragger intercepted it). Scope
+   the cap to the flipped (right-side) resizer only: the name-column divider
+   has no body counterpart and relies on its full-height dragger. */
+.TimelineHeaderRow .VerticalResizer.is-flipped .VerticalResizer--dragger {
+  height: var(--timeline-header-row-height, 38px);
 }
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow/TimelineHeaderRow.css
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow/TimelineHeaderRow.css
@@ -35,12 +35,3 @@ SPDX-License-Identifier: Apache-2.0
 .TimelineHeaderRow--sidePanelCell {
   border-left: 1px solid var(--border-strong);
 }
-
-/* Cap header draggers to the header row so they do not overlay the table's
-   own dividers (issue #3925: the span-details divider was only draggable in
-   the table because the header's full-height dragger intercepted it). Scope
-   the cap to the flipped (right-side) resizer only: the name-column divider
-   has no body counterpart and relies on its full-height dragger. */
-.TimelineHeaderRow .VerticalResizer.is-flipped .VerticalResizer--dragger {
-  height: var(--timeline-header-row-height, 38px);
-}
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow/TimelineHeaderRow.test.jsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow/TimelineHeaderRow.test.jsx
@@ -22,8 +22,14 @@ vi.mock('../TimelineRow', () => {
 });
 
 vi.mock('../../../common/VerticalResizer', () => ({
-  default: ({ position, min, max }) => (
-    <div data-testid="vertical-resizer" data-position={position} data-min={min} data-max={max} />
+  default: ({ position, min, max, rightSide }) => (
+    <div
+      data-testid="vertical-resizer"
+      data-position={position}
+      data-min={min}
+      data-max={max}
+      data-right-side={rightSide ? 'true' : undefined}
+    />
   ),
 }));
 
@@ -152,6 +158,23 @@ describe('<TimelineHeaderRow>', () => {
       const resizer = screen.getAllByTestId('vertical-resizer')[0];
       expect(resizer).toHaveAttribute('data-max', String(1 - sidePanelWidth));
     });
+
+    it('renders two resizers: the name-column divider and the side-panel divider', () => {
+      render(<TimelineHeaderRow {...sidePanelProps} />);
+      const resizers = screen.getAllByTestId('vertical-resizer');
+      expect(resizers).toHaveLength(2);
+      expect(resizers[0]).not.toHaveAttribute('data-right-side');
+      expect(resizers[1]).toHaveAttribute('data-right-side', 'true');
+    });
+
+    it('configures the side-panel divider like the body resizer', () => {
+      render(<TimelineHeaderRow {...sidePanelProps} />);
+      const resizers = screen.getAllByTestId('vertical-resizer');
+      expect(resizers[1]).toHaveAttribute('data-position', String(1 - sidePanelWidth));
+      // Keeps the timeline column at MIN_TIMELINE_COLUMN_WIDTH while respecting SIDE_PANEL_WIDTH_MAX.
+      expect(resizers[1]).toHaveAttribute('data-min', String(1 - Math.min(0.7, 1 - nameColumnWidth - 0.05)));
+      expect(resizers[1]).toHaveAttribute('data-max', String(1 - 0.2));
+    });
   });
 
   it('renders the TimelineCollapser', () => {
@@ -194,11 +217,12 @@ describe('<TimelineHeaderRow>', () => {
         resizerMax: 0.8,
       };
 
-      it('renders the VerticalResizer', () => {
+      it('renders only the name-column resizer (the side-panel divider is gated on bars)', () => {
         render(<TimelineHeaderRow {...treeOnlySidePanelProps} />);
-        const resizer = screen.getAllByTestId('vertical-resizer')[0];
-        expect(resizer).toBeInTheDocument();
-        expect(resizer).toHaveAttribute('data-max', '0.8');
+        const resizers = screen.getAllByTestId('vertical-resizer');
+        expect(resizers).toHaveLength(1);
+        expect(resizers[0]).not.toHaveAttribute('data-right-side');
+        expect(resizers[0]).toHaveAttribute('data-max', '0.8');
       });
     });
   });
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow/TimelineHeaderRow.test.jsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow/TimelineHeaderRow.test.jsx
@@ -149,7 +149,7 @@ describe('<TimelineHeaderRow>', () => {
 
     it('sets resizer max to 1 - sidePanelWidth', () => {
       render(<TimelineHeaderRow {...sidePanelProps} />);
-      const resizer = screen.getByTestId('vertical-resizer');
+      const resizer = screen.getAllByTestId('vertical-resizer')[0];
       expect(resizer).toHaveAttribute('data-max', String(1 - sidePanelWidth));
     });
   });
@@ -196,7 +196,7 @@ describe('<TimelineHeaderRow>', () => {
 
       it('renders the VerticalResizer', () => {
         render(<TimelineHeaderRow {...treeOnlySidePanelProps} />);
-        const resizer = screen.getByTestId('vertical-resizer');
+        const resizer = screen.getAllByTestId('vertical-resizer')[0];
         expect(resizer).toBeInTheDocument();
         expect(resizer).toHaveAttribute('data-max', '0.8');
       });
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow/TimelineHeaderRow.test.jsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow/TimelineHeaderRow.test.jsx
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import React from 'react';
-import { render, screen } from '@testing-library/react';
+import { fireEvent, render, screen } from '@testing-library/react';
 import '@testing-library/jest-dom';
 
 import TimelineHeaderRow from './TimelineHeaderRow';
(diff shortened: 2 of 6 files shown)


packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow/TimelineHeaderRow.css
```diff
@@ -35,3 +35,10 @@ SPDX-License-Identifier: Apache-2.0
 .TimelineHeaderRow--sidePanelCell {
   border-left: 1px solid var(--border-strong);
 }
+
+/* Cap header resizers to the header row so they do not overlay the table's
+   own dividers (issue #3925: the span-details divider was only draggable in
+   the table because the header's full-height dragger intercepted it). */
+.TimelineHeaderRow .VerticalResizer--dragger {
```
jkowall: This selector hits both header resizers. The name-column divider has no body counterpart; its 100vh dragger is how you resize from the table. Restrict this to the new flipped/right-side resizer, e.g. `.TimelineHeaderRow .VerticalResizer.is-flipped .VerticalResizer--dragger`.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow/TimelineHeaderRow.tsx
```diff
@@ -96,6 +98,15 @@ export default function TimelineHeaderRow(props: TimelineHeaderRowProps) {
           max={resizerMax}
         />
       )}
+      {sidePanelVisible && (
```
jkowall: Gate on `timelineBarsVisible` as well, matching the body resizer in `TraceTimelineViewer/index.tsx`. When bars are hidden `timelineColumnWidth` is 0, so this lands on top of the name-column resizer.

`min`/`max` also need the body's extra constraint (`1 - spanNameColumnWidth - MIN_TIMELINE_COLUMN_WIDTH`), not the raw `SIDE_PANEL_WIDTH_*` constants. The body does not use `rightSide`; it flips in `onChange`. Either match that, or keep `rightSide` and pass equivalent flipped bounds.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow/TimelineHeaderRow.test.jsx
```diff
@@ -149,7 +149,7 @@ describe('<TimelineHeaderRow>', () => {
 
     it('sets resizer max to 1 - sidePanelWidth', () => {
       render(<TimelineHeaderRow {...sidePanelProps} />);
-      const resizer = screen.getByTestId('vertical-resizer');
+      const resizer = screen.getAllByTestId('vertical-resizer')[0];
```
jkowall, line 158: Taking `[0]` only avoids the duplicate-test-id failure. Assert two resizers when the panel is open (position `nameColumnWidth + timelineColumnWidth`, min/max, `rightSide`) and still one when it is not. Extend the mock so `rightSide` is visible on the test node. `onSidePanelWidthChange` is also missing from these props.

jkowall (dismissed):
The extra header resizer is the right approach for #3925, but the height cap and the new resizer's visibility/bounds are not safe as written.

Name-column resize currently depends on the header dragger spanning the table (`100vh` in `VerticalResizer.css`). This CSS rule clips every header dragger to the row, so the name-column divider would only be draggable from the header. Scope the cap to the new right-side resizer.

The body span-details resizer is gated on `timelineBarsVisible` and uses a min that keeps `spanNameColumnWidth + MIN_TIMELINE_COLUMN_WIDTH`. This header copy is always on when the panel is open, with fixed `SIDE_PANEL_WIDTH_MIN/MAX`. When bars are hidden both header resizers sit on the same x and the later one steals the drag (`setSidePanelWidth` is a no-op because `effectiveSidePanelWidth` is `1 - spanNameColumnWidth`). When bars are visible a header drag can shrink the timeline below the body minimum.

Tests only switched `getByTestId` to `getAllByTestId()[0]`. They do not assert the second resizer's position, `rightSide`, min/max, or that it is absent when the panel or bars are hidden.

`VerticalResizer` is mouse-only (`onMouseDown`, `aria-hidden`). Pre-existing, not blocking.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow/TimelineHeaderRow.tsx
KKamJi98: All valid, thanks for the careful review. I'll fix all of it: scope the CSS cap to the flipped resizer only, gate the header resizer on `timelineBarsVisible` like the body one, mirror the body's min/max bounds so the timeline column keeps its minimum, and extend the tests to assert the second resizer. Patch to follow.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow/TimelineHeaderRow.css
KKamJi98: Fixed in 0750d6ad: the cap now targets .is-flipped only, so the name-column divider keeps its full-height dragger.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow/TimelineHeaderRow.tsx
KKamJi98: Fixed in 0750d6ad: gated on timelineBarsVisible, and the bounds now mirror the body resizer (timeline keeps its minimum width).

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow/TimelineHeaderRow.test.jsx
KKamJi98: Fixed in 0750d6ad: tests now assert two resizers when the panel is open (position/min/max/rightSide on the second) and exactly one in tree-only mode.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow/TimelineHeaderRow.tsx
```diff
@@ -96,6 +98,15 @@ export default function TimelineHeaderRow(props: TimelineHeaderRowProps) {
...
+        <VerticalResizer
+          position={nameColumnWidth + timelineColumnWidth}
+          onChange={onSidePanelWidthChange}
+          min={SIDE_PANEL_WIDTH_MIN}
+          max={SIDE_PANEL_WIDTH_MAX}
+          rightSide
+        />
+      )}
```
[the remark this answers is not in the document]
KKamJi98: Fixed in 0750d6ad: gated on timelineBarsVisible.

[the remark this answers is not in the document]
KKamJi98: Fixed in 0750d6ad: position is now 1 - sidePanelWidth, matching the rightSide convention used by DetailsPanel.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow/TimelineHeaderRow.css
```diff
@@ -35,3 +35,10 @@ SPDX-License-Identifier: Apache-2.0
...
+  height: var(--timeline-header-row-height);
+}
```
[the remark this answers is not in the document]
KKamJi98: Fixed in 0750d6ad: added a 38px fallback.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow/TimelineHeaderRow.test.jsx
```diff
@@ -149,7 +149,7 @@ describe('<TimelineHeaderRow>', () => {
...
       expect(resizer).toHaveAttribute('data-max', String(1 - sidePanelWidth));
```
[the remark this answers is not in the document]
KKamJi98: Fixed in 0750d6ad: added assertions for the second resizer's configuration.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/TimelineHeaderRow/TimelineHeaderRow.tsx
KKamJi98: @jkowall, could you take another look when you have a chance? I've addressed your feedback on the resizer scope, visibility and bounds, and added assertions for the new divider.

yurishkuro (changes requested):
Thanks for tracking down the header/body divider inconsistency.

**Two independent `VerticalResizer` instances now control the same `sidePanelWidth`, and they visibly disagree.**

Before this PR, the span-details boundary had exactly one draggable implementation: the `VerticalResizer` in `index.tsx:263-269`, which already worked correctly from inside the table. This PR doesn't extend that — it adds a second, separate instance in `TimelineHeaderRow.tsx:106-114`, with its own `DraggableManager`, its own `getBoundingClientRect()`-based bounds, and (via `rightSide`) its own `scaleX(-1)` CSS mirror. Both ultimately call the same `setSidePanelWidth` (`index.tsx:108-114`, wired to the header via `onSidePanelWidthChange={setSidePanelWidth}` at line 217, and to the body via the inline-inverted `onChange` at line 267), so the *value* stays consistent. But the two renderings of the divider don't: at `sidePanelWidth = 0.56`, the header instance's grip icon renders at x=367 while the actual column boundary (the CSS border on `.TimelineHeaderRow--sidePanelCell`) sits at x=906 — a ~540px gap, reproducible any time the width deviates from 0.5. The `.TimelineHeaderRow .VerticalResizer.is-flipped .VerticalResizer--dragger { height: ... }` rule in `TimelineHeaderRow.css:44-46` only stops the two instances from fighting over pointer events in the overlap region; it doesn't address why there are two in the first place.

The name-column divider right next to this one is the existing counterexample: a single `VerticalResizer`, uncapped, living only in the header, doing double duty for both rows because `TimelineHeaderRow` is `position: fixed` and the dragger's `calc(100vh - nav-offset)` height naturally reaches down through the table. That pattern was available here too — remove the body-side instance (`index.tsx:263-269`) and give the header's instance the same uncapped, full-height treatment, mirroring the name-column implementation exactly. That leaves one input surface instead of two, and removes the need for the CSS height-cap coordination between them along with the mirroring-offset bug it currently has.

Functionally, the fix does what it claims — verified in a browser that dragging now works from the header, `cursor: col-resize` is exposed correctly across the header row (it previously wasn't; that region resolved to `TimelineViewingLayer` with `cursor: vertical-text`), and the touched test suites pass (54/54). But I'd ask for the consolidation above rather than landing two coordinated-by-CSS resizer instances for one boundary — worth doing now before `TimelineHeaderRow.test.jsx` locks in assertions against the two-instance shape.

---

KKamJi98:
[Replying to this review](https://github.com/jaegertracing/jaeger-ui/pull/4412#pullrequestreview-5157511127)

Addressed in b60bd71f. Removed the body resizer and made the header divider the single full-height instance, following the name-column pattern.

Browser testing also caught two related issues: the header cell intercepting drags at the exact boundary, and hover-triggered rerenders canceling an active drag. Both are fixed, with regression tests for drag continuity, asymmetric widths, bounds, and visibility.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/index.test.jsx
```diff
@@ -69,16 +69,17 @@ vi.mock('./spanPills', () => ({
 vi.mock('./VirtualizedTraceView', () => mockDefault(() => <div data-testid="virtualized-trace-view-mock" />));
 vi.mock('./SpanDetailSidePanel', () => mockDefault(() => <div data-testid="span-detail-side-panel-mock" />));
 vi.mock('../../common/VerticalResizer', () => ({
```
yurishkuro: `index.tsx` no longer imports `VerticalResizer`, so this mock is dead. It can be removed.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/index.test.jsx
```diff
@@ -298,9 +294,9 @@ describe('<TraceTimelineViewer>', () => {
       expect(screen.getByTestId('virtualized-trace-view-mock')).toBeInTheDocument();
     });
 
-    it('renders a VerticalResizer between main and side panel when timeline bars are visible', () => {
+    it('does not render a second body resizer when timeline bars are visible', () => {
```
yurishkuro: These two tests (here and the one at line 302) assert that a component this file no longer renders is absent, which cannot fail regardless of what `TraceTimelineViewer` does. Drop them along with the mock; the header-side behavior is already covered by the `header-side-panel-change` test below and by `TimelineHeaderRow.test.jsx`.

yurishkuro (commented):
Thanks for the consolidation — one header-owned divider that reaches down through the table is exactly the shape the name-column divider already had, and it removes the offset between the grip and the actual column boundary that the two-instance version had.

Verified on this head with side-panel mode at width 0.56: the grip, the header cell edge, the side panel's left edge and the main column's right edge all sit on the same x, and they stay aligned after dragging from the header row and from the body. Tests, tsc, oxlint and fmt-lint pass.

Two small things before this lands, neither about the code path itself:

1. `index.test.jsx` still mocks `VerticalResizer` and keeps two tests asserting the mock is absent, but `index.tsx` no longer imports the component, so those tests can never fail. Inline comments below.
2. The PR description still describes the earlier design (a second resizer in the header plus a CSS height cap on its dragger). Please rewrite it to describe the branch as it now stands: the body resizer is removed and the header's single full-height resizer serves both rows.

---

KKamJi98:
[Replying to this review](https://github.com/jaegertracing/jaeger-ui/pull/4412#pullrequestreview-5204416967)

Both addressed.

1. Dead test code removed in 79cd1864: the `VerticalResizer` mock and the two side-panel tests that only asserted its absence. The `layout combinations` test carried the same vacuous assertion, so I dropped that line too and kept its side-panel and `VirtualizedTraceView` assertions.
2. PR description rewritten to match the branch as it stands: one header-owned full-height resizer, the body instance and the header height cap both gone, plus the ordering, bounds, visibility gating and drag-continuity details.

Verification on 79cd1864: `index.test.jsx` + `TimelineHeaderRow.test.jsx` 40 passed, full `jaeger-ui` suite 271 files / 3706 tests passed, `pnpm run lint` clean (fmt-lint, tsc-lint, oxlint, check-license, check-copyright-year, check-tsx-naming, check-overrides, knip).

yurishkuro (approved):

PR #4416, opened 2026-08-31
fix: avoid unknown GenAI view

### LINKED ISSUE #4415: [chore]: Do not switch to GenAI span view if there are no recognized attributes
### Requirement

The "is GenAI" decision seems to be made very early based on whether there are __any__ gen-ai attributes on the span, but that view is pointless if all those attributes are being categorized into "Other" category, like this:

Image

It just ends up with low signal collapsed accordion whereas displaying the main Detail tab would be more useful.

The GenAI tab has several dedicated sections for various pre-defined attributes. If those attributes are missing the sections are not displayed and only Other GenAI Attributes section is displayed (as collapsed). A better solution would be to:

### Subtasks

- [x] Do not switch to the GenAI tab if only the Other accordion is present — #4417
- [ ] Auto-expand the Other accordion if it is the only section

Soumoditya:
## Summary
- keep traces with only unrecognized GenAI attributes on the standard timeline view
- retain recognized GenAI classifications for the GenAI view
- add regression coverage for unknown GenAI attributes

## Testing
- 
px --yes pnpm@11.23.0 --filter @jaegertracing/jaeger-ui test src/model/OtelTraceFacade.test.ts`n- 
px --yes pnpm@11.23.0 run fmt-lint`n
Fixes #4415

### DIFF
--- a/packages/jaeger-ui/src/model/OtelTraceFacade.test.ts
+++ b/packages/jaeger-ui/src/model/OtelTraceFacade.test.ts
@@ -94,6 +94,24 @@ describe('OtelTraceFacade', () => {
       };
       expect(new OtelTraceFacade(genAiTrace).isGenAITrace).toBe(true);
     });
+
+    it('returns false when spans only have unrecognized GenAI attributes', () => {
+      const unknownGenAiSpan: Span = {
+        ...mockSpan,
+        spanID: 'unknown-genai-span',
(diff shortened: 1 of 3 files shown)


yurishkuro:
wrong fix. I updated the ticket.

yurishkuro:
closing as this specific part of the ticket is already solved

PR #4417, opened 2026-08-31
fix(genai): Keep details for unknown attributes

conradmr94:

- Part of #4415

## Description of the changes

- Keep the standard span details view when all `gen_ai.*` attributes would be categorized as "Other GenAI Attributes".
- Show the GenAI tab only when the span contains at least one recognized GenAI section.
- Leave trace-wide GenAI classification unchanged.
- Add regression coverage for spans containing only unrecognized GenAI attributes.
- Written with AI assistance.

- `pnpm --filter @jaegertracing/jaeger-ui test src/components/TracePage/TraceTimelineViewer/SpanDetail/index.test.jsx`
- `pnpm run fmt`
- `pnpm run lint`
- `pnpm test`
- `pnpm run build`

## Checklist

- [x] I have run the lint and test steps successfully: `pnpm run lint` and `pnpm test`

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/index.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/index.tsx
@@ -11,6 +11,7 @@ import AccordionLinks from './AccordionLinks';
 import AccordionText from './AccordionText';
 import DetailState from './DetailState';
 import GenAITab from './GenAITab';
+import { extractGenAiSections } from './GenAITab/genAiData';
 import { formatDuration, formatDurationCompact } from '../utils';
 import CopyIcon from '../../../common/CopyIcon';
 import LabeledList from '../../../common/LabeledList';
@@ -88,7 +89,8 @@ export default function SpanDetail(props: SpanDetailProps) {
   ];
   const deepLinkCopyText = `${window.location.origin}${window.location.pathname}?uiFind=${span.spanID}`;
 
-  const showGenAITab = isGenAISpan(span);
+  const showGenAITab =
+    isGenAISpan(span) && extractGenAiSections(span.attributes).some(section => section.type !== 'other');
 
   const detailsContent = (
     <div>
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/index.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/index.tsx
@@ -89,11 +89,8 @@ export default function SpanDetail(props: SpanDetailProps) {
   ];
   const deepLinkCopyText = `${window.location.origin}${window.location.pathname}?uiFind=${span.spanID}`;
 
-  const genAISections = React.useMemo(
-    () => (isGenAISpan(span) ? extractGenAiSections(span.attributes) : []),
-    [span]
-  );
-  const showGenAITab = genAISections.some(section => section.type !== 'other');
+  const showGenAITab =
+    isGenAISpan(span) && extractGenAiSections(span.attributes).some(section => section.type !== 'other');
 
   const detailsContent = (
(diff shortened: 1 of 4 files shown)


packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/index.tsx
```diff
@@ -88,7 +89,8 @@ export default function SpanDetail(props: SpanDetailProps) {
   ];
   const deepLinkCopyText = `${window.location.origin}${window.location.pathname}?uiFind=${span.spanID}`;
 
-  const showGenAITab = isGenAISpan(span);
+  const showGenAITab =
+    isGenAISpan(span) && extractGenAiSections(span.attributes).some(section => section.type !== 'other');
```
yurishkuro: isn't this going to do double work for extracting GenAI sections?

yurishkuro: if we memoize the result than this fix might be ok

conradmr94: good point. updating in rev2

conradmr94: updated SpanDetail to memoize the extracted sections and pass the result to GenAITab, so the visibility check an drendering won't perform separate extractions. i also added coverage for the precomputed-sections path.

yurishkuro:
Pushed a revert of 6d9748e directly to this branch — the memoization request was my mistake, sorry for the detour. The fix in c527d9b stands on its own.

The concern I raised was that `SpanDetail` and `GenAITab` would both extract the sections. That is true, but it does not survive its own cost estimate: it is one extra pass over a single span's `gen_ai.*` attributes, and both call sites already `useMemo` on the span, so it happens once per detail panel — not per row and not per interaction. Nothing like the `VirtualizedTraceView` paths where an extra pass actually matters.

The remedy also cost more than the problem. An optional `sections` prop gives `GenAITab` two forms of invocation to keep working for a single caller, and the `renders precomputed sections` test existed only to cover the second form. Memoizing inside `extractGenAiSections()` — what I suggested — would have been worse still: a module-level cache handing every caller the same mutable array, to save work that is already free.

| | Optional `sections` prop | Memoize inside `extractGenAiSections` | Required `sections` prop | Extract in both places |
| --- | --- | --- | --- | --- |
| Single invocation form | 🔴 two | 🟢 | 🟢 | 🟢 |
| No hidden module state | 🟢 | 🔴 shared mutable result | 🟢 | 🟢 |
| Extractions per detail panel | 🟢 one | 🟢 one | 🟢 one | 🟡 two, both memoized <sup>1</sup> |
| Test churn | 🟡 one prop-only test | 🟢 none | 🔴 33 `<GenAITab>` call sites | 🟢 none |
| Diff beyond the bug fix | 🟡 | 🟡 | 🔴 | 🟢 none |

🟢 good · 🟡 partial · 🔴 poor

<sup>1</sup> One pass over one span's attributes per panel render.

`pnpm run fmt`, `pnpm run lint`, `pnpm test` (3616 + 115 tests) and `pnpm run build` all pass on the reverted state.

---

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/index.tsx
yurishkuro: I reverted, the double work is cheap and not dependent on trace size, so let's keep it simple.

yurishkuro (approved):

yurishkuro:
Changed to `part of issue` because it does not fix the other part from the issue, auto-expanding the accordion

conradmr94:
ack. thanks for the follow-up

PR #4420, opened 2026-09-01
feat(timeline): Add contextual controls to span rows

### LINKED ISSUE #4418: [chore]: Contextual controls in span row in trace timeline view
### Requirement

We could add a few useful actions to the span row, shown only on hover

Image

1. collapse collapsible children of the current span
2. focus subtree - makes the current span look like the root of the trace, with "reset" button (same icon but maybe crossed out and colored as active) added to the span permanently
3. the More actions... menu adds "Copy deep link" and "Copy span ID"

shubhtrek:
- Resolves #4418

## Description of the changes

Adds three contextual action buttons to each span row in the trace timeline view, revealed on hover:

1. **Collapse/Expand subtree** (spans with children only) — toggles visibility of all descendant spans recursively.
2. **Focus subtree** — zooms the timeline to show only the focused span and its descendants, rendering it as the root. An active (highlighted) button remains permanently visible so the user can reset with a single click.
3. **More actions (…) dropdown** — two clipboard actions:
   - *Copy deep link*: copies `<currentURL>?uiFind=<spanID>` to clipboard
   - *Copy span ID*: copies the raw hex span ID to clipboard

### Files changed

| File | Role |
|---|---|
| `SpanRowControls/index.tsx` | New component — three hover buttons backed by Zustand |
| `SpanRowControls/index.css` | Hidden by default; revealed on `.span-row:hover`; active focus button always visible |
| `SpanBarRow.tsx` | Mounts `<SpanRowControls>` inside the span name cell |
| `store.timeline.ts` | Adds `focusedSubtreeSpanID`, `setFocusedSubtreeSpanID`, `toggleSubtreeCollapse`; all reset on `setTrace()` |
| `timeline-utils.ts` | Adds `getSubtreeSpans` and `getDescendantParentSpanIDs` pure helpers |
| `VirtualizedTraceView.tsx` | Filters `trace.spans` to the subtree when `focusedSubtreeSpanID` is active |
| `SpanTreeOffset.tsx` | Trims ancestor indent guides above the focused span so it renders cleanly as root |

## Before / After

https://github.com/user-attachments/assets/65523f0e-2fd6-4011-8f5c-15d638559364

https://github.com/user-attachments/assets/16efdfd1-1688-427c-a333-0eb67a4619c4

- 7 new unit tests in `SpanRowControls/index.test.tsx` covering: render (parent vs leaf span), collapse toggle, focus toggle, copy deep link, copy span ID, event propagation stop
- 2 new store tests in `store.test.ts` covering `setFocusedSubtreeSpanID` and `toggleSubtreeCollapse`
- Full suite: **3622 / 3622 tests pass** (`pnpm test`)
- `pnpm run tsc-lint` — no type errors
- `pnpm run fmt` — no formatting changes
- `pnpm run oxlint` — 0 new errors
- Manually verified in dev server on a real trace

## Checklist

### DIFF
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanRowControls/index.css
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanRowControls/index.css
@@ -1,0 +1,77 @@
+/*
+Copyright (c) 2026 The Jaeger Authors.
+SPDX-License-Identifier: Apache-2.0
+*/
+
+.SpanRowControls {
+  display: inline-flex;
+  align-items: center;
+  gap: 2px;
+  margin-left: 0.35rem;
+  margin-right: 0.25rem;
+  opacity: 0;
+  pointer-events: none;
+  transition: opacity 0.15s ease-in-out;
+  position: relative;
+  z-index: 2;
+  vertical-align: middle;
+}
+
+/* Reveal controls on span row hover */
+.span-row:hover .SpanRowControls {
+  opacity: 1;
+  pointer-events: auto;
+}
+
+/* If a span is focused, its controls remain visible so the reset button is always accessible */
+.SpanRowControls.is-focused {
+  opacity: 1;
+  pointer-events: auto;
+}
+
+.SpanRowControls--button {
+  background: var(--surface-component-background);
+  border: 1px solid var(--border-default);
+  border-radius: 4px;
+  color: var(--text-secondary);
+  cursor: pointer;
+  display: inline-flex;
+  align-items: center;
+  justify-content: center;
+  font-size: 0.85rem;
+  height: 20px;
+  width: 20px;
+  padding: 0;
+  line-height: 1;
+  transition: all 0.15s ease;
+}
+
+.SpanRowControls--button:hover {
+  background: var(--surface-component-background-hover);
+  border-color: var(--interactive-primary);
+  color: var(--interactive-primary);
+}
+
+.SpanRowControls--button:focus-visible {
+  outline: 2px solid var(--interactive-primary);
+  outline-offset: 1px;
+}
+
+/* Active focus / reset button style */
+.SpanRowControls--button.is-active {
+  background: var(--interactive-primary);
+  border-color: var(--interactive-primary);
+  color: var(--text-inverse);
+}
+
+.SpanRowControls--button.is-active:hover {
+  background: var(--interactive-primary-hover);
+  border-color: var(--interactive-primary-hover);
+  color: var(--text-inverse);
+}
+
+.SpanRowControls--menuIcon {
+  font-size: 1rem;
+  vertical-align: -0.15em;
+  margin-right: 0.25rem;
+}
--- a/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanRowControls/index.tsx
+++ b/packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanRowControls/index.tsx
@@ -1,0 +1,144 @@
+// Copyright (c) 2026 The Jaeger Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+import React, { useCallback } from 'react';
+import { Dropdown, Tooltip, message } from 'antd';
+import type { MenuProps } from 'antd';
+import { LuCopy, LuEllipsis, LuFocus, LuLink, LuListCollapse } from 'react-icons/lu';
+import copy from 'copy-to-clipboard';
+import cx from 'classnames';
+
+import { IOtelSpan } from '../../../../types/otel';
+import { useTraceTimelineStore } from '../store';
+
+import './index.css';
+
+type SpanRowControlsProps = {
+  span: IOtelSpan;
+};
+
+export const SpanRowControls: React.FC<SpanRowControlsProps> = ({ span }) => {
+  const focusedSubtreeSpanID = useTraceTimelineStore(s => s.focusedSubtreeSpanID);
+  const setFocusedSubtreeSpanID = useTraceTimelineStore(s => s.setFocusedSubtreeSpanID);
+  const childrenHiddenIDs = useTraceTimelineStore(s => s.childrenHiddenIDs);
+  const toggleSubtreeCollapse = useTraceTimelineStore(s => s.toggleSubtreeCollapse);
+
+  const isFocused = focusedSubtreeSpanID === span.spanID;
+  const isSubtreeCollapsed = childrenHiddenIDs.has(span.spanID);
+
+  const handleCollapseToggle = useCallback(
+    (e: React.MouseEvent) => {
+      e.stopPropagation();
+      e.preventDefault();
+      toggleSubtreeCollapse(span);
+    },
+    [toggleSubtreeCollapse, span]
+  );
+
+  const handleFocusToggle = useCallback(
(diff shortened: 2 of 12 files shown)


shubhtrek:
@yurishkuro PTAL give me feedback.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
```diff
@@ -99,12 +101,20 @@ function generateRowStatesFromTrace(
...
 ): RowState[] {
   if (!trace) {
     return [];
   }
-  return generateRowStates(trace.spans, childrenHiddenIDs, detailStates, detailPanelMode, prunedServices);
+  let spans = trace.spans;
+  if (focusedSubtreeSpanID) {
+    const focusedSpan = trace.spanMap.get(focusedSubtreeSpanID);
+    if (focusedSpan) {
+      spans = getSubtreeSpans(focusedSpan);
```
jkowall, line 114: This filtered array is what `generateRowStates` indexes. `renderSpanBarRow` then uses that `spanIndex` against full `trace.spans` (`spanContainsErredSpan`, `findServerChildSpan`, ScrollManager). Preserve the original full-trace index (lookup by `spanID`) or pass the same filtered array through render and the accessors.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
```diff
@@ -99,12 +101,20 @@ function generateRowStatesFromTrace(
...
+    }
+  }
+  return generateRowStates(spans, childrenHiddenIDs, detailStates, detailPanelMode, prunedServices);
```
jkowall, line 117: Service-filter sanitization only protects the original root services. If the focused span's service is in `prunedServices`, this entire subtree is dropped and the reset control disappears. Protect the focused service or clear focus when it is filtered.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanRowControls/index.css
```diff
@@ -0,0 +1,77 @@
...
+  position: relative;
+  z-index: 2;
+  vertical-align: middle;
+}
+
+/* Reveal controls on span row hover */
+.span-row:hover .SpanRowControls {
+  opacity: 1;
+  pointer-events: auto;
+}
```
jkowall, line 24: Hover is the only reveal. Buttons stay in tab order at `opacity: 0`. Add `.span-row:focus-within .SpanRowControls` (and consider `visibility` so they aren't tabbable while hidden). Touch has no hover.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanRowControls/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+  );
+
+  const menuItems: MenuProps['items'] = [
+    {
+      key: 'copy-deep-link',
+      icon: <LuLink className="SpanRowControls--menuIcon" />,
+      label: 'Copy deep link',
+      onClick: ({ domEvent }) => {
+        domEvent.stopPropagation();
+        const deepLink = `${window.location.origin}${window.location.pathname}?uiFind=${span.spanID}`;
```
jkowall, line 58: This overwrites `location.search`. `uiFind` is the right param, but keep the rest of the query string (see `updateUiFind`). Also check `copy()`'s return value before `message.success`.

jkowall (changes requested):
Requesting changes. This matches #4418, but focus is wrong once you leave the happy path. I also want @yurishkuro to weigh in on the spec questions at the bottom.

1. After focus, `RowState.spanIndex` is an index into the *subtree*, not `trace.spans`. `renderSpanBarRow` still does `const { spans } = trace` and feeds that into `spanContainsErredSpan` / `findServerChildSpan`. ScrollManager maps row↔spanIndex then walks `this._trace.spans`, so next/prev and `scrollToFirstVisibleSpan` jump to the wrong span or throw. Keep the original indices (lookup by spanID) or thread the same filtered array through render and the accessors. Copilot flagged this; I agree, it's a blocker.

2. The focused span's service can still be pruned. `sanitizePrunedServices` only protects the original root, so hiding that service deletes the reset control. Protect the focused service or clear focus.

3. Deep link is `` `${origin}${pathname}?uiFind=` ``. `uiFind` is the right convention, but this drops the rest of `location.search`. Use the same merge as `updateUiFind`. `copy()` can fail; `CopyIcon` only toasts when it returns true.

4. Controls are `opacity: 0` until `.span-row:hover` and stay tabbable. No `:focus-within`. Touch has no hover. The existing `SpanTreeOffset` chevron is keyboard-accessible; these should be too.

5. `toggleSubtreeCollapse` writes Zustand only. `childrenToggle` / `collapseAll` dual-write Redux first for the tracking middleware. Do the same here.

6. Tests miss the failure modes: focus + original `spanIndex` / ScrollManager, nested-parent recursive collapse, clipboard failure, preserving other query params. The VTV focus test asserts `childSpans.length + 1` (direct kids), not `getSubtreeSpans`.

No second PR implements #4418. File-level overlap to watch:
- #4407 keyboard/ARIA on `SpanBarRow` + `SpanTreeOffset` (quota-blocked). Don't ship a hover-only control set next to that.
- #4168 (your other a11y PR) touches the same `SpanBarRow` / `VirtualizedTraceView` files.
- #4317 refactors `SpanTreeOffset` ancestors; this PR trims them for focus. They will conflict.
- #4352 removes dead hover-guide state on `SpanTreeOffset` + VTV.

@chinmaychahar claimed the issue; no PR from them.

Spec questions for Yuri:
(a) Recursive collapse of all descendant parents vs one level. I think recursive is correct, otherwise this duplicates the chevron.
(b) "Look like the root" currently only filters rows and trims indent. Time axis / minimap / header stay full-trace, so a short child is still a sliver. Intentional?
(c) On focus, all three buttons stay visible. The issue asked for the reset icon only.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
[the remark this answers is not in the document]
shubhtrek: I've updated this to preserve the original full-trace indexing

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanRowControls/index.css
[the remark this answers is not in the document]
shubhtrek: I've added `.span-row:focus-within .SpanRowControls` and `.SpanRowControls:focus-within` to `SpanRowControls/index.css`. Now, whenever a user tabs into any of the buttons via keyboard navigation, the controls become fully visible and interactive

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/VirtualizedTraceView.tsx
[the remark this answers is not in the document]
shubhtrek: Updated `generateRowStatesFromTrace` to exclude the focused span's service from `effectivePrunedServices` so the subtree root and reset button are never pruned away

PR #4421, opened 2026-09-02
docs(rfc-0002): Correct wire-format claims against captured v2.19.0 output

u7k4rs6:

Probing Jaeger v2.19.0 turned up five discrepancies with RFC 0002: four claims that do not
match what the server returns, plus one behaviour the RFC does not document at all. Each was
found by probing a local instance and diffing the response against the document, not by
reading the code.

Everything else in the RFC's wire-format section checked out and is left alone — see
"What was verified and left unchanged" at the end.

## Description of the changes

Documentation only; no code changes.

Note that §3.5's table (line 569 on main) already says the spec's `format: bytes` is wrong
and that base64 never appears — plural, correct. So the RFC currently contradicts itself
across §3.5 and §3.6.2, and this change resolves it in favour of the section that was
already right.

### 1. Milestone 3.1 — the field is `spanKind`, not `span_kind`

The note said the endpoint "returns `span_kind` as a lowercase string". The value convention
is right; the field name is not. Captured:

```
$ curl -s 'http://localhost:16686/api/v3/operations?service=otlp-probe'
{"operations":[{"name":"all-zeroes","spanKind":"internal"},
               {"name":"kitchen-sink","spanKind":"internal"},
               {"name":"server-error","spanKind":"server"}]}
```

The OpenAPI spec agrees — `jaeger.api_v3.Operation` declares `spanKind` and lists it in
`required`:

```yaml
jaeger.api_v3.Operation:
    required:
        - name
        - spanKind
    type: object
    properties:
        name:      { type: string }
        spanKind:  { type: string }
```

The RFC's own Zod sketch further down (§3.6.2, `OperationsResponseSchema`) already uses
`spanKind`, so the prose was the outlier.

### 2. Milestone 3.1 — `UNSPECIFIED` comes back as `"internal"`

Not previously documented, and it changes what a shared mapper has to be.

The probe sent three spans. `kitchen-sink` carried an explicit `kind: 0`
(`SPAN_KIND_UNSPECIFIED`), `server-error` carried `kind: 2` (`SERVER`), `all-zeroes` carried
`kind: 0`. On read-back:

| span | `kind` sent | `/api/v3/traces/{id}` | `/api/v3/operations` |
|---|---|---|---|
| `kitchen-sink` | `0` | field **absent** | `"internal"` |
| `all-zeroes` | `0` | field **absent** | `"internal"` |
| `server-error` | `2` | `2` | `"server"` |

So `kind: 0` and `kind: 1` (`INTERNAL`) are indistinguishable on `/api/v3/operations`. A
shared `SpanKind` mapper needs a number→string table and a string→number table that are
**not** inverses, and cannot treat this endpoint as authoritative for a span's kind.

### 3. §3.6.2 — the spec emits `format: bytes`, not `format: "byte"`

The section said the spec uses `format: "byte"`, "the OpenAPI convention for base64". The
actual token is plural:

```yaml
opentelemetry.proto.trace.v1.Span:
    properties:
        traceId:
            type: string
            format: bytes      # <- plural; not the registered OpenAPI `byte`
        spanId:
            type: string
            format: bytes
        parentSpanId:
            type: string
            format: bytes
```

`byte` is the registered OpenAPI format meaning base64. `bytes` is the protobuf type name
emitted verbatim by `protoc-gen-openapi` and is not a recognised OpenAPI format at all.

This is worth correcting rather than filing as a typo for two reasons. The proposed fix
("Option A: fix the Swagger spec") is described against a token that is not in the file. And
the plural token is also *why the generated client works today*: `openapi-zod-client` does
not recognise `bytes`, so it degrades to a plain `z.string()` and never attempts base64
decoding. The hex IDs pass validation by accident, not by design.

Corrected in the prose, the comparison table, and the rejected-alternative bullet.

### 4. §3.6.2 status list — the spec's `200` response contradicts its own description

`GET /api/v3/traces/{traceId}` in `query_service.openapi.yaml`:

```yaml
/api/v3/traces/{traceId}:
    get:
        description: |-
            GetTrace returns a single trace.
             Note that the JSON response over HTTP is wrapped into result envelope "{"result": ...}"
             ...
        responses:
            "200":
                content:
                    application/json:
                        schema:
                            $ref: '#/components/schemas/opentelemetry.proto.trace.v1.TracesData'
```

The description says the response is wrapped; the schema `$ref`s `TracesData` **unwrapped**.
The wrapped form is what the server sends — captured, 3,975 bytes:

```json
{"result":{"resourceSpans":[{"resource":{...},"scopeSpans":[...]}]}}
```

So the schema is the wrong half, and a client generated from this spec looks for
`resourceSpans` at the top level and does not find it. Added to the jaeger-idl status
checklist alongside the existing `format: bytes` item. Fixing it upstream would also let the
generated client type the envelope instead of the UI hand-rolling it.

### 5. §3.6.2 status list — no OTLP schema declares `required:`

Also added, because it bears directly on §3.6.1's "generate types and schemas from the spec"
strategy. Of the 34 schemas in the spec, 14 declare a `required:` array, and every one of
them is Jaeger-defined:

```
jaeger.api_v3.Operation                              ['name', 'spanKind']
jaeger.api_v3.TraceSummary                           ['traceId']
jaeger.api_v3.ServiceSummary                         ['name']
jaeger.api_v3.GetServicesResponse                    ['services']
jaeger.api_v3.GetOperationsResponse                  ['operations']
jaeger.api_v3.DependenciesResponse                   ['dependencies']
jaeger.api_v3.Dependency                             ['parent', 'child', 'callCount']
jaeger.api_v3.Pagination                             ['pageSize']
jaeger.expression.v1.AttributeReference              ['key']
jaeger.expression.v1.Call                            ['op', 'args']
jaeger.expression.v1.FieldReference                  ['name', 'level']
jaeger.expression.v1.List                            ['values']
jaeger.expression.v1.NestedReference                 ['level']
jaeger.expression.v1.Scalar                          ['value']
```

Every `opentelemetry.proto.*` schema declares none — all 13 of them:

```
opentelemetry.proto.trace.v1.Span                    -- none (all optional) --
opentelemetry.proto.trace.v1.Span_Event              -- none (all optional) --
opentelemetry.proto.trace.v1.Span_Link               -- none (all optional) --
opentelemetry.proto.trace.v1.Status                  -- none (all optional) --
opentelemetry.proto.trace.v1.TracesData              -- none (all optional) --
opentelemetry.proto.trace.v1.ResourceSpans           -- none (all optional) --
opentelemetry.proto.trace.v1.ScopeSpans              -- none (all optional) --
opentelemetry.proto.common.v1.AnyValue               -- none (all optional) --
opentelemetry.proto.common.v1.ArrayValue             -- none (all optional) --
opentelemetry.proto.common.v1.KeyValue               -- none (all optional) --
opentelemetry.proto.common.v1.KeyValueList           -- none (all optional) --
opentelemetry.proto.common.v1.InstrumentationScope   -- none (all optional) --
opentelemetry.proto.resource.v1.Resource             -- none (all optional) --
```

The remaining 7 — `jaeger.api_v3.FindTracesRequest`, `FindTraceSummariesRequest`,
`FindTraceSummariesResponse`, `TraceQueryParameters`, `jaeger.expression.v1.Expression`,
`google.protobuf.Any` and `google.rpc.Status` — are request/response wrappers and imported
well-known types, and also declare none.

So a generated schema makes `traceId`, `spanId` and `name` optional, despite each field's
description in the same file reading "This field is required." Required-field enforcement on
the OTLP types has to be layered on top of generation until the IDL carries
`field_behavior = REQUIRED` there too — the same annotation that jaeger-idl PR #204 already
applied to `TraceSummary`.

## What was verified and left unchanged

The rest of §3.5's table was checked against the same capture and is correct, so this PR does
not touch it:

| Claim | Verified |
|---|---|
| `traceId`/`spanId`/`parentSpanId` are lower-case hex, never base64 | yes |
| timestamps are quoted decimal ns strings | yes — `"1754200000000000000"` |
| `kind` is a JSON number and is omitted entirely when `0` | yes |
| `status.code` gets the same enum treatment | yes — `status` becomes `{}` when UNSET |
| the response is a single `{"result": {...}}` document | yes — held at 50k spans (47 MB) and 200k spans (190 MB); the gateway merges the gRPC stream, though transfer is chunked |

The §3.6 warning not to expect the protojson enum *name* (`"SPAN_KIND_SERVER"`) is also
correct — no capture produced that form.

Documentation only.

Claims were verified against `cr.jaegertracing.io/jaegertracing/jaeger:2.19.0`
(`sha256:ede4864215be4cd85bd8c3129a2fea6c5713c5653c7282c429dba123014bc68b`), all-in-one with
default memory storage. A single trace was injected over OTLP/HTTP JSON exercising every
`AnyValue` variant, a span with kind `UNSPECIFIED`, a status of `UNSET`, an int above 2^53,
events, links, dropped counts, `traceState`, `flags` and `schemaUrl` at both levels, then read
back over `/api/v3/traces/{id}`, `/api/v3/operations` and the legacy `/api/traces/{id}`.

OpenAPI claims were checked against `swagger/api_v3/query_service.openapi.yaml` at jaeger-idl
`main`, fetched 2026-09-02 (`dfaaf26f776db565f5729e247f03a823cb4eaf35`).

No formatting change to this file. No code changed, so no test or build run is implicated.

https://claude.ai/code/session_01KCppwxw1hu8N56oa7XXjKw


jkowall (dismissed):
RFC-only. Internal consistency checks out: §3.5 already used `format: bytes` and `OperationsResponseSchema` already used `spanKind`; this PR aligns the outlier prose and the comparison table.

Checked jaeger-idl `query_service.openapi.yaml` at `dfaaf26` (the SHA cited here): `Operation.spanKind` is required with `name`; ID fields use `format: bytes` (plural; no singular `byte`); GetTrace 200 `$ref`s unwrapped `TracesData` while the description documents the `{"result": ...}` envelope; none of the 13 `opentelemetry.proto.*` schemas declare `required:`.

I cannot re-run a v2.19.0 capture from this repo, so the UNSPECIFIED -> `"internal"` operations mapping is taken from the author's probe notes. Yuri should still sanity-check those wire-format facts; this is his RFC.

docs/rfc/0002-otel-native-jaeger-ui.md
```diff
@@ -479,7 +479,14 @@ Introduce a top-level configuration flag `useOpenTelemetryTerms` (defaulting to
 #### Milestone 3.1: Metadata Exploration (Search Page)
 **Goal**: Implement the simplest OTLP endpoints and integrate them into the Search page.
 
-> **Note**: The `/api/v3/operations` endpoint returns `span_kind` as a **lowercase** string (e.g. `"server"`, `"client"`). This is a Jaeger-specific serialization, not the OTLP-JSON proto enum name. `useSpanNames` already normalizes with `.toLowerCase()` when filtering.
+> **Note**: The `/api/v3/operations` endpoint returns the field as `spanKind` (proto3 camelCase), holding a **lowercase** string value (e.g. `"server"`, `"client"`). This is a Jaeger-specific serialization, not the OTLP-JSON proto enum name. `useSpanNames` already normalizes with `.toLowerCase()` when filtering.
+>
+> **The kind mapping here is lossy in one direction.** A span sent with `kind: 0`
```
yurishkuro: instead of documenting it's easier to fix.

docs/rfc/0002-otel-native-jaeger-ui.md
```diff
@@ -571,7 +578,7 @@ The new parser replaces the role of `transformTraceData` for the OTLP route. Cov
...
 | `status.code` | JSON number | same enum treatment |
 
-Do not expect the protojson enum *name* (`"SPAN_KIND_SERVER"`) on this endpoint, as the sketch's comment above assumed. That form would come from gogo protojson, which is not what serialises this response. Note the contrast with `/api/v3/operations` (see Milestone 3.1), which is a Jaeger-defined message rather than OTLP and does return `span_kind` as a lower-case string — the two endpoints genuinely differ, so a shared `SpanKind` mapper needs to handle both.
+Do not expect the protojson enum *name* (`"SPAN_KIND_SERVER"`) on this endpoint, as the sketch's comment above assumed. That form would come from gogo protojson, which is not what serialises this response. Note the contrast with `/api/v3/operations` (see Milestone 3.1), which is a Jaeger-defined message rather than OTLP and returns `spanKind` as a lower-case string — the two endpoints genuinely differ, so a shared `SpanKind` mapper needs to handle both, and must account for `/api/v3/operations` collapsing `UNSPECIFIED` onto `"internal"`.
```
yurishkuro: this paragraph is Claude's annoying habit of leaving changelog-like changes in the docs instead of writing what is "now". Please remove and fix invalid claim above if indeed invalid (which it seems to be, OTLP-JSON is supposed to return enums as numbers).

docs/rfc/0002-otel-native-jaeger-ui.md
```diff
@@ -642,11 +649,13 @@ export const schemas = {
 
 ##### Handling OTEL-JSON vs. Swagger Spec Mismatch
 
-**The Mismatch**: Swagger spec currently defines IDs as `type: "string", format: "byte"` (OpenAPI convention for base64), but OTEL-JSON wire format uses hex encoding per OTLP specification.
+**The Mismatch**: Swagger spec currently defines IDs as `type: "string", format: "bytes"`, but OTEL-JSON wire format uses hex encoding per OTLP specification.
+
+Note the token is `bytes`, plural — **not** the registered OpenAPI `byte` format that means base64. `protoc-gen-openapi` emits the protobuf type name verbatim, so `format: bytes` is not a recognised OpenAPI format at all. That distinction matters for the fix below: correcting the spec means replacing a non-standard token, not swapping one standard format for another. It also explains why the generated client happens to work today — `openapi-zod-client` does not recognise `bytes`, so it degrades to a plain `z.string()` and never attempts base64 decoding.
```
yurishkuro: is bytes even legal value in OpenAPI spec? Is "hex" legal"? Either way sounds like we need to fix OpenAPI doc generation to be accurate.

docs/rfc/0002-otel-native-jaeger-ui.md
```diff
@@ -656,6 +665,8 @@ Correct the Swagger spec post-generation in the jaeger-idl repository to accurat
 - [x] **OpenAPI v3 spec available** at `swagger/api_v3/query_service.openapi.yaml` (generated by gnostic/`protoc-gen-openapi`).
 - [x] `trace_id`/`span_id` fields in `TraceSummary` and `ServiceSummary.name` annotated with `field_behavior = REQUIRED` (jaeger-idl PR #204), causing `required:` arrays to be emitted in the OpenAPI YAML.
 - [ ] `format: bytes` still used for raw OTLP span/trace IDs in the spec — these are not yet used by the UI (pending Milestone 3.2).
+- [ ] The `GET /api/v3/traces/{traceId}` `200` response contradicts its own description. The response schema `$ref`s `opentelemetry.proto.trace.v1.TracesData` **unwrapped**, while the same operation's `description` states *"the JSON response over HTTP is wrapped into result envelope `{"result": ...}`"*. The wrapped form is what the server actually sends (see Milestone 3.2), so the schema is the wrong half. A client generated from this spec expects `resourceSpans` at the top level and will not find it. Fixing this in jaeger-idl would also let the generated client type the envelope instead of the UI hand-rolling it.
```
yurishkuro: > The response schema `$ref`s `opentelemetry.proto.trace.v1.TracesData` **unwrapped**

this is not a valid sentence

yurishkuro: also, where does the description originate and can we fix it?

docs/rfc/0002-otel-native-jaeger-ui.md
u7k4rs6: Agreed, dropping this paragraph.

The default is explicit rather than incidental: http_gateway.go:243-251 rewrites
an empty SpanKind to "internal", pinned by TestHTTPGatewayGetOperationsDefaultSpanKind.
Emitting "" instead is not clean, since span_kind is annotated REQUIRED in the
IDL and jsonpb would omit it. "unspecified" is already produced by
jptrace.ProtoSpanKindToString, so the change would be a line there plus the test.

Happy to open that in jaegertracing/jaeger if you want it. I did not find an
existing issue.

docs/rfc/0002-otel-native-jaeger-ui.md
u7k4rs6: Fair, removing the framing and stating the current behaviour directly.

On the claim itself: /api/v3/operations is not OTLP-JSON. It returns
jaeger.api_v3.Operation, a Jaeger-defined message whose span_kind is declared as
a string in the IDL, and the value is produced by SpanKindToString, which
lowercases. So enums-as-numbers holds for the OTLP endpoints and this one is a
string by design, not by accident. Captured against 2.19.0:

    {"operations":[{"name":"server-error","spanKind":"server"}]}

Let me know if you would rather the note dropped the contrast entirely.

docs/rfc/0002-otel-native-jaeger-ui.md
u7k4rs6: Legal but inert. format is an open field in OpenAPI: any value keeps the document
valid, and tools that do not recognise a format fall back to the base type as
though none were given. "hex" would be equally legal and equally ignored. Only
"byte" is the registered format meaning base64.

I also had the mechanism wrong and am correcting it in this push:
protoc-gen-openapi does not pass the proto type name through here. NewBytesSchema()
hardcodes "bytes" with no argument. The integer and number cases next to it do
pass kind.String() verbatim, which is what misled me.

Agreed that fixing generation is the right target. jaeger-idl's only
post-processing step is prune-openapi, which never touches format, so a fix
belongs in gnostic or in a new override.

docs/rfc/0002-otel-native-jaeger-ui.md
u7k4rs6: Rewriting the sentence.

The description comes from the leading comment on the GetTrace RPC in
jaeger-idl's query_service.proto and lands verbatim in the generated YAML. The
description is the accurate half; the schema is the wrong one. protoc-gen-openapi
builds the 200 response from the output message alone and never receives the
method descriptor, so it cannot see that the RPC is server-streaming and has no
streaming concept anywhere in the generator.

The envelope is hand-written in jaeger's returnTrace using GRPCGatewayWrapper,
which is already a message in the IDL. So the fix may be as small as having the
200 response reference that wrapper rather than TracesData.

docs/rfc/0002-otel-native-jaeger-ui.md
```diff
@@ -479,7 +479,7 @@ Introduce a top-level configuration flag `useOpenTelemetryTerms` (defaulting to
 #### Milestone 3.1: Metadata Exploration (Search Page)
 **Goal**: Implement the simplest OTLP endpoints and integrate them into the Search page.
 
-> **Note**: The `/api/v3/operations` endpoint returns `span_kind` as a **lowercase** string (e.g. `"server"`, `"client"`). This is a Jaeger-specific serialization, not the OTLP-JSON proto enum name. `useSpanNames` already normalizes with `.toLowerCase()` when filtering.
+> **Note**: The `/api/v3/operations` endpoint returns the field as `spanKind` (proto3 camelCase), holding a **lowercase** string value (e.g. `"server"`, `"client"`). This is a Jaeger-specific serialization, not the OTLP-JSON proto enum name. `useSpanNames` already normalizes with `.toLowerCase()` when filtering.
```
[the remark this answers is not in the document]
u7k4rs6: i've fixed it

yurishkuro (approved):

Issue #4278, opened 2026-07-29
[🧑‍🎓 LFX] OpenTelemetry-Native Query and State Layers Migration


yurishkuro:
**Term:** LFX Mentorship 2026 Term 3 (Sep - Nov)
**Mentors:** @yurishkuro, @jkowall
**CNCF Proposal:** https://github.com/cncf/mentoring/issues/2029
**LFX Link:** https://mentorship.lfx.linuxfoundation.org/project/36d66a45-bf3f-4bd2-8032-438436557e11

Please read the **[application guidelines](https://www.jaegertracing.io/mentorship/applying/)** for any application-related questions, and see [#9137](https://github.com/jaegertracing/jaeger/issues/9137) for this term's timeline and the other Jaeger projects on offer.

## Problem Statement

Jaeger UI is two-thirds of the way through three interlocking migrations, at three different stages.

- **OpenTelemetry-native data path** — stopped at a deliberate decision point. The UI now speaks OTEL vocabulary throughout — attributes, resources, events, links — but it obtains that vocabulary by wrapping the legacy Jaeger model in a facade. Service discovery and trace search already run on `/api/v3/`; loading an individual trace does not. Every trace the UI renders still arrives over the legacy `/api/traces/:id` endpoint and is reshaped in the browser. This step was left open because it is the one that requires committing to how the whole OTLP schema enters the UI's query layer — see The Vision below.
- **State management** — progressing steadily, close to done. Phase by phase, and largely through community contributions, Redux has given up traces, search results, services, config, dependencies, and the dependency graph. Two reducers remain, one of which already has a migration open in review. The one part that is not simply more of the same is the trace timeline: it writes to *both* Redux and Zustand on every interaction, because analytics tracking is implemented as a Redux middleware keyed on the old action types, so the tracking has to move before the old slice can go.
- **Unified state architecture** — designed, not yet built. There is a documented target for where each kind of state belongs, and a documented design for resolving layout settings between a shared URL, per-trace heuristics, and stored preferences. Parts of the second one are open in review; neither is finished, so a shared trace link still cannot reproduce what the sender saw.

Several contributors have work in flight against these goals. Landing that work — refereeing competing approaches, defining the seams between them, and shepherding contributors through review — is as much a part of this project as writing new code.

## The Vision

Finish the migrations, and delete what they replaced.

[...]

PR #184, opened 2018-02-06
Misc cleanup - flow, precommit hook, no devtools in prod

tiffon:
Fix #183.

PR #4455, opened 2026-09-07
feat(api/v3): validate v3 trace wire contract at boundary

### LINKED ISSUE #3265: [Feature]: Upgrade to OpenTelemetry terminology

sksingh2005:
- Resolves #3265 Phase 3.5–3.7 and part of #4278 — the validated wire contract for `GET /api/v3/traces/{trace_id}`. `generated-client.ts` already contains full OTLP schemas but every field is `.partial().passthrough()`, so no real validation occurs at the boundary. Consumers like the `getTrace` hook for #4129 (#3835 / Milestone 3.2) need validated wire data before parsing. Builds on #4438 (merged: 13 OTLP schemas with reachable names).

## Description of the changes
- `packages/jaeger-ui/src/api/v3/schemas.ts` layers the refinement boundary over the generated schemas: `traceIdHex`/`spanIdHex` (32/16 hex, rejects `format: bytes` base64, rejects all-zero IDs), `decimalNanoString` (unsigned `/^\d+$/` + uint64 `BigInt` bound for `startTimeUnixNano`/`endTimeUnixNano`) alongside `decimalInt64String` (signed, for `AnyValue.intValue`), `spanKindEnum` / `statusCodeEnum` as open-ended non-negative ints (OTLP forward compatibility — unknown future values pass through instead of breaking), `refinedAnyValue` `z.lazy` + `superRefine` exactly-one-of via `!== undefined` (preserves `""`/`false`/`"0"`/`9223372036854775807`, recursive `arrayValue`/`kvlistValue`), required `spans` / `scopeSpans` arrays (matching the generated base), and `GetTraceResponseSchema = {result: TracesData}` envelope separable for future streaming.
- `packages/jaeger-ui/src/api/v3/client.ts` adds `fetchTrace (traceId)`: validates the request ID with `traceIdHex`, `GET /api/v3/traces/{id}`, `GetTraceResponseSchema.parse` at the boundary, returns `validated.result`.
- `knip.config.ts` adds `src/api/v3/schemas.ts` as entry so refinements aren't flagged before the parser lands.

## Where the manual refinement boundary sits (and why it shrinks, not grows)

**The IDL owns structure; the refinement layer owns wire-encoding behavior that generation provably drops or misstates. Nothing else goes in by hand.**

### Stays fully automatic

Field names, nesting, which objects exist `generated-client.ts`, consumed as-is for services/operations and re-exported by stable alias so codegen renames don't leak into consumers.

### Earns a place in `schemas.ts` only if all three hold

1. **Codegen cannot express it.** The generated schemas are `.partial().passthrough()` throughout they accept `{}` and multi-field `AnyValue` objects without complaint, so "just use the generated schema" currently validates almost nothing on this route. The manual layer covers exactly:
   - Hex IDs instead of base64 (`format: bytes` in the spec vs hex on the wire), plus the OTLP all-zero-ID reservation no OpenAPI keyword expresses.
   - Quoted 64-bit integers with `BigInt` bounds (timestamps uint64, `intValue` signed int64) `Number` would silently truncate, and the spec's `type: string` carries no range.
   - Proto3 zero-omission: `kind` / `status.code` absent at `0`, mapped deliberately instead of failed.
   - The always-present `status: {}` wrapper for UNSET a wire observation, not a schema derivation.
   - The recursive `AnyValue` one-of with falsy preservation (`!== undefined` so `''` / `false` / `0` survive).
   - The `{result}` envelope as a separable schema, so future streaming can replace it without touching enrichment.
2. **Pinned to a captured payload.** Each refinement traces to a case in `schemas.trace-contract.test.ts` valid captures and malicious inputs. No fixture, no rule.
3. **Additive and deletable.** Manual means aliases + `.regex` / `.refine` / `.superRefine` layered over the generated schemas, confined to the boundary module. No parser-private wire types, no output-schema modification — regenerating the client stays a clean overwrite.

### Exit test per rule

When the IDL is corrected upstream (the `format: bytes` token, the unwrapped envelope `$ref`, the missing `required:` arrays  scoped as jaeger-idl follow-up), the corresponding refinement gets **deleted** rather than maintained in parallel. What survives any spec fix is only what OpenAPI/codegen can never express: `BigInt` bounds, omission semantics, envelope separability. The manual layer is a shrinking bridge, not a second contract.

- `pnpm test -w packages/jaeger-ui -- src/api/v3/schemas.trace-contract.test.ts src/api/v3/client.trace-contract.test.ts` -> **28/28**  valid `v3-trace-local-2.13.0.json` envelope, falsy `AnyValue` (`""`/`false`/`"0"`/max int64), nested `kvlist`, omitted `kind`/`status:{}`, `BigInt` duration, envelope missing, base64/non-hex ID, non-decimal timestamp, enum range, link hex, `AnyValue` one-of/recursion, `fetchTrace` envelope + hex request validation.
- `pnpm test -w packages/jaeger-ui -- src/api/v3` -> **105/105 pass**.
- Full task-completion suite per `AGENTS.md`: `pnpm run fmt && pnpm run lint && pnpm test && pnpm run build`

## Checklist


sksingh2005:
@parshipcy can you please enable copilot review comments.

yurishkuro:
please explain in the PR description how you draw the line between automatic schema validation and these manually introduced rules.

packages/jaeger-ui/src/api/v3/client.trace-contract.test.ts
```diff
@@ -0,0 +1,100 @@
...
+    mockFetch.mockResolvedValue({ ok: true, json: async () => validEnvelope });
+    const promise = client.getTrace('0123456789abcdef0123456789abcdef');
+    const data = await promise;
+    expect(data.resourceSpans).toHaveLength(1);
+  });
+
+  it('rejects when request traceId is not 32-char hex', async () => {
+    mockFetch.mockResolvedValue({ ok: true, json: async () => validEnvelope });
+    const promise = client.fetchTrace('NOT_HEX' as any);
+    await expect(promise).rejects.toThrow(z.ZodError);
```
yurishkuro: this expectation is not definitive, it could pass with any error, not necessarily the one you are expecting. If the error messages are something that we internally generate then we can assert on a stable substring of that error message, as we do in most backend tests.

packages/jaeger-ui/src/api/v3/client.ts
```diff
@@ -124,6 +131,28 @@ export class JaegerClient {
...
+    if (!response.ok) {
+      throw new Error(`Failed to fetch trace ${traceId}: ${response.status} ${response.statusText}`);
+    }
+    const data = await response.json();
+    const validated = GetTraceResponseSchema.parse(data);
+    return validated.result;
+  }
+
+  /** Alias for fetchTrace to match the RFC naming. */
+  async getTrace(traceId: string): Promise<TracesDataWire> {
```
yurishkuro: this seems like unnecessary duplication in the API surface. Just pick one name - there is but one consumer of this anyway.

yurishkuro: feel free to update the name in the RFC

packages/jaeger-ui/src/api/v3/schemas.ts
```diff
@@ -19,6 +19,13 @@
  * without reaching into the generated file's internals. This survives codegen
  * renames (e.g. AnyValue → opentelemetry_proto_common_v1_AnyValue) because the
  * bundle key is the stable one.
+ *
+ * For the full trace route (/api/v3/traces/{trace_id}), the generated OTLP
+ * schemas are too permissive (every field .partial().passthrough()). This module
```
yurishkuro, line 24: Do we need to fix that in the IDL? E.g. fields like traceID/spanID are mandatory so we can annotate them as such so that OpenAPI spec lists them as required.

it may not be necessary if we manually validate those same fields below anyway.

packages/jaeger-ui/src/api/v3/schemas.ts
```diff
@@ -62,9 +69,207 @@ export const ArrayValueSchema = opentelemetry_proto_common_v1_ArrayValue;
...
+
+// Protobuf bounds: timestamps are uint64, AnyValue int64. Syntax alone would let
+// out-of-range values cross the boundary, so refine with BigInt after the regex.
+const UINT64_MAX = 18446744073709551615n;
+const INT64_MIN = -9223372036854775808n;
+const INT64_MAX = 9223372036854775807n;
+
+const inUint64 = (s: string): boolean => {
+  try {
+    return BigInt(s) <= UINT64_MAX;
```
yurishkuro: but a negative number is ok?

packages/jaeger-ui/src/api/v3/schemas.ts
```diff
@@ -62,9 +69,207 @@ export const ArrayValueSchema = opentelemetry_proto_common_v1_ArrayValue;
...
+  } catch {
+    return false;
+  }
+};
+
+const inInt64 = (s: string): boolean => {
+  try {
+    const v = BigInt(s);
+    return v >= INT64_MIN && v <= INT64_MAX;
+  } catch {
```
yurishkuro: relying on exceptions hurts performance, and also BigInt(" ") is 0n, not what you want. It's faster to check with regex first before parsing and then not use try/catch
```
const int64Regex = /^-?\d{1,20}$/;

const inInt64 = (s: string): boolean => {
  // 1. Fail fast without throwing an exception
  if (!int64Regex.test(s)) {
    return false;
  }

  // 2. Safe to parse, guaranteed not to throw a SyntaxError
  const v = BigInt(s);
  return v >= INT64_MIN && v <= INT64_MAX;
};
```

packages/jaeger-ui/src/api/v3/schemas.ts
```diff
@@ -62,9 +69,207 @@ export const ArrayValueSchema = opentelemetry_proto_common_v1_ArrayValue;
...
+    return false;
+  }
+};
+
+export const decimalNanoString = z
+  .string()
+  .regex(/^\d+$/, 'Expected decimal uint64 string (quoted nanoseconds)')
```
yurishkuro, line 111: limit length. But do we even need this step if you have refine() below that would fail non-numbers just the same?

packages/jaeger-ui/src/api/v3/schemas.ts
```diff
@@ -62,9 +69,207 @@ export const ArrayValueSchema = opentelemetry_proto_common_v1_ArrayValue;
...
+] as const;
+
+type AnyValueWire = {
+  stringValue?: string;
+  boolValue?: boolean;
+  intValue?: string;
+  doubleValue?: number;
+  arrayValue?: { values?: AnyValueWire[] };
+  kvlistValue?: { values?: { key: string; value: AnyValueWire }[] };
+  bytesValue?: string;
```
yurishkuro, line 146: are bytes safe as a string? What if the byte sequence is not a valid utf8 string?

packages/jaeger-ui/src/api/v3/schemas.ts
```diff
@@ -62,9 +69,207 @@ export const ArrayValueSchema = opentelemetry_proto_common_v1_ArrayValue;
...
+  .string()
+  .regex(/^\d+$/, 'Expected decimal uint64 string (quoted nanoseconds)')
+  .refine(inUint64, 'Expected uint64 nanosecond value');
+
+export const decimalInt64String = z
+  .string()
+  .regex(/^-?\d+$/, 'Expected decimal int64 string')
+  .refine(inInt64, 'Expected int64 range value');
+
+export const spanKindEnum = z.number().int().min(0).max(5);
```
yurishkuro: this violates OTLP-JSON forward compatibility design, which is the main reason for sending enums as numbers
```
// Safer: accepts future OTLP additions without breaking the pipeline
export const spanKindEnum = z.number().int().min(0);
```

packages/jaeger-ui/src/api/v3/schemas.ts
```diff
@@ -62,9 +69,207 @@ export const ArrayValueSchema = opentelemetry_proto_common_v1_ArrayValue;
...
+      // oxlint-disable-next-line no-use-before-define
+      arrayValue: refinedArrayValue.optional(),
+      // oxlint-disable-next-line no-use-before-define
+      kvlistValue: refinedKeyValueList.optional(),
+      // OTLP/JSON encodes bytes as base64.
+      bytesValue: z.string().base64().optional(),
+    })
+    .passthrough()
+    .superRefine((val: Record<string, unknown>, ctx) => {
+      const present = ANY_VALUE_KEYS.filter(k => val[k] !== undefined);
```
yurishkuro: isn't there something like `ANY_VALUE_KEYS.count(k => val[k] !== undefined);`? Using filter when you only need count forces unnecessary heap alloc

packages/jaeger-ui/src/api/v3/schemas.ts
```diff
@@ -62,9 +69,207 @@ export const ArrayValueSchema = opentelemetry_proto_common_v1_ArrayValue;
...
+        ctx.addIssue({
+          code: z.ZodIssueCode.custom,
+          message: `AnyValue must have exactly one value field, got ${present.join(', ')}`,
+        });
+      }
+    })
+);
+
+export const refinedArrayValue: z.ZodType<{ values?: AnyValueWire[] }> = z.lazy(() =>
+  z.object({ values: z.array(refinedAnyValue).optional() }).passthrough()
```
yurishkuro, line 190: not sure I follow. You already have `arrayValue` as optional:
```
arrayValue: refinedArrayValue.optional(),
```
If it is actually present what does `z.array(refinedAnyValue).optional()` inside it mean?

packages/jaeger-ui/src/api/v3/schemas.ts
```diff
@@ -62,9 +69,207 @@ export const ArrayValueSchema = opentelemetry_proto_common_v1_ArrayValue;
...
+    version: z.string().optional(),
+    attributes: z.array(refinedKeyValue).optional(),
+    droppedAttributesCount: z.number().int().min(0).optional(),
+  })
+  .passthrough();
+
+export const refinedSpanEvent = z
+  .object({
+    timeUnixNano: decimalNanoString.optional(),
+    name: z.string().optional(),
```
yurishkuro, line 220: are these two actually optional?

packages/jaeger-ui/src/api/v3/schemas.ts
```diff
@@ -62,9 +69,207 @@ export const ArrayValueSchema = opentelemetry_proto_common_v1_ArrayValue;
...
+export const refinedSpan = z
+  .object({
+    traceId: traceIdHex,
+    spanId: spanIdHex,
+    traceState: z.string().optional(),
+    parentSpanId: spanIdHex.optional(),
+    flags: z.number().int().min(0).optional(),
+    name: z.string().optional(),
+    kind: spanKindEnum.optional(),
+    startTimeUnixNano: decimalNanoString.optional(),
```
yurishkuro, line 250: does the spec say these 3 are optional?

packages/jaeger-ui/src/api/v3/schemas.ts
```diff
@@ -62,9 +69,207 @@ export const ArrayValueSchema = opentelemetry_proto_common_v1_ArrayValue;
...
+    links: z.array(refinedSpanLink).optional(),
+    droppedLinksCount: z.number().int().min(0).optional(),
+    status: refinedStatus.optional(),
+  })
+  .passthrough();
+
+export const refinedScopeSpans = z
+  .object({
+    scope: refinedScope.optional(),
+    spans: z.array(refinedSpan).optional(),
```
yurishkuro: I don't think spans are optional, the rest of the envelope becomes meaningless if they are missing

packages/jaeger-ui/src/api/v3/schemas.ts
```diff
@@ -62,9 +69,207 @@ export const ArrayValueSchema = opentelemetry_proto_common_v1_ArrayValue;
...
+    schemaUrl: z.string().optional(),
+  })
+  .passthrough();
+
+export const refinedResourceSpans = z
+  .object({
+    resource: refinedResource.optional(),
+    scopeSpans: z.array(refinedScopeSpans).optional(),
```
yurishkuro: same as above

packages/jaeger-ui/src/api/v3/schemas.ts
```diff
@@ -92,8 +297,8 @@ const normalizeTraceId = z.preprocess(
   jaeger_api_v3_TraceSummary.extend({
     traceId: traceIdHex,
     // Restrict to decimal digits when present — BigInt() throws SyntaxError on non-decimal strings.
-    minStartTimeUnixNano: z.string().regex(/^\d+$/, 'Expected decimal int64 string').optional(),
-    maxEndTimeUnixNano: z.string().regex(/^\d+$/, 'Expected decimal int64 string').optional(),
+    minStartTimeUnixNano: z.string().regex(/^\d+$/, 'Expected decimal uint64 string').optional(),
+    maxEndTimeUnixNano: z.string().regex(/^\d+$/, 'Expected decimal uint64 string').optional(),
```
yurishkuro: you introduced nanos type above, maybe we can use it here?

yurishkuro (commented):
OTLP-JSON spec says:

> Mandatory unknown field ignorance. OTLP-JSON receivers must silently ignore unrecognized fields and unmarshal the rest of the payload. This enforces binary-protobuf-like backward compatibility when new fields are introduced to the telemetry schema.

Is that satisfied by the validation?

packages/jaeger-ui/src/api/v3/schemas.ts
sksingh2005: I thought the `/^\d+$/` regex pre-filtered negatives, but the function now enforces >= 0n standalone so it holds regardless of caller.

packages/jaeger-ui/src/api/v3/schemas.ts
sksingh2005: Kept a length-limited regex: it keeps syntax-vs-range errors distinct (which the new substring assertions rely on) and fails fast before BigInt consistent with the regex-first pattern you suggested for inInt64. Happy to fold it into the refine if you'd rather have one step.

packages/jaeger-ui/src/api/v3/schemas.ts
sksingh2005: I think yes  the wire carries base64 text and validation never decodes it; `z.string().base64()` checks alphabet only, so non-UTF8 payloads pass through as opaque strings by design (pinned: 'AQID' accepted, '%%%' rejected).

packages/jaeger-ui/src/api/v3/schemas.ts
sksingh2005: The outer optional is field-present, the inner is the backend dropping values for empty arrays (arrayValue: {} is real wire behavior).  I now added a comment for that.

packages/jaeger-ui/src/api/v3/schemas.ts
sksingh2005: Yes kept it optional. proto3-JSON omits default-valued fields and the IDL declares no OTLP required:, so validation must accept what the backend sends. Happy to tighten per-field with a payload showing otherwise.

packages/jaeger-ui/src/api/v3/schemas.ts
sksingh2005: Done — `spans `and `scopeSpans `are required arrays now (empty allowed). This also matches the generated base, which already declares both required. resourceSpans stays optional for the top-level empty edge.

packages/jaeger-ui/src/api/v3/schemas.ts
sksingh2005: Done now both fields now use `decimalNanoString`, which also adds the uint64 range check.

packages/jaeger-ui/src/api/v3/client.ts
sksingh2005: removed `fetchTrace `and kept `getTrace`. It matches the RFC (so no RFC change needed), matches the in-flight #4129 usage, and the alias test was its only caller.

packages/jaeger-ui/src/api/v3/schemas.ts
sksingh2005: Done both enums are now open-ended at min(0). The kind: 99 / code: 99 test is rewritten as forward-compat acceptance, and negatives are still rejected.

packages/jaeger-ui/src/api/v3/schemas.ts
sksingh2005: Agreed IDs are enforced by the refinements below, so no IDL change is needed for this PR. Filing the `required:` annotations as jaeger-idl follow-up (precedent: #184, #204).

packages/jaeger-ui/src/api/v3/schemas.ts
sksingh2005: @yurishkuro Let me know if my reasoning is wrong here

packages/jaeger-ui/src/api/v3/client.ts
```diff
@@ -124,6 +131,23 @@ export class JaegerClient {
     });
   }
 
+  /**
+   * Fetch a single trace by ID from /api/v3/traces/{trace_id}.
+   * Validates the grpc-gateway envelope {"result": TracesData} and all nested
+   * OTLP fields at the network boundary. Envelope handling is kept separate from
+   * enrichment so future streaming can be added without touching the parser.
+   */
+  async getTrace(traceId: string): Promise<TracesDataWire> {
```
yurishkuro: why getTrace, which breaks the pattern?
```
$ rg async packages/jaeger-ui/src/api/v3/client.ts
25:  async fetchServices(): Promise<string[]> {
42:  async fetchSpanNames(service: string): Promise<{ name: string; spanKind: string }[]> {
62:  async fetchTraceSummaries(query: SearchQuery): Promise<TraceSummary[]> {
134:  private async fetchWithTimeout(url: string, timeout = 10000): Promise<Response> {
```

packages/jaeger-ui/src/api/v3/schemas.ts
yurishkuro: I still prefer we fix the OpenAPI spec in jaeger-idl generation process. It makes for a better wire contract documentation, even if we do not explicitly need that validation here.

packages/jaeger-ui/src/api/v3/schemas.ts
yurishkuro: add a comment explaining the representation

packages/jaeger-ui/src/api/v3/schemas.ts
```diff
@@ -62,9 +69,224 @@ export const ArrayValueSchema = opentelemetry_proto_common_v1_ArrayValue;
...
+        const present = ANY_VALUE_KEYS.filter(k => val[k] !== undefined);
+        ctx.addIssue({
+          code: z.ZodIssueCode.custom,
+          message: `AnyValue must have exactly one value field, got ${present.join(', ')}`,
+        });
+      }
+    })
+);
+
+// `values` stays optional: the backend drops it entirely for empty arrays (`arrayValue: {}` on the wire).
```
yurishkuro, line 188: > `arrayValue: {}` on the wire

I don't follow this, if it's dropped then there is no `arrayValue` on the wire, if it is not dropped I'd expect `arrayValue: []`, not `{}`

yurishkuro: or does OTLP wants to send something like this?
```
arrayValue: {
  values: [ .... ]
}
```

yurishkuro: nvm I see it in the fixture

packages/jaeger-ui/src/api/v3/schemas.ts
sksingh2005: Agreed filing it in jaeger-idl via `field_behavior` annotations (same mechanism as #184/#204), so the spec documents the contract on its own merits.

sksingh2005: I looked into the generation path: OTLP fields can't take `field_behavior` (external protos), so the fix is extending `prune-openapi` (`required:` injection + envelope wrap + bytes format), same pipeline as its existing patches. Will be filing it in jaeger-idl

sksingh2005:
> OTLP-JSON spec says:
> 
> > Mandatory unknown field ignorance. OTLP-JSON receivers must silently ignore unrecognized fields and unmarshal the rest of the payload. This enforces binary-protobuf-like backward compatibility when new fields are introduced to the telemetry schema.
> 
> Is that satisfied by the validation?

Yes every refined schema uses .passthrough(), so unrecognized fields are silently ignored and the rest of the payload validates, per the OTLP-JSON mandate. The envelope itself is also .passthrough(). Pinned by a contract test that parses a payload with unknown fields at the envelope, span, and status levels.

packages/jaeger-ui/src/api/v3/schemas.ts
yurishkuro: > I looked into the generation path: OTLP fields can't take field_behavior (external protos)

External doesn't mean we cannot extend them before using for generation, we do it all the time, just see Jaeger backend's proto steps and .sed files.

packages/jaeger-ui/src/api/v3/schemas.ts
```diff
@@ -62,9 +69,226 @@ export const ArrayValueSchema = opentelemetry_proto_common_v1_ArrayValue;
...
+] as const;
+
+type AnyValueWire = {
+  stringValue?: string;
+  boolValue?: boolean;
+  intValue?: string;
+  doubleValue?: number;
+  arrayValue?: { values?: AnyValueWire[] };
+  kvlistValue?: { values?: { key: string; value: AnyValueWire }[] };
+  bytesValue?: string;
```
yurishkuro, line 147:
```suggestion
  // base64 on the wire; z.string().base64() checks alphabet and keeps base64
  bytesValue?: string;
```

yurishkuro (approved):

parshipcy (approved):
lgtm

parshipcy:
ok so once this merges, the next step is #4129 rather than a new parser.

sksingh2005:
@parshipcy yes working towards that now:)

PR #4457, opened 2026-09-08
fix(trace-viewer): correct incomplete-trace tooltip

### LINKED ISSUE #4422: [Bug]: Incomplete-trace tooltip tells users to reload for conditions reloading cannot fix
### What happened?

`getIncompleteTraceTooltip`
([`trace-viewer.ts:69-77`](https://github.com/jaegertracing/jaeger-ui/blob/af73d50b12a9ca8785ca2faa95a4a33eea908224/packages/jaeger-ui/src/model/trace-viewer.ts#L69-L77))
renders:

> This trace may be incomplete: 1 span has missing parent span. This can happen if the trace
> is still being collected when you view it. Try again later by opening or reloading the
> trace to see whether more spans are available.

That copy asserts one cause, but the count driving it admits others. The count comes from
[`OtelTraceFacade.ts:35-39`](https://github.com/jaegertracing/jaeger-ui/blob/af73d50b12a9ca8785ca2faa95a4a33eea908224/packages/jaeger-ui/src/model/OtelTraceFacade.ts#L35-L39):

```ts
// Calculate orphan span count
// A span is orphaned if it has a parentSpanID but the parent is not in the trace
this._orphanSpanCount = this._spans.filter(
  s => s.parentSpanID && !this._spanMap.has(s.parentSpanID)
).length;
```

This counts *any* span whose `parentSpanID` does not resolve inside the returned trace. That
includes conditions no amount of reloading can change:

- **The parent was dropped by sampling.** The child is exported and stored; the parent never
  was. The child arrives with a `parentSpanID` pointing at a span that does not exist in the
  backend at all.
- **The parent aged out of retention.** The parent was stored, then expired while the child
  is still within the retention window.

In both cases the tooltip tells the user to wait for data that is never coming. The badge
never clears, and reloading never changes the count.

The tooltip is shown in two places, both with the same copy:
[`TracePageHeader.tsx:107-123`](https://github.com/jaegertracing/jaeger-ui/blob/af73d50b12a9ca8785ca2faa95a4a33eea908224/packages/jaeger-ui/src/components/TracePage/TracePageHeader/TracePageHeader.tsx#L107-L123)
and
[`ResultItem.tsx:84-91`](https://github.com/jaegertracing/jaeger-ui/blob/af73d50b12a9ca8785ca2faa95a4a33eea908224/packages/jaeger-ui/src/components/SearchTracePage/SearchResults/ResultItem.tsx#L84-L91).
Both were confirmed to render the badge for the trace below.

### Steps to reproduce

1. Start Jaeger all-in-one with default memory storage:

   ```
   docker run --rm -p 16686:16686 -p 4317:4317 -p 4318:4318 \
     cr.jaegertracing.io/jaegertracing/jaeger:2.19.0
   ```

   Verified against digest
   `sha256:ede4864215be4cd85bd8c3129a2fea6c5713c5653c7282c429dba123014bc68b`.

2. Send two spans over OTLP/HTTP JSON: a root span, and a second span whose `parentSpanId`
   references `00f067aa0ba902ff`, an ID that is never ingested. The root keeps the trace
   well-formed, so the orphan count is the only thing under test. (Move the four
   `*UnixNano` values to the present if you also want the trace to fall inside the search
   page's default lookback window.)

   ```
   curl -s -X POST http://localhost:4318/v1/traces \
     -H 'Content-Type: application/json' \
     -d '{
     "resourceSpans": [
       {
         "resource": {
           "attributes": [
             { "key": "service.name", "value": { "stringValue": "orphan-probe" } }
           ]
         },
         "scopeSpans": [
           {
             "scope": { "name": "orphan-probe", "version": "0.0.1" },
             "spans": [
               {
                 "traceId": "5b8aa5a2d2c872e8321cf37308d69df2",
                 "spanId": "051581bf3cb55c13",
                 "name": "root-span",
                 "kind": 2,
                 "startTimeUnixNano": "1788347030000000000",
                 "endTimeUnixNano": "1788347030200000000"
               },
               {
                 "traceId": "5b8aa5a2d2c872e8321cf37308d69df2",
                 "spanId": "00f067aa0ba902b7",
                 "parentSpanId": "00f067aa0ba902ff",
                 "name": "child-of-a-span-that-was-never-sent",
                 "kind": 2,
                 "startTimeUnixNano": "1788347030010000000",
                 "endTimeUnixNano": "1788347030150000000"
               }
             ]
           }
         ]
       }
     ]
   }'
   ```

   The collector returns `HTTP 200` with body `{"partialSuccess":{}}`.

3. Confirm the backend kept the dangling reference. `GET /api/traces/5b8aa5a2d2c872e8321cf37308d69df2`
   — the endpoint the UI actually calls — returns the second span with:

   ```json
   "references": [
     {
       "refType": "CHILD_OF",
       "traceID": "5b8aa5a2d2c872e8321cf37308d69df2",
       "spanID": "00f067aa0ba902ff"
     }
   ],
   "warnings": [
     "parent span ID=00f067aa0ba902ff is not in the trace; skipping clock skew adjustment",
     "parent span ID=00f067aa0ba902ff is not in the trace; skipping clock skew adjustment"
   ]
   ```

   (The identical warning is returned twice; that duplication is incidental to this report.)

4. Open `http://localhost:16686/trace/5b8aa5a2d2c872e8321cf37308d69df2`. The trace renders
   normally — header reads `Services 1`, `Depth 1`, `Total Spans 2` — with an orange
   **Incomplete** badge beside them.

5. Hover the badge. It reads, verbatim:

   ```
   This trace may be incomplete: 1 span has missing parent span. This can happen if the trace is still being collected when you view it. Try again later by opening or reloading the trace to see whether more spans are available.
   ```

   The same badge appears on the search results row for service `orphan-probe`.

6. Reload the trace any number of times. The count stays at 1 and the badge never clears,
   because the referenced parent was never sent and never will be.

### Expected behavior

The tooltip should not assert a single cause it cannot distinguish. It should say that some
spans reference a parent that is not present in this trace, and that this can be because
collection is still in progress **or** because the parent was never stored — dropped by
sampling, or aged out of retention. The "try again later" suggestion belongs as one
possibility rather than as the explanation, so a user looking at a permanently incomplete
trace is not told to keep waiting.

### Additional context

The change looks contained rather than sprawling. The copy lives in that one function, and
exactly three assertions pin it, in
[`trace-viewer.test.ts`](https://github.com/jaegertracing/jaeger-ui/blob/af73d50b12a9ca8785ca2faa95a4a33eea908224/packages/jaeger-ui/src/model/trace-viewer.test.ts#L6-L21):
[lines 7-10](https://github.com/jaegertracing/jaeger-ui/blob/af73d50b12a9ca8785ca2faa95a4a33eea908224/packages/jaeger-ui/src/model/trace-viewer.test.ts#L7-L10)
(singular form),
[lines 12-15](https://github.com/jaegertracing/jaeger-ui/blob/af73d50b12a9ca8785ca2faa95a4a33eea908224/packages/jaeger-ui/src/model/trace-viewer.test.ts#L12-L15)
(plural form) and
[lines 17-20](https://github.com/jaegertracing/jaeger-ui/blob/af73d50b12a9ca8785ca2faa95a4a33eea908224/packages/jaeger-ui/src/model/trace-viewer.test.ts#L17-L20)
(presence of the reload suggestion). Only the third would need rewriting for a copy change;
the first two assert singular/plural agreement, which any replacement should keep.

Worth noting for whoever picks this up: the backend already distinguishes this case. It
attaches `parent span ID=<id> is not in the trace; skipping clock skew adjustment` to the
span's `warnings`, so the UI has a per-span signal available if the copy should ever
differentiate rather than merely soften.

All references above are pinned to `af73d50b12a9ca8785ca2faa95a4a33eea908224`.

badnikhil:
- Fixes #4422

## Description of the changes
- The tooltip blamed collection lag and told the user to reload. The orphan count fires for any unresolved `parentSpanID`, including parents dropped by sampling or aged out of retention, which reloading never recovers.
- Reworded in the shared `getIncompleteTraceTooltip()`. Both renderers already route through it.

- Two tests added in `trace-viewer.test.ts`, verified failing without the change.
- `npx vitest run` on the touched files, green. Note the suite is not green on `main` (11 files fail unmodified).

## Checklist
- [X] I have read https://github.com/jaegertracing/jaeger/blob/main/CONTRIBUTING_GUIDELINES.md
- [X] I have run lint and test steps successfully: `make lint test`

- [X] **Moderate**: AI helped with code generation or debugging specific parts

### DIFF
--- a/packages/jaeger-ui/src/model/trace-viewer.test.ts
+++ b/packages/jaeger-ui/src/model/trace-viewer.test.ts
@@ -18,4 +18,15 @@ describe('getIncompleteTraceTooltip', () => {
     const result = getIncompleteTraceTooltip(1);
     expect(result).toContain('opening or reloading the trace');
   });
+
+  it('names the causes that reloading cannot fix', () => {
+    const result = getIncompleteTraceTooltip(1);
+    expect(result).toContain('dropped by sampling');
+    expect(result).toContain('aged out of retention');
+  });
+
+  it('makes the reload suggestion conditional', () => {
+    const result = getIncompleteTraceTooltip(1);
+    expect(result).toContain('in which case opening or reloading the trace');
+  });
 });
--- a/packages/jaeger-ui/src/model/trace-viewer.ts
+++ b/packages/jaeger-ui/src/model/trace-viewer.ts
@@ -71,8 +71,9 @@ export function getIncompleteTraceTooltip(orphanCount: number): string {
   const verb = orphanCount !== 1 ? 'have' : 'has';
   return (
     `This trace may be incomplete: ${orphanCount} ${noun} ${verb} missing parent ${noun}. ` +
-    `This can happen if the trace is still being collected when you view it. ` +
-    `Try again later by opening or reloading the trace to see whether more spans are available.`
+    `The parents may still be in flight, in which case opening or reloading the trace will ` +
+    `show more spans. They may also never arrive, if they were dropped by sampling or have ` +
+    `aged out of retention.`
   );
 }
 


yurishkuro:
I think the wording is fine as is.

PR #4458, opened 2026-09-09
fix(api): surface backend errors returned with a 200 status

### LINKED ISSUE #3079: [Bug]: UI Shows Infinite Loading for Non-Existent Trace ID

SurefireStudios:
Fixes #3079.

Asking for a trace ID that does not exist answers HTTP 200 with an empty `data` array and an `errors` entry that has no `code` field.

`getJSON` only inspected the body when the status was 400 or above, so that resolved as a success, and `transformTraceData(response.data[0])` then destructured `undefined`:

```
TypeError: Cannot destructure property 'traceID' of 'data' as it is undefined.
```

So the backend's `trace not found` was discarded and the reader got an internal JavaScript error, or the spinner in the issue's Compare screenshot.

## The change

`getJSON` now treats a body as failed when it carries a non-empty `errors` array and no usable payload, whatever the status code.

Bodies carrying both data and errors are partial successes and still pass through untouched, and nothing in the tree currently reads `errors` off a resolved response, so no existing consumer changes.

`getMessageFromError` returns `msg` when there is no `code`, which previously fell through to the stringify branch and produced `{"msg":"trace not found",…}` instead of the sentence.

Error construction is now shared between both paths, so a body-level failure carries the same `httpStatus`, `httpBody` and `httpUrl` fields a status-code failure already did.

## Tests

Three added in `jaeger.test.js`: a 200 that is only errors, a 200 carrying data alongside errors, and `getMessageFromError` with `msg` and no `code`.

The first and third fail without the source change, which I verified by reverting and re-running.

## Verification

Node 24.21.0, pnpm 12.1.0: `vitest run src/api/jaeger.test.js` gives 22 passed, and a full `vitest run` shows no new failures.

The files that do fail fail identically on `4f3b7c9` without this change — `TextEncoder is not a constructor` environment errors on Windows.

`vp lint` reports 21 `no-explicit-any` warnings on `jaeger.ts`, exactly the count on the unmodified file; my first draft added three, so I retyped the new code with a type guard over `unknown` to return to baseline.

## Scope

Fixed in `getJSON` rather than per caller, since the same 200-with-errors shape can come from any endpoint.


packages/jaeger-ui/src/api/jaeger.ts
```diff
@@ -51,8 +70,29 @@ function getJSON(url: string, options: FetchOptions = {}): Promise<any> {
...
     if (response.status < 400) {
-      return response.json();
+      return response.json().then((body: unknown) => {
+        // Jaeger answers 200 with an `errors` array and no payload when the query
+        // itself was fine but produced nothing, such as a trace ID that does not
+        // exist. Treated as success, that empty body reaches the caller and fails
+        // later as an opaque internal error, losing the message the backend gave us.
+        if (isErrorOnlyPayload(body)) {
+          const errorMessage = body.errors.map(err => getMessageFromError(err, response.status)).join('; ');
+          throw makeError(errorMessage, JSON.stringify(body), JSON.stringify(body, null, 2));
```
yurishkuro, line 92: why encode and pass body twice? helper design seems inefficient

yurishkuro (changes requested):
I don't think we need to merge this. The trace loading path is being upgrade to use OTEL data model, it changes the response format. And this change sits inside getJSON() which is used not only for trace loading but for other fetches, we don't know if their data shape follows the same patterns.

packages/jaeger-ui/src/api/jaeger.ts
SurefireStudios: Good catch — it was encoding twice and throwing one away. `httpBody` read `bodyTextFmt || bodyText` and the formatted string was always truthy on that path, so the compact one was dead.

`makeError` now takes the body already formatted, and each caller passes what it has: the pretty-printed JSON on the 200 path, and `bodyTextFmt ?? bodyText` on the status-code path where the raw text is still the fallback for a body that doesn't parse. Same `httpBody` at both sites, one parameter fewer.

Pushed as 252afc1. 22 tests in `jaeger.test.js` pass, `tsc --noEmit` is clean, and lint warning count is unchanged at 21 (all pre-existing `no-explicit-any` below this function).

PR #4463, opened 2026-09-11
chore(lint): remove react-x plugin

### LINKED ISSUE #3685: [chore]: Remove eslint-plugin-react-x dependency
Tracking issue.

In https://github.com/jaegertracing/jaeger-ui/pull/3684 we're migrating to oxlint from vite-plus, but leaving eslint-plugin-react-x for now (running via bridge) because oxlint does not provide similar level of checks. Upstream issue https://github.com/oxc-project/oxc/issues/20791.

akashchamp:
- Resolves #3685

## Description of the changes
- Replace the `eslint-plugin-react-x` bridge (`jsPlugins`) in `vite.config.ts` with Oxlint's native `react/rules-of-hooks` and `react/exhaustive-deps` rules.
- Remove the `eslint-plugin-react-x` devDependency, its Knip `ignoreDependencies` exception, and the ESLint subtree from `pnpm-lock.yaml`.
- Drop the `brace-expansion` override from `pnpm-workspace.yaml`. It was only reachable through the ESLint tree, so keeping it would fail `check-overrides` as a phantom override.
- Rename the two `eslint-disable-next-line react-x/exhaustive-deps` suppressions to `oxlint-disable-next-line react/exhaustive-deps` and update `BUILD.md`.

Lint config, docs and dependency files only; no runtime UI change.

- `pnpm install --frozen-lockfile`
- `pnpm run lint` (oxfmt, tsc, oxlint, knip, check-overrides all pass)
- `pnpm run build`
- `pnpm test`

## Checklist
- [ ] I have added unit tests for the new functionality (not applicable: lint config and docs only)

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -25,7 +25,6 @@ export default defineConfig({
   },
   lint: {
     plugins: ['oxc', 'typescript', 'unicorn', 'react', 'jest', 'import'],
-    jsPlugins: ['eslint-plugin-react-x'],
     categories: {
       correctness: 'warn',
     },
@@ -120,8 +119,8 @@ export default defineConfig({
       'react/no-children-prop': 'error',
       'unicorn/no-useless-spread': 'error',
       'unicorn/no-new-array': 'error',
-      'react-x/rules-of-hooks': 'error',
-      'react-x/exhaustive-deps': 'error',
+      'react/rules-of-hooks': 'error',
+      'react/exhaustive-deps': 'error',
       'jest/no-disabled-tests': 'warn',
       'jest/no-focused-tests': 'error',
       'jest/no-identical-title': 'error',
--- a/BUILD.md
+++ b/BUILD.md
@@ -4,9 +4,9 @@
 
 ### Dependencies (dev and otherwise)
 
-#### `eslint-plugin-react-x`
+#### React hooks rules
 
-Used by Oxlint (via `jsPlugins`) to enforce React hooks rules (`react-x/rules-of-hooks`, `react-x/exhaustive-deps`). Oxlint does not have a native implementation of hooks rules and delegates to this ESLint plugin.
+Oxlint natively enforces React hooks rules through `react/rules-of-hooks` and `react/exhaustive-deps`.
 
 #### `typescript`
 
--- a/knip.config.ts
+++ b/knip.config.ts
@@ -36,11 +36,6 @@ const config: KnipConfig = {
         // directly by any source file so knip cannot detect it.
         '@babel/core',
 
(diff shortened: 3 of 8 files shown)


vite.config.ts
```diff
@@ -120,8 +119,8 @@ export default defineConfig({
       'react/no-children-prop': 'error',
       'unicorn/no-useless-spread': 'error',
       'unicorn/no-new-array': 'error',
-      'react-x/rules-of-hooks': 'error',
-      'react-x/exhaustive-deps': 'error',
+      'react/rules-of-hooks': 'error',
+      'react/exhaustive-deps': 'error',
```
yurishkuro, line 123: this is just moving already existing rules from a different namespace. But what compensates for the removal of rules from eslint-plugin-react-x? If there is no equivalent to its checks why should we be removing it?

jkowall (dismissed):
Thanks — dropping eslint-plugin-react-x for native Oxlint hooks rules looks right.

Before re-review: please rebase onto current main and resolve the merge conflicts (GitHub shows the PR as conflicting).

Optional after that (non-blocking): RFC 0005 still lists react-x/jsPlugin — worth a small doc sync so it matches BUILD.md.

yurishkuro:
> dropping eslint-plugin-react-x for native Oxlint hooks rules looks right.

why? I would like to see comparative analysis

vite.config.ts
akashchamp: `eslint-plugin-react-x` was only wired up for two rules here, `rules-of-hooks` and `exhaustive-deps` (see `vite.config.ts` on main), nothing else from that plugin was enabled. Oxlint's own `react` plugin has had native implementations of those same two checks for a while now, so this just points the config at them instead of bridging out to the ESLint plugin through `jsPlugins`.

To check they're not weaker, I ran oxlint against a small component with a hook called inside an `if` and a `useEffect` missing a dependency, with `eslint-plugin-react-x` removed entirely:
```
error react-hooks(rules-of-hooks): React Hook "useState" is called conditionally...
error react-hooks(exhaustive-deps): React Hook useEffect has a missing dependency: 'y'...
```
Same violations, same messages. `eslint-plugin-react-x` was also the only entry in `jsPlugins`, so dropping it removes the one place this config still shelled out to a real ESLint parse instead of staying on oxlint's native path.

akashchamp:
Rebased onto main (this had gone conflicting) and synced RFC 0005's wording to match BUILD.md, since it still described the old `jsPlugins` bridge. Full `pnpm run lint`, `pnpm test` (plexus + jaeger-ui), and `pnpm run build` all pass on the rebased head. Reply with the rule-coverage comparison is on the `vite.config.ts` thread.

yurishkuro (approved):

PR #4469, opened 2026-09-12
feat(trace): Add copy span summary button

### LINKED ISSUE #4468: [Feature]:  Added a "Copy Span Summary" Button
### What happened?

While investigating traces in Jaeger UI, it would be useful to have a button in the span details that copies a concise plain-text summary of the span, including the service name, operation name, duration, and span ID.
This would make it easier to share relevant span context in Slack, Jira, or incident discussions without manually copying individual fields or sharing a lengthy URL.

### Steps to reproduce

2. Click on a span to expand its details.
3. Look at the bottom right corner of the span details view next to the Span ID.
4. Only a "Copy" button for the deep link URL is present; there is no option to copy a plain-text summary of the span itself.

### Expected behavior

There should be a "Summary" copy button next to the deep-link copy button that copies a concise plain-text one-liner format, such as:
<service-name> | <operation-name> | <duration> | spanID: <spanID>

### Relevant log output

```shell
N/A
```

### Screenshot

N/A

### Additional context

Currently, users have to manually highlight and copy the service name, operation name, duration, and span ID separately to share in chat discussions or tickets. Adding a dedicated "Summary" copy button saves time during active debugging and incident response.

### Jaeger backend version

_No response_

### SDK

OpenTelemetry SDK

### Pipeline

OTEL SDK -> OTEL Collector -> Jaeger

### Stogage backend

Elasticsearch / Memory

### Operating system

Linux

### Deployment model

Docker

### Deployment configs

```shell
N/A
```

JituRewar:
Fixes #4468 

- Adds a convenient plain-text "Summary" copy button to the span details view, allowing users to copy concise span details directly into Slack, Jira, or incident chat without manually copying multiple fields or opening URLs.

## Description of the changes
- Added `IoDocumentTextOutline` icon import from `react-icons/io5`.
- Constructed a `spanSummaryText` one-liner (`<serviceName> | <operationName> | <duration> | spanID: <spanID>`).
- Rendered a second `CopyIcon` button next to the existing deep-link copy button in `SpanDetail/index.tsx`.
- Updated existing unit test in `SpanDetail/index.test.jsx` and added a new test validating that the summary button copies the correct service, operation, and span ID.

- Tested via automated test suites:
  - `pnpm --filter @jaegertracing/jaeger-ui test src/components/TracePage/TraceTimelineViewer/SpanDetail/index.test.jsx`
  - `pnpm test` (all 271 test files )
  - `pnpm run lint` and `pnpm run tsc-lint` passed with 0 errors
  - `pnpm run fmt-lint` passed cleanly
  - `pnpm run build` succeeded

## Checklist


Harizz076:
Please add a before/after screenshot and screencast.

JituRewar:
@Harizz076 Sure, i have attached screen recordings before and after, take a look.
Before -
[Screencast From 2026-09-13 21-34-10.webm](https://github.com/user-attachments/assets/623824a3-831b-4ef6-a416-c6c54cac51e8)

After -
[Screencast From 2026-09-13 21-36-37.webm](https://github.com/user-attachments/assets/15305c78-d85e-4e73-8594-d1ee79f64f14)

Harizz076:
I went through your screencasts, please add them in the PR description. This solution doesn't really save much time or convenience in my opinion, better to wait for maintainer feedback. Here's me extracting almost the same information without using your feature:

https://github.com/user-attachments/assets/962e7df1-7a5b-450b-bdf6-696c1f708d82

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/index.tsx
```diff
@@ -157,6 +160,13 @@ export default function SpanDetail(props: SpanDetailProps) {
           tooltipTitle="Copy deep link to this span"
           buttonText="Copy"
         />
+        <CopyIcon
+          copyText={spanSummaryText}
```
Harizz076, line 164: This footer is currently a right-aligned debug-info `<small>` with no action layout or overflow behavior. Adding another full text button makes the action overflow the narrow span-details panel; “Summary” is clipped in the screenshot, so the control is not reliably usable.
Please move the copy controls into a dedicated actions container and define its narrow-width behavior (wrapping, compact icon-only controls, or an overflow menu). The span ID/debug text should remain separate from those actions.

packages/jaeger-ui/src/components/TracePage/TraceTimelineViewer/SpanDetail/index.test.jsx
```diff
@@ -292,13 +292,23 @@ describe('<SpanDetail>', () => {
...
-    const copyText = copyIcon.getAttribute('data-copy-text');
+    const copyIcons = screen.getAllByTestId('copy-icon');
+    const copyText = copyIcons[0].getAttribute('data-copy-text');
 
-    expect(copyIcon).toBeInTheDocument();
+    expect(copyIcons[0]).toBeInTheDocument();
     expect(copyText).toContain(`?uiFind=${props.span.spanID}`);
   });
 
+  it('renders a span summary copy button with service, operation and spanID', () => {
```
Harizz076, line 302: Could we avoid asserting against `getAllByTestId(...)[1]` here? That couples both copy-action tests to DOM order, and the `CopyIcon` mock renders both controls as “Copy,” so this does not verify that the new action is actually labelled “Summary.”

Please expose `buttonText` in the mock, select the button by its accessible name, and assert the complete copied value, including the formatted duration, rather than three independent substrings.

JituRewar:
Yes, I agree with your point, the feature may not save a significant amount of time in every case. While exploring Jaeger UI, I noticed that having a Copy Span Summary button next to the existing Copy Link button could still be a small convenience for quickly copying the key span details in a single action. That was the main idea behind this feature.

jkowall (changes requested):
Thanks — copying `service | operation | duration | spanID` is a useful affordance, and `formatDurationCompact` / `CopyIcon` wiring looks right.

Before approve:
1. Layout: the footer `<small className="SpanDetail--debugInfo">` is already tight. A second full "Summary" text button clips in a narrow panel (Harizz076's note). Please put the copy controls in a dedicated actions container and define narrow-width behavior (wrap, icon-only, or overflow), keeping the SpanID debug text separate.
2. Tests: avoid `getAllByTestId('copy-icon')[1]`. Expose `buttonText` on the CopyIcon mock, select by accessible name ("Summary"), and assert the full copied string including the formatted duration.
3. Please rebase onto current `main` — Lint is red on copyright-year for unrelated `metrics.ts` / `metrics.test.js` because this branch is behind.

JituRewar:
@jkowall Thanks for the detailed review, i have addressed all three points

1. Layout: Moved the copy controls into a dedicated actions container and added narrow-width handling so the Summary button no longer clips, while keeping the SpanID debug text separate.
2. Tests: Updated the `CopyIcon` mock to expose `buttonText`, selected the Summary action by its accessible name, and updated the assertion to verify the complete copied summary including the formatted duration.
3. Rebase: Rebased the branch onto the latest `main`, which also resolves the unrelated copyright-year lint failures in `metrics.ts` and `metrics.test.js`.

The changes are pushed in the latest commit. Please take another look when you get a chance.

PR #4480, opened 2026-09-16
test(api/v3): Pin v3 wire contract with capture fixture

sksingh2005:

Part of #4278 (OTel-native Jaeger UI).

The v3 boundary validation added in #4455 used a hand-built response literal. Nothing in the repository parsed a genuine Jaeger response, so the tests could miss drift between the schemas and the backend wire format.

This PR replaces that literal with a reproducible Jaeger 2.21.0 capture. A checked-in OTLP input passes through a digest-pinned Jaeger instance, and the captured response becomes the valid case for the v3 contract tests. This follows the agreed fixture-plus-Zod approach without adding a mock server.

## Description of the changes

- Adds `v3-trace-input.json`, a fixed OTLP payload with explicit IDs and timestamps. It covers all seven `AnyValue` variants, omitted and numeric enum values, nested `kvlistValue`, events, links, and status values.
- Adds `v3-trace-output.json`, captured from `GET /api/v3/traces/0123456789abcdef0123456789abcdef`. Its parsed content matches the earlier Jaeger 2.13.0 capture of the same input.
- Replaces the duplicated `validEnvelope` literals in `schemas.trace-contract.test.ts` and `client.trace-contract.test.ts` with the captured response. Malformed cases clone the same fixture.
- Adds positive `traceId` and `spanId` assertions, so the tests cover valid hex encoding as well as rejection cases.
- Strengthens the `fetchTrace` success test to verify the unwrapped trace data, span count, and trace ID.
- Locates spans by name in the schema contract tests instead of depending on backend response order.
- Moves the shared schema parse into `beforeEach`.
- Adds `scripts/v3-fixture/docker-compose.yml`, which pins the Jaeger 2.21.0 image by tag and digest. The fixture filenames remain unversioned, and Renovate can update the image declaration.
- Adds `scripts/generate-v3-fixture.ts`, which:
  - reads the trace ID and expected span count from `v3-trace-input.json`;
  - rejects inputs with missing or mixed trace IDs;
  - starts Jaeger through Docker Compose using its all-in-one defaults;
  - submits the checked-in input;
  - passes `raw_traces=false` explicitly, matching the response used by the UI;
  - waits until every input span is queryable;
  - writes the output or verifies it with `--check`;
  - normalizes only unordered collections when checking for drift;
  - stops and removes the Compose resources after each run.
- Adds `pnpm run generate:v3-fixture` for generating or checking the capture.
- Adds the `Verify v3 Fixture` workflow. It regenerates and compares the response when the input, output, generator, Compose file, or workflow changes.
- Documents the fixture provenance, regeneration process, ordering rules, and drift behavior in `src/api/v3/README.md`.

- `pnpm --filter @jaegertracing/jaeger-ui test src/api/v3`: 106 tests passed.
- `pnpm --filter @jaegertracing/jaeger-ui test`: 273 files and 3,738 tests passed.
- `pnpm run tsc-lint`: passed.
- `Verify v3 Fixture`: passed against a fresh Jaeger 2.21.0 capture.

## Checklist

- [x] I have read [[CONTRIBUTING.md](https://github.com/jaegertracing/jaeger-ui/blob/main/CONTRIBUTING.md)](https://github.com/jaegertracing/jaeger-ui/blob/main/CONTRIBUTING.md).
- [x] I have signed all commits.
- [x] I have added unit tests for the new functionality.
- [x] I have run the required lint, test, build, and fixture-verification checks.

## AI Usage in this PR

See the [[AI Usage Policy](https://github.com/jaegertracing/jaeger/blob/main/AI_POLICY.md)](https://github.com/jaegertracing/jaeger/blob/main/AI_POLICY.md).

- [ ] **None**: No AI tools were used in creating this PR.
- [ ] **Light**: AI provided minor assistance such as formatting or simple suggestions.
- [x] **Moderate**: AI helped with code generation or debugging specific parts.
- [ ] **Heavy**: AI generated most or all of the code changes.


yurishkuro:
I do not follow what 2.13 refers to. If it's Jaeger version why 2.13?

sksingh2005:
@yurishkuro My bad, that pin was stale from the previous capture work. Refreshed the fixture to a 2.21.0 capture.

yurishkuro:
I'm not a fan of introducing a non-reproducible artifact into the test pipeline. How was the trace created before it was retrieved from Jaeger? Which storage was used? All these factors can change the shape of the trace data, and for schema compliance testing we need to control the creation of the fixture end to end.

sksingh2005:
@yurishkuro 
**How the trace was created.** A fixed OTLP JSON payload with explicit trace and span IDs and explicit nanosecond timestamps, posted to `localhost:4318/v1/ traces`, then read back with `GET /api/v3/traces/{id}`.

**Which storage.** The image default. The container ran `jaegertracing/jaeger:2.21.0` with no storage configuration. I also did not pass `raw_traces`, so the read went through the enriched path. 

I now got it that this is not good enough for schema compliance testing. I checked what the backend actually changed, and it was attribute order and an empty `status` object, both of which the spec allows and neither of which the assertions depend on. That is luck rather than control, and it would not survive a storage change.

I will make the creation reproducible from this repo: generate the payload with `telemetrygen`, pin the image tag and storage explicitly, pass `raw_traces=true`, and vendor the command next to the fixture so the capture can be regenerated and diffed.

One question i had. Is a checked-in fixture with a pinned, documented pipeline acceptable, or do you want the backend out of the loop entirely? Only the `{"result": ...}` envelope needs a server response. The span-level assertions could run on generator output, which is what `schemas.test.ts` already does with `otlp2jaeger-in.json`.

sksingh2005:
Pushed the reproducible pipeline. Two corrections to what I wrote above.

I kept `raw_traces=false`, not true. `fetchTrace `sends no `raw_traces `param, so the UI receives the server default, the enriched path. Capturing with true would pin a document the UI never parses. The flag is now passed explicitly so the capture does not depend on the default staying put.

The probe is hand-authored JSON, not telemetrygen output. Fixed IDs and timestamps make it byte-identical on every run with no tool dependency, and it sits next to the fixture so input and output are reviewed together.

scripts/generate-v3-fixture.sh
```diff
@@ -0,0 +1,281 @@
...
+#
+# Usage: scripts/generate-v3-fixture.sh [--check]
+#   --check regenerates into a temporary file and diffs it against the committed
+#   fixture instead of overwriting it. Exits non-zero on drift.
+
+set -euo pipefail
+
+# jaegertracing/jaeger:2.21.0, multi-arch manifest. IMAGE_TAG names the version
+# the digest resolves to and supplies the fixture filename; bump both together.
+readonly IMAGE="jaegertracing/jaeger@sha256:3d0ac795ff98aa04d1be04311d2dac6c25b4bfc8322dc02e53bc5b170c5018c3"
```
yurishkuro: this is not a good solution. I recommend creating a docker compose file instead, which will be recognized by renovate bot and automatically upgraded when new Jaeger releases are published (see similar files in the main repo).

scripts/generate-v3-fixture.sh
```diff
@@ -0,0 +1,281 @@
...
+readonly IMAGE_TAG="2.21.0"
```
yurishkuro: It does not make sense to me to pin the version like that and use it as part of the file name. The pin will happen in docker-compose, the fixture name should not have a version - you are creating maintenance headache by doing so.

scripts/generate-v3-fixture.sh
```diff
@@ -0,0 +1,281 @@
...
+  fi
+  return 0
+}
+trap cleanup EXIT
+
+# Counts the spans in a trace document read from stdin. Accepts both shapes in
+# play here: the bare TracesData the probe submits, and the {"result": ...}
+# envelope the query API answers with. Prints 0 for anything that does not
+# parse, so the readiness loop keeps waiting rather than failing.
+count_spans() {
```
yurishkuro: I think this all can be done with a single jq command, you do not need embedded python script

scripts/generate-v3-fixture.sh
```diff
@@ -0,0 +1,281 @@
...
+# sync when the payload gains or loses a span.
+expected_spans="$(count_spans <"$probe")"
+if [[ "$expected_spans" == "0" ]]; then
+  echo "Could not count spans in $probe." >&2
+  exit 1
+fi
+
+cleanup
+
+# Fail loudly if the digest stops resolving to the version the fixture is named
```
yurishkuro: set -euo pipefail should already cause script to fail if Docker cannot run the task

scripts/jaeger-fixture-config.yaml
```diff
@@ -0,0 +1,57 @@
+# Copyright (c) 2026 The Jaeger Authors.
```
yurishkuro: I don't see how this is different from all-in-one which requires no config

scripts/generate-v3-fixture.sh
```diff
@@ -0,0 +1,281 @@
...
+    echo "Formatted with vp fmt."
+  else
+    {
+      echo
+      echo "The formatter could not run here, so $fixture is still raw wire JSON."
+      echo "Run 'pnpm exec vp fmt' on it before committing, or CI's fmt-lint will fail."
+    } >&2
+    exit 3
+  fi
+fi
```
yurishkuro: 281 LOC seems like a huge overkill for what this is meant to achieve. Also, could we not write this in typescript as the rest of the repo? Let it throw exceptions etc. if it needs to fail.

scripts/generate-v3-fixture.sh
```diff
@@ -0,0 +1,281 @@
...
+readonly IMAGE="jaegertracing/jaeger@sha256:3d0ac795ff98aa04d1be04311d2dac6c25b4bfc8322dc02e53bc5b170c5018c3"
+readonly IMAGE_TAG="2.21.0"
+readonly CONTAINER="jaeger-v3-fixture"
+readonly TRACE_ID="0123456789abcdef0123456789abcdef"
+readonly RAW_TRACES="false"
+
+script_dir="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)"
+api_dir="$script_dir/../packages/jaeger-ui/src/api/v3"
+probe="$api_dir/v3-trace-probe.json"
+fixture="$api_dir/v3-trace-local-$IMAGE_TAG.json"
```
yurishkuro: I would call these two files -input and -output, for consistency

packages/jaeger-ui/src/api/v3/schemas.trace-contract.test.ts
```diff
@@ -119,30 +48,28 @@ describe('GetTrace wire contract', () => {
   });
 
   it('preserves nested kvlist and arrays inside AnyValue', () => {
-    const parsed = GetTraceResponseSchema.parse(validEnvelope);
-    const nested = parsed.result.resourceSpans![0].scopeSpans![0].spans![0].attributes!.find(
-      a => a.key === 'nested'
-    )!;
+    const parsed = GetTraceResponseSchema.parse(capture);
```
yurishkuro: can this not be done in beforeTest?

yurishkuro (changes requested):

scripts/generate-v3-fixture.sh
parshipcy: +1

scripts/generate-v3-fixture.sh
```diff
@@ -0,0 +1,281 @@
...
+echo "Submitted trace $TRACE_ID with $expected_spans spans"
+
+# Wait for the whole trace rather than for a 200, so a capture can never be
+# short a span. Keep the body that satisfied the gate instead of re-fetching, so
+# the captured document is the one that was actually checked.
+echo -n "Waiting for all $expected_spans spans to be queryable"
+body=""
+stored="false"
+for _ in $(seq 1 30); do
+  body="$(curl --fail --silent "http://localhost:16686/api/v3/traces/$TRACE_ID?raw_traces=$RAW_TRACES" 2>/dev/null || true)"
```
parshipcy: see line 48-49
span count is read from the probe file. The id is not. If someone change ids in `v3-trace-probe.json` and forget this constant, otlp submit still succeed, then the script will wait on the old id and times out. So, two sources of truth.

packages/jaeger-ui/src/api/v3/schemas.trace-contract.test.ts
sksingh2005: Done. I moved the shared `GetTraceResponseSchema.parse(capture)` call into `beforeEach`, and the tests now reuse the parsed value.

scripts/generate-v3-fixture.sh
sksingh2005: Fixed. The TypeScript generator now derives both the trace ID and expected span count from `v3-trace-input.json`. It also rejects inputs with missing or mixed trace IDs, so the input fixture is the single source of truth.

.github/workflows/v3-fixture.yml
```diff
@@ -0,0 +1,56 @@
+name: 'Verify v3 Fixture'
```
yurishkuro:
```suggestion
name: 'Verify API v3 Fixtures'
```

yurishkuro: and rename the file to verify-api-v3-fixtures.yml

.github/workflows/v3-fixture.yml
```diff
@@ -0,0 +1,56 @@
...
+
+# The v3 contract tests parse a captured Jaeger response. Regenerating that
```
yurishkuro:
```suggestion
# The API v3 contract tests parse a captured Jaeger response. Regenerating that
```

.github/workflows/v3-fixture.yml
```diff
@@ -0,0 +1,56 @@
...
+  merge_group:
+  pull_request:
+    branches: [main]
+    paths:
+      - 'scripts/generate-v3-fixture.ts'
+      - 'scripts/v3-fixture/docker-compose.yml'
+      - 'packages/jaeger-ui/src/api/v3/v3-trace-input.json'
+      - 'packages/jaeger-ui/src/api/v3/v3-trace-output.json'
+      - '.github/workflows/v3-fixture.yml'
+  push:
```
yurishkuro: I do not think this needs to run on a push to main if it already ran on the PR.

scripts/generate-v3-fixture.ts
```diff
@@ -0,0 +1,184 @@
...
+        throw new Error('The v3 trace fixture differs from a fresh capture');
+      }
+      console.log('The v3 trace fixture is up to date.');
+      return;
+    }
+
+    await writeFile(OUTPUT_FILE, `${JSON.stringify(output, null, 2)}\n`);
+    execFileSync('pnpm', ['exec', 'vp', 'fmt', OUTPUT_FILE], {
+      cwd: path.resolve(import.meta.dirname, '..'),
+      shell: process.platform === 'win32',
```
yurishkuro: ?  Why can this script not be cross-platform?

scripts/generate-v3-fixture.ts
```diff
@@ -0,0 +1,184 @@
...
+  const input = JSON.parse(await readFile(INPUT_FILE, 'utf8')) as JsonValue;
+  const { expectedSpans, traceId } = inputDetails(input);
+
+  compose('down', '--volumes', '--remove-orphans');
+  try {
+    compose('up', '--detach');
+    const output = await captureTrace(input, traceId, expectedSpans);
+    if (checkOnly) {
+      const committed = JSON.parse(await readFile(OUTPUT_FILE, 'utf8')) as JsonValue;
+      if (!isDeepStrictEqual(normalise(committed), normalise(output))) {
```
yurishkuro: Is "normalizing" just doing the sorting? It could use a comment explaining why we need it.

yurishkuro:
Please update PR title (no version). 

Have you considered using [Testcontainers for Node.js](https://testcontainers.com/modules/typescript/) in `scripts/generate-v3-fixture.ts` instead of shelling out manual `child_process` commands? 

It could significantly simplify the fixture generation:

* **Deterministic Readiness:** Replaces arbitrary `setTimeout` delays or bash retry loops with native polling (e.g., `Wait.forHttp("/", 16686)`). The script proceeds the exact millisecond the API is ready, eliminating race conditions.
* **Guaranteed Cleanup:** Eliminates fragile `try...finally` teardown logic. Testcontainers automatically drops all associated containers when the Node process exits, preventing orphaned background instances and port collisions if the script crashes halfway through.
* **Less Boilerplate:** Keeps environment setup entirely within TypeScript, dropping OS-level utilities (bash, curl) and reducing the orchestration code by roughly 70%.

A quick example of how it could look:

```typescript
import { DockerComposeEnvironment, Wait } from "testcontainers";
import path from "path";

// Adjust path to where your docker-compose.yml lives
const composeFilePath = path.resolve(__dirname, ".."); 

const environment = await new DockerComposeEnvironment(composeFilePath, "docker-compose.yml")
  .withWaitStrategy("jaeger", Wait.forHttp("/", 16686)) 
  .up();

// ... generate telemetry ...
// ... fetch fixture from http://localhost:16686/api/v3/traces ...

// Teardown when done
await environment.down();
```

scripts/generate-v3-fixture.ts
sksingh2005: I made that condition for handling the platform difference. On Windows, pnpm resolves to `pnpm.cmd`, which `execFileSync` cannot run without a shell. On Linux and macOS, it executes `pnpm` directly without one. The current revision no longer invokes pnpm from this script, so the conditional is gone.

yurishkuro (approved):

PR #4483, opened 2026-09-17
refactor(TraceStatistics): Migrate to functional component

### LINKED ISSUE #3384: [UI Refactor] Migrate TraceStatistics to Functional Component
This issue is part of the larger effort to modernize the Jaeger UI codebase (Parent Issue #2610). The goal is to refactor the `TraceStatistics` component from a React Class Component to a React Functional Component using Hooks.

**Target Component:** `packages/jaeger-ui/src/components/TracePage/TraceStatistics/index.tsx`

### Acceptance Criteria

| --- | --- | --- |

### Resources

om7057:
Closes #3384

## What

Converts `TraceStatistics` from a class component to a functional component with hooks.

- `searchInTable` is now a standalone exported pure function instead of an instance method, and is unit-tested directly.
- State: `tableValue`, `wholeTable`, `valueNameSelector1`, `showPopup`, `popupContent` all moved to `useState`. `componentDidUpdate`'s search re-application became a `useEffect` on `[uiFindVertexKeys]`, guarded with a ref so it doesn't fire on the initial mount (matching the original, which never ran `componentDidUpdate` on mount either).
- Dropped `sortIndex`, `sortAsc`, and `valueNameSelector2`. All three were write-only state in the original class component (set, never read), so they had no behavior to preserve.
- No JSX/render output changes: the tree is unchanged, only the component shape and internal wiring.

While converting, found and fixed a real bug: `searchInTable` mutates its input array in place and returns the same reference. Passing that straight into `setTableValue(...)` inside a hook is a no-op, since `useState`'s setter bails out on identical object identity, so the DOM would never update on search. Class `setState` doesn't have that bail-out, which is why the bug was invisible in the original code. Fixed by spreading into a new array before setting state.

## Tests

Rewrote `index.test.jsx` (11 tests) so nothing reaches into component internals via ref (`.state`, `.setState`, calling instance methods directly), since functional components don't expose that. Tests now drive the real `TraceStatisticsHeader` UI (dropdown selection, clicks) and assert on rendered output, mocking `./tableValues` and `./generateDropdownValue` the same way `TraceStatisticsHeader.test.jsx` already does. `searchInTable`'s pure-function tests were moved to their own `describe` block and now import it directly rather than going through a rendered component.

- `pnpm run fmt && pnpm run lint && pnpm test && pnpm run build` all pass clean.

## Visual verification

Same trace, same default view (`Group By: Service Name`), before and after. No JSX changes, so the rendered output is identical. (The differing row-color swatches you may notice between separate screenshots are unrelated to this PR: `colorGenerator` in `src/utils/color-generator.ts` is a module-level singleton that assigns colors by first-seen order across the whole browser session, not per-component.)

**Before (main):**
before-change

**After (this branch):**
After-change


om7057:
@jkowall Could you please take a look at this one?

jkowall (changes requested):
Thanks for the TraceStatistics class→hooks migration (f55629b).

The mapping looks right: useState for the live fields, useEffect on uiFindVertexKeys with an isFirstRender skip mirrors componentDidUpdate, searchInTable is exported and the spread-after-mutate fix is an improvement over the class same-ref setState path. Dropping write-only sortIndex/sortAsc/valueNameSelector2 is fine. RTL coverage for search coloring, grouping, sql.query popup, and sort smoke is in good shape. No Redux/URL write-back in this component.

Requesting changes before approve:

1. searchInTable: TNil is null|undefined, but the guard is only `typeof !== 'undefined'`, so null reaches forEach and throws. Please treat null the same as undefined (and add a unit case).

Also please rebase onto main (branch is ~22 behind). Required lint is currently red on check-copyright-year for unrelated files that main already fixed; rebase should clear that.

Nit: a test that uiFind-only changes with a stable uiFindVertexKeys set do not recolor would document the intentional class parity.

om7057:
Thanks for the detailed review.

1. Fixed in e3e8d252: searchInTable's guard now checks truthiness (`if (uiFindVertexKeys)`) instead of `typeof !== 'undefined'`, so a null value is treated the same as undefined instead of reaching `.forEach` and throwing. Added a regression test (a null uiFindVertexKeys with no throw, matching the existing undefined-case coverage). This also resolves Copilot's null-vertex-keys finding.
2. Rebased onto main, which clears the check-copyright-year lint failure.
3. On the uiFind-only-doesn't-recolor nit: I looked at adding it, but I don't think it's cleanly observable from outside the component. The useEffect's dep array is just [uiFindVertexKeys], so a uiFind-only change does skip re-running searchInTable, but onCellFunction reads the uiFind prop directly when choosing what color to render, independent of whether searchInTable reran. So a render-level assertion would actually see colors respond to uiFind either way, for a different reason than the one the test would be trying to pin down. Happy to add it if you had a specific approach in mind.

fmt/lint/tsc/test/build all pass locally.

